PR #4, opened 2020-03-09
plumbing/object: avoid O(N^2) string building when decoding commit message

mcuadros:
Original created by @dsymonds at https://github.com/src-d/go-git/pull/1246

```
Most commits have relatively small messages, so this was never
noticeable. However, there are some repositories that have
semi-automated messages that can get very large (e.g.
github.com/riscv/riscv-clang and its riscv-trunk branch), on the order
of 109k lines. Changing from string += to using a bytes.Buffer reduces
the time for Commit.Decode for that specific case from 35s to 74ms.

Signed-off-by: David Symonds <dsymonds@golang.org>
```

### DIFF
--- a/plumbing/object/commit.go
+++ b/plumbing/object/commit.go
@@ -177,6 +177,7 @@ func (c *Commit) Decode(o plumbing.EncodedObject) (err error) {
 
 	var message bool
 	var pgpsig bool
+	var msgbuf bytes.Buffer
 	for {
 		line, err := r.ReadBytes('\n')
 		if err != nil && err != io.EOF {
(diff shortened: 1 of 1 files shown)


dsymonds:
Oh, there's a new repo? Did you want me to make a new PR against this one?

mcuadros:
Yes, the previous repository is completely blocked you can read about it at [comment](https://github.com/src-d/go-git/issues/1295#issuecomment-592965250). Source{d} is in bankrupts but I co-founded a new company that is willing to maintain this repository but to avoid the current problem in the futere we are setting the project in an independent organization. 

Fill free to create a new PR (if you make it fast), since I am going to merge the PRs and then change the import path and bump a new version based on go modules.

dsymonds:
Okay, check out #7 instead of this.

PR #5, opened 2020-03-09
git Remote: fix  fetching remotes with unrelated incomplete history

mcuadros:
Original created by @NonLogicalDev at https://github.com/src-d/go-git/pull/1152

```
In getHaves, allow the getHavesRef to fail with Object Not found when
iterating over incomplete history, and don't treat it as a fatal error.

Signed-off-by: Oleg Utkin <oleg.utkin@nonlogical.io>
```

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -623,7 +623,11 @@ func getHaves(
 		}
 
 		err = getHavesFromRef(ref, remoteRefs, s, haves)
-		if err != nil {
+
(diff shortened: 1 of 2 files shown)


caarlos0:
I think this is the cause of https://github.com/getantibody/antibody/issues/343

caarlos0:
actually, just tested with this PR, doesn't fix the issue...

PR #11, opened 2020-03-18
plumbing: transport server check for nil Packfile, fixes ref deletes

brandonwestcott:
When a `Remote` receives a push with only delete commands, it sends a `nil` value for `Packfile` ([related code](https://github.com/go-git/go-git/blob/661b1ace8b4d2a387e2ae130bbb034e62cc8ce11/remote.go#L1064-L1079)), which results in a a `panic: runtime error: invalid memory address or nil pointer dereference` in the code below. This simply fixes that by skipping the processing if its `nil`.

It does seem there is already a test case covering deletes inside the `ReceivePackSuite`, however the [test sends an empty packfile](https://github.com/go-git/go-git/blob/661b1ace8b4d2a387e2ae130bbb034e62cc8ce11/plumbing/transport/test/receive_pack.go#L262) rather than sending `nil`.

If it would be better to update `Remote` to send a similar empty packfile please let me know and I'll be happy to make the update.

Also thanks for all the work here, we've been using go-git heavily for the last few weeks and its fantastic.

### DIFF
--- a/plumbing/transport/server/receive_pack_test.go
+++ b/plumbing/transport/server/receive_pack_test.go
@@ -1,8 +1,13 @@
 package server_test
 
 import (
+	"context"
+
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/protocol/packp"
 	"github.com/go-git/go-git/v5/plumbing/transport"
 
+	fixtures "github.com/go-git/go-git-fixtures/v4"
 	. "gopkg.in/check.v1"
 )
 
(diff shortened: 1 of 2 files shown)


mcuadros:
can you provide a test for this?

brandonwestcott:
@mcuadros added a test for this directly inside [`transport/server/receive_pack_test.go`](https://github.com/go-git/go-git/pull/11/files#diff-69e61f0ee72e941c8ab64850b37041c4R40-R63). 

Also the force push is because I had branched from v4, so simply rebased in current v5 master for the PR.

brandonwestcott:
@mcuadros thanks!

PR #12, opened 2020-03-20
options: git, fix typo

jeveleth:


mcuadros:
Thanks for noticing it, its a stupid typo, but the problem is that if we merge the PR we will break the API :man_facepalming:

jeveleth:
@mcuadros , Ah. Oh well.

mcuadros:
But doesn't mean that is not going to be fixed ever :/

PR #20, opened 2020-04-06
plumbing: format, config.Merged to allow access to local and global config

cap10morgan:
...for reading and writing global (`~/.gitconfig`) and reading system (`/etc/gitconfig`) configs in addition to local repo config.

This is heavily based on / inspired by @djgilcrease's PR on the original src-d repo: https://github.com/src-d/go-git/pull/1243. However, I ran into an issue using that code where if you ran `SetConfig` it would write out the merged system, global, and local config params all to your local `./.git/config` file, which was hard to fix under that architecture.

So, I tried to take a simpler starting approach with this PR but still allow it to accomplish my 2 goals:
1. Show a merged view of system (`/etc/gitconfig`), global (`~/.gitconfig`), and local (`./.git/config`) configs with the same precedence ordering that git itself uses (local overrides global overrides system).
1. Allow setting / changing options in any of these files (with the exception of setting system options b/c we typically won't have permission to write to that file).

The way I did this was basically to leave the higher-level config stuff (e.g. `.Remotes`, `.Core.IsBare`) alone and add a new `.Merged` param that works similarly to `.Raw` but brings in these other configs. So now you can see the merged view with things like `.Merged.Section("foo").Option("bar")` (which is all read-only) or set options with things like `.Merged.GlobalConfig().AddOption(...)`. `.Raw` now delegates to `.Merged.LocalConfig()`.

I added a config example to demonstrate some of the usage and a couple of tests of marshalling and unmarshalling the different configs.

### DIFF
--- a/plumbing/format/config/merged.go
+++ b/plumbing/format/config/merged.go
@@ -1,0 +1,215 @@
+package config
+
+type Scope int
+
+const (
+	LocalScope Scope = iota
+	GlobalScope
+	SystemScope
+)
+
+type ScopedConfigs map[Scope]*Config
+
+type Merged struct {
+	scopedConfigs ScopedConfigs
+}
+
+type MergedSection struct {
+	backingSection *Section
+}
+
+type MergedSubsection struct {
+	backingSubsection *Subsection
+}
+
+type MergedSubsections []*MergedSubsection
+
+func NewMerged() *Merged {
+	cfg := &Merged{
+		scopedConfigs: make(ScopedConfigs),
+	}
+	for s := LocalScope; s <= SystemScope; s++ {
+		cfg.scopedConfigs[s] = New()
+	}
+
+	return cfg
+}
+
+func (m *Merged) ResetScopedConfig(scope Scope) {
+	m.scopedConfigs[scope] = New()
+}
+
+func (m *Merged) ScopedConfig(scope Scope) *Config {
+	return m.scopedConfigs[scope]
+}
+
+func (m *Merged) LocalConfig() *Config {
+	return m.ScopedConfig(LocalScope)
+}
+
+func (m *Merged) GlobalConfig() *Config {
+	return m.ScopedConfig(GlobalScope)
+}
+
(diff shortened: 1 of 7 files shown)


mcuadros:
Can you fix the "Test" suite? Something is missing on the PR

djgilcrease:
Thank you for picking this up @cap10morgan

cap10morgan:
> Thank you for picking this up @cap10morgan

Thank you for the shoulders to stand on!

cap10morgan:
> Can you fix the "Test" suite? Something is missing on the PR

Yep, sorry about that.

mcuadros:
Ok, the code for the merge thing looks complex but fair. My only concern is about the example and promulgating the usage of the `Config.Merged`, this should be an advanced pattern (last resort) to access data not defined on the `Config` struct. So I suggest included a big note in the example explaining that is an advanced use case. 

Also, if we are planning to use the author and committer info from the config, we should add the fields to it, as we have `Core.IsBare` and all other values. This can be done in this PR or in the following PRs.

plumbing/format/config/merged.go
```diff
@@ -0,0 +1,223 @@
+package config
+
+type Scope int
+
+const (
```
mcuadros, line 5: Documentation, in general, will be good, at least in the structs and consts.

cap10morgan:
> Ok, the code for the merge thing looks complex but fair. My only concern is about the example and promulgating the usage of the `Config.Merged`, this should be an advanced pattern (last resort) to access data not defined on the `Config` struct. So I suggest included a big note in the example explaining that is an advanced use case.

Done in 4d4e2b1.

> Also, if we are planning to use the author and committer info from the config, we should add the fields to it, as we have `Core.IsBare` and all other values. This can be done in this PR or in the following PRs.

OK. I'm working on some code to do that, but will post a separate PR for it.

b4nst:
Thanks for the PR ! Any idea of when the next release with this feature will be published ?

PR #39, opened 2020-04-23
storage/filesystem: dotgit, sanity check provided reference path 

kevans91:
Described in finer detail in the commit messages, readReferenceFile() where the path resolves to a directory succeeds on FreeBSD. The root of the problem is that on the other platforms go-git operates on, read() on a dir fd will fail and thus readReferenceFile() will fail.

This small series amends an existing test to demonstrate the problem (the assert at dotgit_test.go:121 will fail the test on FreeBSD), then fixes it by sanity-checking with Stat() that we're not looking at a directory. The second commit fixes up the test to ensure that readReferenceFile() is returning the specific error we're adding, rather than just any error.

This was noticed when Gitea on FreeBSD would 500 while attempting to read a branch with a `/` in any part of the name, as it would split and readReferenceFile() at each component until it got success; which always happened immediately here, despite the first component being a directory.

### DIFF
--- a/storage/filesystem/dotgit/dotgit.go
+++ b/storage/filesystem/dotgit/dotgit.go
@@ -59,6 +59,9 @@ var (
 	// targeting a non-existing object. This usually means the repository
 	// is corrupt.
 	ErrSymRefTargetNotFound = errors.New("symbolic reference target not found")
+	// ErrIsDir is returned when a reference file is attempting to be read,
+	// but the path specified is a directory.
+	ErrIsDir = errors.New("reference path is a directory")
 )
 
 // Options holds configuration for the storage.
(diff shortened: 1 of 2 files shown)


storage/filesystem/dotgit/dotgit.go
```diff
@@ -946,6 +949,14 @@ func (d *DotGit) addRefFromHEAD(refs *[]*plumbing.Reference) error {
 
 func (d *DotGit) readReferenceFile(path, name string) (ref *plumbing.Reference, err error) {
 	path = d.fs.Join(path, d.fs.Join(strings.Split(name, "/")...))
+	st, err := d.fs.Stat(path)
```
cemeyer, line 952: Might be better to open first and then `fstat()`, to avoid going through path lookup twice?  Also prevents TOCTOU type issues.  Maybe not a big deal.

kevans91: My initial patch was to attempt an `f.Stat()` after open, but it turns out the `File` involved here isn't an `os.File` but part of this osfs astraction layer from $otherModule, and it doesn't offer an equivalent that will effectively `fstat(File)` in a platform-independent way. =-(

PR #42, opened 2020-04-25
plumbing: object, make renames diff default

mcuadros:
This makes the renames detect general available in the `Commit` and `Tree` objects. Making the old behavior optimal. This is a change of behavior than maybe consider breaking changes, but it's an improvement over the previous diff and aligns it with the expected behavior of the function. If still some need access to the old logic can be used.

### DIFF
--- a/plumbing/object/commit.go
+++ b/plumbing/object/commit.go
@@ -78,6 +78,9 @@ func (c *Commit) Tree() (*Tree, error) {
 
 // PatchContext returns the Patch between the actual commit and the provided one.
 // Error will be return if context expires. Provided context must be non-nil.
+//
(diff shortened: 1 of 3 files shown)


erizocosmico:
I don't think we should make it default if it breaks compatibility, even if the feature is useful. People who still want it can just use `DiffContextWithOptions` instead, it's not much of a burden.

mcuadros:
This is not a very strong breaking change, and to be honest, is hard that no one expects no having the renames detected. Also, the function `DiffContextWithOptions` is deep on the project and not exposed to the plumbing make it hard to use and to the newcomers, this will propagate the problem of not detecting renames more and more. 

Forcing the people to update to v6 just for this sound worst.

PR #47, opened 2020-04-29
plumbing: diff, refresh internals

twpayne:
This is a work-in-progress PR on top of #46 to update the `plumbing/diff` package. It is not ready to be considered for merge yet.

Goals:

* There should be no changes to the API, this is an internal clean-up.
* Port the code to use `strings.Builder` more and `fmt.Fprintf`/`bytes.Buffer` less, which should improve both readability and performance. Note that `strings.Builder` did not exist when the code was originally written.
* Fix some minor issues around colored diffs (notably that color resets should occur before newlines if possible, not after).

### DIFF
--- a/plumbing/format/diff/unified_encoder.go
+++ b/plumbing/format/diff/unified_encoder.go
@@ -1,70 +1,53 @@
 package diff
 
 import (
-	"bytes"
 	"fmt"
 	"io"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/go-git/go-git/v5/plumbing"
 )
 
-const (
-	diffInit = "diff --git a/%s b/%s\n"
+// DefaultContextLines is the default number of context lines.
+const DefaultContextLines = 3
 
-	chunkStart  = "@@ -"
-	chunkMiddle = " +"
-	chunkEnd    = " @@"
-	chunkCount  = "%d,%d"
+var (
+	splitLinesRegexp = regexp.MustCompile(`[^\n]*(\n|$)`)
 
-	noFilePath = "/dev/null"
-	aDir       = "a/"
-	bDir       = "b/"
+	operationChar = map[Operation]byte{
+		Add:    '+',
+		Delete: '-',
+		Equal:  ' ',
+	}
 
-	fPath  = "--- %s\n"
-	tPath  = "+++ %s\n"
-	binary = "Binary files %s and %s differ\n"
-
-	addLine    = "%s+%s%s%s"
-	deleteLine = "%s-%s%s%s"
-	equalLine  = "%s %s%s%s"
-	noNewLine  = "\n\\ No newline at end of file\n"
-
-	oldMode         = "old mode %o\n"
-	newMode         = "new mode %o\n"
-	deletedFileMode = "deleted file mode %o\n"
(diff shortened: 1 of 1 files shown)


twpayne:
The original `UnifiedEncoder` code accumulated diffs in `bytes.Buffer` which was never `Reset()` so the memory associated with it was not garbage collected until the `UnifiedEncoder` itself was garbage collected. Consequently long term use of a `UnifiedEncoder` had a memory leak.  This PR removes that.

mcuadros:
This was much needed it, thanks a lot @twpayne

twpayne:
This is now ready for review. Notes:

* There are no API changes.
* There are no changes to the output.
* The existing tests (approximately 97% coverage) all pass and are deliberately not changed by this PR. In fact the coverage statistic is slightly improved as there is a slight reduction in the number of lines of code.
* The only feature change should be an elimination of the memory leak described above.
* The PR moves the format strings passed to `Fprintf` from global constants to literals at the callsites themselves. In my opinion, this makes the code easier to read at the expense of some duplication. The diff format changes rarely, so I think this compromise is acceptable.
* There are a number of comment additions and other stylistic changes (e.g. received names being changed to better match the type).

For sure, there are more cleanups that can be done, but I think this is a reasonable improvement for now.

plumbing/format/diff/unified_encoder.go
```diff
@@ -73,102 +57,78 @@ func (e *UnifiedEncoder) SetColor(colorConfig ColorConfig) *UnifiedEncoder {
 	return e
 }
 
+// Encode encodes patch.
 func (e *UnifiedEncoder) Encode(patch Patch) error {
-	e.printMessage(patch.Message())
+	buf := &bytes.Buffer{}
```
mcuadros, line 62: why it's not `strings.Builder` at the end?

twpayne: Good question!

The honest answer is that the code uses `fmt.Fprintf` a lot and I didn't realize that this could write to `strings.Builder` so I stuck with `bytes.Buffer`.

The post-hoc justification is that eventually we call `buf.WriteTo` to write the contents to the `io.Writer` associated with the `UnifiedEncoder`. `strings.Builder` has no such method, so we'd have to add a `string` to `[]byte` conversion.

I did also consider writing to the `io.Writer` associated with the `UnifiedEncoder` directly, but rejected it because:
* We use a lot of `WriteByte` and `WriteString` calls, which `io.Writer` does not implement.
* The accumulate-everything-in-a-`bytes.Buffer` before writing it has nice behavior when errors occur. Specifically, you don't risk writing a partial diff (unless `buf.WriteTo` fails, of course).

mcuadros: `strings.Builder` was designed ad-hoc for this use case. And it's way more gentle with the memory than `bytes.Buffer`. In `strings.Builder` you have WriteByte, WriteRune, and WrityString, so replace it shouldn't be a problem. 

I don't see where are the WriteTo calls, just the method ` writeTo(buf *bytes.Buffer, color ColorConfig)` we we can change the args to the builder.

twpayne: Thanks for the info, I've updated the PR to use `strings.Builder`.

Issue #56, opened 2020-05-06
Fetching commits directly via "git fetch ... SHA:..."


tianon:
See https://github.com/git/git/commit/f8edeaa05d8623a9f6dad408237496c51101aad8 for an example of the functionality I'm talking about -- a prominent example of usage is in https://github.com/actions/checkout (v2) for being able to fetch the relevant commit for a workflow even if the branch has been deleted (GitHub didn't support doing this via `git fetch` previously, but I think that GitHub Actions has prompted a change, especially since "dangling" commits have always been accessible from the web UI if you had or could craft an appropriate URL to them).

I've tried a couple permutations of `RefSpec` syntax to see if it was already supported (and read through docs), but couldn't get anything working so I *think* it's not supported? (but happy to be proved wrong! :sweat_smile: :heart:)

For now I've gone with the workaround of shelling out to `git fetch` for the place that I need this. :+1:

Thanks for your work on this awesome Git implementation! :heart: :+1:

tianon:
FWIW, I found the request this is a "duplicate" of (but it's on the old archived repo): https://github.com/src-d/go-git/issues/628

There's some useful context there too (including confirmation that this was not supported in 2017, and probably still not in 2018, so very likely still not today). :+1:

mcuadros:
You were lucky and I had time to implement it. :champagne:

PR #58, opened 2020-05-10
Remote.Fetch: support exact SHA1 refspecs

mcuadros:
This implements the support of refspec with exact SHA1 in `Remote.Fetch`.

This may be only used on sever with the `AllowReachableSHA1InWant` or `AllowTipSHA1InWant`, otherwise `Remote.Fetch` returns `ErrExactSHA1NotSupported`. (GitHub supports it)

Closes https://github.com/go-git/go-git/issues/56

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -29,6 +29,7 @@ var (
 	NoErrAlreadyUpToDate     = errors.New("already up-to-date")
 	ErrDeleteRefNotSupported = errors.New("server does not support delete-refs")
 	ErrForceNeeded           = errors.New("some refs were not updated")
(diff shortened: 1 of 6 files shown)


remote.go
```diff
@@ -850,6 +861,26 @@ func (r *Remote) newUploadPackRequest(o *FetchOptions,
 	return req, nil
 }
 
+func (r *Remote) isSupportedRefSpec(refs []config.RefSpec, ar *packp.AdvRefs) error {
+	var containtsIsExact bool
```
erizocosmico, line 865: s/containtsIsExact/containsIsExact/

PR #63, opened 2020-05-14
_examples: submodule, adding an example with submodule update

pyaillet:
Hi !

Thanks for this project 👍 

The goal of this pull request is to add an example of how to make
the equivalent of `git submodule update --remote` with `go-git`


mcuadros:
Can you and it to the examples index? Thanks!

Issue #76, opened 2020-05-25
Does go-git support "insteadOf" replacements for repo urls (in gitconfig)


kostyay:
I can't seem to find any reference to this in the code.
Does go-git know how to handle configurations such as `git config --global url.ssh://git@github.com/.insteadOf https://github.com/`?

Thanks in advance.

mcuadros:
It's not supported. But it could be an easy one.

kostyay:
Let me know what you think
https://github.com/go-git/go-git/pull/79

gcstang:
@kostyay is there an example on how to use this with git.PlainClone ?

PR #79, opened 2020-05-29
config: support insteadOf for remotes' URLs

kostyay:
Added support for insteadOf configuration for remote's urls. 
Can also be applied to submodules, but I didn't implement that.

This fixes: https://github.com/go-git/go-git/issues/76

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,8 +1,0 @@
-# Default ignored files
-/shelf/
-/workspace.xml
-# Datasource local storage ignored files
-/dataSources/
-/dataSources.local.xml
-# Editor-based HTTP Client requests
-/httpRequests/
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
+# Editor-based HTTP Client requests
+/httpRequests/
--- a/.idea/go-git.iml
+++ b/.idea/go-git.iml
@@ -1,9 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<module type="WEB_MODULE" version="4">
-  <component name="Go" enabled="true" />
-  <component name="NewModuleRootManager">
-    <content url="file://$MODULE_DIR$" />
-    <orderEntry type="inheritedJdk" />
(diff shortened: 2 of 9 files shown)


mcuadros:
Great! I have a big concern if you check how behaves git, with instead, it is straight forward. The `insteadOf` just happens in the clone operation, so once that the clone happens, the original URL is gone. So you don't need to keep track of it in the remote or something similar.

So I believe that we need to check the global/system configuration in the `git.Clone` and apply the URL, and the left the remote as it is right now.

Additionally, we should at an option in `CloneOptions` to allow enable or disable this behavior. (By default disable)

kostyay:
ok, i will update the pr

aca:
@kostyay 
This will be definitely useful. Any updates on this?

kostyay:
I didn't have time to rewrite it yet.
I will see if I can get to it this weekend.

kostyay:
> Great! I have a big concern if you check how behaves git, with instead, it is straight forward. The `insteadOf` just happens in the clone operation, so once that the clone happens, the original URL is gone. So you don't need to keep track of it in the remote or something similar.
> 
> So I believe that we need to check the global/system configuration in the `git.Clone` and apply the URL, and the left the remote as it is right now.
> 
> Additionally, we should at an option in `CloneOptions` to allow enable or disable this behavior. (By default disable)

From what I see, ~/.gitconfig is not loaded when `Clone` function is called. So that will have to be fixed as well, because in normal use case AFAIK, the `insteadOf` is configured on the global scope before a repository is cloned.
Any tips how to approach that?
I saw the readme states `Reading and modifying per-repository configuration (.git/config) is supported. Global configuration ($HOME/.gitconfig) is not.`

kostyay:
@mcuadros I've updated the implementation following your comments.
Let me know what you think.

kamaln7:
Hey, sorry that I'm being this person, but based on the discussion I feel like this slipped through the cracks. Any chance it could be reviewed again? Thank you!

kostyay:
> Hey, sorry that I'm being this person, but based on the discussion I feel like this slipped through the cracks. Any chance it could be reviewed again? Thank you!

The repo maintainer never bothered reviewing it

mcuadros:
LGTM can you fix the conflicts @kostyay ?

kostyay:
Fixed

kostyay:
Not sure why tests are failing.. flaky?

mcuadros:
Yup is a flaky test.

PR #80, opened 2020-05-31
examples: tag, Add create and push new tag

lcostea:
Trying to reduce the cost of adopting the library for jobs happening on ci/cd systems. This is the one of the things we are doing before releasing into prod, set a common git tag for all our repos. It took me a couple of hours to make this example working, so hopefully it will save some time for others.

### DIFF
--- a/_examples/tag-create-push/main.go
+++ b/_examples/tag-create-push/main.go
@@ -1,0 +1,168 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"time"
+
+	"github.com/go-git/go-git/v5"
+	. "github.com/go-git/go-git/v5/_examples"
+	"github.com/go-git/go-git/v5/config"
+	"github.com/go-git/go-git/v5/plumbing/object"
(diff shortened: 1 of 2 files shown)


lcostea:
I added the logs with the git commands. This helped me track down one problem: I was calling existsTag twice. Thanks

_examples/tag-create-push/main.go
```diff
@@ -0,0 +1,161 @@
...
+			return nil
+		}
+		log.Printf("push to remote origin error: %s", err)
+		return err
+	}
+
+	return nil
+}
+
+func defaultSignature(name, email string) *object.Signature {
```
mcuadros, line 155: Since the last version, the signature is not needed since is loaded from the global config

lcostea: It is not working for me, I get the "tagger field is required" error and I do have user.email and user.name set in /Users/current-user/.gitconfig  
Am I missing something?

mcuadros: #84 fix the issue

PR #86, opened 2020-06-07
feat: add file with using .gitignore, fixed src-d/go-git#1219

jk2K:


jk2K:
same to `git add -A`

mcuadros:
Thanks for the contribution, since `git add` has a lot of options, can you implement the Options pattern? Adding instead of `AddAll` a new `AddWithOptions(&AddOptions{All:true})`, in this case, if we add in the future support for more options we don't need to create more variations of `Add` function.

jk2K:
ok, I try to change the code

PR #88, opened 2020-06-08
storage: filesystem, support .git/commondir repository layout

distorhead:
Adopted already existing older patch: https://github.com/src-d/go-git/pull/1098 — but with changes:
 - Has submodules support:
   - use WORKTREE_GIT_DIR/modules instead of COMMON_GIT_DIR/modules (!);
   - use COMMON_GIT_DIR/objects;
   - use COMMON_GIT_DIR/config.
 - More straithforward naming: explicitly use commonfs or fs in DotGit.

Possibly there can be more cases to catch. But our testing shows it works ok.

We use go-git library in [werf gitops tool](https://github.com/werf/werf) and your library is awesome. Hope it will receive future support!

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -20,6 +20,8 @@ github.com/go-git/go-billy/v5 v5.0.0 h1:7NQHvd9FVid8VL4qVUMm8XifBK+2xCoZ2lSk0agR
 github.com/go-git/go-billy/v5 v5.0.0/go.mod h1:pmpqyWchKfYfrkb/UVH4otLvyi/5gJlGI4Hb3ZqZ3W0=
 github.com/go-git/go-git-fixtures/v4 v4.0.1 h1:q+IFMfLx200Q3scvt2hN79JsEzy4AmBTp/pqnefH+Bc=
 github.com/go-git/go-git-fixtures/v4 v4.0.1/go.mod h1:m+ICp2rF3jDhFgEZ/8yziagdT1C+ZpZcrJjappBCDSw=
+github.com/go-git/go-git-fixtures/v4 v4.0.2-0.20200613231340-f56387b50c12 h1:PbKy9zOy4aAKrJ5pibIRpVO2BXnK1Tlcg+caKI7Ox5M=
+github.com/go-git/go-git-fixtures/v4 v4.0.2-0.20200613231340-f56387b50c12/go.mod h1:m+ICp2rF3jDhFgEZ/8yziagdT1C+ZpZcrJjappBCDSw=
 github.com/google/go-cmp v0.3.0 h1:crn/baboCvb5fXaQ0IJ1SGTsTVrWpDsCWC8EGETZijY=
 github.com/google/go-cmp v0.3.0/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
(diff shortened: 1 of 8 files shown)


distorhead:
Finally, after testing, I decided to rewrite this PR and came up with completely different and clean approach.

In this PR a new dotgit.RepositoryFilesystem struct is defined, which is billy.Filesystem interface compatible object-wrapper, that can handle commondir and dispatch all operations to the correct file path.

`git.PlainOpen` remain unchanged, but `git.PlainOpenWithOptions` has a new option: `PlainOpenOptions.EnableDotGitCommonDir=true|false` (which is false by default). When `EnableDotGitCommonDir=true` repository-open procedure will read `.git/commondir` (if it exists) and then create dotgit.RepositoryFilesystem object initialized with 2 filesystems. This object then passed into storage and then into dotgit.DotGit as `billy.Filesystem` interface. This object will catch all filesystem operations and dispatch to the correct repository-filesystem (dot-git or common-dot-git) according to the rules described in the doc: https://git-scm.com/docs/gitrepository-layout#Documentation/gitrepository-layout.txt.

Also worktree_test.go has been adopted from an older, already existing existing PR: https://github.com/src-d/go-git/pull/1098.

This PR needs new fixtures added in the following PR: https://github.com/go-git/go-git-fixtures/pull/1.

distorhead:
All tests are passing now, and our project (werf) also works ok with this PR, so I consider this ready for the review.

This PR with fixtures can be accepted only after https://github.com/go-git/go-git-fixtures/pull/1, because it uses new fixture data.

Then https://github.com/go-git/go-git/pull/88/commits/45f2d6a6e1100b0e5f8ace0cb73016bbb2495f3c commit needs to be changed to use actual go-git-fixtures revision.

go.sum
```diff
@@ -8,6 +8,8 @@ github.com/creack/pty v1.1.9/go.mod h1:oKZEueFk5CKHvIhNR5MUki03XCEU+Q6VDXinZuGJ3
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/distorhead/go-git-fixtures/v4 v4.0.2-0.20200610084137-b7883837af3d h1:zrxi8RVOcqobSh23wOli79q3envi/5Rkd74PoSpw0oA=
```
mcuadros, line 11: you can already use the original repository

options.go
```diff
@@ -577,6 +577,8 @@ type PlainOpenOptions struct {
 	// DetectDotGit defines whether parent directories should be
 	// walked until a .git directory or file is found.
 	DetectDotGit bool
+	// Enable .git/commondir support (see https://git-scm.com/docs/gitrepository-layout#Documentation/gitrepository-layout.txt)
+	EnableDotGitCommonDir bool
```
mcuadros, line 581: Can you document that only works with the `filesystem` storage?

go.sum
distorhead: Done

options.go
distorhead: Done

PR #93, opened 2020-06-12
add r.LogRange() similar to: git log ref1..ref2

### LINKED ISSUE #69: how to git log between two branches
Is there a way to do the equivalent of git log master...my/branch?
Very new to this package, but really only seeing timestamps as a way to limit log results. 

My goal is to iterate through all the commit messages between my current branch and master

ilius:
This fixes #69

### DIFF
--- a/options.go
+++ b/options.go
@@ -369,6 +369,13 @@ type LogOptions struct {
 	Until *time.Time
 }
 
+type LogRangeOptions struct {
+	LogOptions
+
+	// End commit hash, exclusive (will not be included in iterator)
+	End plumbing.Hash
+}
+
 var (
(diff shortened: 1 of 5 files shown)


ilius:
I think I made a mistake in the algorithm. Closing this for now.

marcusholl:
@ilius : I have a use-case where I need the equivalent to `git log ref1..ref2`. Is there any other option with the current state of the coding in go-git how this can be done?

ilius:
I'm not sure.
Someone suggested something on
https://github.com/go-git/go-git/issues/69

marcusholl:
@ilius : thanks for the fast response ...

PR #95, opened 2020-06-14
Implement filters

### LINKED ISSUE #89: Implement smudge and clean filters
I'd like to get LFS working with go-git and the first step to that is smudge and clean filters.

I'm not that familiar with the codebase or how Git invokes clean/smudge filters but with some direction I'd be up for implementing it.

Southclaws:
Resolves #89 

Work in progress.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -98,6 +98,8 @@ type Config struct {
 	// Branches list of branches, the key is the branch name and should
 	// equal Branch.Name
 	Branches map[string]*Branch
+	// Filters list of filters for smudge/clean. Key is the pattern to match.
+	Filters map[string]*Filter
 	// Raw contains the raw information of a config file. The main goal is
 	// preserve the parsed information from the original format, to avoid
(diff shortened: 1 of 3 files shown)


alkar:
Thanks for working on this!
Is there anything I can help with to push this forward?

Southclaws:
This is definitely something I still need to do (need it for a project) but I've not touched it for a while.

I've only messed around with it really, not really written anything proper. I think the main issue is understanding exactly at what points in the codebase care about invoking a filter. Unfortunately I didn't write any notes but I think I found two (commit and diff)

I will get back to this at some point and may have some more useful work to show!

alkar:
I _think_ there's three points: checkout and staging (commit) defined by `filter=` and diff as you mentioned.
Reference: https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes

Southclaws:
Ah, thanks, I forgot about staging!

PR #109, opened 2020-06-22
Worktre: Pull, report "Already up to date" when local repository ahead of remote

mruediger:
If you run 'git pull', do a commit and run 'git pull' again git will
report "Already up to date" whereas go-git would report a reports
non-fast-forward update. This commit changes the behavior of go-git to
match that of git.


mruediger:
Can someone help me with the two failing tests? I don't know what the issue with macos-latest on 1.13.x and also have no idea how to provide test coverage for a failing `isFastForward()`

PR #120, opened 2020-07-05
plumbing: publish Tree.Entry(), TreeEntryIter, minor TreeWalker optimization

riking:
A TreeWalker in non-recursive mode is a backwards-compatible equivalent to TreeEntryIter, but is problematic for VFS implementations due to the eager fetching of subtree objects. Disable this eager fetching in non-recursive mode.


riking:
Also exposed Tree.Entry as that's needed for the FUSE Lookup() call efficiency.

mcuadros:
Can you provide some test?

riking:
Oops, just realized that .Entries is exposed.

PR #121, opened 2020-07-05
plumbing: memoryobject, make blob reader seekable

riking:
Replace the bytes.Buffer with a bytes.Reader wrapped in a custom NopCloser, so that the extra reading methods are sill accessible.


mcuadros:
Can you provide some tests?

riking:
Done

riking:
Test failures look fake

PR #125, opened 2020-07-06
*: minor linter fixes

cristaloleg:


worktree_bsd.go
```diff
@@ -12,7 +12,7 @@ import (
 func init() {
 	fillSystemInfo = func(e *index.Entry, sys interface{}) {
 		if os, ok := sys.(*syscall.Stat_t); ok {
-			e.CreatedAt = time.Unix(int64(os.Atimespec.Sec), int64(os.Atimespec.Nsec))
```
mcuadros, line 15: this is need it in bsd.

cristaloleg: oh, that's tricky :)

PR #136, opened 2020-07-15
Repository.ResolveRevision, support partial hashes

### LINKED ISSUE #135: Resolve short hashes
Related to https://github.com/src-d/go-git/issues/599.

Repository.ResolveRevision works right now with a bunch of formats, including branch/tag names and full (40 char) hashes. Unlike `git rev-parse`, though, it doesn't support partial hashes.

dsymonds:
Like `git rev-parse <prefix>`, this enumerates the hashes of objects
with the given prefix and adds them to the list of candidates for
resolution.

This has an exhaustive slow path, which requires enumerating all objects
and filtering each one, but also a couple of fast paths for common
cases. There's room for future work to make this faster; TODOs have been
left for that.

Fixes #135.

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -3,6 +3,7 @@ package git
 import (
 	"bytes"
 	"context"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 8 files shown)


repository.go
```diff
@@ -1445,10 +1446,34 @@ func (r *Repository) ResolveRevision(rev plumbing.Revision) (*plumbing.Hash, err
 
 			var tryHashes []plumbing.Hash
 
-			maybeHash := plumbing.NewHash(string(revisionRef))
-
-			if !maybeHash.IsZero() {
-				tryHashes = append(tryHashes, maybeHash)
+			// Handle complete and partial hashes.
```
mcuadros, line 1449: can we move this to a private function?

storage/filesystem/dotgit/dotgit.go
```diff
@@ -427,12 +476,16 @@ func (d *DotGit) genObjectList() error {
 	}
 
 	d.objectMap = make(map[plumbing.Hash]struct{})
-	return d.forEachObjectHash(func(h plumbing.Hash) error {
+	if err := d.forEachObjectHash(func(h plumbing.Hash) error {
```
mcuadros, line 479: can we get the func definition outside of the if statement?

storage/filesystem/object.go
```diff
@@ -518,6 +519,36 @@ func (s *ObjectStorage) findObjectInPackfile(h plumbing.Hash) (plumbing.Hash, pl
 	return plumbing.ZeroHash, plumbing.ZeroHash, -1
 }
 
+func (s *ObjectStorage) HashesWithPrefix(prefix []byte) ([]plumbing.Hash, error) {
```
mcuadros, line 522: This is just tested indirectly? Will be great add a test.

mcuadros:
Thanks for your contribution! I added a minor comments

dsymonds:
PTAL

PR #139, opened 2020-07-21
Worktree.Commit: speed up commit with all: true option

### LINKED ISSUE #113: worktree.Add(".") does not work as expected and CommitOptions{All: true} is too slow
Hello.

Thank you for your package.
I started to use it recently with version `v5.1.0` and found some strange things.
First of all `worktree.Add(".")` doesn't work same way as `git add .`. In my case it doesn't delete files from index.

Also if I try to use `worktree.Commit("message", &CommitOptions{All: true})` it takes a lot of time although my repo is relatively small.

So if I rename files and then run `worktree.Add(".")` and `worktree.Commit("message", &CommitOptions{})` it works relatively fast but there are old and new files persist in my repo.
With `worktree.Commit("message", &CommitOptions{All: true})` it works as expected but takes about 70 seconds on my repo with ~1500 files (~15 Mb overall), and about 30 seconds if I deleting biggest project's folder: ~650 files (~10 Mb)

Any suggestions here?

StupidScience:
Hello.

Partially fixes #113

Speeding up `worktree.Commit("message", &CommitOptions{All: true})` due to invoking `worktree.doAddFile(..)` directly and avoid invoking `worktree.Status()` every time.

In my case it speeds up commit from 70 sec to 2-5 seconds

### DIFF
--- a/worktree_commit.go
+++ b/worktree_commit.go
@@ -58,17 +58,23 @@ func (w *Worktree) autoAddModifiedAndDeleted() error {
 		return err
 	}
 
+	idx, err := w.r.Storer.Index()
+	if err != nil {
+		return err
+	}
+
 	for path, fs := range s {
 		if fs.Worktree != Modified && fs.Worktree != Deleted {
 			continue
 		}
 
-		if _, err := w.Add(path); err != nil {
+		if _, _, err := w.doAddFile(idx, s, path, nil); err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


StupidScience:
@mcuadros ca you please take a look here?

StupidScience:
Failed test doesn't seem to be connected to changes.

PR #146, opened 2020-08-02
chore: add an exmaple: clone with ssh

fatelei:
RT

### DIFF
--- a/_examples/clone/auth/ssh/main.go
+++ b/_examples/clone/auth/ssh/main.go
@@ -1,0 +1,52 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	git "github.com/go-git/go-git/v5"
+	. "github.com/go-git/go-git/v5/_examples"
+	"github.com/go-git/go-git/v5/plumbing/transport/ssh"
+)
+
+func main() {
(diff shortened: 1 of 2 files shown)


fatelei:
@mcuadros  Add an example for git clone with ssh. Failed test because net dial.

_examples/clone/auth/ssh/main.go
```diff
@@ -0,0 +1,52 @@
...
+		password = os.Args[4]
+	}
+
+	_, err := os.Stat(privateKeyFile)
+	if err != nil {
+		Warning("read file %s failed %s\n", privateKeyFile, err.Error())
+		return
+	}
+
+	// Clone the given repository to the given directory
```
mcuadros, line 26: this should be the equivalent git command, not a comment

_examples/clone/auth/ssh/main.go
```diff
@@ -0,0 +1,52 @@
...
+	Info("git clone %s %s using ssh", url, directory)
```
mcuadros, line 27: this should be the equivalent git command, not a comment

fatelei: ok

fatelei: resolved

PR #153, opened 2020-08-24
Add test-case for file not matching a folder glob

stippi2:
The test-case currently fails. I will link to this PR in the issue I plan to open.


pjbgf:
@stippi2 happy to review this PR in case you are still keen on fixing the issue you have reported.

PR #165, opened 2020-09-07
advrefs.addRefs: it's okay if HEAD can't be resolved.

apenwarr:
The intention of this function is to add all references that exist, not
complain about references that don't exist. If HEAD can't be resolved
because it points at an invalid reference, just move on.

In original git, this situation prints a warning message but doesn't fail.

### DIFF
--- a/plumbing/protocol/packp/advrefs.go
+++ b/plumbing/protocol/packp/advrefs.go
@@ -88,7 +88,13 @@ func (a *AdvRefs) addRefs(s storer.ReferenceStorer) error {
 		return a.addSymbolicRefs(s)
 	}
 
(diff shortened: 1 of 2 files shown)


mcuadros:
At that time the decision made was... It's an error that you can handle so if you are ok, with an unresolved ref, just ignore the error.
The problem now even if we agree to ignore the error is that we can't change the API without a major update.

apenwarr:
The problem was that there are other functions that call addRefs() and abort halfway through if the HEAD ref is inaccessible. That's not desirable either.

mcuadros:
I understand, so the the modification should be done in this functions,
allowing the full execution and return the error at the same time

PR #178, opened 2020-09-30
config: add init.defaultBranch to the config

tomlazar:
Adds a new struct to the Config struct that includes the `init.defaultBranch` variable as mentioned [here](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-config.html#FILES)

>init.defaultBranch
>    Allows overriding the default branch name e.g. when initializing a new repository or when cloning an empty repository.
>

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -89,6 +89,13 @@ type Config struct {
 		Window uint
 	}
 
+	Init struct {
+		// DefaultBranch Allows overriding the default branch name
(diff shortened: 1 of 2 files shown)


mcuadros:
Can you provide some tests? Thanks!

tomlazar:
added a couple more tests for other functions in the `config` package,  but most of the functions in there are pretty well covered, let me know if you want me to something  for one of the other pacakges!

PR #184, opened 2020-10-13
git: worktree, Support relative submodule URL.

mikyk10:
Original issue created by @tyru at https://github.com/src-d/go-git/issues/737

This PR supports relative URLs in `.gitmodules` which refers to the superproject's remote origin.

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -716,7 +716,33 @@ func (w *Worktree) readGitmodulesFile() (*config.Modules, error) {
 	}
 
 	m := config.NewModules()
-	return m, m.Unmarshal(input)
+	if err := m.Unmarshal(input); err != nil {
+		return m, err
+	}
+
+	w.resolveRelativeSubmodulePaths(m)
+
+	return m, nil
+}
+
+
+// resolveRelativeSubmodulePaths is to replace any relative submodule URL (../foobar.git) in .gitmodules
+// to the accessible repository URL
+func (w *Worktree) resolveRelativeSubmodulePaths(m *config.Modules) {
+	origin, err := w.r.Remotes()
+	// remote is not associated with this worktree. we don't need to process any futher more
+	if err != nil {
+		return
+	}
+
+	parentURL := filepath.Dir(origin[0].c.URLs[0])
+
+	for i := range m.Submodules {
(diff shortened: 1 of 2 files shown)


mikyk10:
I'm not sure why only _1.14.x, macos-latest_ failed while others succeeded. Please let me know what you think about this PR.

mcuadros:
This is a feature supported by git?

mikyk10:
@mcuadros 
Yes, the feature is supported by git. According to the [official documentation](https://git-scm.com/docs/git-submodule).

> <repository> is the URL of the new submodule’s origin repository. This may be either an absolute URL, or (if it begins with ./ or ../), the location relative to the superproject’s default remote repository (Please note that to specify a repository foo.git which is located right next to a superproject bar.git, you’ll have to use ../foo.git instead of ./foo.git - as one might expect when following the rules for relative URLs - because the evaluation of relative URLs in Git is identical to that of relative directories).
> 
> The default remote is the remote of the remote-tracking branch of the current branch. If no such remote-tracking branch exists or the HEAD is detached, "origin" is assumed to be the default remote. If the superproject doesn’t have a default remote configured the superproject is its own authoritative upstream and the current working directory is used instead.

cbauernhofer:
This fix actually contains 2 bugs. 

First parentURL := filepath.Dir(origin[0].c.URLs[0]) produces a wrong parentURL in case you use https. The resulting URL will have the format http:/xyz... - so there is one slash missing because filepath.Dir() is removing it. 

The second issue is that multiple relative paths are not handled correctly - so ../../submodule.git will not work as well.

As this PR is already merged new issues should be created. Will this be done by you guys?

dikkini:
> This fix actually contains 2 bugs.
> 
> First parentURL := filepath.Dir(origin[0].c.URLs[0]) produces a wrong parentURL in case you use https. The resulting URL will have the format http:/xyz... - so there is one slash missing because filepath.Dir() is removing it.
> 
> The second issue is that multiple relative paths are not handled correctly - so ../../submodule.git will not work as well.
> 
> As this PR is already merged new issues should be created. Will this be done by you guys?

need fix. any date?

dikkini:
How i can get this fixes in my project?

Issue #157, opened 2020-08-26
SSH cloning returning "empty git-upload-pack given"


bmiller2807:
I am not sure if this is an issue, bug , or lack of understanding, but, when im trying to clone a repo using ssh , i am recieving this error "empty git-upload-pack given". Any ideas why ? A snippet :

                          sshPath := privateKeyFile
			sshKey, _ := ioutil.ReadFile(sshPath)
			publicKey, keyError := ssh.NewPublicKeys(user, []byte(sshKey), password)
			if keyError != nil {
				fmt.Println(keyError)
			}

			repo, err = git.PlainClone(cloneDir, false, &git.CloneOptions{
				URL:               scanRequest.RepoURL,
				Progress:          os.Stdout,
				Auth:              publicKey,
				ReferenceName:     plumbing.ReferenceName(fmt.Sprintf("refs/heads/%s", scanRequest.Branch)),
				SingleBranch:      true,
				RecurseSubmodules: git.DefaultSubmoduleRecursionDepth,
			})
			if err != nil {
				fmt.Sprintf("%s %s %s", "07", "Clone failed: %s\n", err.Error())
				return
			}

ChandlerSwift:
You don't happen to be cloning from Azure Devops do you? There's a known issue (they only implement the git v2 protocol -- https://github.com/go-git/go-git/issues/64) that seems to give the "empty git-upload-pack" error you've been seeing.

bmiller2807:
Thank you @ChandlerSwift for the reply, hopefully this will be added soon.

PR #200, opened 2020-11-01
Advertise multi ack capabilities

### LINKED ISSUE #64: Git Clone does not work with Azure DevOps repos
I'm trying to do a git clone on a git repo hosted on Azure DevOps. This fails with "error: unexpected client error: unexpected requesting "https://PATH-TO-REPO/git-upload-pack" status code: 400"

Seems to be the same error as in https://github.com/src-d/go-git/issues/1058

If you need more information please let me know

phillebaba:
This PR hopes to fix a long standing issue for go-git when communicating with servers that require multi-ack capability. The biggest source of that problem right now is Azure DevOps. Meaning that any project that uses go-git currently can't support it.

This first PR does not fully implement multi-ack but instead advertises the capability and then does the normal transaction. My hope is to get multi ack fully implemented in a later PR, but this is a first step to enable projects to work with Azure DevOps.

I had to make a small change to how ssh connections are closed as Azure DevOps will close the connection after sending all data, instead of waiting for the client to close the connection.

Fixes #64 #157 

The PR depends on @bashims PR #111, so it would be great if we could get that merged before to make sure he gets credit for his work.

### DIFF
--- a/plumbing/format/packfile/parser.go
+++ b/plumbing/format/packfile/parser.go
@@ -287,6 +287,7 @@ func (p *Parser) resolveDeltas() error {
 				if err := p.resolveObject(stdioutil.Discard, child, content); err != nil {
 					return err
 				}
+				p.resolveExternalRef(child)
 			}
 
 			// Remove the delta from the cache.
@@ -297,6 +298,16 @@ func (p *Parser) resolveDeltas() error {
 	}
 
 	return nil
+}
+
+func (p *Parser) resolveExternalRef(o *objectInfo) {
+	if ref, ok := p.oiByHash[o.SHA1]; ok && ref.ExternalRef {
+		p.oiByHash[o.SHA1] = o
+		o.Children = ref.Children
+		for _, c := range o.Children {
+			c.Parent = o
+		}
+	}
 }
 
 func (p *Parser) get(o *objectInfo, buf *bytes.Buffer) (err error) {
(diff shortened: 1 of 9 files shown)


phillebaba:
I dont really understand the failing test. @mcuadros do you have some insight in why it is failing only on macos?

mcuadros:
The test is just a flaky one, sorry for that.

About the PR, I really don't understand how it is going to work; if you anounce the capabilities, nothing is going to solve the problem since it's not implemented. So I rather wait to have the full functionallity before merge it.

phillebaba:
Ah good, as I was a bit confused if it was something I was doing.

So let me start off by saying that full support is the end goal, but I do not know how long it would take for me to get it done.
I have been following the steps suggested by @smola https://github.com/src-d/go-git/issues/335#issuecomment-310791062 to get this first feature done.

My understanding of multi_ack and correct me if I am wrong, is that it is an optimization when the client sends its haves/wants to avoid having the client walk back the full branch. This is done by the server responding with a continue when the client has walked far enough back for the server to have the data. So if the client does not listen to these responses all its doing is doing more processing and sending more data than is actually needed, it is just extra work. As long as we can parse the data format returned by the server we should be fine.
https://git-scm.com/docs/protocol-capabilities#_multi_ack

As the capabilities are determined by the server we will not advertise multi_ack unless the server can support it.

djoyahoy:
I want to chime in here and say that Azure DevOps is very important to our users and even partial support would be useful to users of [kpack](https://github.com/pivotal/kpack).

mcuadros:
The problem is that Azure DevOps, only supports multi_ack, I suppose Azure has developed his own git library and they didn't implement the classic protocol. So if you announce the support of multi_ack, then go-git is going to fail elsewhere. The only way to make it work is by implementing multi_ack support.

Multi_ack requires a refactor of the protocol and some other packages, which is a middium task that will require quite amount of hours and knowledge, who know this better is @smola 

@djoyahoy since version November 2019 this project has lost his unique supporter, so many of the contributors are out of time to work on this, like my case. If big comporations are using this library will be great if they can support it.

phillebaba:
I will close this PR now as we have added support for both libgit2 and go-git in the flux project for those who requires multi_ack capabilities.

PR #205, opened 2020-11-05
Handle Wine Symlink Bug

abergmeier:
Wine currently implements Symlink syscalls as Stubs. Sadly they also
always return TRUE. So for windows validate, that the first successful
Symlink call was indeed successful.
This does introduce a false positive if Symlinks get removed immediately
after creation. The error is very straight forward about that then.

### DIFF
--- a/filesystem_wine.go
+++ b/filesystem_wine.go
@@ -1,0 +1,146 @@
+package git
+
+import (
+	"errors"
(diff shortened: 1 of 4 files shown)


mcuadros:
Can you contribute this to go-billy? Thanks

PR #214, opened 2020-11-21
config:  Added signingkey & gpgsign fields.

devpotamus:
Update to config/config.go to support:
- user.signingkey 
- commit.gpgsign

This adds a new section `Commit` to the config that needs to be handled by the marshal/unmarshalling logic (included in PR).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -66,6 +66,8 @@ type Config struct {
 		Name string
 		// Email is the email of the author and the commiter of a commit.
 		Email string
+		// SigningKey is the GPG key used to sign commits.
+		SigningKey string
 	}
 
 	Author struct {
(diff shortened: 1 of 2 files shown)


devpotamus:
Is there additional output on the following error?

```
Error: FAIL: repository_test.go:2632: RepositorySuite.TestResolveRevision

repository_test.go:2665:
    c.Assert(err, IsNil, Commentf("while checking %s", rev))
... value *errors.errorString = &errors.errorString{s:"reference not found"} ("reference not found")
... while checking 918c48b
```

mcuadros:
This adds the fields, but doesn't sign the commits right?

devpotamus:
Correct, these are only the fields that support signing. The changes allow for these fields to be populated from the git configuration file. 

It looks like there is already support for signing keys through Option types like CommitOptions. Diving deeper there, it looks like this could be incorporated there as well to pull these fields automatically if set. If desirable, I can update this PR with that change or generated a new PR with that change.

Thoughts?

mcuadros:
Ideally, the signature should be configured automatically in the `CommitOptions` in the same way the Author is loaded from the global config.

PR #222, opened 2020-12-08
Adding "object-format" and "filter" capabilities

rofc:
Adding the following capabilities, which seem to be part of the upload-pack in [latest git version](https://git-scm.com/docs/protocol-capabilities/):
- object-list
- filter


mcuadros:
Thanks!

PR #224, opened 2020-12-10
git: worktree_commit, just store objects not already stored

jsteuer:
Hi! 

If I do a simple single file commit using go-git, lots of files within .git/objects/ are touched - they all has the same last modified timestamp. If I do that same simple commit using the regular git CLI, only few files are touched. 

The change of this PR seems to fix this behavior by simply checking if the object is already stored. But maybe I'm missing something?

However, the current gain in performance is impressive: on my test repository (25k files, 1k commits, total size <100MB) the time to commit drops from 3.4s down to 2.2s.

### DIFF
--- a/worktree_commit.go
+++ b/worktree_commit.go
@@ -230,5 +230,9 @@ func (h *buildTreeHelper) copyTreeToStorageRecursive(parent string, t *object.Tr
 		return plumbing.ZeroHash, err
 	}
 
+	hash := o.Hash()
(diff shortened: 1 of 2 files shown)


mcuadros:
Hi,

Can you provide a test checking the files are saved?

jsteuer:
Hi, sure, I added a unit test. It checks, that the objects are saved and by comparing the timestamps, it ensures that there is no write to the same object over and over again.

Issue #140, opened 2020-07-22
Ignore SSL Verification


kgrvamsi:
How can we turn off SSL Verify with CloneOptions?

Here is how we do it with git

```
You can tell Git and Git LFS to ignore SSL cert verification errors too. You can disable it for individual commands:

$ GIT_SSL_NO_VERIFY=1 git clone ...

You can also configure individual repositories to ignore verification errors.

$ git config http.sslverify false
$ git clone ...

```

riyaz-ali:
`go-git` doesn't seem to support `GIT_SSL_NO_VERIFY` or `http.sslverify` .. 

A quick work-around would be to override the default `https` transport and provide an `http.Client` that's configured with `tls.Config{InsecureSkipVerify: true}` .. you can do that with something similar to the following snippet -

```golang
package main

import (
    "net/http"
    "crypto/tls"

    "github.com/go-git/go-git/v5/plumbing/transport/client"
    httptransport "github.com/go-git/go-git/v5/plumbing/transport/http"
)

func init(){
    var transport = httptransport.NewClient(&http.Client{
        transport: &http.Transport{
            TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
        }
    })

    client.InstallProtocol("https", transport)
}
```

But be careful! **disabling security checks is dangerous and should be avoided** .. you can read about alternate ways on [this stack overflow post](https://stackoverflow.com/q/12122159/6611700)

StrongMonkey:
Hey guys, I put a PR https://github.com/go-git/go-git/pull/228 to add support for this. Can anyone take a look at it?

kgrvamsi:
@StrongMonkey i will give a try this week

pjbgf:
Closing this as #228 introduced an easy API for it. Additionally, the example [above](https://github.com/go-git/go-git/issues/140#issuecomment-704057479) should also work.

PR #228, opened 2020-12-18
Add insecureSkipTLS and cabundle

StrongMonkey:
This PR add insecureSkipTLSVerify and cabundle to any remote http calls
so that https repo with private CA signed can be used. This is the
equivalent of https.sslVerify and GIT_SSL_CAINFO


StrongMonkey:
https://github.com/go-git/go-git/issues/140

PR #242, opened 2021-01-14
stage deleted files in doAddDirectory

### LINKED ISSUE #223: worktree.AddWithOptions not staging deleted files
According to docs this should also stage deleted files, but sadly it does not.

```
err := worktree.AddWithOptions(&git.AddOptions{
	All: true,
})
```

```
// AddOptions describes how a add operation should be performed
type AddOptions struct {
	// All equivalent to `git add -A`, update the index not only where the
	// working tree has a file matching `Path` but also where the index already
	// has an entry. This adds, modifies, and removes index entries to match the
	// working tree.  If no `Path` nor `Glob` is given when `All` option is
	// used, all files in the entire working tree are updated.
	All bool
	// Path is the exact filepath to a the file or directory to be added.
	Path string
	// Glob adds all paths, matching pattern, to the index. If pattern matches a
	// directory path, all directory contents are added to the index recursively.
	Glob string
}
```

hensur:
I just ran into #223 

This seems to fix the issue for deleted files.
Directories however, probably won't work correctly.

I'm happy to receive some feedback on this solution.

I might get around to implementing directories as well. Some further input on that would be great as well.

(partly) fixes #223

### DIFF
--- a/worktree_status.go
+++ b/worktree_status.go
@@ -316,8 +316,7 @@ func (w *Worktree) doAddDirectory(idx *index.Index, s Status, directory string, 
 		var foundMatch bool
 		for _, file := range files {
 			if file.IsDir() {
-				foundMatch = true
-				break
+				continue
 			}
 
 			if e.Name == path.Join(directory, file.Name()) {
--- a/worktree_status.go
+++ b/worktree_status.go
@@ -283,10 +283,16 @@ func (w *Worktree) doAddDirectory(idx *index.Index, s Status, directory string, 
 		}
 	}
 
+	var entries []*index.Entry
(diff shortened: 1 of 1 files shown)


gedw99:
thanks @hensur for this fix.

 I am also hit by this. Can this be merged ?

ltdeoliveira:
Hello guys, I'm really in need of this fix, do you guys have any idea of when this will be merged?

mcuadros:
We need cover this with some test, to avoid hit again on this issue

AriehSchneier:
This looks like it was fixed in #652

Issue #51, opened 2020-04-30
CI tests look flaky


twpayne:
`go-git`'s CI often fails with errors like:

```
----------------------------------------------------------------------
##[error]FAIL: <autogenerated>:1: ReceivePackSuite.TestAdvertisedReferencesEmpty

/Users/runner/runners/2.169.1/work/go-git/go-git/plumbing/transport/test/receive_pack.go:35:
    c.Assert(err, IsNil)
... value *net.OpError = &net.OpError{Op:"dial", Net:"tcp", Source:net.Addr(nil), Addr:(*net.TCPAddr)(0xc0000f6c90), Err:(*os.SyscallError)(0xc00015a000)} ("dial tcp [::1]:49628: connect: connection refused")

OOPS: 28 passed, 1 FAILED
--- FAIL: Test (9.96s)
FAIL
coverage: 10.6% of statements in ./...
FAIL	github.com/go-git/go-git/v5/plumbing/transport/git	11.449s
```

Example: https://github.com/go-git/go-git/pull/47/checks?check_run_id=634865225

This looks like a race between a test server starting up and a client trying to connect to it.

Presumably this is already known (although I could not find any existing open issue for it either here or at https://github.com/src-d/go-git).

Questions:

* Is there any existing effort to fix these tests or any background information about them?
* Would you appreciate help fixing them? I can't promise any solutions, but might be able to have a poke around.

mcuadros:
Sorry about the late response, this was falling since a long time, a no one was able to fix it :/

twpayne:
Closing due to age.

PR #246, opened 2021-01-18
plumbing: wire up contexts for Transport.AdvertisedReferences

asuffield:
This is primarily for the sake of the http transport, which otherwise makes its first call without a context, causing context timeouts to not work if the remote is unresponsive.

Fixed up several tests which were getting this wrong. I am particularly fond of the ones which canceled the context and then proceeded to test that the function worked after being canceled.

The AdvertisedReferences function can be removed after this change, but this is technically breaking the API, since library users could call it, so it should really be deferred to v6.

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -109,7 +109,7 @@ func (r *Remote) PushContext(ctx context.Context, o *PushOptions) (err error) {
 
 	defer ioutil.CheckClose(s, &err)
 
-	ar, err := s.AdvertisedReferences()
+	ar, err := s.AdvertisedReferencesContext(ctx)
 	if err != nil {
 		return err
 	}
@@ -316,7 +316,7 @@ func (r *Remote) fetch(ctx context.Context, o *FetchOptions) (sto storer.Referen
 
 	defer ioutil.CheckClose(s, &err)
 
-	ar, err := s.AdvertisedReferences()
+	ar, err := s.AdvertisedReferencesContext(ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -1034,7 +1034,7 @@ func (r *Remote) List(o *ListOptions) (rfs []*plumbing.Reference, err error) {
 
 	defer ioutil.CheckClose(s, &err)
 
-	ar, err := s.AdvertisedReferences()
+	ar, err := s.AdvertisedReferencesContext(context.TODO())
 	if err != nil {
 		return nil, err
 	}
--- a/plumbing/transport/common.go
+++ b/plumbing/transport/common.go
@@ -58,6 +58,11 @@ type Session interface {
 	// If the repository does not exist, returns ErrRepositoryNotFound.
 	// If the repository exists, but is empty, returns ErrEmptyRemoteRepository.
 	AdvertisedReferences() (*packp.AdvRefs, error)
+	// AdvertisedReferencesContext retrieves the advertised references for a
+	// repository.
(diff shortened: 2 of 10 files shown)


asuffield:
Failed checks are a flaky test (network access from the test to the internet) and fractional coverage decrease from the shifting error paths, which I'm inclined to ignore.

sevki:
Perhaps someone should disable them all together

https://github.com/go-git/go-git/issues/51

remote.go
```diff
@@ -1032,7 +1032,7 @@ func (r *Remote) List(o *ListOptions) (rfs []*plumbing.Reference, err error) {
 
 	defer ioutil.CheckClose(s, &err)
 
-	ar, err := s.AdvertisedReferences()
+	ar, err := s.AdvertisedReferencesContext(context.TODO())
```
xiujuan95, line 1035: Hi, I was using `go-git` package to do `git ls-remote` check. But I found this command doesn't have a timeout configuration. I saw you were doing this action. That would be nice! 

But why do you pass `context.TODO()` to it? As far as I know, `context.TODO()` ia an empty context. This means, `git ls-remote` also doesn't have timeout. Could you explain it for me pls? Thanks in advance!

asuffield: FetchContext and PushContext exist, but there's no ListContext method. That's a reasonable thing to add, it's just not part of this PR, which is doing the plumbing.

xiujuan95: Thanks for your reply! Could you please also help add `ListContext` func? Or may I submit a new PR based on your changes? But this requires your PR is merged, not sure how long it will take?

asuffield: I'd put the API change in a separate MR. No idea how long it will take somebody to get around to merging this one.

mcuadros:
Sorry for the late review, I will happy to merge this if you add a test to ensure the context propagation.

remote.go
xiujuan95: Hope this PR can be merged soon!!!! By thw way, about API change PR, do you want to own it? If not, I can do it.

xiujuan95: Hi, could you please add a test as @mcuadros said so that this PR can be merged? Because in our environment, this feature is needed and lack of it iss causing some problems for us. Thanks in advance!

asuffield:
> Sorry for the late review, I will happy to merge this if you add a test to ensure the context propagation.

The existing tests already covered this reasonably well - I had just fixed them to ensure that context propagates, rather than ensuring that it does not. I made a second pass and added a couple more cases, did you have any more specific in mind?

remote.go
xiujuan95: ok, thanks!

xiujuan95:
Any updates about this PR?

mcuadros:
> The existing tests already covered this reasonably well - I had just fixed them to ensure that context propagates, rather than ensuring that it does not. I made a second pass and added a couple more cases, did you have any more specific in mind?

We need a test that fails the old behavior, to validate that we pass properly the context.

asuffield:
> We need a test that fails with the old behavior, to validate that we pass properly the context.

The existing tests do this. For example, TestPlainCloneContextNonExistingOverExistingGitDirectory. Fixing those tests to defer cancel() instead of calling it immediately is functionally inverting the test so that it verifies the context is passed, instead of verifying that the context is not passed.

PR #256, opened 2021-02-16
plumbing: gitignore, Fix gitconfig path in LoadSystemPatterns doc

andrewarchi:
This patch fixes an inconsistent gitconfig location documented in the `plumbing/format/gitignore` package.


andrewarchi:
By the way, the test failure is unrelated to my change and it can be safely merged (#51).

PR #265, opened 2021-03-09
diff: Allow srcPrefix and dstPrefix to be configured

yabberyabber:
The default behavior here remains the same, but this change does
allow consumers of the UnifiedEncoder to set their own path prefixes
which will override the defaults of a/ and b/.

Setting this should be analogous to setting `--src-prefix` and
`--dst-prefix` in `git diff`.

See https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---src-prefixltprefixgt

### DIFF
--- a/plumbing/format/diff/unified_encoder.go
+++ b/plumbing/format/diff/unified_encoder.go
@@ -38,6 +38,10 @@ type UnifiedEncoder struct {
 	// a change.
 	contextLines int
 
+	// srcPrefix and dstPrefix are prepended to file paths when encoding a diff.
+	srcPrefix string
+	dstPrefix string
+
 	// colorConfig is the color configuration. The default is no color.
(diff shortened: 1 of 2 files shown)


yabberyabber:
I believe these failures are unrelated to my change. Ref #51 

Is there a way I can configure this change to re-check without submitting an arbitrary change?

mcuadros:
Can you provide some tests? 
Thanks

yabberyabber:
> Can you provide some tests?
> Thanks

Yep - latest patchset has a test which covers non-default prefixes

yabberyabber:
@mcuadros Can we cut of a new release of go-git? Quite a few things have changed since 5.2 and it would be nice to be able to import these changes with go modules

mcuadros:
done

PR #275, opened 2021-03-28
plumbing: make the config entirely case insensitive for keys

MichaelMure:
According to https://git-scm.com/docs/git-config/2.31.1:

> The variable names are case-insensitive, [...]

Other code path has been changed to be case insensitive but this one seems to have been overlooked.

### DIFF
--- a/plumbing/format/config/section.go
+++ b/plumbing/format/config/section.go
@@ -139,9 +139,9 @@ func (s *Section) RemoveOption(key string) *Section {
 	return s
 }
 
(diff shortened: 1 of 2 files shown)


MichaelMure:
... turns out that go-git's behavior match what git is doing. I suppose that make some sense, even though that feels wrong. I'll let you judge.

mcuadros:
Sorry, I prefer to follow git on this kind of thanks. Anyway thanks for the PR.

PR #278, opened 2021-04-01
Remote: new ListContext function

xiujuan95:
It's from this discussion: https://github.com/go-git/go-git/pull/246#discussion_r592130087

When a user calls `go-git` package to do `git ls-remote` check, we should provide a timeout. Otherwise, this will be stuck in some situation, and cause a controller which is using `List` alway waits again and again.. This is not expected I think.

This pull request add a `ListContext` for `List` func and pass `context`. For context, I set `10s` timeout. If exceed 10s, then `List` func will exit. Maybe `10s` is too short, we can discuss a reasonable value about it.

Also I add a unit test to test timeout case. Locally, all unit tests passed.

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"time"
 
 	"github.com/go-git/go-billy/v5/osfs"
 	"github.com/go-git/go-git/v5/config"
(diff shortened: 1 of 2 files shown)


xiujuan95:
Does anyone know how can I re-run my tests directly rather than change my codes and push again. (Because these failed tests are not related with my changes)

xiujuan95:
Could anyone pls help review this PR? Maybe @mcuadros, can you help review it pls? Thanks in advance!

mcuadros:
Looks great can you rebase it?

xiujuan95:
> Looks great can you rebase it?

@mcuadros Sure, done! Please take a look again, thanks a lot!

go.sum
```diff
@@ -25,24 +25,16 @@ github.com/go-git/go-billy/v5 v5.0.0 h1:7NQHvd9FVid8VL4qVUMm8XifBK+2xCoZ2lSk0agR
 github.com/go-git/go-billy/v5 v5.0.0/go.mod h1:pmpqyWchKfYfrkb/UVH4otLvyi/5gJlGI4Hb3ZqZ3W0=
 github.com/go-git/go-billy/v5 v5.1.0 h1:4pl5BV4o7ZG/lterP4S6WzJ6xr49Ba5ET9ygheTYahk=
 github.com/go-git/go-billy/v5 v5.1.0/go.mod h1:pmpqyWchKfYfrkb/UVH4otLvyi/5gJlGI4Hb3ZqZ3W0=
-github.com/go-git/go-git-fixtures/v4 v4.0.1 h1:q+IFMfLx200Q3scvt2hN79JsEzy4AmBTp/pqnefH+Bc=
```
mcuadros, line 28: please don't update the go.sum

xiujuan95: @mcuadros May I know why?

mcuadros: Because the purpose of this PR is to add the ListContext function not update the dependencies.

xiujuan95: I see, done! @mcuadros Please take a look again, thanks!

PR #284, opened 2021-04-02
Fetch submodules pointing to orphaned but still reachable commits

edigaryev:
Sometimes submodules are not updated for long periods of time and the tree that holds the commit they point to gets removed at some point, which results in commits that are orphaned, but still reachable using [`allow-reachable-sha1-in-want`](https://git-scm.com/docs/protocol-capabilities#_allow_reachable_sha1_in_want) Git's wire protocol capability.

Here's a real case: https://github.com/cirruslabs/cirrus-ci-docs/pull/830#issuecomment-812618218.

This change makes sure go-git will behave similarly to Git in such cases, which [does a subsequent fetch](https://github.com/git/git/blob/080627942802e767b068a491f3d349227f4e2cdc/submodule.c#L1571-L1577) using a commit hash.

### DIFF
--- a/submodule.go
+++ b/submodule.go
@@ -254,6 +254,24 @@ func (s *Submodule) fetchAndCheckout(
 		return err
 	}
 
+	// Handle a case when submodule refers to an orphaned commit that's still reachable
+	// through Git server using a special protocol capability[1].
+	//
+	// [1]: https://git-scm.com/docs/protocol-capabilities#_allow_reachable_sha1_in_want
+	if !o.NoFetch {
+		if _, err := w.r.Object(plumbing.AnyObject, hash); err != nil {
+			refSpec := config.RefSpec("+" + hash.String() + ":" + hash.String())
+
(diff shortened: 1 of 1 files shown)


fkorotkov:
@mcuadros, gentle ping. I'll appreciate if you could take a look. This fix will help FreeBSD folks test FreeBSD on Cirrus CI faster.

mcuadros:
any chance to do a test?

edigaryev:
> any chance to do a test?

Sure, if you don't mind accepting a PR with a single commit first and then dropping the tree it belongs to.

I assume I'll need to create a PR in https://github.com/git-fixtures/basic?

mcuadros:
Yup

edigaryev:
> Yup, there is a submodule fixture but maybe is not enough.

Well, OK 🤷 

It seems that the `TestResolveRevision` is failing on `master`, any idea if this might be related to this PR?

PR #298, opened 2021-04-14
plumbing: transport/ssh, support more formats in `NewPublicKeys` SSH helper

### LINKED ISSUE #297: SSH transport does not support newer encryption formats
During the consumption of this library we ran into a limitation of the [`NewPublicKeys` helper](https://github.com/go-git/go-git/blob/bf3471db54b0255ab5b159005069f37528a151b7/plumbing/transport/ssh/auth_method.go#L125) present in the SSH transport library, which does not seem to support newer encryption formats like PEM blocks with a [password protected ED25519 key](https://github.com/golang/crypto/blob/master/ssh/testdata/keys.go#L260-L272).

Mitigation of this issue, even as a consumer, is fairly easy by creating our own helper that makes use of `ParsePrivateKey` and `ParsePrivateKeyWithPassphrase` from `crypto/ssh`, but it would likely be better if the library itself would switch to using these functions.

hiddeco:
Fixes #297

### DIFF
--- a/plumbing/transport/ssh/auth_method.go
+++ b/plumbing/transport/ssh/auth_method.go
@@ -1,8 +1,6 @@
 package ssh
 
 import (
-	"crypto/x509"
-	"encoding/pem"
 	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


mcuadros:
can you rabase?

hiddeco:
Done.

PR #303, opened 2021-04-18
plumbing: format/packfile, prevent large objects from being read into memory completely

zeripath:
This PR adds code to prevent large objects from being read into memory from packfiles or the filesystem.

Objects greater than 1Mb are now no longer directly stored in the cache
or read completely into memory.

Signed-off-by: Andrew Thornton <art27@cantab.net>


plumbing/format/packfile/patch_delta.go
```diff
@@ -73,6 +75,123 @@ func PatchDelta(src, delta []byte) ([]byte, error) {
...
+		if err == io.EOF {
+			return nil, ErrInvalidDelta
+		}
+		return nil, err
+	}
+	remainingTargetSz := targetSz
+
+	dstRd, dstWr := io.Pipe()
+
+	go func() {
```
jfontan, line 102: Why a goroutine instead of returning a reader that calls this code? Could it happen that if you don't read the whole object the goroutine is left behind?

jfontan:
@zeripath Thanks for the PR. I like the changes but I'm a little worried of using a goroutine for the reader. Do you think it can be done without using it?

plumbing/format/packfile/patch_delta.go
zeripath: If you don't read the whole object - then you will `Close()` the returned `io.ReadCloser`. That reader is `dstRd` a `*io.PipeReader` so closing it will `Close()` the `dstWr *io.PipeWriter`.

This goroutine should only be blocked in a `dstWr.Write(...)` function call - either as part of an `io.Copy` or other `Write`. On `dstRd.Close()` the `dstWr.Write(...)` will fail and return an error, which will then cause `_ = dstWr.CloseWithError(err)` to be called (which does nothing in this case) and then finally `return` killing the go-routine.

If the goroutine gets blocked somewhere else, say L140-L172 - then when it does become unblocked then at the time of writing to dstWr the closed pipe will be seen and the goroutine closed at that point. However, most likely at this point `dstRd` reader is actually going to be blocked in `Read()`. 

(Note: I noticed I had missed a few error-checks and returns below which I have put in.)

Whilst it should be possible to unroll this to get away from the goroutine - that would be much more complicated and require much more state which the pipe makes simpler.

jfontan (approved):

zeripath:
So before you merge this, there are a couple of questions that we should think about:

1. Should we really be caching these incompletely read objects? - Could the object files be changed under our feet by a repack etc? Could you end up with a broken object that it would not be easy to get a gogit repository to clean out from the cache? I don't know go-git's architecture completely so I'm not certain if that's possible.
2. Is the 1Mb limit too low for the delta'd objects? It feels appropriate for the raw object db cases as those are simple file reads - whereas a delta'd object is a little more involved and perhaps the limit should be raised to account for the complexity?

Many thanks for your work on this

zeripath:
Sorry I've taken so long to answer my own questions here. 

> 1. Should we really be caching these incompletely read objects? - Could the object files be changed under our feet by a repack etc? Could you end up with a broken object that it would not be easy to get a gogit repository to clean out from the cache? I don't know go-git's architecture completely so I'm not certain if that's possible.

This is fine - it's only when you get the reader that you actually do the reading.

> 2. Is the 1Mb limit too low for the delta'd objects? It feels appropriate for the raw object db cases as those are simple file reads 
- whereas a delta'd object is a little more involved and perhaps the limit should be raised to account for the complexity?

This is also probably fine.

PR #311, opened 2021-05-07
git: Add support for deepening shallow clones

marwatk:
We'd previously get an error "object not found" when attempting to fetch a new depth on a shallow clone. This PR fixes the issue and adds a relevant test.

We could not get untouched tests to fully pass in our environment, with or without our changes. We're hoping your CI is properly configured to run the entire suite, if not we can investigate further.

Let me know if there are any changes you'd like to see before accepting this PR.

Thanks!

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -350,6 +350,13 @@ func (r *Remote) fetch(ctx context.Context, o *FetchOptions) (sto storer.Referen
 		return nil, err
 	}
 
+	if !req.Depth.IsZero() {
+		req.Shallows, err = r.s.Shallow()
+		if err != nil {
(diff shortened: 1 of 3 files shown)


remote.go
```diff
@@ -367,13 +374,43 @@ func (r *Remote) fetch(ctx context.Context, o *FetchOptions) (sto storer.Referen
 		return nil, err
 	}
 
+	if !updated {
+		updated, err = depthChanged(req.Shallows, r.s)
+		if err != nil {
+			return nil, fmt.Errorf("error checking depth change: %v", err)
+		}
+	}
+
```
mcuadros, line 383: There is two equal conditions together :P

PR #322, opened 2021-05-24
make ListOptions timeout configurable

khrisrichardson:
I appreciate the recent addition of `context.WithTimeout` in `remote.go`, however since then I have been regularly running into situations where `0.6s` was insufficient for the equivalent of `ls-remote`, so this PR bumps the context deadline to `10s` and makes it configurable should that not be sufficient.


khrisrichardson:
Ha ha. Just saw #321 :)

mcuadros:
Since there is no other method where you can configure the timeout I rather leaving like this. Thanks anyway

PR #324, opened 2021-05-25
internal: url, support more scp-like URLs

loderunner:
SCP-like URLs do not necessarily need to have a prefix followed by a `/` in the path.

For example, if a git repository is located at `/test-repo.git` on a remote, the following command is valid.

```shell
git clone user@remote:/test-repo.git
```

As long as `user` can use SSH to log into the remote (which might be described in SSH config), `git` will attempt to clone the repo at the given path.

----

I removed the `/` as separator for the port as it interfered with this URL scheme, but I might have missed the rationale for this. Feel free to explain, and I will update the PR accordingly.

### DIFF
--- a/internal/url/url.go
+++ b/internal/url/url.go
@@ -6,7 +6,7 @@ import (
 
 var (
 	isSchemeRegExp   = regexp.MustCompile(`^[^:]+://`)
-	scpLikeUrlRegExp = regexp.MustCompile(`^(?:(?P<user>[^@]+)@)?(?P<host>[^:\s]+):(?:(?P<port>[0-9]{1,5})(?:\/|:))?(?P<path>[^\\].*\/[^\\].*)$`)
+	scpLikeUrlRegExp = regexp.MustCompile(`^(?:(?P<user>[^@]+)@)?(?P<host>[^:\s]+):(?:(?P<port>[0-9]{1,5})(?::))?(?P<path>[^\\]+)$`)
 )
 
 // MatchesScheme returns true if the given string matches a URL-like
--- a/internal/url/url.go
+++ b/internal/url/url.go
@@ -6,7 +6,7 @@ import (
 
 var (
 	isSchemeRegExp   = regexp.MustCompile(`^[^:]+://`)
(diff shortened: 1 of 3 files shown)


mcuadros:
can you take a look to the testS?

loderunner:
I rewrote some things in the tests.

* Fixed the regex to support port number in scp-like URL better
* Fixed the regex to support root paths better
* Fixed an assertion in `internal/url/url_test.go`
* Changed a test in `plumbing/transport/common_test.go`

----

I tried my bust, but I'm worried about breaking compatibility, here.

Investigating a bit deeper, I found that `git` does not support port numbers in SCP-like URLs.

On github (connection should be refused):

```
$ git clone git@github.com:2222:loderunner/go-git.git
Cloning into 'go-git'...
fatal: remote error:
   is not a valid repository name
  Visit https://support.github.com/ for help
```

With `scrt.test` defined as `127.0.0.1` in `/etc/hosts`:

```
$ git clone scrt@scrt.test:2222:repos/toto.git
Cloning into 'toto'...
ssh: connect to host scrt.test port 22: Connection refused
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

```
$ git clone scrt@127.0.0.1:2222:repos/toto.git
Cloning into 'toto'...
ssh: connect to host 127.0.0.1 port 22: Connection refused
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Ran on my MacBook Pro, with an sshd running at port 2222

```
$ git --version
git version 2.24.3 (Apple Git-128)
```

----

According the the `scp` manpage,  `git@github.com:9999/user/repository.git` is not a valid path anyway:

```
The source and target may be specified as a local pathname, a remote host with optional path in the form [user@]host:[path], or a URI in the form scp://[user@]host[:port][/path].
```

What should I do: stick to git specification or maintain `go-git` backwards compatibility?

mcuadros:
Support backward compatibility is the best in this case.

b3n4kh:
My PR to this issue got closed as duplicate - this one is apparently dead. What is the way forward here?

pjbgf:
Closing as this was fixed by #715.

Issue #105, opened 2020-06-19
Pushing a hash to a branch is not working?


eelco:
I’m trying to push a commit hash to a branch, this is a fairly common pattern for us (to do deployments):

```
        refSpecs := []config.RefSpec{
		config.RefSpec(fmt.Sprintf("%s:refs/heads/release/%s", commit.Hash, env)),
	}
	err = r.Push(&git.PushOptions{
		RemoteName: "origin",
		RefSpecs:   refSpecs,
        }
```

However, this results in an `already up-to-date` error, which is definitely not true. The equivalent command on the CLI works without problem, so I’m assuming it’s not the RefSpec that’s wrong here.

yurifrl:
I'm having the exact same issue

polothy:
Had same problem, this worked for me:
```go
		headRef, err := repo.Head()
		if err != nil {
			return err
		}
		refSpec = fmt.Sprintf("%s:refs/heads/branch-name", headRef.Name())
		err = repo.Push(&git.PushOptions{
			RemoteName: "origin",
			RefSpecs: []config.RefSpec{config.RefSpec(refSpec)},
			Auth: &http.BasicAuth{...},
			Force: true,
		})
		if err != nil {
			return err
		}
```

The `refSpec` ends up having a value of just `HEAD:refs/heads/branch-name`

martin-sucha:
The refspec is ignored here: https://github.com/go-git/go-git/blob/c66023ad305ee56a38781cc5d3ca96c7115bd2e5/remote.go#L476

hermanbanken:
I also have this issue when I directly specify the refspec to be a commit hash:

```
	err = outputRepo.Push(&git.PushOptions{
		RefSpecs:          []config.RefSpec{config.RefSpec("38fed38fa8f8f4f9a5564a5727cd6778b97a2c75:refs/heads/somebranch")},
		Auth:              gitAuth,
		Progress:          prefixw.New(os.Stderr, "> "),
	})
```

I was trying to use the `RequireRemoteRefs` as an alternative to `Force`, but that doesn't work. If I set `Force`, somehow the push goes through. Seems that there is a bug deciding if a push needs to happen (internally `len(req.Commands) == 0` while refs need to be updated...)

PR #325, opened 2021-05-26
Remote: Push, add support to push commits per hashes

tjamet:
Using plain git, the command `git push ${sha}:refs/heads/some-branch`
actually ensures that the remote branch `some-branch` points to the
commit `${sha}`.

In the current version of go-git, this results in an "everything is
up to date" error.

When a source reference is not found, check the object storage to find
the sha. If it is found, consider pushing this exact commit.

fixes: #105

### DIFF
--- a/options.go
+++ b/options.go
@@ -198,8 +198,14 @@ type PushOptions struct {
 	RemoteName string
 	// RemoteURL overrides the remote repo address with a custom URL
 	RemoteURL string
-	// RefSpecs specify what destination ref to update with what source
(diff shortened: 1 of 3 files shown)


mcuadros:
Can you add some documentation on the public method?

tjamet:
Sure! I have just added a commit to document the reference, as stated in `git help push`

tjamet:
Hi @mcuadros is there any other place I need to add the documentation to?

PR #326, opened 2021-05-27
Fix object not found error in OFS deltas

### LINKED ISSUE #323: latest "prevent large objects from being read into memory" commit caused "object not found" error
@zeripath Before applying commit "[720c192](https://github.com/go-git/go-git/commit/720c192831a890d0a36b4c6720b60411fa4a0159)", these codes works:

    _, err := git.PlainClone(pkgDir, false, &git.CloneOptions{
      URL:               "https://github.com/rime/rime-cantonese",
      ReferenceName:     plumbing.NewBranchReferenceName("master"),
      RecurseSubmodules: git.DefaultSubmoduleRecursionDepth,
      Depth:             1})
    fmt.Println(err)
    nil

After the commit, these codes return a "object not found" error.

Because now the 3mb file "jyut6ping3.dict.yaml" in commit tree object can't not have a blob Reader().

That is, the b.obj.Reader() function in plumbing/object/object.go can't return a valid io.ReadCloser and nil error now.

zeripath:
Unfortunately the code in #303 was incorrect for large objects in OFS deltas.

The underlying fault is that the seeker in the packfile was seeking away whilst the
delta was being read.

This PR simply reads the delta into a buffer.

Fix #323

Signed-off-by: Andrew Thornton <art27@cantab.net>

### DIFF
--- a/plumbing/format/packfile/packfile.go
+++ b/plumbing/format/packfile/packfile.go
@@ -293,24 +293,28 @@ func (p *Packfile) getReaderDirect(h *ObjectHeader) (io.ReadCloser, error) {
 	case plumbing.CommitObject, plumbing.TreeObject, plumbing.BlobObject, plumbing.TagObject:
 		return p.s.ReadObject()
 	case plumbing.REFDeltaObject:
-		deltaRC, err := p.s.ReadObject()
+		deltaBuf := bufPool.Get().(*bytes.Buffer)
+		defer bufPool.Put(deltaBuf)
+		deltaBuf.Reset()
+		_, _, err := p.s.NextObject(deltaBuf)
 		if err != nil {
 			return nil, err
 		}
-		r, err := p.readREFDeltaObjectContent(h, deltaRC)
+		r, err := p.readREFDeltaObjectContent(h, deltaBuf)
 		if err != nil {
-			_ = deltaRC.Close()
 			return nil, err
 		}
 		return r, nil
 	case plumbing.OFSDeltaObject:
-		deltaRC, err := p.s.ReadObject()
+		deltaBuf := bufPool.Get().(*bytes.Buffer)
+		defer bufPool.Put(deltaBuf)
+		deltaBuf.Reset()
+		_, _, err := p.s.NextObject(deltaBuf)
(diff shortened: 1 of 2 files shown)


starkers:
I just tested this in my project and it appears to have resolved the issue, thanks for the quick turn around @zeripath

puerco:
Hi @zeripath, thanks a lot for working on this fix! I'm writing on behalf of the Kubernetes Release Engineering team. It seems that the issue this PR fixes broke the Kubernetes release process and we are unable to cut new versions right now :upside_down_face: 

I was wondering If I can help you to merge this faster. Do you need help testing the patch or can we assist in any other way from our side? Thanks!!

zeripath:
@starkers agh. Ok so the packfile must still be being used elsewhere. I guess we need to duplicate the packfile fd for each of these. I'll take a look. Would you be willing to help test this functionality if I adjust this PR so that it switches this off by default?

@puerco - sorry about this. I'll adjust this pr to switch this off by default until we're sure it's working.

hiddeco:
@zeripath we have been testing this PR in https://github.com/fluxcd/image-automation-controller/pull/173 in the hope it would also solve another issue we have been running into around `empty git-upload-pack given` errors that are thrown since `v5.4.0` (#328).

Is it possible there was another bug introduced somewhere that in some conditions sends empty packages? As this patch does not seem to have an effect.

mcuadros:
I just created a new release with #329: https://github.com/go-git/go-git/releases/tag/v5.4.2

zeripath:
@mcuadros I'm deeply sorry about this.

Yeah we do still need to do something about the large files - especially to do with stating of files causing these files to end up memory.

I'll have another go at this. @starkers do you have a simple test case that can be used to replicate the issues?

puerco:
Thank you very much @zeripath and everyone involved, I can confirm we are able to clone large repos again. I am attaching here the test program I used to diagnose the problem when it hit us. I hope it helps. Let me know if we can help further :)

[test-clone.go.gz](https://github.com/go-git/go-git/files/6585213/test-clone.go.gz)

PR #329, opened 2021-06-02
Revert "plumbing: format/packfile, prevent large objects from being read into memory completely (#303)"

zeripath:
This reverts commit 720c192831a890d0a36b4c6720b60411fa4a0159.

It appears that this just doesn't work and is causing major problems.

I think the changes in storage/filesystem/object.go are fine but the delta and OFS changes would need more thought. (Likely changing the readers to always use ReadAt(...) or something shadowing that with a stored offset.

I deeply apologise for breaking workflow - I had thought that the tests were catching this - clearly not.


mcuadros:
Thanks, I was about to do it.  Don't worry these things happens.

Issue #344, opened 2021-06-25
scpLikeUrlRegExp requires `/` character


b3n4kh:
scpLikeUrlRegExp `^(?:(?P<user>[^@]+)@)?(?P<host>[^:\s]+):(?:(?P<port>[0-9]{1,5})(?:\/|:))?(?P<path>[^\\].*\/[^\\].*)$`

Requires that the url includes a  `/` in the <path> component. This should be optional as per [rfc3986](https://datatracker.ietf.org/doc/html/rfc3986).

This prevents access URIs like `git.example.com:repo.git` which would be a valid git repo / scp URI.

A relevant rfc draft to consider could also be [draft-ietf-secsh-scp-sftp-ssh-uri-04](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-scp-sftp-ssh-uri-04#section-4.3) which also referes to rfc3986.

This bug was newly introduced by: https://github.com/go-git/go-git/commit/db0e226d48285a0ae8cda7db05d2ca20d9000dc6

pjbgf:
Closing as this was fixed by #715.

PR #345, opened 2021-06-25
fix rexexp to match https://datatracker.ietf.org/doc/html/rfc3986#pag…

b3n4kh:
Resolves Issue: https://github.com/go-git/go-git/issues/344


mcuadros:
We are already working on this: https://github.com/go-git/go-git/pull/324

PR #354, opened 2021-07-24
plumbing: config, Branch name with hash can be cloned. Fixes #309

### LINKED ISSUE #309: Can't clone branch ref name with hash/number sign (#)
I am cloning a repository in single branch mode with a branch name that contains a hash/number sign (#) into name. And get an error.

Steps to reproduce (created a public github repository for that):

```
_, err := git.PlainClone("/tmp/repo", false, &git.CloneOptions{
    URL:           "git@github.com:jkblume/gogit-hashtag-bug.git",
    ReferenceName: plumbing.ReferenceName("refs/heads/branch#name"),
    SingleBranch:  true,
})

fmt.Println(err.Error())
```

The command fails with error message: 'malformed refspec, separators are wrong' in file config/refspec.go line 35.

dowy:
This is a fix for https://github.com/go-git/go-git/issues/309 'Can't clone branch ref name with hash/number sign (#)'

The problem is occurring because the # character in a git config file is interpreted as a comment. In fact, the config-file writer needs to escape a number of characters that would cause it to create an invalid config-file.

Following example copied from #611

When a repository is cloned a config file is created in the .git directory. I have mocked up an example below by hand:

```ini
[core]
	bare = false # This is just a comment so is ignored
[remote "origin"]
	url = https://https://github.com/myrepo.git
	fetch = +refs/heads/release/mybranchname:refs/remotes/origin/release/mybranchname

```
However, if there is a # character in the branch name like this:

```ini
[core]
	bare = false # This is just a comment so is ignored
[remote "origin"]
	url = https://https://github.com/myrepo.git
	fetch = +refs/heads/release/my#branchname:refs/remotes/origin/release/my#branchname

```
When go-git reads back the file it ignores everything after the # character, so the value read for the fetch option, rather than looking like this:

```
+refs/heads/release/my#branchname:refs/remotes/origin/release/my#branchname

```
It is read back from the file as this:

```
+refs/heads/release/my

```
go-git then returns an error 'malformed refspec, separators are wrong' because it is expecting at least one ':' separator character in that value.

The fix is to add quotes when encoding that file, so it looks like this:

```ini
[core]
	bare = false # This is just a comment so is ignored
[remote "origin"]
	url = https://https://github.com/myrepo.git
	fetch = "+refs/heads/release/my#branchname:refs/remotes/origin/release/my#branchname"

```
Then when reading and decoding that file go-git gets the correct value.

### DIFF
--- a/plumbing/format/config/encoder.go
+++ b/plumbing/format/config/encoder.go
@@ -11,6 +11,10 @@ type Encoder struct {
 	w io.Writer
 }
 
+var (
+	subsectionReplacer = strings.NewReplacer(`"`, `\"`, `\`, `\\`)
+	valueReplacer = strings.NewReplacer(`"`, `\"`, `\`, `\\`, "\n", `\n`, "\t", `\t`, "\b", `\b`)
+)
 // NewEncoder returns a new encoder that writes to w.
 func NewEncoder(w io.Writer) *Encoder {
 	return &Encoder{w}
@@ -48,8 +52,7 @@ func (e *Encoder) encodeSection(s *Section) error {
 }
 
 func (e *Encoder) encodeSubsection(sectionName string, s *Subsection) error {
-	//TODO: escape
-	if err := e.printf("[%s \"%s\"]\n", sectionName, s.Name); err != nil {
+	if err := e.printf("[%s \"%s\"]\n", sectionName, subsectionReplacer.Replace(s.Name)); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


mcuadros:
How does this help to the issue? Since this is only related to the config file. 
Have you tested this with the example code from the original issue?

dowy:
> 
> 
> How does this help to the issue? Since this is only related to the config file.
> Have you tested this with the example code from the original issue?

Yes, I have tested it.
The "PlainClone" code writes a bunch of stuff to a Config file, then reads it back wrongly, ignoring the "comment" after the hash.
I don't understand why the code does that but fixing the Config file handling fixes the problem.

dowy:
@mcuadros
Are you ok to approve this?

shiqi-synctera:
We are also running into this, would love if the fix can be merged in :))

erizocosmico (approved):

PR #356, opened 2021-07-27
plumbing: object/patch, Support renames in FileStat. Fixes #261

### LINKED ISSUE #261: Renamed files have an empty Name field in FileStat
Take for example `https://xenbits.xen.org/git-http/xen.git`, commit `9872d99215ba9234b927d2c55625b70b4f555c60`.  Two of the files in that commit have been detected as a rename; the git output for one of those files looks like this:

```diff
diff --git a/xen/arch/x86/microcode.c b/xen/arch/x86/cpu/microcode/core.c
similarity index 99%
rename from xen/arch/x86/microcode.c
rename to xen/arch/x86/cpu/microcode/core.c
index 27a88c6826..ac5da6b2fe 100644
--- a/xen/arch/x86/microcode.c
+++ b/xen/arch/x86/cpu/microcode/core.c
[...]
```

The `FileStat` entry that matches that hunk has `Addition` set to 4, `Deletion` set to 11, but `Name` is the empty string.

It would be good to be able to say something usefully about renames: either make Name a `[]Name`, or (perhaps better and more backwards-compatible), add a `NewName` field which, if non-null, would indicate that `Name` was renamed to `NewName`.

BTW, thanks so much for this module -- it's been quite helpful!

bassettb:


kubaracek:
@mcuadros Does this look conceptually ok? I might take over the PR and add some tests.

PR #360, opened 2021-08-04
Support building with GopherJS

paralin:
This PR adds supports building with "gopherjs build."


paralin:
Not stale

pjbgf:
Are you using go-git with GopherJS? If so, we could add a new CI workflow to test that and ensure we don't break support going forwards. Something around the lines of the [JS](https://github.com/go-git/go-billy/blob/master/.github/workflows/test_js.yml) and/or [wasm](https://github.com/go-git/go-billy/blob/master/.github/workflows/test_wasip1.yml) workflows we have in go-billy could work here.

paralin:
I am not currently as it does not yet support go1.20, go1.21, go1.22

However, there is effort underway to get it up to spec with go1.22 and once that completes I will test it & likely use it.

If you want, we can close this for now and I'll reopen it if I use gopherjs again.

pjbgf:
@paralin thanks for the quick reply. Let's close it for now as we are due a bump out of Go1.19, and reopen once they support the latest Go versions.

PR #368, opened 2021-08-23
Fix nil pointer panic when omitting FetchOptions

schrej:
When omitting `FetchOptions` when calling any of the fetch methods, a nil pointer dereference occurs. This PR fixes this by creating `FetchOptions` object when the parameter is `nil`.


mcuadros:
Options are never optional.

schrej:
Then why does the function expect a pointer?

PR #374, opened 2021-09-10
examples: added "tag find if head is tagged"

snebel29:
I was trying to find if there was a tag referencing HEAD, it took me a while playing with examples which were key to be successful, so I've decided to add this example in case this is useful for someone else in the future.

Cheers.


mcuadros:
Can you add it to the index (the markdown) file from the folder?

snebel29:
@mcuadros done.

PR #378, opened 2021-09-17
Limiting search space, don't continue to earlier commits than the sin…

laszlocph:
The date based filter should just return with EOF if it sees the first commit that is earlier than the since limit.

Before this change, even with the since filter, all commits in the repo were examined. Causing a huge wait time for large repos to be traversed.


mcuadros (changes requested):
can you provide a test?

PR #375, opened 2021-09-15
Add RemoteURL to {Fetch,Pull,Push}Options

noerw:
Adds an option `o.RemoteURL`, that can be used to override the URL of the remote repo to operate on:
`o.RemoteName` will be ignored for the actual fetch.

fixes #173

PR #379, opened 2021-09-25
git: Allow supplying a custom transport when performing various operations

cynix:
Currently, custom transports are registered by protocol name, therefore it is cumbersome to use different transports for different remotes if they have the same protocol. This PR makes that easier by introducing an optional transport field in the option structs of various operations.


noerw:
Do I understand correctly, that this would also allow to selectively operate on https or ssh, the same way as #375 does?
In that case you provided a more generic solution, and we could close #375 in favor of this.

PR #383, opened 2021-10-01
Correct parsing of git-ssh url without breaking backward compatibility

Feresey:
I really miss the correct url handling in the git scp format.

I found this #324  merge request, but it complained about a lack of backward compatibility.

This is my attempt not to break backward compatibility and add the correct parsing of git url


mcuadros (changes requested):
Can you refactor it to make it work with a global variable instead of a struct field?

PR #385, opened 2021-10-05
git: add --follow-tags option for pushes

john-cai:
This PR adds support for the --follow-tags option for pushes.


john-cai:
@mcuadros could you grant workflow permission for me? thanks!

PR #404, opened 2021-11-02
Remote: Push, add ForceWithLease Push Option

john-cai:
--force-with-lease allows a push to force push with some safety
measures. If the ref on the remote is what we expect, then the force
push is allowed to happen.
See https://git-scm.com/docs/git-push#Documentation/git-push.txt---force-with-leaseltrefnamegt
for more information


john-cai:
just saw https://github.com/go-git/go-git/pull/258/commits/a39eddcb06323a22f2ce9b07b710c428f38bbd52. might need to refactor

PR #416, opened 2021-11-26
adjust path extracted from file: url on Windows

### LINKED ISSUE #415: file: protocol support on Windows doesn't appear to work
I'm trying to use go-git with a file: protocol on Windows.

So my URL is something like "file:///c:/directory".
This parses correctly, host is "" (correctly), but the curiosity with windows file uRLs is that the path is then "/c:/directory".

This path is then passed to the git command line directly, and git then fails to clone.

It gives the appearance of working if you use a file url of "file://c:/directory", so only two slashes.
It only appears to work though, as the URL is parsed as "c" being the host, and "/directory" being the path. So it works as long as your  current drive is "c:".

The logic for converting file: URLs into actual usable windows file paths has to be something like "if windows and path starts slash, letter, colon, slash, then remove the leading slash".

Note that you can't use windows file paths in go-git without using the URL form, as it interprets it as an scp like URL and not a file path.

tomqwpl:
Fixes #415

A file: url for windows looks something like `file:///d:/directory/file.txt` . Once parsed, the Path from the URL has `/d:/directory/file.txt`. This is not a valid path to the operating system. In order for it to be meaningful, the leading slash needs to be removed.

### DIFF
--- a/plumbing/transport/file/client.go
+++ b/plumbing/transport/file/client.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/go-git/go-git/v5/plumbing/transport"
@@ -96,7 +97,23 @@ func (r *runner) Command(cmd string, ep *transport.Endpoint, auth transport.Auth
 		}
 	}
 
-	return &command{cmd: execabs.Command(cmd, ep.Path)}, nil
+	return &command{cmd: execabs.Command(cmd, adjustPathForWindows(ep.Path))}, nil
+}
+
+func isDriveLetter(c byte) bool {
+	return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z')
+}
+
+// On Windows, the path that results from a file: URL has a leading slash. This
+// has to be removed if there's a drive letter
+func adjustPathForWindows(p string) string {
+	if runtime.GOOS != "windows" {
+		return p
+	}
+	if len(p) >= 3 && p[0] == '/' && isDriveLetter(p[1]) && p[2] == ':' {
+		return p[1:]
+	}
+	return p
 }
 
 type command struct {
--- a/repository_windows_test.go
+++ b/repository_windows_test.go
@@ -38,6 +38,7 @@ func (s *RepositorySuite) TestCloneFileUrlWindows(c *C) {
 	c.Assert(err, IsNil)
 
 	url := "file:///" + strings.ReplaceAll(dir, "\\", "/")
+	c.Assert(url, Matches, "file:///[A-Za-z]:/.*")
 	_, err = Clone(memory.NewStorage(), nil, &CloneOptions{
 		URL: url,
 	})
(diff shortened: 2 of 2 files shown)


mcuadros:
Can you provide some test failing on windows?

tomqwpl:
I've added a test. Let me know if you think I've put it in the wrong place, or if it should be done differently. I couldn't see any obvious similar example, so I write a test that created a new repo, committed a change and then tried to clone it using a file:/// URL. Without the change the test fails.

hairyhenderson:
I'm hitting this bug and would love to see this merged - anything remaining?

plumbing/transport/file/client.go
```diff
@@ -96,7 +97,23 @@ func (r *runner) Command(cmd string, ep *transport.Endpoint, auth transport.Auth
...
+
+// On Windows, the path that results from a file: URL has a leading slash. This
+// has to be removed if there's a drive letter
+func adjustPathForWindows(p string) string {
+	if runtime.GOOS != "windows" {
+		return p
+	}
+	if len(p) >= 3 && p[0] == '/' && isDriveLetter(p[1]) && p[2] == ':' {
+		return p[1:]
+	}
```
pjbgf, line 115: Should we not rely on [filepath.VolumeName](https://pkg.go.dev/path/filepath#VolumeName) instead?

jfontan: Maybe [filepath.Clean](https://pkg.go.dev/path/filepath#Clean) is all we need. I could not test it in Windows.

pjbgf: +1 on the `filepath.Clean`.

hairyhenderson: `filepath.Clean` changes `/` to `\` so that isn't ideal here - this is dealing with a URL path rather than a filesystem path.

As for `filepath.VolumeName`, I'm not sure that's appropriate either because it not only deals with drive letters, but also UNC paths like `\\host\share`.

hairyhenderson: For what it's worth, I've tested changing this to `filepath.Clean` on a Windows VM and the test fails.

I also played around with `filepath.VolumeName` and this works:

```go
	if p != "" && p[0] == '/' && filepath.VolumeName(p[1:]) != "" {
		return p[1:]
	}
```

but it's less efficient than the existing `isDriveLetter` code (because it checks for UNC as well).

I'd love if this PR could just be merged as-is. I've been using a fork of this repo with this commit applied for the past 7 months without issue.

pjbgf: IMO, the stdlib `filepath.VolumeName` would be better suited than having our own implementation. But the current state of this PR is working and resolves a legit problem users are having.

We can revisit this at a later stage, so will be merging this as-is.

pjbgf (approved):
@tomqwpl thanks for working on this. :bow:

pjbgf:
@hairyhenderson no worries at all. As for the release, only a few `go-git` maintainers are allowed to cut new releases, and unfortunately I am not one of them. Given that we have no paid maintainers, availability can be a problem which is why it is hard to tell when a new version will be cut.

My current recommendation is for folks to pull the pseudo version from `master` instead.

hairyhenderson:
@pjbgf ok, fair enough. Based on history, it looks like @mcuadros is the only one who cuts releases... @mcuadros, when you have time... 😉

FWIW, and while it does involve a bit of up-front effort, but I've found using [Release Please](https://github.com/googleapis/release-please) to make the release process almost a non-event.

PR #419, opened 2021-11-28
NoErrAlreadyUpToDate: Rename to ErrAlreadyUpToDate

abhinav:
It's unclear why NoErrAlreadyUpToDate is named that
instead of following conventions from the rest of the code
and being named ErrAlreadyUpToDate.

This change renames NoErrAlreadyUpToDate to ErrAlreadyUpToDate.

In the interest of backwards compatibility,
we keep the NoErrAlreadyUpToDate variable around,
but mark it as deprecated in favor of ErrAlreadyUpToDate.

---

Feel free to reject this change if the name is intentional
and I've just been unable to find the reason why.

### DIFF
--- a/_examples/tag-create-push/main.go
+++ b/_examples/tag-create-push/main.go
@@ -136,7 +136,7 @@ func pushTags(r *git.Repository, publicKeyPath string) error {
 	err := r.Push(po)
 
 	if err != nil {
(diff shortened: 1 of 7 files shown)


mcuadros:
The reason is that it's not an error, but we want to notify the fact that is already up to date. As git does.
Thanks

PR #437, opened 2022-01-03
Worktree: use syscall.Timespec.Unix

tklauser:
Use the syscall method instead of repeating the type conversions for
the syscall.Stat_t Atim/Atimespec/Ctim members.


tklauser:
@mcuadros friendly ping. Mind to take a look at this?

Issue #370, opened 2021-08-27
How to create orphan branch


afdecastro879:
Hi All!

I'm struggling understanding how can I create an orphan branch using this library. Normally you can achieve this with git doing:

```
git checkout --orphan <new_branch>
```

john-cai:
https://github.com/go-git/go-git/pull/439

lacarvalho91:
@afdecastro879 

I've been looking into this and it turns out that all you need to do to create an orphaned branch is:
```golang
repo.Storer.SetReference(plumbing.NewSymbolicReference(plumbing.HEAD, myNewBranch))
```

This replicates `git checkout --orphan myNewBranch`

If you want the `git switch --orphan myNewBranch` behaviour instead (which is to create a new orphaned branch with a clear index) then you can do:
```golang
wt, err := repo.Worktree()
wt.RemoveGlob("*")
```

jmgilman:
@lacarvalho91 I was unable to get your example to work. Are you doing anything before that step? Any subsequent step after running `SetReference` fails with various errors. I'm assuming it's because something internally is getting messed up.

PR #439, opened 2022-01-03
Add CreateOrphanBranch to create an orphaned branch

john-cai:
Add a function to create a new branch according to git branch --orphan.
This creates a new branch that points to an initial commit without any
parents.

fixes: https://github.com/go-git/go-git/issues/370

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/go-git/go-billy/v5"
 	"github.com/go-git/go-billy/v5/util"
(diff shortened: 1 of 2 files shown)


janosmiko:
+1 for this feature. Could someone verify and accept this PR?

worktree.go
```diff
@@ -188,6 +189,67 @@ func (w *Worktree) Checkout(opts *CheckoutOptions) error {
 
 	return w.Reset(ro)
 }
+
+// CreateOrphanBranch creates a new orphan branch with an initial commit with msg as the commit message
+func (w *Worktree) CreateOrphanBranch(branch plumbing.ReferenceName, msg string, co *CommitOptions) error {
```
lacarvalho91, line 194: @john-cai when I use this it seems to keep the worktree from the current branch, I'm looking for something that essentially replicates: `git switch --orphan my-branch` but this implementation replicates `git checkout --orphan my-branch` instead (a flag maybe? `keepWorkingTree`). 

I'm also wondering if you could allow creating an orphan branch without committing to it, so we can add whatever files ourselves before making a commit on the orphan branch.

lacarvalho91: i've been hacking around with this and it seems if we clear the index and then construct the tree using that index instead of taking the tree from head then we can make the initial commit empty, e.g. :
```golang
	wt, err := w.r.Worktree()

	if err := w.RemoveGlob("*"); err != nil {
		return err
	}

	idx, err := w.r.Storer.Index()
	if err != nil {
		return err
	}

	h := &buildTreeHelper{
		fs: w.Filesystem,
		s:  w.r.Storer,
	}

	tree, err := h.BuildTree(idx)
	if err != nil {
		return err
	}
```

and then pass `tree` to `buildCommitObject`

lacarvalho91: which sounds like what `switch --orphan` does according to the help text:
```
Create a new orphan branch, named <new-branch>. All tracked files are removed.
```

lacarvalho91: maybe instead we can have a flag to say whether the tree is built from index or head, that way users can get the index in whatever state they want before the initial commit is made

lacarvalho91: I've been playing around with this more today and realised that we don't actually need any of this to create an orphaned branch.

`repo.Storer.SetReference(plumbing.NewSymbolicReference(plumbing.HEAD, branch))` is all you need, and it won't create any commits so is closer to the `git checkout --orphan` behaviour. To replicate the `git switch --orphan` before you can just do `worktree.RemoveAll("*")` after.

PR #443, opened 2022-01-09
Repository: don't crash accessing invalid pathinfo

muesli:
When fs.Stat returns an error, pathinfo may be nil. In such situations
the only safe response seems to be to return the error to the caller.

Without this fix, accessing pathinfo.IsDir() below would lead to a crash
dereferencing a nil pointer.

This crash can be reproduced by trying to initialize a Git repo with an
invalid path name.

Also see: https://github.com/muesli/gitty/issues/36

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -280,6 +280,9 @@ func dotGitToOSFilesystems(path string, detect bool) (dot, wt billy.Filesystem, 
 
 		pathinfo, err := fs.Stat("/")
 		if !os.IsNotExist(err) {
+			if pathinfo == nil {
+				return nil, nil, err
+			}
 			if !pathinfo.IsDir() && detect {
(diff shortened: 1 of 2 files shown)


mcuadros:
can you provide a test?

muesli:
> can you provide a test?

I thought about it, but it's actually a little tricky: it's triggered by an invalid path name that can't be `Stat`'d, but what constitutes an invalid path name very much depends on the platform *and* filesystem the test is being run on. I can write a test that works fine on GitHub's CI, but it may not produce the expected error on your local machine. I'll see if there's an easy way to provoke an error consistent across platforms.

muesli:
Ok, turns out this was easier than anticipated: a path `\000` should be invalid everywhere. I've added a test that would segfault without the proposed fix.

Issue #424, opened 2021-12-04
Add example how to check if one branch is merged to another or not


mih-kopylov:
There's a `git for-each-ref --no-merged` command that shows references that are not merged to current branch (commit)

Would be nice to have an example how to check if a reference is merged to current commit or not

fatelei:
i have add an example in https://github.com/go-git/go-git/pull/461

PR #461, opened 2022-01-29
chore: add example for for-each-ref-no-merge

fatelei:
add example for #424

### DIFF
--- a/_examples/for-each-ref-no-merge/main.go
+++ b/_examples/for-each-ref-no-merge/main.go
@@ -1,0 +1,67 @@
+package main
+
+import (
+	"errors"
+	"fmt"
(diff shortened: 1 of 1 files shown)


_examples/for-each-ref-no-merge/main.go
```diff
@@ -0,0 +1,67 @@
...
+		if errors.Is(err, io.EOF) {
+			break
+		}
+		if ref.Hash().String() == headRef.Hash().String() || ref.Name() == "HEAD" {
+			continue
+		}
+
+		if ref != nil {
+			flag := false
+			for _, object := range objects {
```
mih-kopylov, line 52: I'm curious if `objects` is a slice with all the objects of HEAD, then will the complexity be O(n^n) ?

PR #472, opened 2022-02-10
Repository: Export Clone method

nanoslayer:
Export the Repository.Clone method, which was made private in a052498f23198ad1f2e9ca8f9943d2a48e274c06.

We want to be able to initialize a Git repository and schedule the clone to happen at a later time. Currently, that is not possible with go-git as the clone method is private. Making the clone method public would allow initializing a repository before initiating a clone.

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -207,7 +207,7 @@ func CloneContext(
 		return nil, err
 	}
 
-	return r, r.clone(ctx, o)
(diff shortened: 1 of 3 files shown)


ddevault:
Bump.

ddevault:
This is still something we need.

Issue #490, opened 2022-03-17
Git credentials support


kassio:
## What are credentials?

A git credential is the data required to access a git repository:

- url, which can be split in
    - protocol, like "https"
    - host, like "github.com"
    - path, like "go-git/go-git"
- username
- password

This data can be fetched/saved using configurations in the `gitconfig`.

## What can be configured?

git permits the user to configure some credentials data to be used by default on every repository or some repositores (based on url). The following options are available:

- `username` A default username, if one is not provided in the URL.
- `useHttpPath` By default, Git does not consider the "path" component of an http URL to be worth matching via external helpers. This means that a credential stored for `https://example.com/foo.git` will also be used for `https://example.com/bar.git`. If you do want to distinguish these cases, set this option to `true`.
- `helper` The name of an external credential helper, and any associated options. If there are multiple instances of the `credential.helper` configuration variable, each helper will be tried in turn, and may provide a username, password, or nothing. Once Git has acquired both a username and a password, no more helpers will be tried.
If `credential.helper` is configured to the empty string, this resets the helper list to empty (so you may override a helper set by a lower-priority config file by configuring the empty-string helper, followed by whatever set of helpers you would like). The helper can be:
    - a shell snippet, when it starts with `!`
    
    ```
    # shell snippet
    [credential "https://example.com"]
    	username = your_user
    	helper = "!f() { test \"$1\" = get && echo \"password=$(cat $HOME/.secret)\"; }; f"
    ```
    
    - an absolute path
    
    ```
    # you can also use an absolute path, which will not use the git wrapper
    [credential]
    	helper = "/path/to/my/helper --with-arguments"
    ```
    
    - otherwise, the string `git credential-` is prepended to the string to be executed.
    
    ```
    # run "git credential-foo"
    [credential]
    	helper = foo
    
    # same as above, but pass an argument to the helper
    [credential]
    	helper = "foo --bar=baz"
    
    # the arguments are parsed by the shell, so use shell
    # quoting if necessary
    [credential]
    	helper = "foo --bar='whitespace arg'"
    ```
    
    When a helper is executed, it will have one "operation" argument appended to its command line, which is one of:
    
    - `get` Return a matching credential, if any exists.
    - `store` ****Store the credential, if applicable to the helper.
    - `erase` ****Remove a matching credential, if any, from the helper’s storage.
    
[...]

PR #491, opened 2022-03-17
Parse git credential configuration

kassio:
Related to: #490

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -54,7 +54,13 @@ func (s *ConfigSuite) TestUnmarshal(c *C) {
 [init]
 		defaultBranch = main
 [url "ssh://git@github.com/"]
-	insteadOf = https://github.com/
+		insteadOf = https://github.com/
+[credential]
+		helper = osxkeychain
+		helper = "!f() { echo 'password=multiple-helpers'; }; f"
+		helper = "/bin/get-my-git-password"
+[credential "https://kernel.org"]
+		username = foo
 `)
 
 	cfg := NewConfig()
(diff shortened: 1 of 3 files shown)


config/config_test.go
```diff
@@ -54,7 +54,13 @@ func (s *ConfigSuite) TestUnmarshal(c *C) {
...
 		defaultBranch = main
 [url "ssh://git@github.com/"]
-	insteadOf = https://github.com/
+		insteadOf = https://github.com/
+[credential]
+		helper = osxkeychain
+		helper = "!f() { echo 'password=multiple-helpers'; }; f"
+		helper = "/bin/get-my-git-password"
+[credential "https://kernel.org"]
+		username = foo
```
MxKaras, line 63: Say the config file did not have a default entry i.e. only this
```
[credential "https://kernel.org"]
		username = foo
```

What would be populated for `Credentials["default"]`? Should there be a default entry in this case?

config/config_test.go
```diff
@@ -89,6 +95,15 @@ func (s *ConfigSuite) TestUnmarshal(c *C) {
...
 	c.Assert(cfg.Init.DefaultBranch, Equals, "main")
+	c.Assert(cfg.Credentials["default"].Helper, DeepEquals, []string{
+		"osxkeychain",
+		"!f() { echo 'password=multiple-helpers'; }; f",
+		"/bin/get-my-git-password"})
+	c.Assert(cfg.Credentials["default"].Username, Equals, "")
+	c.Assert(cfg.Credentials["default"].UseHttpPath, Equals, false)
+	c.Assert(cfg.Credentials["https://kernel.org"].Helper, DeepEquals, []string{})
+	c.Assert(cfg.Credentials["https://kernel.org"].Username, Equals, "foo")
+	c.Assert(cfg.Credentials["https://kernel.org"].UseHttpPath, Equals, false)
```
pjbgf, line 106: Add a case in which `UseHttpPath` is `true`.

pjbgf (commented):
@kassio please rebase after addressing the few nits below. Then ping me for a review.

kassio:
Unfortunately, I won't be able to finish this. If anyone wants/can take over it I appreciate.

PR #507, opened 2022-03-31
go.mod: update kevinburke/ssh_config dependency

kevinburke:
Previously if you tried to read a HostName in a config file that
looked like this:

```
Host github
    HostName github.com        # This is the host for code review
```

DefaultUserSettings.Get("HostName") would return "github.com        ",
which I think is unintuitive and unexpected.

This behavior is fixed in v1.2 which would return "github.com" in the
above example.


6543 (approved):

pjbgf:
No longer relevant after https://github.com/go-git/go-git/commit/6629ba68b5b9df17999b470f37d7e4070b2bf885.

PR #514, opened 2022-04-11
Implement Ref Pruning on Fetch

martinmaly:
An implementation of post-fetch prune of refs
that no longer exist on the remote.
Support for boolean config value `remotes.*.prune`.
Does not yet support the `fetch.prune` config value.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -238,6 +238,7 @@ const (
 	initSection      = "init"
 	urlSection       = "url"
 	fetchKey         = "fetch"
+	pruneKey         = "prune"
 	urlKey           = "url"
 	bareKey          = "bare"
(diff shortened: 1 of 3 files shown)


efiacor:
@martinmaly Hi, we are looking to revive this commit. Could you assist with this?

martinmaly:
Sure. I'll rebase it onto current head and see if I can get the PR working again.

efiacor:
> Sure. I'll rebase it onto current head and see if I can get the PR working again.

Thanks for taking the time to reply.
Looks like a straight upgrade to latest go-git is working fine - https://github.com/nephio-project/porch/pull/46
We can prob close this so.

Issue #411, opened 2021-11-17
git-ssh: try multiple algorithms for known_hosts


franklouwers:
Hi,

When using git over ssh, it seems that if a host (eg: GitHub.com) has multiple keys (eg: an rsa and an ed25519 ones), but only one of those keys is listed in the `$HOME/.ssh/known_hosts`, and the one that's not listed has a higher preference (e.g. ed25519 over rsa), then the git clone operation will fail.

I feel it should try to be more intelligent and try to use the rsa one.

steffakasid:
I experienced the same. As workaround you could do the following: `ssh-keyscan github.com >> known_hosts`

mojotx:
I've been struggling for several days, trying to get go-git to work, and I kept thinking it had something to do with ssh-agent, or that I had missed some minor thing in my code. 

THIS was the problem, and running `ssh-keyscan github.com >> ~/.ssh/known_hosts` fixed it for me.

mojotx:
I did some digging, to see how easy this would be to fix, and I don't think this is an issue in [go-git](https://github.com/go-git/go-git)'s code

In [plumbing/transport/ssh/common.go](https://github.com/go-git/go-git/blob/f0b111ab70e4e90013658b0835929b2083902017/plumbing/transport/ssh/common.go#L158), in the `dial()` func, the code is calling ssh.NewClientConn(). 

```go
	c, chans, reqs, err := ssh.NewClientConn(conn, addr, config)
```

That function is actually defined in [golang/x/crypto/ssh](https://pkg.go.dev/golang.org/x/crypto/ssh#NewClientConn). I didn't follow the rabbit trail all the way down into the crypto code, but I suspect that's where it needs to be fixed. 

If I had the time, I would try writing a trivial Go program that makes a SSH connection to Github,, disabling the pseudo-tty allocation, which would be the equivalent of `ssh -T git@github.com`. 

Currently if you do that, and you have your SSH keys set up correctly, it prints something like this:

```
$ ssh -T git@github.com
Hi mojotx! You've successfully authenticated, but GitHub does not provide shell access.
```

If a trivial Go implementation of `ssh -T git@github.com` also gives you the same host-key error, then we know for sure that the problem is in the golang/x/crypto/ssh code. 

For what it's worth, I tried forking the go-git repo, and then updating the crypto module, to see if it's been fixed already, and it hasn't. 

```
$ go get -v -u golang.org/x/crypto
[...]

PR #517, opened 2022-04-13
Configure ssh.clientOptions using environment variables. Fixes go-git#411.

smveloso:
I stumbled upon https://github.com/go-git/go-git/issues/411 (which is actually caused by https://github.com/golang/go/issues/29286) and I did not like the idea of relying solely on editing the "known_hosts" file as a workaround. So I tried to see if I could set ssh client parameters in code.

Although a work in progress, I think this might be a worthy (very small) contribution.

I chose to do it with environment variables at first to minimize changes in existing code, but I also think this approach may be better suited for applications that use go-git in container environments (as opposed to a configuration file) and ones that deal with a limited set of remote repositories.

Maybe a more general approach might be reading a standard ssh client (ssh_config) file; the file format already allows setting different parameters for different remote hosts. However, in containers/kubernetes scenarios, creating a file and making it available as a volume seems to be too much, unless you really need many different "profiles" in your use case.

### DIFF
--- a/plumbing/transport/ssh/auth_method.go
+++ b/plumbing/transport/ssh/auth_method.go
@@ -56,12 +56,10 @@ func (a *KeyboardInteractive) String() string {
 }
 
 func (a *KeyboardInteractive) ClientConfig() (*ssh.ClientConfig, error) {
-	return a.SetHostKeyCallback(&ssh.ClientConfig{
-		User: a.User,
-		Auth: []ssh.AuthMethod{
-			a.Challenge,
-		},
-	})
+	clientConfig := DefaultSSHClientConfig
+	clientConfig.User = a.User
+	clientConfig.Auth = []ssh.AuthMethod{a.Challenge}
+	return a.SetHostKeyCallback(clientConfig)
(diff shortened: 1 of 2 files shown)


smveloso:
I think the other approaches would be:

1. Add a field to every struct that is used as argument in functions that might result in ssh operations So, for example, [CloneOptions](https://github.com/go-git/go-git/blob/master/options.go) would have an "SSHClientOptions" field.

2. Wait for golang's ssh library to implement auto-configuration (set "HostKeyAlgorithms" according to what is available in the known_hosts file(s)).

3. Implement auto-configuration directly in go-git. 

Option "3" unfortunately won't be able to leverage code from golang's ["x/crypto/ssh/knownhosts"](https://pkg.go.dev/golang.org/x/crypto@v0.0.0-20220411220226-7b82a4e95df4/ssh/knownhosts) package, because the relevant parts that process "known_hosts" files are not public.

mcuadros:
Fixed by https://github.com/go-git/go-git/pull/548

Issue #500, opened 2022-03-24
`worktree.Status()` not respecting .gitignore


acumino:
When we have multi nested directory and the `.gitignore` file is inside that. The `wroktree.Status()` is not respecting `patterns` inside `.gitignore` that can cause the git status to dirty.

## How to reproduce
Check this repo https://github.com/acumino/git-status. When we run `$ go run ./test.go` which is basically checking the status and returning an error if `status` is not clean.

```
wt, err := repo.Worktree()
	if err != nil {
		fmt.Printf("failed to get git worktree: %v", err)
	}
if stat, err := wt.Status(); err != nil {
		fmt.Printf("failed to get git status: %v", err)
	} else if !stat.IsClean() {
		fmt.Printf("%v\n%v", err, stat)
	}
```
It throws an error and ideally it should not.
```
?? pp/hh/hack1/configs/authwebhook.yaml
```

## Explanation

The issue is happening because `append()` is being used for the `slice` in recursion at line `64`.
https://github.com/go-git/go-git/blob/c785af3f4559ebac52c42f12d17cb118aac383ad/plumbing/format/gitignore/dir.go#L49-L76

The passed value of `fs` is being changed with appends and is being used as `domain` here https://github.com/go-git/go-git/blob/c785af3f4559ebac52c42f12d17cb118aac383ad/plumbing/format/gitignore/pattern.go#L42 

So when this `fs` is updated next time with append it is causing  overwrite to previously written `domain` values leading to previously added `pattern`'s domain to overwritten. So git status basically at the end store wrong pattern to ignore, leading to dirty git status.

An explanation of how slices are passed in function and are used internally `golang` https://stackoverflow.com/questions/39993688/are-slices-passed-by-value.

## Solution 
Instead to passing `append(path, fi.Name())` in `ReadPatterns()` every times a new copy of slice should be passed.

sourcec0de:
I was trying to follow along here because I had a similar theory.
https://go.dev/play/p/pjtX_coRiQ8
```go
package main

import "fmt"

func recurse(a []string) []string {
	if len(a) == 3 {
		return a
	}

	return recurse(append(a, "test"))
}

func main() {
	a := []string{"a"}
	b := recurse(a)
	fmt.Println(a, b)
}
```

My assumption was that because Go slices are pass-by-value that `a` and `b` would have the same contents.
They do not in this case. This program evaluated to.
```
[a] [a test test]

Program exited.
```


[...]

PR #522, opened 2022-04-23
Copy slice before appending

acumino:
Fixes: #500

### DIFF
--- a/plumbing/format/gitignore/dir.go
+++ b/plumbing/format/gitignore/dir.go
@@ -61,7 +61,9 @@ func ReadPatterns(fs billy.Filesystem, path []string) (ps []Pattern, err error) 
 	for _, fi := range fis {
 		if fi.IsDir() && fi.Name() != gitDir {
 			var subps []Pattern
-			subps, err = ReadPatterns(fs, append(path, fi.Name()))
+			tempPath := make([]string, len(path))
+			copy(tempPath, path)
+			subps, err = ReadPatterns(fs, append(tempPath, fi.Name()))
 			if err != nil {
 				return
 			}


TopherIsSwell:
I'm not associated with this repository or this MR, but as a third-party, I can confirm that this patch improved my issue.

With upstream go-git in a repo with a 350M folder that was listed in the .gitignore,  worktree.Status() returned in ~ 1.6s at best.
With this patch, in the same situation, worktree.Status() returns in <500ms at worst.

mcuadros:
can you provide a test?

acumino:
> can you provide a test?

There is no new functionality, so no need for a new unit test.

plumbing/format/gitignore/dir.go
```diff
@@ -61,7 +61,9 @@ func ReadPatterns(fs billy.Filesystem, path []string) (ps []Pattern, err error)
 	for _, fi := range fis {
 		if fi.IsDir() && fi.Name() != gitDir {
 			var subps []Pattern
-			subps, err = ReadPatterns(fs, append(path, fi.Name()))
+			tempPath := make([]string, len(path))
```
easterf, line 64: A simpler fix is `subps, err = ReadPatterns(fs, append(path[0:len(path):len(path)], fi.Name()))`

plumbing/format/gitignore/dir.go
```diff
@@ -61,7 +61,9 @@ func ReadPatterns(fs billy.Filesystem, path []string) (ps []Pattern, err error)
...
+			copy(tempPath, path)
+			subps, err = ReadPatterns(fs, append(tempPath, fi.Name()))
```
emcfarlane, line 66: Think it would be cleaner to have the copy done by ParsePattern(), the object that takes ownership of the slice. Or in the readIgnoreFile func to share ownership. Here you would still get bugs if calling ReadPatterns incorrectly, e.g:

```go
path := []string{"a", "b"}
ps, _ := ReadPatterns(fs, path)
path[1] = "change" // ps now corrupted
```

TopherIsSwell:
Is there any interest in merging this or another fix upstream? This is a matter of compatibility with canonical git functionality. This issue is literally pushing me out of the go ecosystem and back into C, because go-git is wrong (with regard to .gitignore handling) and libgit2 isn't. I don't think that's the world that anyone wants to live in.

So I'm wondering if there's any interest in helping prevent my devolution into madness by giving this another look. Whichever of the three patterns in this conversation is preferable is probably fine (I've only tested the one in the commit), but it seems like we just need a decision maker to call which one to merge in.

Please let me know if I can help.

Thanks,
Topher

AriehSchneier:
@acumino a test is not only needed for new functionality, a test is useful for fixes like this for multiple reasons include:
- ensure that in future there is no regression (ie someone can't cause the same bug to come up later)
- prove that there was a bug and that it is now fixed (the test should fail without the fix, but pass with the fix)

PR #558, opened 2022-08-01
Search for object in git alternates

### LINKED ISSUE #433: Cloning a repository which uses git alternates over http transport gives "object not found" error
I tried debugging it and found that go-git was not looking for objects in alternates repositories. I patched it by looking at the EncodedObject [implementation](https://github.com/go-git/go-git/blob/b4368b2a2ca4103b1ff4e37c34a963127342747e/storage/filesystem/object.go#L328)

This is my [patch](https://github.com/gitopia/go-git/commit/3301af33c1f79dfea08c4cc8d137eb43781fa8bd)

Let me know if this is the right resolution

faza:
Fixes #433


mcuadros:

PR #560, opened 2022-08-03
Remove new line character from hash

### LINKED ISSUE #559: Git Hash Length - 40 characters vs. 41
Not sure per se what has an error but thought I would start here with my findings. I had the understanding and also from looking around I'm getting the confirmation that the length of a git hash should be 40 characters.

However, while troubleshooting an issue where all commit hashes were being reported to be invalid by the `IsHash` function in the `plumbing` package, I noticed my commit hashes were 41 characters long.

```
➜  git:(watch_branch_heads) ✗ HASH=$(echo "testing something random" | git hash-object --stdin)
➜  git:(watch_branch_heads) ✗ echo ${HASH}                                             
01d7507b634a3a85c32de659732c7b037c2cb9b1
➜  git:(watch_branch_heads) ✗ echo ${HASH} | wc -c
41
➜  git:(watch_branch_heads) ✗
```
I have further confirmed this with the length of the commit at the `HEAD` of the repository and get the same length.

```
➜  git:(watch_branch_heads) ✗ wc -c .git/refs/heads/master 
41 .git/refs/heads/master
➜  git:(watch_branch_heads) ✗ 
```

What are you all seeing and what should be the best way around this? We could accept both values or ...

https://github.com/go-git/go-git/blob/bc1f419cebcf7505db31149fa459e9e3f8260e00/plumbing/hash.go#L76-L83

OchiengEd:
This commit removes a new line character at the end of the file before
checking if it is a valid hash

Signed-off-by: Edmund Ochieng <ochienged@gmail.com>

Closes #559

### DIFF
--- a/plumbing/hash.go
+++ b/plumbing/hash.go
@@ -7,6 +7,7 @@ import (
 	"hash"
 	"sort"
 	"strconv"
+	"strings"
 )
 
 // Hash SHA1 hashed content
@@ -74,6 +75,10 @@ func (p HashSlice) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
 
 // IsHash returns true if the given string is a valid hash.
 func IsHash(s string) bool {
+	if strings.Contains(s, "\n") {
+		s = strings.Replace(s, "\n", "", -1)
+	}
+
 	if len(s) != 40 {
 		return false
 	}


mcuadros:

Issue #572, opened 2022-08-24
Missing files when iterating with go-git when repository is cloned with filter flag.


Preet-786:
We have found an issue when repository is cloned with flag ` --filter=blob:limit=<size>m `. Some files (which are not filtered/omitted) were missing while iterating files with go-git API. Descriptive information is below. 

consider this example repository: https://github.com/Accel-Byte/exp.git ,
there are total 18 files, 3 of them is more than 1mb 

1. [demo/ODEIntroduction.ipynb](https://github.com/Accel-Byte/exp/blob/main/demo/ODEIntroduction.ipynb)
2. [demo2/ODEIntroduction.ipynb](https://github.com/Accel-Byte/exp/blob/main/demo2/ODEIntroduction.ipynb)
3. [demo3/ODEIntroduction.ipynb](https://github.com/Accel-Byte/exp/blob/main/demo3/ODEIntroduction.ipynb)

The code to reproduce the issue is [main.go](https://gist.github.com/Accel-Byte/62535ca509573bf400bdf9884d274c4e).

**Normal behaviour:** 
This is working as expected
```
$ git clone https://github.com/Accel-Byte/exp.git  --no-checkout
$ go run main.go ./exp
1, blob 88b5e7eed5204583fcd1f4e608a71d80a62eae71, 1mbcheck.txt
2, blob d645695673349e3947e8e5ae42332d0ac3164cd7, demo/LICENSE
…
18, blob f8b9a1acca2ae4e73edcd8c563c5bc4ae73456d1, top_pull.txt
TOTAL FILES 18
```
This will show us the result of total 18 files which is expected behaviour.

**Probematic Behaviour:**
Now remove exp repo, and clone again with filter of 1mb
```
$ git clone https://github.com/Accel-Byte/exp.git --filter=blob:limit=1m --no-checkout 
$ go run main.go ./exp
1, blob 88b5e7eed5204583fcd1f4e608a71d80a62eae71, 1mbcheck.txt
2, blob d645695673349e3947e8e5ae42332d0ac3164cd7, demo/LICENSE
TOTAL FILES 2
```
This lists only 2 files, which is incorrect. We have excluded only 3 files which are greater than 1mb. But iterating with go-git will show only 2 files.

The problem lies [here](https://github.com/go-git/go-git/blob/master/plumbing/object/file.go#L100), as we have cloned with blob filter the git database won't have the corresponding data and hence `GetBlob` method fails. This first failure while iteration stops  the whole listing procedure.
[...]

PR #573, opened 2022-08-24
Update file.go

Preet-786:
plumbing : object/file.go,  Logging for missing blob and Returning empty blob in case of GetBlob error.

Fix for https://github.com/go-git/go-git/issues/572


Accel-Byte:
Can anyone merge this pr? this is useful for partially cloned repos. (https://git-scm.com/docs/partial-clone)

db47h:
With partial clones (i.e. `--filter=blob:none`), git just downloads missing blobs as needed.  This PR just returns empty blobs. A proper fix for #572 would be to implement filter support (at least for blobs), including a CheckoutContext() since a checkout may end up downloading GBs of blobs.

Issue #395, opened 2021-10-16
support global config (i.e. $HOME/.gitconfig)


davidalpert:
Per the `config` section of the [Supported Capabilities](https://github.com/go-git/go-git/blob/master/COMPATIBILITY.md#supported-capabilities):

Feature | Status | Notes
-- | -- | --
config | ✔ | Reading and modifying per-repository configuration (.git/config) is supported. Global configuration ($HOME/.gitconfig) is not.

This can cause some unexpected behaviour for people who have certain settings configured globally (i.e. in `$HOME/.gitconfig`).

I propose enhancing `go-git` to respect globally configured git settings

davidalpert:
it looks like this may have been partially implemented with the `ConfigScoped` method: https://github.com/go-git/go-git/blob/4ec1753b4e9324d455d3b55060020ce324e6ced2/repository.go#L489-L492

davidalpert:
After some more research I suspect this may point to a larger issue with the current semantics of the `v5` config subsystem that may even be classified as a bug.

I see that some notion of a read-only `Merged` config was added in #20 and enhanced in #31 and apparently released in `v5.0.0` but reverted in `v5.1.0` by @mcuadros in #75 in which:
- c80979499a2177f3e172b6d89dabaa0bf64e81d1 reverted #20
- 936faa404ef02c88bde6a62663a6e1a4831c1fb3 reverted #31 
- `_examples/config/main.go` was removed
- the [current implementation](https://github.com/go-git/go-git/pull/75/files#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R40-R48) (as of `v5.4.2`) of config scopes added

While #75 greatly improved the ability to read git configuration from different scopes the current implementation (as of `v5.4.2`) still diverges from what is expected behaviour.
- this issue where I was tripped up because reading config doesn't merge values in the way that I expected
- #508
- #110

I wonder @mcuadros (as the author of #75) if you might be willing to entertain another overhaul of the config subsystem to address all three of these open config issues and better align with the semantics of the `git config` implementation? If we align on direction here I would be happy to contribute.

From the [git-config docs](https://git-scm.com/docs/git-config/2.37.2) (as of git 2.37.2):

    When reading, the values are read from the system, global and repository local configuration files by default, and options --system, --global, --local, --worktree and --file <filename> can be used to tell the command to read from only that location (see FILES).

[...]

PR #574, opened 2022-08-27
tests for #395

davidalpert:
These test cases demonstrate the behaviour I expect for Repository.Config
and Repository.ConfigScoped, based on the semantics of the git-config
docs at git 2.37.2: https://git-scm.com/docs/git-config/2.37.2

    When reading, the values are read from the system, global and
    repository local configuration files by default, and options
    --system, --global, --local, --worktree and --file <filename>
    can be used to tell the command to read from only that location
    (see FILES).

By name, I assume that:
- `repo.Config()` is the equivalent of `git config` to load and merge system, global, and local configuration by default; while
- `repo.ConfigScoped(scope config.Scope)` is the equivalent of `git config --{scope}` reading config only from that location.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -12,6 +12,7 @@ import (
 	"sort"
 	"strconv"
 
+	"github.com/go-git/go-billy/v5"
 	"github.com/go-git/go-billy/v5/osfs"
 	"github.com/go-git/go-git/v5/internal/url"
 	format "github.com/go-git/go-git/v5/plumbing/format/config"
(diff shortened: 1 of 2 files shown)


davidalpert:
NOTE: this branch is expected to have failing tests which illustrate my expectations wrt #395 

the commit at the HEAD of https://github.com/davidalpert/go-git/tree/395-fix contains my proposed fix, although I am waiting to hear from @mcuadros and others if this can be merged now as a bug fix or constitutes a breaking change and needs to wait for a future major release.

davidalpert:
rebased again on top of `upstream/master`

davidalpert:
rebased again on top of `upstream/master` and have tests failing to demonstrate my expectation based on the git config documentation.

PR #579, opened 2022-09-09
fix: ignore git meta dir or file when add

dpc11:
In most situations `.git` is directory, but in submodules `.git` is a file containing path to root repository.
Both of them should be ignored when `add`.


mcuadros:

DPC11:
Add a unit test for submodule added and commit, which will not pass before this pull request.

Go-git is a great git library for golang, but still need more work on submodule to completely performed as well as in git shell :]

PR #584, opened 2022-09-24
feat(repo): Add ability to get latest tag from commit

### LINKED ISSUE #451: "git describe --tags"
There is no way to call "git describe --tags" using it as module from another dependent.

MisterMX:
This adds a new function `LatestTag` to `Repository` that allows getting the latest tag in the commit history from a given commit. It allows achieving something that `git describe --tag` does.

Fixes #451

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -750,21 +750,20 @@ func (r *Repository) buildTagSignature(tag *object.Tag, signKey *openpgp.Entity)
 // If you want to check to see if the tag is an annotated tag, you can call
 // TagObject on the hash of the reference in ForEach:
 //
-//   ref, err := r.Tag("v0.1.0")
-//   if err != nil {
-//     // Handle error
-//   }
+//	ref, err := r.Tag("v0.1.0")
+//	if err != nil {
+//	  // Handle error
+//	}
 //
-//   obj, err := r.TagObject(ref.Hash())
-//   switch err {
-//   case nil:
-//     // Tag object present
-//   case plumbing.ErrObjectNotFound:
-//     // Not a tag object
-//   default:
-//     // Some other error
-//   }
-//
+//	obj, err := r.TagObject(ref.Hash())
+//	switch err {
+//	case nil:
+//	  // Tag object present
+//	case plumbing.ErrObjectNotFound:
+//	  // Not a tag object
+//	default:
+//	  // Some other error
+//	}
 func (r *Repository) Tag(name string) (*plumbing.Reference, error) {
 	ref, err := r.Reference(plumbing.ReferenceName(path.Join("refs", "tags", name)), false)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


mcuadros:
can you provide tests? thanks

romainreignier:
Using this branch with this example produces an infinite loop in `buildTagRefMap`:

```go
package main

import (
	"fmt"
	"os"

	"github.com/go-git/go-git/v5"
	. "github.com/go-git/go-git/v5/_examples"
)

// Basic example of how to list tags.
func main() {
	CheckArgs("<path>")
	path := os.Args[1]

	// We instantiate a new repository targeting the given path (the .git folder)
	r, err := git.PlainOpen(path)
	CheckIfError(err)

	// Get latest tag
	Info("git describe --tags")
	ref, err := r.Head()
	CheckIfError(err)
	tag, distance, err := r.LatestTag(ref.Hash())
	CheckIfError(err)
	fmt.Printf("%s-%d-%s", ref.Name().Short(), distance, tag.Hash().String())
}
```

romainreignier:
What I don't understand is that the `Next()` does not seem to work.
If I add this line:
```diff
diff --git a/repository.go b/repository.go
index a70cc49..50ecf98 100644
--- a/repository.go
+++ b/repository.go
@@ -820,6 +820,7 @@ func (r *Repository) buildTagRefMap() (map[plumbing.Hash]*plumbing.Reference, er
        }
        tags := map[plumbing.Hash]*plumbing.Reference{}
        for t, err := iter.Next(); err == nil; {
+               fmt.Println(t)
                to, err := r.TagObject(t.Hash())
                if err == plumbing.ErrObjectNotFound {
                        // t is a lighweight tag
```

I have an infinite log output with always the same tag (ran on this repo):
```
6f43e8933ba3c04072d5d104acc6118aac3e52ee refs/tags/v1.0.0
6f43e8933ba3c04072d5d104acc6118aac3e52ee refs/tags/v1.0.0
6f43e8933ba3c04072d5d104acc6118aac3e52ee refs/tags/v1.0.0
6f43e8933ba3c04072d5d104acc6118aac3e52ee refs/tags/v1.0.0
```
So `t` is never updated in the for loop. This seems quite strange.

arnarg:
It doesn't have a post statement.

It needs to be like this:
```go
for t, err := iter.Next(); err == nil; t, err = iter.Next() {
  // ...
}
```

repository.go
```diff
@@ -779,6 +778,62 @@ func (r *Repository) Tag(name string) (*plumbing.Reference, error) {
...
+		return nil, 0, err
+	}
+
+	tags, err := r.buildTagRefMap()
+	if err != nil {
+		return nil, 0, err
+	}
+
+	distance := 0
+	for obj, err := log.Next(); err == nil; {
```
arnarg, line 807:
```suggestion
	for obj, err := log.Next(); err == nil; obj, err = log.Next() {
```

repository.go
```diff
@@ -779,6 +778,62 @@ func (r *Repository) Tag(name string) (*plumbing.Reference, error) {
...
+	return nil, 0, plumbing.ErrReferenceNotFound
+}
+
+func (r *Repository) buildTagRefMap() (map[plumbing.Hash]*plumbing.Reference, error) {
+	iter, err := r.Tags()
+	if err != nil {
+		return nil, err
+	}
+	tags := map[plumbing.Hash]*plumbing.Reference{}
+	for t, err := iter.Next(); err == nil; {
```
arnarg, line 822:
```suggestion
	for t, err := iter.Next(); err == nil; t, err = iter.Next() {
```

arnarg (commented):
Both the for loops need a post statement.

jufemaiz:
QQ was there a reason this was closed without being merged? We currently have the following:

```bash
git describe --abbrev=9 --dirty --tag --long --always
```

To build a version detail for tagging. We'd love to have something like this available here and ditch our use of shell out commands.

MisterMX:
It was closed because we no longer needed it and I don't have to time work on it. Anyways, feel free to open a separate MR with my changes.

PR #589, opened 2022-09-29
object not found when trying to pull a repository cloned with Depth: 1

### LINKED ISSUE #305: `object not found` when trying to pull a repository cloned with `Depth: 1`
Hi,

Cloning:
```go
	URL: gitURL,
	Depth: 1,
	})
```

Pulling:
```go
// Open repo
if err != nil {
}
// Get worktree
tree, err := r.Worktree()
if err != nil {
}

// Pull
	// do nothing
} else if err != nil {
}
```

rotty3000:
fixes #305

In fact setting `Depth` to any value which results in the clone being partial will lead to the same error.

The iterator used in the fastforward check wasn't prepared for the case when a commit might have a missing parent (as in shallow clones). Handling that case solved the problem of pulling on shallow clones.

Signed-off-by: Raymond Augé <raymond.auge@liferay.com>

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -859,6 +859,15 @@ func getHavesFromRef(
 	toVisit := maxHavesToVisitPerRef
 	return walker.ForEach(func(c *object.Commit) error {
 		haves[c.Hash] = true
+
+		if s, _ := s.Shallow(); len(s) > 0 {
+			for _, sh := range s {
+				if sh == c.Hash {
+					return storer.ErrStop
+				}
+			}
+		}
+
 		toVisit--
 		// If toVisit starts out at 0 (indicating there is no
 		// max), then it will be negative here and we won't stop


rotty3000:
I think I've actually found a better solution to this problem. It seems that the concept of `haves` v.s `wants` is not balanced in that shallowness is only checked in the `wants` side of the checks but not in the `haves` side.

This iteration should check if `c.Hash` is a shallow reference, something like this:

```patch
@@ -691,6 +691,15 @@ func getHavesFromRef(
 	toVisit := maxHavesToVisitPerRef
 	return walker.ForEach(func(c *object.Commit) error {
 		haves[c.Hash] = true
+
+		if s, _ := s.Shallow(); len(s) > 0 {
+			for _, sh := range s {
+				if sh == c.Hash {
+					return storer.ErrStop
+				}
+			}
+		}
+
 		toVisit--
 		// If toVisit starts out at 0 (indicating there is no
 		// max), then it will be negative here and we won't stop
```

yarinm:
Who can review this and merge? We're dealing with similar issues and we believe this can solve our issues.

@rotty3000

pjbgf:
Do you mind adding a test to make it explicit what issue is being resolved?
I tested this against the example given on #305 and it did not seem to fix the issue.

rotty3000:
sorry, accidentally closed.. Please see the updated solution which is the one we've been using for the past few months.

If someone wants to contribute a test to the PR in order to make it complete solution I wouldn't mind.

AriehSchneier:
Fixed by PR #778

pjbgf:
Closing this in favour of the already merged #778.

PR #595, opened 2022-10-10
Fix type error in example

sholsapp:
This change passes the type `plumbing.ReferenceName` to the `plumbing.NewHashReference`.

I stumbled upon this error while using the example. If this isn't
correct, I'm curious what I should do instead. It appears the type
`plumbing.ReferenceName` is just an type alias to `string`.


erizocosmico:
It's fine to pass a string literal to `plumbing.NewHashReference` as literals are untyped. If you passed `string("some str")` instead of `"some str"` it would not compile. So most likely the error you encountered has to do with something else. Can you open an issue with the error you encountered so we can look into it?

PR #598, opened 2022-10-12
plumbing: gitattributes, Avoid index out of range 

To1ne:
When a path is deeper than the single asterisk pattern the code would
crash with a "index out of range".

This change checks the length of the remaining pattern before it
references an element of that slice.

With a single trailing asterisk paths deeper than the pattern should not
get the attributes.

For example with the following `.gitattributes` file:

    thirdparty/* linguist-vendored

This is how git handles it:

    $ git check-attr --all thirdparty/README.md

    thirdparty/README.md: diff: markdown
    thirdparty/README.md: linguist-vendored: set

    $ git check-attr --all thirdparty/package/README.md

    thirdparty/package/README.md: diff: markdown

### DIFF
--- a/plumbing/format/gitattributes/pattern.go
+++ b/plumbing/format/gitattributes/pattern.go
@@ -52,6 +52,11 @@ func (p *pattern) Match(path []string) bool {
 	var match, doublestar bool
 	var err error
 	for _, part := range path {
+		// path is deeper than pattern
+		if len(pattern) == 0 {
+			return false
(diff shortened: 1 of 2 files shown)


To1ne:
@mcuadros Would you consider this small bug fix?

erizocosmico (approved):

erizocosmico:
Good catch, thank you for the contribution!

Issue #473, opened 2022-02-15
Trouble cloning submodules when referencing different github endpoints


nBerg:
Hi, 

I'm running into problem with git submodules. 

I have a repository hosted in an enterprise github server. That repository references some submodules from the public github.com. These are open source and public repositories.

When trying to clone this repository with the submodules, I receive a error: `error: authentication required`
If I set `RecurseSubmodules` to 0 it succeeds.

```
package main

import (
	"fmt"
	"os"

	"github.com/go-git/go-git/v5"
	. "github.com/go-git/go-git/v5/_examples"
   "github.com/go-git/go-git/v5/plumbing/transport/http"
)

func main() {
  url := "https://enterprise.github.url/org/repo"
  directory := "/tmp/foo"
  token := "token for authorized user on enterprise github"

  // Clone the given repository to the given directory
	Info("git clone %s %s", url, directory)

	r, err := git.PlainClone(directory, false, &git.CloneOptions{
		// The intended use of a GitHub personal access token is in replace of your password
		// because access tokens can easily be revoked.
		// https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/
		Auth: &http.BasicAuth{
			Username: "username", // yes, this can be anything except an empty string
			Password: token,
		},
		URL:      url,
		Progress: os.Stdout,

        // If set to 0 this succeeds, if set to 1 (or higher) this fails with authentication error
        RecurseSubmodules: 1,
	})
	CheckIfError(err)

	// ... retrieving the branch being pointed by HEAD
	ref, err := r.Head()
	CheckIfError(err)
	// ... retrieving the commit object
	commit, err := r.CommitObject(ref.Hash())
	CheckIfError(err)

	fmt.Println(commit)
}
```

Sudipta-Priyadarshani1:
Who is working on this issue? Do we have any prioritisation?

Sudipta-Priyadarshani1:
tanya.shanker@ibm.com - I know, you were looking into this issue. Its been more than 3months but we don't have a permanent fix for this. We have a customer facing GA in June mid as our E2E integration testing has started, code supposed to be freezed. Need help asap
^^ pradeepb@in.ibm.com

pjbgf:
A similar issue was answered [here](https://github.com/go-git/go-git/issues/943#issuecomment-1841410779). If only the main repository is private, and the submodules are public, the auth for submodules can be disabled via:

```
	s, err := w.Submodules()
	if err != nil {
		return err
	}
[...]

PR #600, opened 2022-10-21
Add retry logic for git clone

shahulsonhal:
Git clone was failing for public repository if a wrong authentication was provided which ideally should not happen because public repositories does not require authentication.

This change port the same behavior of git CLI that is always try cloning the repository without authentication, if it fails with 401 retry with authentication.

fixes https://github.com/go-git/go-git/issues/473

Signed-off-by: shahulsonhal <shahulsonhal@gmail.com>

### DIFF
--- a/plumbing/transport/http/common.go
+++ b/plumbing/transport/http/common.go
@@ -44,9 +44,8 @@ func advertisedReferences(ctx context.Context, s *session, serviceName string) (
 		return nil, err
 	}
 
-	s.ApplyAuthToRequest(req)
 	applyHeadersToRequest(req, nil, s.endpoint.Host, serviceName)
-	res, err := s.client.Do(req.WithContext(ctx))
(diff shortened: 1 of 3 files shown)


shahulsonhal:
@mcuadros Is there any way to know if the repo we are cloning is a submodule or not? Does go-git already have a method to check that?

onee-only:
> @mcuadros Is there any way to know if the repo we are cloning is a submodule or not? Does go-git already have a method to check that?

Calling `(*Worktree).Submodules()` should let you know if repository contains submodules or not.

Seems like when running git clone, and if option is not `NoRecurseSubmodules`, go-git calls method above to retrieve all the submodules. And for each submodule, go-git calls the method below to fetch submodule's data. I think we can add retry logic here.

https://github.com/go-git/go-git/blob/c7feada87363f96c224214ba309efd5cf5a886fc/submodule.go#L242-L250

PR #602, opened 2022-10-27
Parse optimisations

pjbgf:
[Recently we worked](https://github.com/fluxcd/go-git/pull/5) on some optimisations that could benefit upstream.

The changes improved benchmarks results across number of executions (`+60%`), time (`-4300000 ns/op`), memory (`-23200000 B/op`) and allocations (`-800 allocs/op`).

#### Parse changes Benchmark
```
name       old time/op    new time/op    delta
Parser-16    11.9ms ± 1%     7.4ms ± 1%  -37.77%  (p=0.008 n=5+5)

name       old alloc/op   new alloc/op   delta
Parser-16    30.7MB ± 0%     4.8MB ± 2%  -84.42%  (p=0.008 n=5+5)

name       old allocs/op  new allocs/op  delta
Parser-16     4.19k ± 0%     3.48k ± 0%  -16.84%  (p=0.008 n=5+5)
```

Comparing the memory profile of both executions, shows that the gains are primarily in the zlib writing code.
image

#### Reference.String() changes Benchmark
For bench mark delta on the ReferenceName optimisations:
```
name                        old time/op    new time/op    delta
ReferenceStringSymbolic-16     140ns ± 4%      40ns ± 9%  -71.19%  (p=0.008 n=5+5)
ReferenceStringHash-16         174ns ±14%      85ns ± 4%  -51.13%  (p=0.008 n=5+5)
ReferenceStringInvalid-16     48.9ns ± 2%     1.5ns ± 3%  -96.96%  (p=0.008 n=5+5)

name                        old alloc/op   new alloc/op   delta
ReferenceStringSymbolic-16     88.0B ± 0%     32.0B ± 0%  -63.64%  (p=0.008 n=5+5)
ReferenceStringHash-16          176B ± 0%      144B ± 0%  -18.18%  (p=0.008 n=5+5)
ReferenceStringInvalid-16      0.00B          0.00B          ~     (all equal)

name                        old allocs/op  new allocs/op  delta
ReferenceStringSymbolic-16      4.00 ± 0%      1.00 ± 0%  -75.00%  (p=0.008 n=5+5)
ReferenceStringHash-16          5.00 ± 0%      3.00 ± 0%  -40.00%  (p=0.008 n=5+5)
ReferenceStringInvalid-16       0.00           0.00          ~     (all equal)
```

PR #608, opened 2022-11-07
Optimise zlib reader and consolidate sync.Pools

pjbgf:
Expands on the optimisations made on https://github.com/go-git/go-git/pull/602 and ensures that zlib reader does not need to recreate a deflate dictionary at every use.

The use of sync pools was consolidated into a new `sync` utils package, for improved code structure.
```
name       old time/op    new time/op    delta
Parser-16    7.51ms ± 3%    7.71ms ± 6%     ~     (p=0.222 n=5+5)

name       old alloc/op   new alloc/op   delta
Parser-16    4.65MB ± 3%    1.90MB ± 3%  -59.06%  (p=0.008 n=5+5)

name       old allocs/op  new allocs/op  delta
Parser-16     3.48k ± 0%     3.32k ± 0%   -4.57%  (p=0.016 n=5+4)
```

:warning: Requires rebase after https://github.com/go-git/go-git/pull/602 is merged.


erizocosmico:
Thank you for the contribution!

PR #613, opened 2022-11-16
Allow unsupported `multi_ack` capability

pjbgf:
Azure DevOps requires capabilities `multi_ack` / `multi_ack_detailed`, which are not fully implemented and by default are included in `transport.UnsupportedCapabilities`.

The initial clone operations require a full download of the repository, and therefore those unsupported capabilities are not as crucial, so by removing them from that list allows for the first clone to work successfully.

Additional fetches will yield issues, therefore to support that repository users have to work from a clean clone until those capabilities are fully supported. Commits and pushes back into the repository have also been tested and work fine.

This PR adds an example for cloning Azure DevOps repositories.

### DIFF
--- a/_examples/azure_devops/main.go
+++ b/_examples/azure_devops/main.go
@@ -1,0 +1,56 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	git "github.com/go-git/go-git/v5"
+	. "github.com/go-git/go-git/v5/_examples"
+	"github.com/go-git/go-git/v5/plumbing/protocol/packp/capability"
+	"github.com/go-git/go-git/v5/plumbing/transport"
+	"github.com/go-git/go-git/v5/plumbing/transport/http"
+)
+
+func main() {
+	CheckArgs("<url>", "<directory>", "<azuredevops_username>", "<azuredevops_password>")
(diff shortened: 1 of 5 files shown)


codablock:
> Additional fetches will yield issues

@pjbgf Can you explain in more detail how this will manifest? Will a fetch result in a detectable error or will it silently corrupt the repo?

I'm currently trying to fix Azure DevOps support and my though was that I would modify `transport.UnsupportedCapabilities` according to your example and then react on `Fetch()` errors with a fresh re-clone of the whole repository. This would however require that errors are always properly reported.

pjbgf:
@codablock It won't corrupt the repository, but any follow-up fetch operations will fail - I don't recall the specific error now.

The most optimal approach I found for this without making anything Azure DevOps specific was to use ls-remote to detect changes, and if so do a new shallow clone.

codablock:
@pjbgf Thanks for the response, that reassures that my approach should work.

I'm also using ls-remote to pre-detect if a fetch is required (for different reasons however, but it played into my cards now). I however decided to check for the existence of `multi_ack` in the returned `err.String()` of `Fetch()` and then assume that https://github.com/go-git/go-git/pull/613/files#diff-7d88eba281e953fbedac05d3dc4f71400d8621be2a110964b17a44b1b4c3424bR59 kicked in. In that case, I perform a fresh clone.

PR #618, opened 2022-11-24
sha1: Add collision resistent implementation

pjbgf:
Implement the same SHA1 collision resistent algorithm used by both the Git CLI and libgit2.

Only commits with input that match the unavoidable bit conditions will be further processed, which will result in different hashes.
Which is the same behaviour experienced in the Git CLI and Libgit2.

Users can override the hash algorithm used with:

`hash.RegisterHash(crypto.SHA1, sha1.New)`

xref links:
- https://github.com/libgit2/libgit2/commit/2dfd1294f7a694bfa9e864a9489ae3cb318a5ed0
- https://github.com/git/git/commit/28dc98e343ca4eb370a29ceec4c19beac9b5c01e


pjbgf:
The build failure is related to network intermittency and unrelated to the PR changes.

PR #623, opened 2022-11-30
CommitOptions: AllowEmptyCommits, return error instead of creating empty commits

pjbgf:
BuildTree now returns an `ErrEmptyCommit` error, when there are no changes to be committed, which can be opted-out via `CommitOptions.AllowEmptyCommits`.

This is a breaking change which enables applications to detect when empty commits are to be created.

Some instances in which this can occur is when the fs (e.g. `billy/osfs`) make changes to the underlying files, causing a conflict between what the previous Git worktree state was, and the current state. Changes to the fs implementations are orthogonal to this, and will be dealt with separately.

The new behaviour aligns with the Git CLI, in which empty commits returns the error message: `nothing to commit, working tree clean`.

### DIFF
--- a/options.go
+++ b/options.go
@@ -458,6 +458,10 @@ type CommitOptions struct {
 	// All automatically stage files that have been modified and deleted, but
 	// new files you have not told Git about are not affected.
 	All bool
+	// AllowEmptyCommits enable empty commits to be created. An empty commit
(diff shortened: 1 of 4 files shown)


mcuadros:
Ok, but make a way to opt out of this feature by adding `CommitOptions.AllowEmptyCommits` or something like this.

Issue #624, opened 2022-12-02
Error building without CGO


dornimaug:
Trying to built a program using go-git v5.5.0 with CGO_ENABLED set to 0 results in an error:
```
 	imports github.com/go-git/go-git/v5
 	imports github.com/go-git/go-git/v5/config
 	imports github.com/go-git/go-git/v5/plumbing
 	imports github.com/go-git/go-git/v5/plumbing/hash
 	imports github.com/pjbgf/sha1cd/cgo: build constraints exclude all Go files in /home/jenkins/go/pkg/mod/github.com/pjbgf/sha1cd@v0.2.0/cgo
```

This seems to have been added/changed in #618.

Using 
```
hash.RegisterHash(crypto.SHA1, sha1.New)
```
as suggested in the PR, does not fix the problem.

pjbgf:
@dornimaug thanks for reporting this. I will take a look at it later on today.

In the mean time, you may want to replace the go-git version with a version that preceded the change:
```
replace github.com/go-git/go-git/v5 => github.com/go-git/go-git/v5 v5.4.3-0.20221117033540-c798d4a42004
```

eddycharly:
FWIW, same issue here:
```
package github.com/kyverno/kyverno/cmd/cli/kubectl-kyverno
	imports github.com/kyverno/kyverno/cmd/cli/kubectl-kyverno/test
	imports github.com/go-git/go-git/v5
	imports github.com/go-git/go-git/v5/config
	imports github.com/go-git/go-git/v5/plumbing
	imports github.com/go-git/go-git/v5/plumbing/hash
	imports github.com/pjbgf/sha1cd/cgo: build constraints exclude all Go files in /home/runner/go/pkg/mod/github.com/pjbgf/sha1cd@v0.2.0/cgo
```

sithembiso:
@pjbgf is this fix related to this:
```
# github.com/pjbgf/sha1cd/cgo
vendor/github.com/pjbgf/sha1cd/cgo/sha1.go:3:11: fatal error: 'lib/sha1.h' file not found
 #include <lib/sha1.h>
          ^~~~~~~~~~~~
1 error generated.
```
I'm still getting the same error if I set `CGO_ENABLED=1` (my project required).

If I set it to `CGO_ENABLED=0`, I get the following still:
```
package github.com/lepsta/uju/api
        imports github.com/go-git/go-git/v5
        imports github.com/go-git/go-git/v5/config
[...]

PR #625, opened 2022-12-02
Fix error when building with `CGO_ENABLED=0`

pjbgf:
Fixes regression in which applications that depend on go-git could no longer build with `CGO_ENABLED=0`.
Adds a new step on CI workflow to ensure support is not broken in the future.

Closes https://github.com/go-git/go-git/issues/624 https://github.com/go-git/go-git/issues/626.


martinlindhe:
We fixed one cgo issue yesterday in https://github.com/pjbgf/sha1cd/pull/10

I am able to build without cgo with that patch.

Issue #610, opened 2022-11-08
How i to get add multiplies Files in go-git


nnowaf:
I'm wondering how to use syntax git add file-1 file-2 file-3 and I'm trying to use ` _, err = w.Add(".gitlab src")`. And give an error like this `entry not found`

pjbgf:
@nnowaf you need to pass a file at a time to `Add`, here's an example:

```
	wt, err := g.repository.Worktree()
	if err != nil {
		return "", err
	}

	status, err := wt.Status()
	if err != nil {
		return "", err
	}

	for file := range status {
		_, _ = wt.Add(file)
	}
```

yfwz100:
> @nnowaf you need to pass a file at a time to `Add`, here's an example:
> 
> ```
> 	wt, err := g.repository.Worktree()
> 	if err != nil {
> 		return "", err
> 	}
> 
> 	status, err := wt.Status()
> 	if err != nil {
> 		return "", err
> 	}
> 
> 	for file := range status {
> 		_, _ = wt.Add(file)
> 	}
> ```

It does work. However, calling `wt.Add()` will call `wt.Status()`, which is a costly operation. It'd better expose a method that can add multiple paths.

yfwz100:
It seems that #993 may help adding files efficiently in a large repo.

pjbgf:
@yfwz100 thanks for that, so I will be closing this for the time being. We can always open a new issue in the future if needed.

PR #639, opened 2022-12-22
Support adding multiple paths

yfwz100:
It's mentioned in #610. Reusing status info when adding multiple paths may improve performance to some extent.

### DIFF
--- a/options.go
+++ b/options.go
@@ -439,6 +439,8 @@ type AddOptions struct {
 	All bool
 	// Path is the exact filepath to the file or directory to be added.
 	Path string
+	// Paths are the exact filepaths to the files or directories to be added.
+	Paths []string
 	// Glob adds all paths, matching pattern, to the index. If pattern matches a
 	// directory path, all directory contents are added to the index recursively.
 	Glob string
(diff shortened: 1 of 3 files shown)


pjbgf:
The total cost of adding multiple files (test with 200 unique files) is reasonably smaller:

```
goos: linux
goarch: amd64
BenchmarkAdd/memfs/add/200-16         	1000000000	         0.2941 ns/op	       1 B/op	       0 allocs/op
BenchmarkAdd/memfs/add-range/200-16   	1000000000	         0.002613 ns/op	       0 B/op	       0 allocs/op
BenchmarkAdd/osfs/add/200-16           	1000000000	         0.7767 ns/op	       1 B/op	       0 allocs/op
BenchmarkAdd/osfs/add-range/200-16     	1000000000	         0.01517 ns/op	       0 B/op	       0 allocs/op
```

pjbgf:
I am not sure about the changes in `AddOptions`, as having both `Path` and `Paths` is a bit confusing from an user perspective.
It would be nice to get the views of other maintainers on it.

yfwz100:
> I am not sure about the changes in `AddOptions`, as having both `Path` and `Paths` is a bit confusing from an user perspective. It would be nice to get the views of other maintainers on it.

Thanks for the test data. I've observed the cost difference in my project if we can reuse the status info, especially when IO performance is poor. I admit that the implementation may be confusing for two path parameters. That's because I don't want to break the existing API. It would be better if Golang provides union types like TypeScript. Or should it be changed to a new API?

yfwz100:
Is there anything I can do for this PR?

yfwz100:
Actually I applied this patch in my application months ago. However, it seems that having a SkipStatus option in #993 can resolve this issue too. The basic idea here is the same, to skip status checks when adding multiple files. I will close this PR and try to use that instead.

PR #659, opened 2023-01-14
dotgit: fix a filesystem race in Refs/walkReferencesTree

MichaelMure:
In walkReferencesTree(), the filesystem is browsed recursively. After a folder is listed, each child element (directory, file) are inspected.

What can happen is that by the time we get to operate on the child element, it might have been deleted from the filesystem and we would get a PathError.

In the case of directories there was already a case to avoid bubbling up the error (we consider that there is no ref there, which is correct), but that was missing for files. This commit just apply the same logic.

### DIFF
--- a/storage/filesystem/dotgit/dotgit.go
+++ b/storage/filesystem/dotgit/dotgit.go
@@ -943,6 +943,7 @@ func (d *DotGit) walkReferencesTree(refs *[]*plumbing.Reference, relPath []strin
 	files, err := d.fs.ReadDir(d.fs.Join(relPath...))
 	if err != nil {
 		if os.IsNotExist(err) {
+			// a race happened, and our directory is gone now
 			return nil
 		}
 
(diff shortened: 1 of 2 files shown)


MichaelMure:
Hmmm, that CI failure is quite odd. That failing test does indeed call the function I changed, but only when doing a git fetch. I don't see how my change could affect anything there.

MichaelMure:
Any chance to merge this?

jfontan:
It LGTM. I suppose creating a test for it may mean creating a billy filesystem wrapper that errors with certain files.

jfontan:
The following commit in my fork contains the test for both deleted references and directories in references. Feel free to cherry pick it into your branch. I've tested your change on top of it and fixes the test.

https://github.com/jfontan/go-git/commit/6a904d63a401b6326603bf1f038cb125d5f69856

MichaelMure:
@jfontan that's very cool, thank you!

pjbgf (approved):
Great stuff @MichaelMure @jfontan! :bow:

PR #675, opened 2023-02-03
Fix SetRef ignoring packed references

ZNixian:
Previously, setting a reference that was previously packed would ignore
the packed reference, and leave the reference both as a file and in
packed form.

It also couldn't read the packed refs when verifying the old commit:
this meant that any references that are packed can't be changed by code
that supplies an old reference.

This could easily be triggered by cloning a Git repo with C git, then
updating the repo with go-git. C git would leave most of the upstream
branches packed, and if they were updated on the upstream go-git would
be unable to update their references.

One can create such a repository as follows:

```sh
mkdir test-upstream
(cd test-upstream && git init)
(cd test-upstream && git commit -m first --allow-empty)
git clone test-upstream test-downstream
(cd test-downstream && git pack-refs --all)
(cd test-upstream && git commit -m second --allow-empty)
```

Fetching test-downstream prior to this patch would them exhibit this
issue.

This PR also fixes PackRefs packing symbolic refs, which C git doesn't do, and
go-git's parser can't handle. That patch is part of this PR since it's needed for
the main commit's tests.

### DIFF
--- a/storage/filesystem/dotgit/dotgit.go
+++ b/storage/filesystem/dotgit/dotgit.go
@@ -68,6 +68,8 @@ var (
 	// ErrIsDir is returned when a reference file is attempting to be read,
 	// but the path specified is a directory.
 	ErrIsDir = errors.New("reference path is a directory")
+	// ErrEmptyRefFile is returned when a reference file is an empty file.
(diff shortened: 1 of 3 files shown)


mcuadros:
Can you take a look at the tests?

PR #682, opened 2023-02-17
fix: don't use the `firstErrLine` when it is empty

ThinkChaos:
Returning `nil` causes `handleAdvRefDecodeError` to fall back to `io.ErrUnexpectedEOF`.


mcuadros:
can you provide a test?

ThinkChaos:
There's no existing tests that manipulate a `session`, so I don't want to spend too much time on that.  
I pushed something very minimal, honestly not sure it's that useful, but at least it's there.

pjbgf:
The new test assess only the change in behaviour. There was no previous tests around that part of the code, so it should be a good enough first step forwards.

pjbgf (approved):
@ThinkChaos thank you! :bow:

Issue #400, opened 2021-10-26
Support x509 and SSH key signed commits and tags


hiddeco:
As `git` started to support [SSH signing since yesterday](https://github.com/git/git/pull/1041), I am polling interest to start supporting this in `go-git`. While we are at it, it would likely be beneficial to add support for x509 signed commits as well (possible using e.g. [`smimesign`](https://github.com/github/smimesign#installation)).

Changes required, based on a quick analysis:

- Detect signature "headers" other than PGP in
  + https://github.com/go-git/go-git/blob/bd662b091fad176e225b39e7f05b73ecba37cd72/plumbing/object/commit.go#L170
  + https://github.com/go-git/go-git/blob/bd662b091fad176e225b39e7f05b73ecba37cd72/plumbing/object/tag.go#L138
- Ensure the signatures are made available in `Commit` and `Tag`
- Adapt `Verify` logic to work with multiple signature types.

The latter seems to be the hardest part, as the current contract assumes an armored key ring. One way to make this backwards compatible would be to just assume the string data is the correct data to verify the type, but I am open to alternatives.

hiddeco:
With [GitHub now supporting SSH signatures](https://github.blog/changelog/2022-08-23-ssh-commit-verification-now-supported/), I am wondering if this could get any input from maintainers. As it now has started to become a more pressing issue, given user adoption will probably grow.

If some guidance is provided to get to a workable solution, I am willing to put in the required work.

hiddeco:
Since `v5.6.0` the `plumbing` package will now properly recognize x509 and SSH signatures of tags. This means manual verification (without the `Verify` methods) is now possible.

For easy support of this using `go-git` API itself, I am working on finding a decent way to introduce this without breaking the API and/or adding obnoxious `VerifySSH|VerifyX509` methods.

wiktor-k:
Thanks for working on this! (I got here from [the fluxcd discussion](https://github.com/fluxcd/flux2/discussions/4145)).

I'm especially interested in [SSH signatures](https://github.com/wiktor-k/ssh-signing) that look vastly simpler than OpenPGP ones (e.g. they already embed the public key for verification - no additional files needed - and the `allowed_signers` file is a concise format for specifying which keys are authorized).

na-jakobs:
Thanks @hiddeco for your work on this. It would be nice to get ssh signing added to the library.

wiktor-k:
[...]

PR #690, opened 2023-02-27
plumbing: support SSH/X509 signed tags

hiddeco:
This PR enables support for extracting the SSH and X509 signatures from (annotated) Git tags, as an initial step to support the verification of more signatures than just PGP in go-git.

The ported logic from Git further ensures that we look for a signature at the tail of an annotation, instead of the first signature we find in the annotation, as this could theoretically result in a faulty signature getting detected if part of a an annotation itself (e.g. by being placed in the middle as part of an inherited message).

For commits, no further change is required as the current extraction of any signature (format) from `gpgsig` in the commit header is sufficient for manual verification.

In a future iteration, we could add `signature/ssh` and `signature/x509` packages to further enable people to deal with verifying other signatures than PGP. As well as adding additional methods to `Commit` and `Tag` to provide glue between the pacakges and the most prominent user-facing APIs.

Start of solving #400

### DIFF
--- a/plumbing/object/signature.go
+++ b/plumbing/object/signature.go
@@ -1,0 +1,101 @@
+package object
+
+import "bytes"
+
+const (
+	signatureTypeUnknown signatureType = iota
+	signatureTypeOpenPGP
+	signatureTypeX509
+	signatureTypeSSH
+)
+
+var (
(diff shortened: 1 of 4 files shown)


pjbgf (approved):
lgtm

PR #557, opened 2022-07-26
examples: add config example back. Fixes #523

cmiller01:
* previously an example existed for working with `config.Config`, but changes in implementation made the example not usable. (see #75)
* this mostly puts back the original example with some modifications to make it work with current implementation

PR #692, opened 2023-02-28
_examples: README.md , Remove broken Config link.

pat-nel87:
This link is broken and doesn't appear to map to anything current.  I propose We remove this link.


cmiller01:
What about merging mine?! https://github.com/go-git/go-git/pull/557 I was going through my old open PRs, looks like I had great timing 😄 , agree it would be nice to cleanup

Issue #674, opened 2023-02-02
Crash on resolving empty string


9seconds:
Please consider this snippet:

```go
package main

import (
	"fmt"

	"github.com/go-git/go-git/v5"
	"github.com/go-git/go-git/v5/plumbing"
)

func main() {
	repo, err := git.PlainClone("/tmp/foo", false, &git.CloneOptions{
		URL: "https://github.com/go-git/go-git",
	})
	if err != nil {
		panic(err)
	}

	fmt.Println(repo.ResolveRevision(plumbing.Revision("")))
}
```

If I run it:

```console
$ go run main.go
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x73b8e6]

goroutine 1 [running]:
github.com/go-git/go-git/v5.(*Repository).ResolveRevision(0xc0001a4450, {0x0?, 0x7fc52f?})
	/home/vagrant/.asdf/installs/golang/1.19.5/packages/pkg/mod/github.com/go-git/go-git/v5@v5.5.2/repository.go:1560 +0x86
main.main()
	/repos/tt/repro.go:18 +0x9c
exit status 2
```

PR #696, opened 2023-03-02
Fix panic for empty revisions

pjbgf:
Fixes https://github.com/go-git/go-git/issues/674.


jfontan (approved):
I would change the commit to comply with the contribution guide. Otherwise LGTM!

PR #697, opened 2023-03-02
Bump dependencies

pjbgf:
**Bumped Dependencies**
- github.com/ProtonMail/go-crypto to version 0.0.0-20230217124315-7d5c6f04bbb8.
- github.com/acomagu/bufpipe to version 1.0.4.
- github.com/go-git/go-billy/v5 to version 5.4.1.
- golang.org/x/crypto to version 0.6.0.

This PR includes an orthogonal commit that reduces the GitHub token permission used within the GH Actions.

### DIFF
--- a/.github/workflows/git.yml
+++ b/.github/workflows/git.yml
@@ -1,5 +1,8 @@
 on: [push, pull_request]
 name: Git Compatibility
+permissions:
+  contents: read
+
 jobs:
   test:
     strategy:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,8 @@
 on: [push, pull_request]
 name: Test
+permissions:
+  contents: read
+
 jobs:
   version-matrix:
     strategy:
--- a/go.mod
+++ b/go.mod
@@ -1,13 +1,15 @@
 module github.com/go-git/go-git/v5
 
+go 1.13
+
 require (
-	github.com/ProtonMail/go-crypto v0.0.0-20221026131551-cf6655e29de4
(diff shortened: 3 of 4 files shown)


jfontan (approved):

jfontan:
There must have been a hickup in the previous tests. I've rerun the failed ones and they work. Quite strange.

pjbgf:
Thank you. I noticed that some tests are failing intermittently, but haven't got to the bottom of it yet.

pjbgf:
Yes, that version is not backwards compatible and also includes a [bug](https://github.com/sergi/go-diff/issues/123) that is yet to be fixed.

aymanbagabas:
> Yes, that version is not backwards compatible and also includes a [bug](https://github.com/sergi/go-diff/issues/123) that is yet to be fixed.

https://github.com/sourcegraph/go-diff could be a good candidate to replace https://github.com/sergi/go-diff

jfontan:
@aymanbagabas I've checked the sourcegraph library but it seems it does not generate diffs, only parses them

aymanbagabas:
@jfontan I think it does.
> The library exports a function GeneratePatch which computes edits between the original and updated file contents and produces a unified diff for them as a string. This patch can then be applied to a valid git repository containing the specified file(s).

It looks like its based on an older version of https://github.com/golang/tools/tree/master/internal/diff similar to this library https://github.com/hexops/gotextdiff/

pjbgf:
Created a discussion on this to get the conversation going: https://github.com/go-git/go-git/discussions/747.

PR #655, opened 2023-01-09
fix(ssh): unable to pass a custom HostKeyCallback func

aymanbagabas:
Don't overwrite HostKeyCallback if one is provided.

Fixes: c35b8082c863 ("plumbing: transport/ssh, auto-populate ClientConfig.HostKeyAlgorithms. Fixes #411")
Fixes: https://github.com/go-git/go-git/issues/654
Signed-off-by: Ayman Bagabas <ayman.bagabas@gmail.com>

PR #698, opened 2023-03-03
plumbing/transport/ssh: do not override HostKeyCallback if set

SVilgelm:
We use the custom Auth method that returns `ClientConfig` with custom `HostKeyCallback` to check a host against known fingerprints (kind of white list), but the `SetConfigHostKeyFields` function always overrides the `HostKeyCallback` to read the fingerprints from the files and etc...

Added a simple check fixes the issue by calling the `SetConfigHostKeyFields` only if the `HostKeyCallback` is `nil`.

### DIFF
--- a/plumbing/transport/ssh/common.go
+++ b/plumbing/transport/ssh/common.go
@@ -122,9 +122,11 @@ func (c *command) connect() error {
 		return err
 	}
 	hostWithPort := c.getHostWithPort()
-	config, err = SetConfigHostKeyFields(config, hostWithPort)
-	if err != nil {
-		return err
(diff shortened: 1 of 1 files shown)


hiddeco:
Can you maybe extend the tests in `plumbing/transport/ssh/common_test.go` to cover for this scenario? Other than this, it looks good to me.

pjbgf:
Duplicate of https://github.com/go-git/go-git/pull/655.

SVilgelm:
cool, then I close this in favor of #655, that implementation looks better than mine, thank you!

PR #705, opened 2023-03-07
object: introduce signature abstraction

hiddeco:
This is work in progress to further facilitate #400. As this introduces (non-breaking) but important contract changes, I would like to have eyes on this early on.


hiddeco:
Talked with @pjbgf about this last week, and due to the lack of time I have to finalize the "nice" user part of SSH signatures on user systems (i.e. by taking the `allowed_signers` into account).

We thought it may be an option to introduce the abstraction without providing an implementation with a "keyring" like functionality. As the abstraction itself seems sound, and works as is.

Any objections against this @mcuadros? If not, I will rebase the PR and spend one more time thinking about the details of the abstraction itself (e.g. the `EntityType` versus just type checking the implementation of the interface).

PR #708, opened 2023-03-09
plumbing/transport/http: NewErr should always return Err

SVilgelm:
* Change the `NewErr` function to always return the instance of Err struct for any responses with included responses body in `Err.ResponseBody` field.
* Extend the `Err` struct to wrap another error. It is needed to keep backward compatibility: `errors.Is(err, transport.ErrAuthenticationRequired)`.
* Extend the `UnexpectedError` and `PermanentError` with `Unwrap` function, so the `errors.As` can unwrap the error.
* Add `ResponseBody` field to `Err` struct and TestNewErrResponseBody test

For debbugging purpose we need to have as much as possible info from the response, some servers can include a valuable information in body in case of 4XX responses.

### DIFF
--- a/plumbing/transport/http/common.go
+++ b/plumbing/transport/http/common.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"strconv"
@@ -249,7 +250,9 @@ func (a *TokenAuth) String() string {
 
 // Err is a dedicated error to return errors based on status code
 type Err struct {
-	Response *http.Response
+	Response     *http.Response
+	ResponseBody []byte
+	WrappedError error
 }
 
 // NewErr returns a new Err based on a http response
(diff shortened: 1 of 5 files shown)


plumbing/transport/http/common.go
```diff
@@ -260,23 +262,37 @@ func NewErr(r *http.Response) error {
...
 }
 
 // StatusCode returns the status code of the response
 func (e *Err) StatusCode() int {
 	return e.Response.StatusCode
 }
 
+// Body returns the body of the response
+func (e *Err) Body() io.ReadCloser {
+	return e.Response.Body
```
pjbgf, line 282: At present, all `NewErr` calls ensure that  the body is closed, which is required in order to avoid resource leaks, as per docs:

> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close. If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.

I am not sure there is a benefit of exposing a `Body()` that is already (or soon to be) closed.

SVilgelm: The function can read the body and store in local cache or add to the error message. 
I have another workaround by intercepting the response in custom round tripper, but I really don't like that solution

SVilgelm: > I am not sure there is a benefit of exposing a `Body()`

as I said we need it to analyze and print the full response, including the body. For example the GitLab can return this response: `403 Forbidden  - Your account has been blocked.` In the case of such response, we don't have information about the reason of 403 response, it can be, like in that example, the account is blocked, or invalid token/credentials.

SVilgelm:
Hi @pjbgf Could you please revisit this PR?

PR #715, opened 2023-03-16
internal: Fix regression in scp-like match

jotadrilo:
I found that a regression was added in this commit https://github.com/go-git/go-git/commit/db0e226d48285a0ae8cda7db05d2ca20d9000dc6.

Specifically, it dropped support for repositories without multipath (ex. `git@host:repo.git`). The package fails cloning this kind of repository with this error:

```
$ go run -- "git@host:repo.git"
Cloning "git@host:repo.git"...
Error: repository not found
```

This PR fixes it and it adds test cases to cover the intended functionality.

### DIFF
--- a/internal/url/url_test.go
+++ b/internal/url/url_test.go
@@ -13,11 +13,27 @@ type URLSuite struct{}
 var _ = Suite(&URLSuite{})
 
 func (s *URLSuite) TestMatchesScpLike(c *C) {
+	// See https://github.com/git/git/blob/master/Documentation/urls.txt#L37
 	examples := []string{
+		// Most-extended case
 		"git@github.com:james/bond",
-		"git@github.com:007/bond",
+		// Most-extended case with port
 		"git@github.com:22:james/bond",
+		// Most-extended case with numeric path
+		"git@github.com:007/bond",
+		// Most-extended case with port and numeric "username"
 		"git@github.com:22:007/bond",
+		// Single repo path
+		"git@github.com:bond",
+		// Single repo path with port
+		"git@github.com:22:bond",
+		// Single repo path with port and numeric repo
+		"git@github.com:22:007",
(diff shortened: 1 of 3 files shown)


internal/url/url_test.go
```diff
@@ -14,10 +14,20 @@ var _ = Suite(&URLSuite{})
 
 func (s *URLSuite) TestMatchesScpLike(c *C) {
 	examples := []string{
+		// Most-extended case
 		"git@github.com:james/bond",
-		"git@github.com:007/bond",
+		// Most-extended case with port
 		"git@github.com:22:james/bond",
+		// Most-extended case with numeric "username"
```
pjbgf, line 21: "username" seems confusing, this would be a component of the repository path regardless of its meaning to the server (e.g. org, user, etc).

pjbgf (commented):
Please also add test cases in which the path ends with `.git` and starts with `_`. They all work with your changes, however that will improve coverage against regression.

Otherwise LGTM.

internal/url/url_test.go
jotadrilo: Actually, I used "username" because the commit that added the regression (https://github.com/go-git/go-git/commit/db0e226d48285a0ae8cda7db05d2ca20d9000dc6) used "username". I thought there was a good reason.

I will remove that "username" and use path, which is these cases' semantics (and link to upstream docs too).

pjbgf:
Nice one @jotadrilo, please squash the commits.

pjbgf (approved):
@jotadrilo thank you for your contribution! :1st_place_medal:

jotadrilo:
@mcuadros do you have any estimation on when will these changes land in a new release? thank you!

pjbgf:
@jotadrilo we should be cutting a release at some point in next 2-4 weeks.

PR #722, opened 2023-03-27
storage: filesystem, Populate index before use. Fixes #148

### LINKED ISSUE #148: Repository.Resolution with short hash fails with repository opened by PlainOpen
I tried to use the short hash resolution feature merged in https://github.com/go-git/go-git/pull/136.

When I clone the repository using `PlainClone`, short hash revision resolution succeed.
But short hash revision resolution failed with the repository opened by `PlainOpen`.
```
func main() {
	r, err := git.PlainClone("/tmp/foo", false, &git.CloneOptions{
		URL:      "https://github.com/go-git/go-git",
		Progress: os.Stdout,
	})
	if err != nil {
		log.Println(err.Error())
		return
	}
	// SUCCESS
	h, err := r.ResolveRevision("cded5b6")
	if err != nil {
		log.Println("ResolveRevision", err)
		return
	}
	log.Println(h)
	r, err = git.PlainOpen("/tmp/foo")
	if err != nil {
		log.Println(err.Error())
		return
	}
	// FAILED：reference not found
	h, err = r.ResolveRevision("cded5b6")
	if err != nil {
		log.Println("ResolveRevision", err)
		return
	}
	log.Println(h)
}
```

AriehSchneier:


AriehSchneier:
@pjbgf Any chance I could get this reviewed?

jfontan:
LGTM! The only nitpick is to signoff the commit (`git commit -s`).

pjbgf:
@AriehSchneier the test `TestHashesWithPrefix` is [failing](https://github.com/go-git/go-git/actions/runs/4532947460/jobs/8457195706?pr=722):

```
FAIL: object_test.go:393: FsSuite.TestHashesWithPrefix

object_test.go:405:
    c.Assert(hashes, HasLen, 1)
... obtained []plumbing.Hash = []plumbing.Hash{plumbing.Hash{0xf3, 0xdf, 0xe2, 0x9d, 0x26, 0x83, 0x3, 0xfc, 0x6e, 0x1b, 0xbc, 0xe2, 0x68, 0x60, 0x5f, 0xc9, 0x95, 0x[73](https://github.com/go-git/go-git/actions/runs/4532947460/jobs/8457195414?pr=722#step:6:74), 0x40, 0x6e}, plumbing.Hash{0xf3, 0xdf, 0xe2, 0x9d, 0x26, 0x83, 0x3, 0xfc, 0x6e, 0x1b, 0xbc, 0xe2, 0x68, 0x60, 0x5f, 0xc9, 0x95, 0x73, 0x40, 0x6e}}
... n int = 1
```

Which makes me think that the changes will regress e28d9c90aad624596d20d1a59c8371d81b69190b. Therefore we need to have an option to toggle this, and update the tests accordingly. This would be disabled by default, so the existing tests will pass without changes.

AriehSchneier:
> LGTM! The only nitpick is to signoff the commit (`git commit -s`).

Happy to do this, which I have now done and force-pushed the amended commit (no other change other than to signoff).

Just wondering if this is actually a requirement or a preference? I didn't see it as a step in the [Contributing Guidelines](https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md)?

AriehSchneier:
> @AriehSchneier the test `TestHashesWithPrefix` is [failing](https://github.com/go-git/go-git/actions/runs/4532947460/jobs/8457195706?pr=722):
> 
> ```
> FAIL: object_test.go:393: FsSuite.TestHashesWithPrefix
> 
> object_test.go:405:
>     c.Assert(hashes, HasLen, 1)
> ... obtained []plumbing.Hash = []plumbing.Hash{plumbing.Hash{0xf3, 0xdf, 0xe2, 0x9d, 0x26, 0x83, 0x3, 0xfc, 0x6e, 0x1b, 0xbc, 0xe2, 0x68, 0x60, 0x5f, 0xc9, 0x95, 0x[73](https://github.com/go-git/go-git/actions/runs/4532947460/jobs/8457195414?pr=722#step:6:74), 0x40, 0x6e}, plumbing.Hash{0xf3, 0xdf, 0xe2, 0x9d, 0x26, 0x83, 0x3, 0xfc, 0x6e, 0x1b, 0xbc, 0xe2, 0x68, 0x60, 0x5f, 0xc9, 0x95, 0x73, 0x40, 0x6e}}
> ... n int = 1
> ```
> 
> Which makes me think that the changes will regress [e28d9c9](https://github.com/go-git/go-git/commit/e28d9c90aad624596d20d1a59c8371d81b69190b). Therefore we need to have an option to toggle this, and update the tests accordingly. This would be disabled by default, so the existing tests will pass without changes.

Is this a regression, or has it just shown an existing bug? You can get the test to fail with the existing code by adding this on line 401:
```_ = o.HasEncodedObject(plumbing.ZeroHash)```

It is not failing to find the hash, it is now finding the hash twice (which it already does if you call something that populates the index cache, like `HasEncodedObject`). Should I change the failing test to check for length >= 1 instead of == 1 and we can open a new bug report that `HashesWithPrefix` can report the same hash twice?

PS I'm not exactly sure what you meant by adding this with a toggle, where would it be toggled?

**Comment by AriehSchneier** [comment]
> @AriehSchneier the test `TestHashesWithPrefix` is [failing](https://github.com/go-git/go-git/actions/runs/4532947460/jobs/8457195706?pr=722):
> 
> ```
> FAIL: object_test.go:393: FsSuite.TestHashesWithPrefix
> 
> object_test.go:405:
>     c.Assert(hashes, HasLen, 1)
> ... obtained []plumbing.Hash = []plumbing.Hash{plumbing.Hash{0xf3, 0xdf, 0xe2, 0x9d, 0x26, 0x83, 0x3, 0xfc, 0x6e, 0x1b, 0xbc, 0xe2, 0x68, 0x60, 0x5f, 0xc9, 0x95, 0x[73](https://github.com/go-git/go-git/actions/runs/4532947460/jobs/8457195414?pr=722#step:6:74), 0x40, 0x6e}, plumbing.Hash{0xf3, 0xdf, 0xe2, 0x9d, 0x26, 0x83, 0x3, 0xfc, 0x6e, 0x1b, 0xbc, 0xe2, 0x68, 0x60, 0x5f, 0xc9, 0x95, 0x73, 0x40, 0x6e}}
> ... n int = 1
> ```
> 
> Which makes me think that the changes will regress [e28d9c9](https://github.com/go-git/go-git/commit/e28d9c90aad624596d20d1a59c8371d81b69190b). Therefore we need to have an option to toggle this, and update the tests accordingly. This would be disabled by default, so the existing tests will pass without changes.

@pjbgf  Is this a regression, or has it just shown an existing bug? You can get the test to fail with the existing code by adding this on line 401:
```_ = o.HasEncodedObject(plumbing.ZeroHash)```

It is not failing to find the hash, it is now finding the hash twice (which it already does if you call something that populates the index cache, like `HasEncodedObject`). Should I change the failing test to check for length >= 1 instead of == 1 and we can open a new bug report that `HashesWithPrefix` can report the same hash twice?

PS I'm not exactly sure what you meant by adding this with a toggle, where would it be toggled?

File: storage/filesystem/object.go [hunk 1]
```diff
@@ -545,6 +545,9 @@ func (s *ObjectStorage) HashesWithPrefix(prefix []byte) ([]plumbing.Hash, error)
 
 	// TODO: This could be faster with some idxfile changes,
 	// or diving into the packfile.
```
pjbgf, line 547: This comment is related to the `for` loop, therefore the changes must be placed before it.

storage/filesystem/object.go
```diff
@@ -545,6 +545,9 @@ func (s *ObjectStorage) HashesWithPrefix(prefix []byte) ([]plumbing.Hash, error)
...
+	if err := s.requireIndex(); err != nil {
+		return nil, err
+	}
```
pjbgf, line 550: For the failing test, before `s.requireIndex()` is executed the hash is already in `hashes`. Now with `requireIndex()` the follow up code appends the same hash again, hence the duplication. The original version of `TestHashesWithPrefix` was correct.

Please add a short-circuit ahead of the `requireIndex` as that fixes the problem and yield faster results in some execution paths:

```suggestion
	// short-circuit if hashes were already found.
	if len(hashes) > 0 {
		return hashes, nil
	}

	if err := s.requireIndex(); err != nil {
		return nil, err
	}
```

This is based on the `TestHashesWithPrefix` example:
```
BenchmarkHashesWithPrefix_ReturnIfFound-16         	   15084	     78833 ns/op	   67402 B/op	    2090 allocs/op
BenchmarkHashesWithPrefix_RequireIndexOnly-16    	    1092	   1072770 ns/op	  198198 B/op	    2941 allocs/op
```

storage/filesystem/object_test.go
```diff
@@ -402,10 +402,25 @@ func (s *FsSuite) TestHashesWithPrefix(c *C) {
 	prefix, _ := hex.DecodeString("f3dfe2")
 	hashes, err := o.HashesWithPrefix(prefix)
 	c.Assert(err, IsNil)
-	c.Assert(hashes, HasLen, 1)
+	c.Assert(len(hashes) >= 1, Equals, true)
```
pjbgf, line 405:
```suggestion
	c.Assert(hashes, HasLen, 1)
```

pjbgf (changes requested):
> Is this a regression, or has it just shown an existing bug?

@AriehSchneier I wasn't completely sure so I dug into it and found the problem as per comment below. My understanding now is that the change introduced a regression in both performance and behaviour, but you can mitigate both adding a short-circuit.

storage/filesystem/object.go
pjbgf: Good questions. From what I understood from the code, in the `s.index` we have the full index - which is being reloaded by your changes. The former, I am not completely sure as in some manual tests I could confirm it does not contain some of the references, and therefore would not return all refs.

Due to that, I think you should remove:

```
	hashes, err := s.dir.ObjectsWithPrefix(prefix)
	if err != nil {
		return nil, err
	}
```

And replace it with the creation of the `hashes` array instead. IMO, the potential performance hit is overshadowed by the accuracy improvement on that case.

pjbgf: That would therefore remove the need for the short-circuit as well.

AriehSchneier: Looking at `EncodedObject(...)` It looks like its possible that there are objects that have not been packed, I think the only way to guarantee it returns all the possible values is to search both and deduplicate the results. Are you happy for me to implement that?

AriehSchneier: Similar story with `IterEncodedObjects`, it first iterates the encoded objects, and then iterates the pack files, but passes in a list of objects that it has already `seen`.

AriehSchneier: I've pushed an update where it is checking both, but removes duplicates the same way the rest of the code does.

storage/filesystem/object.go
AriehSchneier: I think the comment applies to both sections together, if the code was switched to dive into packfile then it wouldn't require populating the index?
(I'm still happy to move it if you still think its better between the population and usage)

storage/filesystem/object.go
pjbgf: Looks good to me. Do you mind writing some comments to that effect please?

AriehSchneier: Done, followed the comments on the other functions

pjbgf:
@AriehSchneier please squash the commits and rebase.

AriehSchneier:
> @AriehSchneier please squash the commits and rebase.

Done

PR #727, opened 2023-03-29
Update default ref to use refs/heads/main

techknowlogick:
When creating new repositories, use refs/heads/main as the default ref

Related: https://github.com/charmbracelet/soft-serve/pull/198

### DIFF
--- a/options.go
+++ b/options.go
@@ -323,7 +323,7 @@ func (o *CheckoutOptions) Validate() error {
 	}
 
 	if o.Branch == "" {
-		o.Branch = plumbing.Master
+		o.Branch = plumbing.Main
 	}
 
 	return nil
--- a/plumbing/reference.go
+++ b/plumbing/reference.go
@@ -126,6 +126,7 @@ func (r ReferenceName) Short() string {
 const (
 	HEAD   ReferenceName = "HEAD"
 	Master ReferenceName = "refs/heads/master"
+	Main   ReferenceName = "refs/heads/main"
 )
 
 // Reference is a representation of git reference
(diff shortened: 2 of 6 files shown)


pjbgf:
This is a breaking change, I would rather tackle it in a way that would keep backwards compatibility.

Additionally, I think the behaviour you are trying to achieve could be done using https://github.com/go-git/go-git/blob/ce62f3e9ff86270538a514a68d3bd5563a733e3b/config/config.go#L96.

techknowlogick:
Thanks @pjbgf, yes it is indeed breaking, although it might be considered as breaking to align with changes to git client itself. I understand you have many downstreams you need to consider so I won't advocate for this breaking change past what I have already done.

Sadly, I had also considered the config setting, but the specific issue I am facing can't be set by the default branch config setting. The `Init()` function: https://github.com/go-git/go-git/blob/master/repository.go#L92 hardcodes `master` branch when creating a new repo.

I'd be happy to contribute a PR, perhaps to add a similar function, `InitWithBranch`, where `Init` would alias to that function but pass in the default of `master`?

Something like:

```go
func Init(s storage.Storer, worktree billy.Filesystem) (*Repository, error) {
    InitWithBranch(s, worktree, lumbing.Master)
}
func InitWithBranch(s storage.Storer, worktree billy.Filesystem, branch string) (*Repository, error) {
// ...
h := plumbing.NewSymbolicReference(plumbing.HEAD, branch)
// ...
```

pjbgf:
@techknowlogick apologies for the delay. Happy to review a PR on extending `Init` to support other default branches. I would just make sure you follow the naming convention we use and call it `InitWithOptions`.

techknowlogick:
Thanks for that suggestion re: existing conventions @pjbgf. I have submitted #764, which is hopefully more suitable :)

Issue #68, opened 2020-05-18
git grep on a bare repository


ddevault:
Would be nice to have this.

dmaharana:
git grep 'search string' HEAD

PR #728, opened 2023-03-30
fix: git grep bare repositories

aymanbagabas:
Perform grep on both `*Repository` or `*Worktree`.

Fixes: https://github.com/go-git/go-git/issues/68

### DIFF
--- a/options.go
+++ b/options.go
@@ -642,7 +642,13 @@ var (
 )
 
 // Validate validates the fields and sets the default values.
+//
+// TODO: deprecate in favor of Validate(r *Repository) in v6.
 func (o *GrepOptions) Validate(w *Worktree) error {
+	return o.validate(w.r)
+}
+
+func (o *GrepOptions) validate(r *Repository) error {
 	if !o.CommitHash.IsZero() && o.ReferenceName != "" {
 		return ErrHashOrReference
 	}
@@ -650,7 +656,7 @@ func (o *GrepOptions) Validate(w *Worktree) error {
 	// If none of CommitHash and ReferenceName are provided, set commit hash of
 	// the repository's head.
 	if o.CommitHash.IsZero() && o.ReferenceName == "" {
-		ref, err := w.r.Head()
+		ref, err := r.Head()
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


pjbgf:
@aymanbagabas I will take a proper look at this later on this week.

options.go
```diff
@@ -642,15 +642,15 @@ var (
 )
 
 // Validate validates the fields and sets the default values.
-func (o *GrepOptions) Validate(w *Worktree) error {
+func (o *GrepOptions) Validate(r *Repository) error {
```
pjbgf, line 645: Would this not break backwards compatibility?

aymanbagabas: It's an API change, so probably yes. I overlooked this one. We could change it to `interface{}` and then check the type and return an error if type is not one of `*Repository` or `*Worktree`. AFAICT, `Validate()` is only used once in the codebase.

pjbgf: I think it is easier if we simply revert this change, and then get `func (r *Repository) Grep(opts *GrepOptions)` to call `r.wt.Grep(opts)`, instead of the other way around. WDYT?

worktree.go
```diff
@@ -836,6 +836,11 @@ func (w *Worktree) Grep(opts *GrepOptions) ([]GrepResult, error) {
 	return findMatchInFiles(fileiter, treeName, opts)
 }
 
+// Grep performs grep on a worktree.
+func (w *Worktree) Grep(opts *GrepOptions) ([]GrepResult, error) {
+	return w.r.Grep(opts)
```
pjbgf, line 841:
```suggestion
// Grep performs grep on a repository.
func (r *Repository) Grep(opts *GrepOptions) ([]GrepResult, error) {
	return r.wt.Grep(opts)
```

options.go
aymanbagabas: `r.wt` is `nil` for bare repositories. We can't do that if we want to support grep on both worktree and bare repos

pjbgf: To support this while not breaking the current API, we could:

```suggestion
// TODO: deprecate in favor of Validate(r *Repository) in v6.

// Validate validates the fields and sets the default values.
func (o *GrepOptions) Validate(w *Worktree) error {
	return o.validateRepository(w.r)
}

func (o *GrepOptions) validateRepository(r *Repository) error {
```

pjbgf (approved):
Great job, @aymanbagabas! :1st_place_medal:

Issue #566, opened 2022-08-16
does go-git support ProxyCommand or ProxyJump?


rafiqkattangere:
I see the host name and port over rides are picked by the go-git client. But I dont think ProxyCommand or ProxyJump are considered.
Does anyone have alternatives how we can use ssh over 443 and bypass the proxy behind the firewall using ProxyCommand or ProxyJump.

Nomango:
The latest version seems to have added support for SSH proxy, allowing you to specify a proxy when cloning.

e.g.

```go
git.CloneContext(ctx, memory.NewStorage(), nil, &git.CloneOptions{
    // ...
    ProxyOptions: transport.ProxyOptions{ URL: "socks5://127.0.0.1:1234" },
})
```

However, it is not possible to specify a custom proxy dialer. The proxy must be a URL.

hegde-prashant:
SSH connection to GitHub happening via port 443. Is there a way to specify 'http' proxy?

PR #729, opened 2023-03-30
feat(transport): support custom ssh proxy dialer

Nomango:
see #566

SSH transport defaults to using the proxy.Dialer loaded from environment variables. Making it a configurable option may be a better choice.

### DIFF
--- a/plumbing/transport/ssh/common.go
+++ b/plumbing/transport/ssh/common.go
@@ -4,6 +4,7 @@ package ssh
 import (
 	"context"
 	"fmt"
+	"net"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


hegde-prashant:
@Nomango we are waiting for this change. any idea when this will be merged and released? thanks.

Nomango:
> @Nomango we are waiting for this change. any idea when this will be merged and released? thanks.

The latest version seems to have added support for SSH proxy, allowing you to specify a proxy when cloning.

e.g.

```go
git.CloneContext(ctx, memory.NewStorage(), nil, &git.CloneOptions{
    // ...
    ProxyOptions: transport.ProxyOptions{ URL: "socks5://127.0.0.1:1234" },
})
```

PR #744, opened 2023-04-18
Add support for custom proxy settings

aryan9600:
This PR adds support for specifying a proxy per remote operation, as opposed to the current global way via environment variables.
It adds a new struct `transport.ProxyOptions` which is then added as a new field to all other APIs used for specifying options for remote operations such as `FetchOptions`, `PushOptions`, etc. Furthermore, the http transport has been slightly refactored to optionally cache `net/http.Transport` objects (using a LRU cache) based on the transport specific settings (if any). Currently the below fields are considered to be transport specific settings:
* `ep.CaBundle`
* `ep.InsecureSkipTLS` (only if its true)
* `ep.ProxyOptions`

Any HTTP operation involving an endpoint which specifies any of these options will cache a transport (configured with these settings) and in the future reuse the cached transport for these specific options. This also follows what is recommended by the [`net/http.Transport` docs](https://pkg.go.dev/net/http#Transport):
> Transports should be reused instead of created as needed. Transports are safe for concurrent use by multiple goroutines.

By default caching is turned off. Users can opt in by creating a new Git HTTP transport and installing it before using the library
```go
func init(
    transport, err := http.NewClientWithOptions(nil, http.ClientOptions{
        CacheMaxSize: 100 // must be a positive integer
    })
    client.InstallProtocol("http", transport)
    client.InstallProtocol("https", transport)
)
```


plumbing/transport/ssh/common.go
```diff
@@ -166,10 +168,36 @@ func dial(network, addr string, config *ssh.ClientConfig) (*ssh.Client, error) {
...
-	conn, err := proxy.Dial(ctx, network, addr)
+	var conn net.Conn
+	var err error
+
+	if proxyOpts.URL != "" {
+		proxyUrl, err := url.Parse(proxyOpts.URL)
+		if err != nil {
+			return nil, err
+		}
+		if proxyOpts.Username != "" && proxyOpts.Password != "" {
```
pjbgf, line 179: Password does not seem to be a required field for proxy auth purposes.

```suggestion
		if proxyOpts.Username != "" {
```

plumbing/transport/ssh/internal/test/test_utils.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func buildCommand(c []string) (cmd *exec.Cmd, stdin io.WriteCloser, stderr, stdout io.ReadCloser, err error) {
+	if len(c) != 2 {
+		err = fmt.Errorf("invalid command")
+		return
+	}
+
+	// fix for Windows environments
+	path := strings.Replace(c[1], "/C:/", "C:/", 1)
```
pjbgf, line 58:
```suggestion
	// fix for Windows environments
	if runtime.GOOS == "windows" {
	  path := strings.Replace(c[1], "/C:/", "C:/", 1)
	}
```

plumbing/transport/http/testdata/certs/server.crt
```diff
@@ -0,0 +1,22 @@
+-----BEGIN CERTIFICATE-----
```
pjbgf, line 1: What are your thoughts on just generating the cert files on demand before running the tests?
Main idea was to make it clear how it is being generated, so it is easier for folks to maintain it in the future. Happy to be shot down.

pjbgf (commented):
Really good stuff @aryan9600! The main concern I have is around the caching as per below. Maybe we can split this into two PRs, and deal with the caching as a follow-up. WDYT?

Please re-word the commits in-line with the [commit message guidelines](https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md#format-of-the-commit-message). E.g. `plumbing: transport/ssh, ...`.

aryan9600:
The reason why I included the caching changes in this PR is because its related to the way proxy support is added to the http transport. But if you feel like we should have this PR focus _only_ on proxy support, lets move the caching stuff to another PR but instead of treating it as a follow up, I feel like it'd be better to treat it as a prerequisite to this PR? That way adding proxy support will be straightforward.

plumbing/transport/http/testdata/certs/server.crt
aryan9600: i think that might increase the time to run tests considerably as generating these certs does take a non-negligible amount of time. maybe we could add a README or help.txt in the folder containing the commands that were used to generate them?

pjbgf: Sure, or maybe add as part of the commit description.

plumbing/transport/http/transport.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+func addTransport(opts transportOptions, transport *http.Transport) {
+	m.Lock()
+	if len(transports) >= TransportCacheSize {
+		for key := range transports {
+			delete(transports, key)
+			break
+		}
+	}
```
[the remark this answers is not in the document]
aryan9600: `removeTransport` doesn't need to be called anywhere since we are setting a hard cap on the cache size. i just added for the sake of completeness.

plumbing/transport/ssh/common.go
aryan9600: is this true for ssh only or also for http?

pjbgf: AFAIK both. I could not find any official documentation to back that up either way though.

pjbgf (changes requested):
As per off-line chat, the caching behaviour needs further documenting and potentially a few changes.

plumbing/transport/http/common.go
```diff
@@ -74,40 +79,83 @@ func advertisedReferences(ctx context.Context, s *session, serviceName string) (
...
+	cl := &client{
 		c: c,
 	}
+
+	if opts != nil {
+		if opts.CacheMaxEntries > 0 {
+			cl.transports = lru.New(opts.CacheMaxEntries)
+		} else if opts.CacheMaxEntries < 0 {
+			cl.transports = nil
+		}
```
pjbgf, line 144: The `else if` seems to be redundant:
```suggestion
		}
```

plumbing/transport/http/common.go
```diff
@@ -74,40 +79,83 @@ func advertisedReferences(ctx context.Context, s *session, serviceName string) (
...
+	// cache will hold at any given point of time.
+	// Calling `client.addTransport()` after the cache has reached the specified
+	// size, will result in the least recently used transport getting deleted
+	// before the provided transport is added to the cache.
+	CacheMaxEntries int
+}
+
+var (
+	// TransportCacheSize is the default capacity of the transport objects cache.
+	TransportCacheSize = 100
```
pjbgf, line 99: Let's make this an opt-in feature:

```suggestion
	// TransportCacheSize is the default capacity of the transport objects cache.
	TransportCacheSize = 0
```

pjbgf (changes requested):

plumbing/transport/ssh/common.go
```diff
@@ -166,10 +167,33 @@ func dial(network, addr string, config *ssh.ClientConfig) (*ssh.Client, error) {
...
+		// Try to use a ContextDialer, but fall back to a Dialer if that goes south.
+		ctxDialer, ok := dialer.(proxy.ContextDialer)
+		if ok {
+			conn, err = ctxDialer.DialContext(ctx, "tcp", addr)
+		} else {
+			conn, err = dialer.Dial("tcp", addr)
+		}
+	} else {
+		conn, err = proxy.Dial(ctx, network, addr)
+	}
```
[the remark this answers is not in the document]
aryan9600: I view it as a tradeoff. With your suggested change, we would enforce the context but we would NOT use the custom proxy addr provided. IMO, its better to use the proxy and let go of the context.
Furthermore, this is just a fallback. looking at the actual object returned by `proxy.FromURL()`,  I don't think the `ctxDialer, ok := dialer.(proxy.ContextDialer)` type assertion will actually fail

plumbing/transport/http/transport.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func (c *client) fetchTransport(opts transportOptions) (*http.Transport, bool) {
+	c.m.RLock()
+	t, ok := c.transports.Get(opts)
+	c.m.RUnlock()
+	transport, ok := t.(*http.Transport)
+	if !ok {
+		return nil, false
+	}
```
pjbgf, line 35: We should short-circuit if the `.Get()` failed:
```suggestion
	if !ok {
		return nil, false
	}
	transport, ok := t.(*http.Transport)
	if !ok {
		return nil, false
	}
```

plumbing/transport/ssh/common.go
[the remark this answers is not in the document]
pjbgf: You are right, based on the execution path the `dialer` should always implement `proxy.ContextDialer`. So I think we are better off erroring when that isn't the case, so that ensures we always enforce the context timeout, but also can debug if things ever went awry:

```go
		ctxDialer, ok := dialer.(proxy.ContextDialer)
		if !ok {
			return nil, fmt.Errorf("...")
		}
		conn, err = ctxDialer.DialContext(ctx, "tcp", addr)
```

pjbgf (commented):
Just a few additional nits, otherwise LGTM.

pjbgf (approved):
@aryan9600 great job, thanks for working on this! :trophy:

PR #750, opened 2023-04-26
git: Add support to ls-remote with peeled references. Fixes #749

### LINKED ISSUE #749: How to find the commit referenced by a tag in a remote repository?
I have a related question/problem to #87 but with a remote repository (i.e. from `NewRemote`)

The tags returned from `List` resolve to the hashes of the tags, not the underlying commits, and `ResolveRevision` isn't an option on a remote repository.

For example, examining https://github.com/aide/aide, with:

```
.
.
.
5ee3529c3014b4238231885b1403faa3e1affb5c        refs/tags/v0.18.1
de5bb24b9b24df7598161a1ce19dc2ce15afa9c6        refs/tags/v0.18.1^{}
.
.
.
```

The references returned from `List` only include the one reference for the `v0.18.1` tag, and `5ee3529c3014b4238231885b1403faa3e1affb5c` is the hash of the tag itself, not the underlying commit `de5bb24b9b24df7598161a1ce19dc2ce15afa9c6`

https://github.com/google/osv.dev/blob/bf67aafde6fa58618e02a016a505b4a8116f6ca2/vulnfeeds/git/repository.go#L60 is where I'm using the library currently.

pjbgf:
A new `PeelingOption` field was introduced into `ListOptions`. The new options
include the default (and backwards compatible) IgnorePeeled. Plus another
two variations which either only returns peeled references (`OnlyPeeled`), or
append peeled references to the list (`AppendPeeled`).

The ls-remote example was updated to align with upstream, in which peeled
references are appended to the results by default.

A new `ErrEmptyUrls` error is now returned when `List` or `ListContext` do not
receive a URL to work with, to improve overall execution flow.

### DIFF
--- a/remote_test.go
+++ b/remote_test.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"time"
 
 	"github.com/go-git/go-git/v5/config"
@@ -865,6 +866,7 @@ func (s *RemoteSuite) TestPushForceWithLease_success(c *C) {
 		c.Assert(sto.SetReference(newCommit), IsNil)
 
 		ref, err := sto.Reference("refs/heads/branch")
+		c.Assert(err, IsNil)
 		c.Log(ref.String())
 
 		url := dstFs.Root()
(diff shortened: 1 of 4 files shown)


remote_test.go
```diff
@@ -1210,6 +1212,41 @@ func (s *RemoteSuite) TestList(c *C) {
 	}
 }
 
+func (s *RemoteSuite) TestListPeeling(c *C) {
+	remote := NewRemote(memory.NewStorage(), &config.RemoteConfig{
+		Name: DefaultRemoteName,
+		URLs: []string{"https://github.com/aide/aide.git"},
```
hiddeco, line 1218: This has the potential of becoming abandoned and/or flaky.

andrewpollock, line 1218:
```suggestion
		URLs: []string{"https://github.com/git-fixtures/tags.git"},
```

Running `go test -v . -check.v -check.f=RemoteSuite.TestListPeeling` with https://github.com/git-fixtures/tags.git instead still passes.

andrewpollock: Figured out how to run the test, see https://github.com/go-git/go-git/pull/750/files#r1183171371

pjbgf: Thank you for the review. Replaced the repository with the suggestion. PTAL

Issue #603, opened 2022-10-27
git blame panic: contents and commits have different length


williamfzc:
```go
func TestAA(t *testing.T) {
	abs, _ := filepath.Abs("F:\\workspace\\github\\go-git")
	repo, err := git.PlainOpen(abs)
	if err != nil {
		panic(err)
	}
	head, err := repo.Head()
	if err != nil {
		panic(err)
	}
	targetCommit, err := repo.CommitObject(head.Hash())
	if err != nil {
		panic(err)
	}

	_, err = git.Blame(targetCommit, "remote.go")
	if err != nil {
                // PANIC!!!
		panic(err)
	}
}
```

panic happened when calling `git.Blame`

```text
panic: contents and commits have different length [recovered]
	panic: contents and commits have different length
```

It will happened also in `work_tree_test.go`.
Looks like it can not handle extra empty line at the end of file.

In git blame:

![image](https://user-images.githubusercontent.com/13421694/198298366-2e5a72dc-5cb7-4f6a-8acc-c2f4efab9967.png)

But actually:

![image](https://user-images.githubusercontent.com/13421694/198298475-cc7e1919-ea1b-4505-b26d-9e4f96066d4f.png)

Tested on Windows and WSL.

## guess

```go
func newLines(contents []string, commits []*object.Commit) ([]*Line, error) {
	lcontents := len(contents)
	lcommits := len(commits)

	if lcontents != lcommits {
		fmt.Printf("c vs c: %d, %d", lcontents, lcommits)
...
```

Output:

```go
=== RUN   TestAA
c vs c: 1430, 1397--- FAIL: TestAA (1.97s)
```

lcommits seems not correct.

williamfzc:
ok it causes by: 
https://github.com/go-git/go-git/blob/08cffa1efade914020497a73907763e8d3707a77/references.go#L105

tarimoe:
I also meet this issue too, any official fix for this ?

tarimoe:
> ok it causes by:
> 
> https://github.com/go-git/go-git/blob/08cffa1efade914020497a73907763e8d3707a77/references.go#L105
> 
> When it has multi parents, itself will not be added to result list, which causes some commits lost.
> 
> [williamfzc@188480d](https://github.com/williamfzc/go-git/commit/188480dfdd4e88057d4493e6c85d5ce4f0e65d23)
> 
> This fix looks fine.

This fix me too~
Thanks

pjbgf:
[...]

PR #751, opened 2023-04-27
Fix git blame panic: contents and commits have different length

williamfzc:
Details: #603


pjbgf:
@williamfzc please squash the commits and reword the commit message to align with the [contributing guidelines](https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md) (e.g. PR's new title).

williamfzc:
> @williamfzc please squash the commits and reword the commit message to align with the [contributing guidelines](https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md) (e.g. PR's new title).

done

references.go
```diff
@@ -104,6 +104,16 @@ func walkGraph(result *[]*object.Commit, seen *map[plumbing.Hash]struct{}, curre
 	default: // more than one parent contains the path
 		// TODO: detect merges that had a conflict, because they must be
 		// included in the result here.
+
```
pjbgf, line 107: Most of `case 1` and `default` are the same. This switch could be simplified with:

```go
	case 0:
		*result = append(*result, current)
		return nil
	default: // or or more than one parent contains the path
		different, err := differentContents(path, current, parents)
		if err != nil {
			return err
		}
		if len(different) > 0 {
			*result = append(*result, current)
		}

		for _, p := range parents {
			err := walkGraph(result, seen, p, path)
			if err != nil {
				return err
			}
		}
	}
```

pjbgf:
@williamfzc @lonegunmanb the tests are failing as now we are returning more references than the tests expect. Please amend [`go-git/utils/revlist2humantest.bash` ](https://github.com/go-git/go-git/blob/master/utils/revlist2humantest.bash#L26) so that the `git rev-list` returns the new references. From a [quick look](https://git-scm.com/docs/git-rev-list) we may need to add `--full-history`.

Then you can use that script to update the test cases [here](https://github.com/go-git/go-git/blob/1dbd729a387edb61c89f088cd68040085e6c81bb/references_test.go#L21).

williamfzc:
> @williamfzc @lonegunmanb the tests are failing as now we are returning more references than the tests expect. Please amend [`go-git/utils/revlist2humantest.bash` ](https://github.com/go-git/go-git/blob/master/utils/revlist2humantest.bash#L26) so that the `git rev-list` returns the new references. From a [quick look](https://git-scm.com/docs/git-rev-list) we may need to add `--full-history`.
> 
> Then you can use that script to update the test cases [here](https://github.com/go-git/go-git/blob/1dbd729a387edb61c89f088cd68040085e6c81bb/references_test.go#L21).

It seems that references_test using some external repos for test.
Should I clone/checkout all these repos and run this script on them one by one?

pjbgf:
When the references generated by `walkGraph` align with the ones returned from [revlist2humantest.bash](https://github.com/go-git/go-git/blob/master/utils/revlist2humantest.bash#L26) the tests will pass. I had another quick look and different test cases yield slightly mismatches, so it would take a combination of further digging into the `git rev-list` [docs](https://git-scm.com/docs/git-rev-list#Documentation/git-rev-list.txt---simplify-merges) and potentially amending our logic to get this fixed.

Apologies but at the moment I don't have the free time to dig in and help more. Feel free to continue trying and ping me if you need another review (or a tests kick off).

lonegunmanb:
I'm trying to fix the broken tests, but I've met a problem. I was fixing this [case](https://github.com/williamfzc/go-git/blob/master/references_test.go#L102-L114):

```go
{"https://github.com/spinnaker/spinnaker.git", "b32b2aecae2cfca4840dd480f8082da206a538da", "pylib/spinnaker/validate_configuration.py", []string{
		"a24001f6938d425d0e7504bdf5d27fc866a85c3d",
		"1e14f94bcf82694fdc7e2dcbbfdbbed58db0f4d9",
		"edf909edb9319c5e615e4ce73da47bbdca388ebe",
		"89eac7e84400db93b750414a5d52569694b9ed13",
		"1e3d328a2cabda5d0aaddc5dec65271343e0dc37",
		"b0c2aa0faeacbf9fbc144e1a375571f3d3031715",
		"e3f31c79bba1da4d60f259089199036f6dde355d",
		"427af6949a88f076bb0cd6925071c21be66b41a5",
		"9a8fda3ad2dd768bda41b7f1d6ed5d2eca464f24",
		"b5d999e2986e190d81767cd3cfeda0260f9f6fb8",
		"be54b2b7561c3d9fb8768f524fa197ca1dad981f",
	}},
```

If we checkout `https://github.com/spinnaker/spinnaker.git`, run the following command:

```shell
git rev-list --full-history --remove-empty --reverse b32b2aecae2cfca4840dd480f8082da206a538da -- pylib/spi
nnaker/validate_configuration.py | awk '{print $1}'| sed 's/^/"/;s/$/",/'
"a24001f6938d425d0e7504bdf5d27fc866a85c3d",
"1e14f94bcf82694fdc7e2dcbbfdbbed58db0f4d9",
"edf909edb9319c5e615e4ce73da47bbdca388ebe",
"89eac7e84400db93b750414a5d52569694b9ed13",
"1e3d328a2cabda5d0aaddc5dec65271343e0dc37",
"b0c2aa0faeacbf9fbc144e1a375571f3d3031715",
"e3f31c79bba1da4d60f259089199036f6dde355d",
"427af6949a88f076bb0cd6925071c21be66b41a5",
"9a8fda3ad2dd768bda41b7f1d6ed5d2eca464f24",
"b5d999e2986e190d81767cd3cfeda0260f9f6fb8",
"be54b2b7561c3d9fb8768f524fa197ca1dad981f",
```

My problem is `89eac7e84400db93b750414a5d52569694b9ed13` was not returned by `references` function. After dug into:

```go
                different, err := differentContents(path, current, parents)
		if err != nil {
			return err
		}
		if len(different) > 0 {
			*result = append(*result, current)
		}
```

It compare the file in current commit with the one in parent commits, if there's no difference then the current commit won't be added to the reference list.

I've checked `pylib/spinnaker/validate_configuration.py`'s hash:

```shell
 ~/spinnaker  $ git checkout 89eac7e84400db93b750414a5d52569694b9ed13 
Note: switching to '89eac7e84400db93b750414a5d52569694b9ed13'.
...
HEAD is now at 89eac7e Merge branch 'ewiseblatt-scripting'
 ~/spinnaker  @89eac7e8  git ls-files -s pylib/spinnaker/validate_configuration.py 
100644 6f550ff32189ba3d2c2e1eb9a1b827d13068c665 0       pylib/spinnaker/validate_configuration.py
 ~/spinnaker  $ git checkout d14f793a6cd7169ef708a4fc276ad876bd3edd4e        
Previous HEAD position was 89eac7e Merge branch 'ewiseblatt-scripting'
HEAD is now at d14f793 Merge branch 'scripting' of github.com:ewiseblatt/spinnaker into ewiseblatt-scripting
 ~/spinnaker  @d14f793a  git ls-files -s pylib/spinnaker/validate_configuration.py 
100644 6f550ff32189ba3d2c2e1eb9a1b827d13068c665 0       pylib/spinnaker/validate_configuration.py
```

So the files in two commits are identical.

Currently our `references` will filter out some commits that are identical to the parents, but it seems like `git rev-list --full-history --remove-empty` won't. How can we get the right expected test cases? @pjbgf any advice or instructions would be appreciated, the tool I'm using now has been blocked by this issue.

lonegunmanb:
Hi @pjbgf, we‘ve tried to fix the broken tests and updated the pr, would you please kick off the tests? If there're still test errors I'd like to submit my own pr based on @williamfzc's branch so we can fix this panic bug asap, thanks!

references.go
```diff
@@ -90,20 +89,15 @@ func walkGraph(result *[]*object.Commit, seen *map[plumbing.Hash]struct{}, curre
 	case 0:
```
AriehSchneier, line 89: No longer need the switch here, just change to `if len(parents) == 0 {`?

references.go
```diff
@@ -114,6 +108,19 @@ func walkGraph(result *[]*object.Commit, seen *map[plumbing.Hash]struct{}, curre
 	return nil
 }
 
+func derivedFromOneParent(path string, current *object.Commit, parents []*object.Commit) (bool, error) {
```
AriehSchneier, line 111: What does `derivedFromOneParent` mean? You are going through all of the parents?

references.go
```diff
@@ -114,6 +108,19 @@ func walkGraph(result *[]*object.Commit, seen *map[plumbing.Hash]struct{}, curre
...
+	for _, parent := range parents {
+		different, err := differentContents(path, current, []*object.Commit{parent})
```
AriehSchneier, line 113: `differentContents` already iterates the list? Why do you need to pass each of the parents individually? Can't you just pass `parents`?

references.go
lonegunmanb: Not all parents, maybe I should rename the function to `derivedFromAnyParent`.

lonegunmanb:
Hi @pjbgf this pr is no longer updated, I've opened a new one #780 based on @williamfzc's work.

williamfzc:
Thanks @lonegunmanb .
Let's continue in #780

PR #753, opened 2023-05-02
git: remote, add support for a configurable timeout.

andrewpollock:
The previous hard-coded 10 second value is too short for listing large repositories like https://gitlab.com/gitlab-org/gitlab

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -1258,7 +1258,15 @@ func (r *Remote) ListContext(ctx context.Context, o *ListOptions) (rfs []*plumbi
 }
 
 func (r *Remote) List(o *ListOptions) (rfs []*plumbing.Reference, err error) {
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	timeout := o.Timeout
+	// Default to the old hardcoded 10s value if a timeout is not explicitly set.
+	if timeout == 0 {
+		timeout = 10
+	}
+	if timeout < 0 {
+		return nil, fmt.Errorf("invalid timeout: %d", timeout)
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(timeout)*time.Second)
 	defer cancel()
 	return r.ListContext(ctx, o)
 }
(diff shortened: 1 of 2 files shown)


pjbgf (changes requested):
Please amend the `TestListTimeout` test to confirm the default value and that it can be amended (e.g. `1s`).

andrewpollock:
> Please amend the `TestListTimeout` test to confirm the default value and that it can be amended (e.g. `1s`).

Working on it. I think I need a bit of help though. The current contents of `TestListTimeout` are... confusing me because they don't appear to be working as expected, in that deelay.me seems to not be working when I try to use it manually (expired SSL certificate, instant (unexpected) 502 error) and I'm not quite sure how to test positively and negatively that timeouts have and have not occurred.

Note to future self: `go test -v . -check.v -check.f=RemoteSuite.TestListTimeout`

pjbgf:
@andrewpollock I went through `TestListTimeout` and unfortunately that test does not seem to do what it is supposed to.

remote.go
```diff
@@ -1253,7 +1253,12 @@ func (r *Remote) ListContext(ctx context.Context, o *ListOptions) (rfs []*plumbi
 }
 
 func (r *Remote) List(o *ListOptions) (rfs []*plumbing.Reference, err error) {
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	timeout := o.Timeout
```
pjbgf, line 1256: Worth adding a check here, `if o.Timeout < 0` return an error.

andrewpollock: Done in 11f41a7

andrewpollock:
> On the test front, would you be keen to refactor it to cover the new behaviour? Happy for that to be a follow-up PR.

I can but try...

pjbgf:
Please squase the commits, otherwise LGTM.

andrewpollock:
> Please squase the commits

Sorry, I'm not a big Git power user, are you asking me to rebase so this PR only contains one commit? I take you don't squash-merge PRs?

pjbgf:
@andrewpollock no worries, I actually meant squashing the commits, so that this PR only has a single commit.

andrewpollock:
> squashing the commits, so that this PR only has a single commit.

Got it, done!

pjbgf (approved):
@andrewpollock nice one, thank you for the follow up. :bow:

PR #755, opened 2023-05-03
git: fix cloning with branch name

AriehSchneier:
This fixes calling clone with a branch name rather than a full spec.

This is to match calling `git clone -b <name> <repository>`


AriehSchneier:
@pjbgf Another PR if you wouldn't mind taking a quick look

PR #756, opened 2023-05-03
git: fix the issue with submodules having the SCP style URL fail due to the wrong URL parsing

matejrisek:
When doing update for a submodule that has an SCP style URL I'm getting the following error:
```
first path segment in URL cannot contain colon
```

The offender seems to be [this line of code](https://github.com/go-git/go-git/blob/7e345bb5e163a4badefb0c54da3a057dcde50ed6/submodule.go#L136). It tries to use the go standard library method for URL parsing which cannot parse SCP style URLs.

Testing:
I'll need a bit of help with writing an integration test (if needed) for this usecase - it seems we'll have to add an additional (change existing) submodule in one of the fixtures to use the SCP style URL.

Edit:
I've added a unit test that should be enough of a safety net for this small use case.

### DIFF
--- a/submodule.go
+++ b/submodule.go
@@ -5,13 +5,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net/url"
 	"path"
 
 	"github.com/go-git/go-billy/v5"
 	"github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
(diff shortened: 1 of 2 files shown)


pjbgf:
@matejrisek please squash the commits.

matejrisek:
> @matejrisek please squash the commits.

@pjbgf sorry, forgot to do it - done

pjbgf:
Thanks for working on this @matejrisek! :bow:

PR #757, opened 2023-05-04
plumbing: move `plumbing/transport/internal/common` to `plumbing/transport/common`

Eun:
## Description
The internal transport common package claims it can be reused for other transport protocols:
```
// Package common implements the git pack protocol with a pluggable transport.
// This is a low-level package to implement new transports. Use a concrete
// implementation instead (e.g. http, file, ssh).
```

However due to the nature of being an internal package, external packages (outside of go-git) cannot use it.

## TLDR
This pull request moves the package `plumbing/transport/internal/common` to `plumbing/transport/common`, so it is accessible outside of go-git

## Why is this needed?
When implementing an new transport (outside of the go-git package) we want to reuse as much code as possible.
With this change it would be possible to reuse the git pack protocol implemented in the package.


pjbgf:
I am not opposed to this change. But I think we may need to go through all the funcs that would be accessible externally and confirm they have the correct level of documentation and input validation.

PR #758, opened 2023-05-04
git: Clone HEAD should not force master. Fixes #363

### LINKED ISSUE #249: Error with single branch clone with non-master default branch
[All new GitHub repos][1] are created with a default branch of `main`, instead of the older `master`.

If I try to clone such a repo with the `SingleBranch` option enabled, I see an error like `couldn't find remote ref "refs/heads/master"`.

This code demonstrates the issue (the repo https://github.com/hairyhenderson/hitron_coda.git has a default branch of `main`):

```go
package main

import (
	"fmt"

	"github.com/go-git/go-git/v5"
)

func main() {
	fs := memfs.New()
	storer := memory.NewStorage()

	opts := &git.CloneOptions{
		URL:          "https://github.com/hairyhenderson/hitron_coda.git",
		SingleBranch: true,
	}

	_, err := git.Clone(storer, fs, opts)
	if err != nil {
		fmt.Println(err)
	}
}
```

```console
$ go run main.go
couldn't find remote ref "refs/heads/master"
```

If I set `SingleBranch` to `false` it succeeds, but this isn't desirable for my use-case since I want to minimize what I'm pulling down.

The issue seems rooted in [this code](https://github.com/go-git/go-git/blob/v5.2.0/repository.go#L923), which defaults to the [`plumbing.Master`](https://github.com/go-git/go-git/blob/v5.2.0/plumbing/reference.go#L128) constant when `SingleBranch` is `true` and `ReferenceName` defaults to `HEAD`.

[1]: https://github.blog/changelog/2020-10-01-the-default-branch-for-newly-created-repositories-is-now-main/

### LINKED ISSUE #363: Cloning with options single branch and ref = HEAD is broken for repo with main base branch
Hey y'all ✌️🤠!

Repositories are shifting over to using `main` as their base branch. 
Today `go-git` does not work as expected when cloning a repo with main = head and the following options:
```Golang
&git.CloneOptions{
 	URL:          url,
 	Depth:        1,
 	SingleBranch: true,
 	ReferenceName: plumbing.HEAD,
 }
``` 
I _kind_ of dug around and noticed that `go-git` hardcodes to `plumbing.Master.Short()` here: 
https://github.com/go-git/go-git/blob/4ec1753b4e9324d455d3b55060020ce324e6ced2/repository.go#L905-L908

This is the git command equivalent: 
`git clone --single-branch --depth=1 {REPO_URL_WITH_MAIN_BRANCH_AS_BASE}`
The expected behavior is that it clones the repository with `HEAD` pointing to `main`. `go-git` returns an error stating `couldn't find remote ref "refs/heads/master"`.

Is this something that y'all are aware of or on y'all's road map?
I am more than happy to help contribute a fix! 
Just would need some guidance & more time to figure out how `config.RefSpec` works and why it's hardcoded to `plumbing.Master.Short()` in the switch statement above.

AriehSchneier:
Also fixes #249

### DIFF
--- a/repository_test.go
+++ b/repository_test.go
@@ -1119,6 +1119,49 @@ func (s *RepositorySuite) testCloneSingleBranchAndNonHEADReference(c *C, ref str
 	c.Assert(branch.Hash().String(), Equals, "e8d3ffab552895c19b9fcf7aa264d277cde33881")
 }
 
+func (s *RepositorySuite) TestCloneSingleBranchHEADMain(c *C) {
+	r, _ := Init(memory.NewStorage(), nil)
+
+	head, err := r.Head()
+	c.Assert(err, Equals, plumbing.ErrReferenceNotFound)
+	c.Assert(head, IsNil)
+
+	err = r.clone(context.Background(), &CloneOptions{
+		URL:          s.GetLocalRepositoryURL(fixtures.ByTag("no-master-head").One()),
+		SingleBranch: true,
+	})
+
+	c.Assert(err, IsNil)
+
+	remotes, err := r.Remotes()
+	c.Assert(err, IsNil)
+	c.Assert(remotes, HasLen, 1)
+
+	cfg, err := r.Config()
+	c.Assert(err, IsNil)
+	c.Assert(cfg.Branches, HasLen, 1)
+	c.Assert(cfg.Branches["main"].Name, Equals, "main")
+	c.Assert(cfg.Branches["main"].Remote, Equals, "origin")
+	c.Assert(cfg.Branches["main"].Merge, Equals, plumbing.ReferenceName("refs/heads/main"))
+
+	head, err = r.Reference(plumbing.HEAD, false)
+	c.Assert(err, IsNil)
+	c.Assert(head, NotNil)
+	c.Assert(head.Type(), Equals, plumbing.SymbolicReference)
+	c.Assert(head.Target().String(), Equals, "refs/heads/main")
+
+	branch, err := r.Reference(head.Target(), false)
+	c.Assert(err, IsNil)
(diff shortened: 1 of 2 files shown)


repository_test.go
```diff
@@ -1071,6 +1071,49 @@ func (s *RepositorySuite) testCloneSingleBranchAndNonHEADReference(c *C, ref str
...
 
+func (s *RepositorySuite) TestCloneSingleBranchHEADMain(c *C) {
+	r, _ := Init(memory.NewStorage(), nil)
+
+	head, err := r.Head()
+	c.Assert(err, Equals, plumbing.ErrReferenceNotFound)
+	c.Assert(head, IsNil)
+
+	err = r.clone(context.Background(), &CloneOptions{
+		URL:          "https://github.com/AriehSchneier/main.git",
```
AriehSchneier, line 1082: @pjbgf I would appreciate if you created me a repo in [`git-fixtures`](https://github.com/git-fixtures) to use instead

AriehSchneier:
Rebased

repository_test.go
pjbgf: @mcuadros do you mind creating a new repository where default branch is `main` please?

AriehSchneier: I found a fixture that has a `main` and no `master`. I forced pushed an update, the only difference is the clone URL is now from fixtures and the hash check is updated.

AriehSchneier: @pjbgf I found a fixture that has a main and no master. I forced pushed an update, the only difference is the clone URL is now from fixtures and the hash check is updated.

pjbgf (approved):
@AriehSchneier great job, thanks for working on this. :bow:

Issue #508, opened 2022-04-04
ConfigScoped lose Raw section during merge


b4nst:
When using `repo.ConfigScoped(config.GlobalScope)` or `repo.ConfigScoped(config.SystemScope)` any Raw sections will be lost during the merge.

E.g:

```txtar
> git config --global custom.foo bar
> go run main.go
-- main.go --
package main

import (
	"fmt"

	"github.com/go-git/go-git/v5"
	"github.com/go-git/go-git/v5/config"
)

func main() {
	r, _ := git.PlainOpen(".")
	c, _ := r.ConfigScoped(config.GlobalScope)
	c2, _ := config.LoadConfig(config.GlobalScope)

	fmt.Println("merged:", c.Raw.Section("custom").Option("foo"))
	fmt.Println("direct:", c2.Raw.Section("custom").Option("foo"))
}
```

gives:
```shell
merged:
direct: bar
```

wlynch:
Also ran into this - it looks like it's because mergo doesn't know to merge the values Raw section because its only merging things if the dst value is the zero-value - https://go.dev/play/p/ru9OxXgUh2K

adityasaky:
@wlynch in your example, passing in `mergo.WithOverride` seems to solve this issue. https://go.dev/play/p/XgyYmDsgKL7

Documented in the second example here: https://pkg.go.dev/github.com/imdario/mergo#hdr-Usage

PR #760, opened 2023-05-05
*: Add 'WithOverride' when merging configs

adityasaky:
Possible fix for #508

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -557,9 +557,7 @@ func (r *Repository) ConfigScoped(scope config.Scope) (*config.Config, error) {
 		return nil, err
 	}
 
-	_ = mergo.Merge(global, system)
-	_ = mergo.Merge(local, global)
-	return local, nil
+	return mergeConfigs(local, global, system)
 }
 
 // Remote return a remote if exists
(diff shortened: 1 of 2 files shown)


pjbgf:
@adityasaky please add a unit test to ensure in the future we don't regress this behaviour. Also, re-word the commit in line with the contributing guidelines - as per PR title.

pjbgf:
@adityasaky apologies for the delay getting back to you. I would simply refactor that code so that the actual merge happens on a func that receives two different configs. Then the test only focus on the new func.

adityasaky:
@pjbgf, I moved the merge into a helper. I also noticed that because we're overriding values, we must reverse the direction of overrides. So `local` is merged into `global`, overriding any collisions with the `local` value, which is then merged into `system`. I also added some tests based on how the rest are implemented, though I haven't used gopkg.in/check.v1 before, so I've probably not done it right.

Locally, I'm having issues with the last test case. I'm unable to tell if I've got something wrong with the test case or if I'm invoking the wrong checker, as the behavior seems right to me based off https://go.dev/play/p/XgyYmDsgKL7. Any thoughts?

adityasaky:
After some more tweaking, I realized the issue is with how slices are overwritten / merged. Life would get simpler if plumbing/format/config.Option were just a map[string]string instead of a struct, as the issue is with merging config.Options. mergo provides some options for merging slices but none of them really address this as it's hard to know the semantics of merging arbitrary slices...

PR #761, opened 2023-05-07
plumbing: parse the encoding header of the commit object

liwenqiu:
project may use legacy encodings, not utf-8, if that, it will keep the encoding info in commit 'encodng' header. 

when we decode the commit object, we need to parse the encoding header, other part can re-code the commit message according to this header.

### DIFF
--- a/plumbing/object/commit.go
+++ b/plumbing/object/commit.go
@@ -17,18 +17,24 @@ import (
 )
 
 const (
-	beginpgp  string = "-----BEGIN PGP SIGNATURE-----"
-	endpgp    string = "-----END PGP SIGNATURE-----"
-	headerpgp string = "gpgsig"
+	beginpgp       string = "-----BEGIN PGP SIGNATURE-----"
+	endpgp         string = "-----END PGP SIGNATURE-----"
+	headerpgp      string = "gpgsig"
+	headerencoding string = "encoding"
 
 	// https://github.com/git/git/blob/bcb6cae2966cc407ca1afc77413b3ef11103c175/Documentation/gitformat-signature.txt#L153
 	// When a merge commit is created from a signed tag, the tag is embedded in
 	// the commit with the "mergetag" header.
 	headermergetag string = "mergetag"
+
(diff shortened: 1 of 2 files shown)


plumbing/object/commit.go
```diff
@@ -46,6 +46,8 @@ type Commit struct {
 	TreeHash plumbing.Hash
 	// ParentHashes are the hashes of the parent commits of the commit.
 	ParentHashes []plumbing.Hash
+	// Encoding is the encoding of the commit.
+	Encoding string
```
pjbgf, line 50: Please create a type that represents this, and also its default value (a.k.a `UTF-8`).

```suggestion
	Encoding MessageEncoding
```

pjbgf (changes requested):
Please add some tests to ensure that:

- The right encoding is being picked up.
- The default encoding is `UTF-8`, when none is provided.

plumbing/object/commit_test.go
```diff
@@ -219,6 +220,7 @@ func (s *SuiteCommit) TestCommitEncodeDecodeIdempotent(c *C) {
 				plumbing.NewHash("f000000000000000000000000000000000000006"),
 				plumbing.NewHash("f000000000000000000000000000000000000007"),
 			},
+			Encoding: MessageEncoding("ISO-8859-1"),
```
pjbgf, line 223: We need to verify the correct encoding was set. That can be done around line `234`.

plumbing/object/commit_test.go
```diff
@@ -274,6 +278,7 @@ change
 		c.Assert(err, IsNil)
 		commit.Hash = obj.Hash()
 		c.Assert(newCommit, DeepEquals, commit)
+		newCommit.Encoding = commit.Encoding
```
pjbgf, line 281: Apologies, but we don't need this. Otherwise we should be good to merge.

```suggestion
```

pjbgf (approved):
@liwenqiu thanks for working on this. :bow:

PR #763, opened 2023-05-08
git: don't add to wants if exists, shallow and depth 1

AriehSchneier:
Depth of 1 is common, add an optimisation where we don't add the object to wants if we already have it and the required depth is 1.

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -457,7 +457,7 @@ func (r *Remote) fetch(ctx context.Context, o *FetchOptions) (sto storer.Referen
 		}
 	}
 
-	req.Wants, err = getWants(r.s, refs)
+	req.Wants, err = getWants(r.s, refs, o.Depth)
 	if len(req.Wants) > 0 {
 		req.Haves, err = getHaves(localRefs, remoteRefs, r.s, o.Depth)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


remote.go
```diff
@@ -1001,7 +1001,7 @@ func getWants(localStorer storage.Storer, refs memory.ReferenceStorage) ([]plumb
 			return nil, err
 		}
 
-		if !exists || shallow {
+		if !exists || (shallow && depth != 1) {
```
pjbgf, line 1004: @AriehSchneier I am not entirely sure this would be the case when `Tags: git.AllTags`. Do you mind expanding on why this would be the right thing to do?

AriehSchneier: The `|| shallow` was added in #311 to enable deepening, my take is that this was trying to say that if the repo is currently shallow then we need to add all commits, even the ones we already have to the list of wants because we might be trying to deepen this commit.
I think the logic could possibly be more complex here to figure out exactly when we do and when we don't need to add it, however in the case where we are trying to fetch a depth of 1 we definitely don't need to deepen the commit.
I ran a fetch across many repos and it was calling upstream and getting nothing back:
```Total 0 (delta 0), reused 0 (delta 0), pack-reused 0```
 This is a simple optimisation for the common case of `depth==1` without adding more complex logic, where the upstream commit hasn't changed we know that we don't need anything and don't need to make an extra upstream call.

AriehSchneier:
I rebased, added a comment and moved the check slightly higher to where we check shallow, this makes the code a little easier to understand

PR #764, opened 2023-05-09
git: Allow Initial Branch to be configurable

techknowlogick:
Allow the initial branch of the repo to be configurable.

This also adds `plumbing.Main` as a convenience constant so that it could be used for anyone wishing to, instead of having to define themselves.

Replaces #727 as this is non-breaking.

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -71,12 +71,28 @@ type Repository struct {
 	wt billy.Filesystem
 }
 
+type InitOptions struct {
+	// The default branch (e.g. "refs/heads/master")
+	DefaultBranch plumbing.ReferenceName
+}
+
 // Init creates an empty git repository, based on the given Storer and worktree.
 // The worktree Filesystem is optional, if nil a bare repository is created. If
 // the given storer is not empty ErrRepositoryAlreadyExists is returned
 func Init(s storage.Storer, worktree billy.Filesystem) (*Repository, error) {
+	options := InitOptions{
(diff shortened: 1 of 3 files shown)


repository.go
```diff
@@ -71,10 +71,22 @@ type Repository struct {
 	wt billy.Filesystem
 }
 
+type InitOptions struct {
+	// The default branch (e.g. "refs/heads/master")
+	InitialBranch plumbing.ReferenceName
```
pjbgf, line 76:
```suggestion
	DefaultBranch plumbing.ReferenceName
```

pjbgf:
Please add at least two tests, one that confirms that by default the `master` branch is used (for backwards compatibility), and another that uses any other branch name.

repository.go
techknowlogick: Thanks, I've updated this

techknowlogick:
Thanks @pjbgf :) I've added the tests requested

pjbgf:
@techknowlogick thank you  for adding the tests. Please squash the commits and reword them, according to the [contributing guidelines](https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md#format-of-the-commit-message) (e.g. as per PR's new title).

techknowlogick:
Thanks @pjbgf, it's now squashed and the commit message has been updated :)

pjbgf (approved):
@techknowlogick great job! Thanks for working on this. :bow:

PR #765, opened 2023-05-09
git: add a clone option to allow for shallow cloning of submodules

matejrisek:
This option matches the git clone option --shallow-submodules. 
https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---no-shallow-submodules

### DIFF
--- a/options.go
+++ b/options.go
@@ -62,6 +62,9 @@ type CloneOptions struct {
 	// within, using their default settings. This option is ignored if the
 	// cloned repository does not have a worktree.
 	RecurseSubmodules SubmoduleRescursivity
+	// ShallowSubmodules limit cloning submodules to the 1 level of depth.
+	// It matches the git command --shallow-submodules.
+	ShallowSubmodules bool
 	// Progress is where the human readable information sent by the server is
 	// stored, if nil nothing is stored and the capability (if supported)
 	// no-progress, is sent to the server to avoid send this information.
--- a/repository.go
+++ b/repository.go
@@ -900,7 +900,13 @@ func (r *Repository) clone(ctx context.Context, o *CloneOptions) error {
 		if o.RecurseSubmodules != NoRecurseSubmodules {
 			if err := w.updateSubmodules(&SubmoduleUpdateOptions{
 				RecurseSubmodules: o.RecurseSubmodules,
-				Auth:              o.Auth,
+				Depth: func() int {
+					if o.ShallowSubmodules {
+						return 1
+					}
+					return 0
+				}(),
+				Auth: o.Auth,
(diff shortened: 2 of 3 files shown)


repository.go
```diff
@@ -900,7 +900,14 @@ func (r *Repository) clone(ctx context.Context, o *CloneOptions) error {
 		if o.RecurseSubmodules != NoRecurseSubmodules {
 			if err := w.updateSubmodules(&SubmoduleUpdateOptions{
 				RecurseSubmodules: o.RecurseSubmodules,
-				Auth:              o.Auth,
+				Depth: func() int {
+					if o.ShallowSubmodules {
+						return 1
+					} else {
+						return 0
+					}
```
pjbgf, line 908: `else` is redundant in this case:

```suggestion
					}
					return 0
```

pjbgf (changes requested):
Small nit, otherwise LGTM.

options.go
```diff
@@ -62,6 +62,9 @@ type CloneOptions struct {
 	// within, using their default settings. This option is ignored if the
 	// cloned repository does not have a worktree.
 	RecurseSubmodules SubmoduleRescursivity
+	// ShallowSubmodules limit cloning submodules to the 1 level of depth.
+	// It matches the git command --shallow-submodules.
+	ShallowSubmodules bool
```
pjbgf, line 67: Please change this to `Depth` instead, to align with the API exposed in other options (e.g. `SubmoduleUpdateOptions.Depth`). Apologies for not picking this up on the first review.

matejrisek: Just to understand the philosophy behind the project better - do we strive to provide the 1 to 1 mapping between features in `git` vs `go-git`? Or do we try to provide the same functionality, but we also make the functionality more general if possible?

Why I'm asking this question is because the code above is a direct mapping of the `--shallow-submodules` from git which only allows for the depth of 1 [[git source code](https://github.com/git/git/blob/fe86abd7511a9a6862d5706c6fa1d9b57a63ba09/builtin/clone.c#L760-L761)]. To the best of my understanding there's no option in git to define the depth level of submodules.

pjbgf: My understanding is that `go-git` aims to reflect the behaviour and naming convention of upstream `git`, but sometimes this is not a 1-to-1 mapping given the different characteristics of a CLI vs a library.

I think this is quite a specific case. As `git submodule` do have a `--depth`, but when it comes down to `git clone` it has `--shallow-submodules` instead. But realistically they do slightly different things, so based on that, let's go ahead with your changes as is.

pjbgf (approved):
@matejrisek nice one, thanks for working on this. :bow:

PR #766, opened 2023-05-10
git: Fix fetching after shallow clone. Fixes #305

### LINKED ISSUE #207: Pull returns object not found
There are some related issues already though. I can reproduce the issue and may find the cause so would like to fix it.

### How to reproduce

First clone a remote repository by shallow clone with depth 1.

```
	r, err := git.CloneContext(ctx, storage, fs, &git.CloneOptions{
		URL:           _urlToRemoteRepository_,
		Auth:          ...,
		ReferenceName: plumbing.NewBranchReferenceName("master"),
		SingleBranch:  true,
		Depth:         1, // shallow clone
		Tags:          git.NoTags,
	})
```

Next, add a new commit to the master branch of the repository.  Now let's assume that we have these refs:

* `refs/remotes/origin/master`: `zzzzz`
* `refs/heads/master`: `yyyyy`
* commits history
   * `zzzzz`
   * `yyyyy`
   * `xxxxx`
   * ....

And the storer contains a commit of `yyyyy` only. With this storer, pull the latest commits:

```
	w.PullContext(ctx, &git.PullOptions{
		Auth:          ...,
		ReferenceName: plumbing.NewBranchReferenceName("master"),
		Depth:         0,
		Force:         true,
	})
```

This returns `plumbing.ErrObjectNotFound` as `object not found` error as a result.

### Expected behavior

The pull works without any errors.

### Where the error happens

1.Worktree.PullContext()
2. Remote.fetch()
3. getHaves()
4. getHavesFromRef()
5. commitPreIterator.ForEach()
6. commitPreIterator.Next() => returning error

It seems `getHaves()` function checks local storage has what commits from local refs. So `getHavesFromRef()` function tries to traverse commits from the ref hash in local but the iterator refer a commit because the storage contains a partial commits. So in the case above, the iterator tries to refer `xxxxx` commit as next of `yyyyy` but `xxxxx` is not found in the storage.

I think this is unexpected behavior but I'm not sure what is the best to fix this.

Option 1: Fix the iterator to returns `io.EOF` instead of `plumbing.ErrObjectNotFound` on `Next()` when the object is not found.
This is easier to fix the issue because no need to fix callers but it may cause another bug.

Option 2: Fix callers to ignore `plumbing.ErrObjectNotFound` where using `object.NewCommitPreorderIter`.
This may reduce to break backward incompatibility.

I tried option2 by fixing `getHavesFromRef` and `isFastForward` function in remote.go like this.
```
	err = iter.ForEach(func(c *object.Commit) error {
		...
	})
	if err != nil && err != plumbing.ErrObjectNotFound {
		return err
	}
	return nil
```

With this, the `object nof found` never happen again.

### LINKED ISSUE #305: `object not found` when trying to pull a repository cloned with `Depth: 1`

AriehSchneier:
Also fixes #207

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -837,6 +837,7 @@ func getHavesFromRef(
 	remoteRefs map[plumbing.Hash]bool,
 	s storage.Storer,
 	haves map[plumbing.Hash]bool,
+	shallows *map[plumbing.Hash]struct{},
 ) error {
 	h := ref.Hash()
 	if haves[h] {
@@ -864,6 +865,21 @@ func getHavesFromRef(
 	toVisit := maxHavesToVisitPerRef
 	return walker.ForEach(func(c *object.Commit) error {
 		haves[c.Hash] = true
+
+		// initialise the shallows map
+		if *shallows == nil {
+			shallowList, _ := s.Shallow()
+			*shallows = make(map[plumbing.Hash]struct{}, len(shallowList))
+			for _, sh := range shallowList {
+				(*shallows)[sh] = struct{}{}
+			}
+		}
+
+		// If this is a shallow reference don't try to iterate further
+		if _, ok := (*shallows)[c.Hash]; ok {
+			return storer.ErrStop
+		}
+
 		toVisit--
 		// If toVisit starts out at 0 (indicating there is no
 		// max), then it will be negative here and we won't stop
@@ -890,6 +906,9 @@ func getHaves(
 		return nil, err
 	}
 
+	// create a map of all the shallow references, but only populate on first use
+	var shallows map[plumbing.Hash]struct{}
+
 	for _, ref := range localRefs {
 		if haves[ref.Hash()] {
 			continue
@@ -899,7 +918,7 @@ func getHaves(
 			continue
 		}
 
-		err = getHavesFromRef(ref, remoteRefs, s, haves)
+		err = getHavesFromRef(ref, remoteRefs, s, haves, &shallows)
 		if err != nil {
 			return nil, err
 		}
--- a/common_test.go
+++ b/common_test.go
@@ -3,10 +3,12 @@ package git
 import (
 	"os"
 	"testing"
+	"time"
 
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/cache"
 	"github.com/go-git/go-git/v5/plumbing/format/packfile"
+	"github.com/go-git/go-git/v5/plumbing/object"
 	"github.com/go-git/go-git/v5/storage/filesystem"
 	"github.com/go-git/go-git/v5/storage/memory"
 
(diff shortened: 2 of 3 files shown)


remote.go
```diff
@@ -864,6 +865,21 @@ func getHavesFromRef(
 	toVisit := maxHavesToVisitPerRef
 	return walker.ForEach(func(c *object.Commit) error {
```
AriehSchneier, line 866: Another option would be just to completely ignore any errors returned from this `walker.ForEach` call.

pjbgf (changes requested):
The `boolean` is redundant, so I would use `struct{}` instead for a slight smaller memory profile.

remote.go
```diff
@@ -837,6 +837,7 @@ func getHavesFromRef(
 	remoteRefs map[plumbing.Hash]bool,
 	s storage.Storer,
 	haves map[plumbing.Hash]bool,
+	shallows *map[plumbing.Hash]bool,
```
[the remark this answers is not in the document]
AriehSchneier: Done!

remote.go
```diff
@@ -864,6 +865,21 @@ func getHavesFromRef(
 	toVisit := maxHavesToVisitPerRef
 	return walker.ForEach(func(c *object.Commit) error {
 		haves[c.Hash] = true
+
+		// initialise the shallows map
+		if *shallows == nil {
```
pjbgf, line 870: Deferencing a `nil` pointer leads to panic.

```suggestion
		if shallows == nil {
```

remote.go
```diff
@@ -864,6 +865,21 @@ func getHavesFromRef(
...
+			shallowList, _ := s.Shallow()
+			*shallows = make(map[plumbing.Hash]struct{}, len(shallowList))
```
pjbgf, line 872:
```suggestion
			shallows = make(map[plumbing.Hash]struct{}, len(shallowList))
```

remote.go
```diff
@@ -864,6 +865,21 @@ func getHavesFromRef(
...
+			for _, sh := range shallowList {
+				(*shallows)[sh] = struct{}{}
```
pjbgf, line 874:
```suggestion
				shallows[sh] = struct{}{}
```

pjbgf (changes requested):
Had another go at it and found a few other nits below.

pjbgf (changes requested):
Had another go at it and found a few other nits below.

remote.go
AriehSchneier: This is a private function that we only call in 1 location, it will also contain a value (I would have passed by reference if go had that feature but it doesn't). We need to check whether the location that this variable is pointing to is nil or not. I could change this to `if shallows != nil && *shallows != nil {` but IMHO that would add an extra unnecessary check.

remote.go
```diff
@@ -837,6 +837,7 @@ func getHavesFromRef(
 	remoteRefs map[plumbing.Hash]bool,
 	s storage.Storer,
 	haves map[plumbing.Hash]bool,
+	shallows *map[plumbing.Hash]struct{},
```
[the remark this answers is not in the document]
AriehSchneier: No it wont: https://dave.cheney.net/2017/04/29/there-is-no-pass-by-reference-in-go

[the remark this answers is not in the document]
AriehSchneier: PS if you test your suggested change you will find that `nil` will always be passed in and we will recalculate the map for every iteration of `range localRefs`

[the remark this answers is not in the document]
pjbgf: You are absolutely right, thanks for sharing the link. :+1:

PR #769, opened 2023-05-10
git: Testing, Fix tests not cleaning temp folders

AriehSchneier:
Also cleaned up a few Code Analysis warnings

### DIFF
--- a/common_test.go
+++ b/common_test.go
@@ -37,7 +37,7 @@ func (s *BaseSuite) TearDownSuite(c *C) {
 	s.Suite.TearDownSuite(c)
 }
 
-func (s *BaseSuite) buildBasicRepository(c *C) {
+func (s *BaseSuite) buildBasicRepository(_ *C) {
 	f := fixtures.Basic().One()
 	s.Repository = s.NewRepository(f)
 }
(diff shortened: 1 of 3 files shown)


common_test.go
```diff
@@ -133,14 +136,17 @@ func (s *BaseSuite) GetLocalRepositoryURL(f *fixtures.Fixture) string {
 
 func (s *BaseSuite) TemporalDir() (path string, clean func()) {
 	fs := osfs.New(os.TempDir())
-	path, err := util.TempDir(fs, "", "")
+	relPath, err := util.TempDir(fs, "", "")
 	if err != nil {
 		panic(err)
 	}
 
-	return fs.Join(fs.Root(), path), func() {
```
AriehSchneier, line 141: The bug was here, the first return parameter is named `path` so we were assigning the full path to the variable `path` as well as returning it, then when we called `RemoveAll` it was appending the full path to the base base (instead of just the relative path)

submodule_test.go
```diff
@@ -15,16 +14,16 @@ import (
...
+	clean    func()
 }
 
 var _ = Suite(&SubmoduleSuite{})
 
 func (s *SubmoduleSuite) SetUpTest(c *C) {
 	path := fixtures.ByTag("submodule").One().Worktree().Root()
 
-	dir, clean := s.TemporalDir()
-	defer clean()
```
AriehSchneier, line 27: This was also a bug (but wasn't causing any issues because `clean()` wasn't actually cleaning. We don't want to clean at the end of the set up, we want to clean after the test has finished (during tear down)

pjbgf (approved):

PR #778, opened 2023-05-24
git: Fix fetching after shallow clone. Fixes #305

### LINKED ISSUE #305: `object not found` when trying to pull a repository cloned with `Depth: 1`

AriehSchneier:
Also fixes https://github.com/go-git/go-git/issues/207

(Note this is another option instead of https://github.com/go-git/go-git/pull/766, but doesn't require populating the shallows map, it also adds an optimisation of not adding extra hashes to the have map that we know we aren't going to need for a shallow request)

### DIFF
--- a/common_test.go
+++ b/common_test.go
@@ -3,10 +3,12 @@ package git
 import (
 	"os"
 	"testing"
+	"time"
 
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/cache"
 	"github.com/go-git/go-git/v5/plumbing/format/packfile"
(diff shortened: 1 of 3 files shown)


pjbgf (approved):

pjbgf:
@AriehSchneier thank you for working through this and finding a better solution. :1st_place_medal:

PR #780, opened 2023-05-29
git: blame, Fix panic when contents and commits have different length

### LINKED ISSUE #603: git blame panic: contents and commits have different length

lonegunmanb:
Based on @williamfzc's work on #751, since he's [busy recently](https://github.com/williamfzc/go-git/pull/1#issuecomment-1566148925), I open this pr to continue the fix work.

This pr has solved the [issues](https://github.com/go-git/go-git/pull/751#discussion_r1208681011) that @AriehSchneier pointed out in his awesome review (thanks a lot!), and should fix #603.

I've run the tests by using [act](https://github.com/nektos/act) on my laptop, the only failed test is [`TestCheckoutIndexOS`](https://github.com/go-git/go-git/blob/master/worktree_test.go#L648), but the same test failed on `master` branch on my laptop too, so I guess it's irrelevant to this pr.

@AriehSchneier @pjbgf would you please give this pr another review? Thanks!

### DIFF
--- a/references.go
+++ b/references.go
@@ -6,8 +6,6 @@ import (
 
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/object"
-	"github.com/go-git/go-git/v5/utils/diff"
-
 	"github.com/sergi/go-diff/diffmatchpatch"
 )
 
@@ -35,7 +33,7 @@ func references(c *object.Commit, path string) ([]*object.Commit, error) {
 	sortCommits(result)
 
 	// for merges of identical cherry-picks
-	return removeComp(path, result, equivalent)
+	return removeRedundant(path, result)
 }
 
 type commitSorterer struct {
@@ -82,36 +80,42 @@ func walkGraph(result *[]*object.Commit, seen *map[plumbing.Hash]struct{}, curre
 	if err != nil {
 		return err
 	}
-	switch len(parents) {
-	// if the path is not found in any of its parents, the path was
-	// created by this commit; we must add it to the revisions list and
-	// stop searching. This includes the case when current is the
-	// initial commit.
-	case 0:
+	if len(parents) == 0 {
+		// if the path is not found in any of its parents, the path was
+		// created by this commit; we must add it to the revisions list and
+		// stop searching. This includes the case when current is the
+		// initial commit.
 		*result = append(*result, current)
 		return nil
-	case 1: // only one parent contains the path
-		// if the file contents has change, add the current commit
-		different, err := differentContents(path, current, parents)
+	}
+	skip, err := derivedFromAnyParent(path, current, parents)
+	if err != nil {
+		return err
+	}
+	if !skip {
+		*result = append(*result, current)
+	}
+
+	for _, p := range parents {
+		err := walkGraph(result, seen, p, path)
 		if err != nil {
 			return err
 		}
-		if len(different) == 1 {
-			*result = append(*result, current)
-		}
-		// in any case, walk the parent
-		return walkGraph(result, seen, parents[0], path)
-	default: // more than one parent contains the path
-		// TODO: detect merges that had a conflict, because they must be
-		// included in the result here.
-		for _, p := range parents {
-			err := walkGraph(result, seen, p, path)
-			if err != nil {
-				return err
-			}
-		}
 	}
 	return nil
+}
+
+func derivedFromAnyParent(path string, current *object.Commit, parents []*object.Commit) (bool, error) {
+	for _, parent := range parents {
+		same, err := sameContentWithParent(path, current, parent)
+		if err != nil {
+			return false, err
+		}
+		if same {
+			return true, nil
+		}
+	}
+	return false, nil
 }
 
 func parentsContainingPath(path string, c *object.Commit) ([]*object.Commit, error) {
@@ -133,20 +137,13 @@ func parentsContainingPath(path string, c *object.Commit) ([]*object.Commit, err
 	}
 }
 
-// Returns an slice of the commits in "cs" that has the file "path", but with different
-// contents than what can be found in "c".
-func differentContents(path string, c *object.Commit, cs []*object.Commit) ([]*object.Commit, error) {
-	result := make([]*object.Commit, 0, len(cs))
+func sameContentWithParent(path string, c, parent *object.Commit) (bool, error) {
 	h, found := blobHash(path, c)
 	if !found {
-		return nil, object.ErrFileNotFound
+		return false, object.ErrFileNotFound
 	}
-	for _, cx := range cs {
-		if hx, found := blobHash(path, cx); found && h != hx {
-			result = append(result, cx)
-		}
-	}
-	return result, nil
+	hx, found := blobHash(path, parent)
+	return found && h == hx, nil
 }
 
 // blobHash returns the hash of a path in a commit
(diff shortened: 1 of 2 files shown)


references.go
```diff
@@ -159,84 +155,28 @@ func blobHash(path string, commit *object.Commit) (hash plumbing.Hash, found boo
...
-
 // Returns a new slice of commits, with duplicates removed.  Expects a
-// sorted commit list.  Duplication is defined according to "comp".  It
-// will always keep the first commit of a series of duplicated commits.
-func removeComp(path string, cs []*object.Commit, comp contentsComparatorFn) ([]*object.Commit, error) {
+// sorted commit list.  Duplication is defined according to the blob hash.
+// It will always keep the first commit of a series of duplicated commits.
+func removeRedundant(path string, cs []*object.Commit) ([]*object.Commit, error) {
 	result := make([]*object.Commit, 0, len(cs))
 	if len(cs) == 0 {
```
AriehSchneier, line 163: I think you should add this at the start of this function and then remove the check here (saves extra unnecessary work):
```
	if len(cs) <= 1 {
		return cs, nil
	}
```

references.go
```diff
@@ -82,41 +80,46 @@ func walkGraph(result *[]*object.Commit, seen *map[plumbing.Hash]struct{}, curre
...
+		// stop searching. This includes the case when current is the
+		// initial commit.
 		*result = append(*result, current)
 		return nil
-	case 1: // only one parent contains the path
-		// if the file contents has change, add the current commit
-		different, err := differentContents(path, current, parents)
+	}
+	// or more than one parent contains the path
+	skip, err := derivedFromAnyParent(path, current, parents)
```
AriehSchneier, line 92: I still think this function should have a better name, or at least an explanation. Just looking at this line, I have no idea what is going on, `skip = derivedFromAnyParent`, huh?

AriehSchneier: Ok, I think I figured it out, maybe something along the lines of `same, err := sameContentAsAnyParent`?

references.go
```diff
@@ -82,41 +80,46 @@ func walkGraph(result *[]*object.Commit, seen *map[plumbing.Hash]struct{}, curre
...
+		if same {
+			return true, nil
 		}
 	}
-	return nil
+	return false, nil
 }
 
 func parentsContainingPath(path string, c *object.Commit) ([]*object.Commit, error) {
-	// TODO: benchmark this method making git.object.Commit.parent public instead of using
```
AriehSchneier, line 118: why have you removed this TODO?

references.go
```diff
@@ -133,20 +136,13 @@ func parentsContainingPath(path string, c *object.Commit) ([]*object.Commit, err
...
+func sameContent(path string, c1, c2 *object.Commit) (bool, error) {
+	h, found := blobHash(path, c1)
 	if !found {
-		return nil, object.ErrFileNotFound
-	}
-	for _, cx := range cs {
-		if hx, found := blobHash(path, cx); found && h != hx {
-			result = append(result, cx)
-		}
+		return false, object.ErrFileNotFound
```
AriehSchneier, line 142: Does this mean that if 1 parent had the file and the other didn't it will error out with file not found?

AriehSchneier (commented):
I am not reviewing the actual logic completely as I don't know enough to confirm if these changes are correct. I'm just giving my comments on code style and if I think there could be some issues that could potentially require looking at based on my read of the code (without extra context).

Another thing to check, is how expensive is `blobHash`, are we calculating it when walking the tree, and then recalculating it in the remove duplicates? Is it possible with the new logic to add 2 commits with the same `blobHash`?

references.go
lonegunmanb: It could be a merged commit so the file comes from one parent directly, so I think `derived` marks that the file is from a parent not by coincidence.

lonegunmanb:
Thanks @AriehSchneier for the further review.

>Another thing to check, is how expensive is blobHash, are we calculating it when walking the tree, and then recalculating it in the remove duplicates? Is it possible with the new logic to add 2 commits with the same blobHash?

Yes, it's possible. Two different branches with different change history (diffs in each commit are different) but with the same blob hash, the latter commit would be removed by `git rev-list` command. We need the blob hash to remove redundant commit like git cli does.

references.go
AriehSchneier: Ahh, I just realised that we are only going through the parents that we already confirmed contained the file

pjbgf:
@lonegunmanb do you mind adding a test that reproduces the issue reported by #603?

references.go
```diff
@@ -159,84 +156,28 @@ func blobHash(path string, commit *object.Commit) (hash plumbing.Hash, found boo
...
 	for i := 1; i < len(cs); i++ {
-		equals, err := comp(path, cs[i], cs[i-1])
-		if err != nil {
-			return nil, err
-		}
-		if !equals {
+		if hashes[i] != hashes[i-1] {
 			result = append(result, cs[i])
 		}
 	}
```
pjbgf, line 177: Instead of two loops and loading a range of hashes into memory:

```go
	previous := plumbing.ZeroHash
	for i, c := range cs {
		current, _ := blobHash(path, c)
		if i > 0 && previous != current {
			result = append(result, cs[i])
		}
		previous = current
	}
```

references.go
```diff
@@ -159,84 +156,28 @@ func blobHash(path string, commit *object.Commit) (hash plumbing.Hash, found boo
...
+func removeRedundant(path string, cs []*object.Commit) ([]*object.Commit, error) {
 	if len(cs) == 0 {
-		return result, nil
+		return cs, nil
 	}
+	result := make([]*object.Commit, 0, len(cs))
 	result = append(result, cs[0])
+	var hashes []plumbing.Hash
+	for _, c := range cs {
+		h, _ := blobHash(path, c)
```
pjbgf, line 170: As we are ignoring the `found` value, I would add a comment to explain why that isn't a problem.

references.go
AriehSchneier: I find this even slightly cleaner:
```
	previous, _ := blobHash(path, cs[0])
	for _, c := range cs[1:] {
		current, _ := blobHash(path, c)
		if current != previous {
			result = append(result, c)
		}
		previous = current
	}
```

lonegunmanb:
> @lonegunmanb do you mind adding a test that reproduces the issue reported by #603?

Good idea! Will do

lonegunmanb:
Hi @pjbgf I found that if I'd like to add a new test case to reproduce #603's issue, I have to open a new pr to `github.com/go-git/go-git-fixtures/v4` to add a new fixture first.

To reproduce this issue is easy, just blame `main.tf` file on [`1fca004730058eca6f11ff84e3fad1b2adf5098a`](https://github.com/Azure/terraform-azurerm-aks/commit/1fca004730058eca6f11ff84e3fad1b2adf5098a).

I'm not familiar with  `go-git-fixtures` nor git packfile, so I just guess I need commit my test fixture first by the following steps:

1. Checkout `1fca004730058eca6f11ff84e3fad1b2adf5098a`
2. `find .git/objects -type f`
3. Add `.git/objects/pack/pack-xxx.pack` file to my forked `go-git-fixtures`'s `data` folder
4. Add new fixture to `fixtures.go` file, like:

```go
{
	Tags:         []string{"packfile"},
	URL:          "https://github.com/Azure/terraform-azurerm-aks",
	Head:         "1fca004730058eca6f11ff84e3fad1b2adf5098a",
	PackfileHash: "xxx",
}
```

Please correct me if I'm wrong, thanks!

lonegunmanb:
Hi @pjbgf, a kindly ping, since I cannot find any document about how to contribute new fixture, would you kindly give me some hints on that please? Thanks!

pjbgf:
@lonegunmanb To be completely honest I wasn't sure how the `go-git-fixtures` projects worked. I had a proper look at it Today and did some housekeeping. Please take a look at the `README.md` changes I added on https://github.com/go-git/go-git-fixtures/pull/8.

The final step will take quite some time, I think for me it was around 5min.

lonegunmanb:
Hi @pjbgf thanks for the update, I've submitted a new fixture pr to the fixture repo. However, when I tried to run the new blame test on my machine I met an issue.

The repo I used to reproduce the blame error is [this](https://github.com/lonegunmanb/TestMergeRepo), it's history graph looks like:

image

The log history:

```text
18d3576
12e8a05
e275548
e8d6963
504c9f7
```

Run `git blame` on the latest commit:

```text
e8d6963c -3
e8d6963c -2
e8d6963c -1
e8d6963c 0
^504c9f7 1
^504c9f7 2
^504c9f7 3
^504c9f7 4
^504c9f7 5
e2755487 6
e2755487 7
```

But the first four lines' hash returned by `Blame` function is `18d357697e5843933e6eaef0e2fae2d4591bd707`, `[Merge pull request](https://github.com/lonegunmanb/TestMergeRepo/commit/18d357697e5843933e6eaef0e2fae2d4591bd707) https://github.com/lonegunmanb/TestMergeRepo/pull/2 [from lonegunmanb/b2](https://github.com/lonegunmanb/TestMergeRepo/commit/18d357697e5843933e6eaef0e2fae2d4591bd707)`

It looks like when git cli calculate the latest change, it didn't use the merge commit. I suspect this behavior was caused by different `diff` operations. 

This pr does solve the panic caused by the incorrect `rev-list` function, but I cannot get `Blame` work as same as git cli. Do you have any suggestions? Thanks!

lonegunmanb:
Hi @AriehSchneier, I'm reading [ngit](https://github.com/mono/ngit)'s [blame implementation](https://github.com/mono/ngit/blob/master/NGit/NGit.Blame/BlameGenerator.cs), and it seems far more complex than go-git.

I think this pr fixed the `references` function, but the `Blame` function seems still incorrect. Could we rename this pr as a `references` function patch and merge it first, and open a new pr for the `Blame` function? It seems like two bugs to me. WDYT @pjbgf @AriehSchneier ?

AriehSchneier:
`references` is only used by `blame`, IMHO it should be deleted as it is giving you a linear list of all commits for a single file removing the tree structure, which I'm not sure could be useful for anything?

If you were to commit the current PR, it would not fix all the cases of the error, and those where it doesn't return an error would almost certainly contain incorrect results. If you want to fix all the cases, then you could just push last commit every time (or at least when the number of lines don't match), but again the results would likely be incorrect.
 
I had a quick look at [gits blame implementation](https://github.com/git/git/blob/master/blame.c) but found it hard to follow, to copy would take a lot of time to read the code taking notes to fully understand what each function is actually doing (and there is quite a lot of code). I hadn't looked at ngits implementation but it does look a lot shorter.

That said, I did come up with my own algorithm last night and have already started coding it. First pass is to just get a working blame, a couple of parts I felt could definitely be optimised, but I just wanted to get a working version. (I also decided for code that was added in multiple branches of the tree that it would be best to get both and pick the older, but for first pass i would just pick the older parent when reaching a fork and not to get the candidates from both - again can be improved later).

lonegunmanb:
I'm closing this pr in favor of #789.

PR #781, opened 2023-05-29
plumbing: gitignore, fix incorrect parsing. Fixes #500

### LINKED ISSUE #500: `worktree.Status()` not respecting .gitignore

AriehSchneier:
(this replaces PR #522)

### DIFF
--- a/plumbing/format/gitignore/dir_test.go
+++ b/plumbing/format/gitignore/dir_test.go
@@ -64,6 +64,27 @@ func (s *MatcherSuite) SetUpTest(c *C) {
 	err = fs.MkdirAll("vendor/gopkg.in", os.ModePerm)
 	c.Assert(err, IsNil)
 
+	err = fs.MkdirAll("multiple/sub/ignores/first", os.ModePerm)
+	c.Assert(err, IsNil)
+	err = fs.MkdirAll("multiple/sub/ignores/second", os.ModePerm)
+	c.Assert(err, IsNil)
+	f, err = fs.Create("multiple/sub/ignores/first/.gitignore")
+	c.Assert(err, IsNil)
+	_, err = f.Write([]byte("ignore_dir\n"))
+	c.Assert(err, IsNil)
+	err = f.Close()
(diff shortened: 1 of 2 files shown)


pjbgf (approved):
@AriehSchneier nice one, thanks for working on this. :1st_place_medal:

Issue #773, opened 2023-05-20
Concurrency Issues


pjbgf:
`go-git` [is not thread-safe](https://github.com/go-git/go-git/issues/48#issuecomment-622090868). In some cases, concurrent reads may lead to repositories becoming corrupted or actions to yield invalid results (e.g. not finding an existing ref).

There is quite a bit of work until we can answer: [What operations are concurrency safe and I can call from different Goroutines? ](https://github.com/go-git/go-git/issues/48)

I created a branch to start amassing fixes around concurrency issues. As this has a performance cost, I won't propose merging it into `master` just yet. Folks insterested on using it can consume the changes via:

`go get github.com/go-git/go-git/v5@read-concurrency`

##### Current status
- [ ] Support read concurrency
  - [ ] Map concurrency issues
    - [ ] `repository.CommitObjects().Foreach()`
    - [ ] `repository.Tags().Foreach()`
  - [ ] Create concurrency tests to avoid regression
- [ ] Off-set performance costs for using mutex

Please share in the comments any specific operations you may having issues to use concurrently.

Relates to: https://github.com/go-git/go-git/pull/186 https://github.com/go-git/go-git/issues/48 https://github.com/go-git/go-git/pull/175

pjbgf:
The loading of object / pack maps (f3311ea24673e5eabb3f7b4ba0d585730548979c) and the index (377731d38d23dc9f7857757271ca90d4679d1c0a) seem to be big culprits of read concurrency. After the changes, I could no longer observe "object not found" when going through all commits and tags concurrently.

mpldr:
Thanks for the info, at least for my usecase (showing various commits from the Kernel tree), the impact was within what I'd consider to be noise. This could obviously be different for more complex operations, but I do not have the insight into the codebase to properly benchmark this.

d4x1:
I want to collect commits details with go-git/v5, but it spent too much time, so I want to collect commits concurrently.

@pjbgf  Are you still working on it?   Let me know if I can do some help.

pjbgf:
@d4x1 happy for you to take over, if you can, as I have been quite busy of late. Here are three commits (not merged) that I worked on:

https://github.com/go-git/go-git/commit/f3311ea24673e5eabb3f7b4ba0d585730548979c
https://github.com/go-git/go-git/commit/377731d38d23dc9f7857757271ca90d4679d1c0a
[...]

PR #783, opened 2023-06-02
storage: Add storage limit

### LINKED ISSUE #215: limit in memory store
It would be useful if the in memory store could be limited to a maximum amount of bytes,
so that it can be safely run on untrusted user input (otherwise they can OOM your stuff by just giving it an oversized repo)

mbialon:
Adds storage limit.

Fixes #215

### DIFF
--- a/storage/limit.go
+++ b/storage/limit.go
@@ -1,0 +1,36 @@
+package storage
+
+import (
+	"errors"
+
+	"github.com/go-git/go-git/v5/plumbing"
+)
+
+var ErrLimitExceeded = errors.New("limit exceeded")
+
+// Limited wraps git.Storer to limit the number of bytes that can be stored.
+type Limited struct {
+	Storer
+	N *int64
+}
+
+// Limit returns a git.Storer limited to the specified number of bytes.
+func Limit(s Storer, n int64) *Limited {
+	return &Limited{Storer: s, N: &n}
+}
+
+func (s *Limited) SetEncodedObject(obj plumbing.EncodedObject) (plumbing.Hash, error) {
+	*s.N -= obj.Size()
+	if *s.N < 0 {
+		return plumbing.ZeroHash, ErrLimitExceeded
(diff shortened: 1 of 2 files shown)


storage/limit.go
```diff
@@ -0,0 +1,36 @@
...
+// Limit returns a git.Storer limited to the specified number of bytes.
+func Limit(s Storer, n int64) *Limited {
+	return &Limited{Storer: s, N: &n}
+}
+
+func (s *Limited) SetEncodedObject(obj plumbing.EncodedObject) (plumbing.Hash, error) {
+	if *s.N - obj.Size() < 0 {
+		return plumbing.ZeroHash, ErrLimitExceeded
+	}
+	*s.N -= obj.Size()
```
pjbgf, line 26: [Atomic](https://pkg.go.dev/sync/atomic) counters would remove concurrency concerns here.

I understand that there are [concurrency issues](https://github.com/go-git/go-git/issues/773) within go-git already, but it would be better if new code was not affected by them as well.

jfontan (approved):

pjbgf:
It is always tricky to support features that are not supported upstream, this specific case means the result repository could be corrupted:
image

I think that _could_ be a fair price to pay if the limit is being enforced at storage level. Then I would make that extra clear on its documentation, so that users don't have unexpected results.

An upstream feature that would be useful in combination of this is [--filter](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---filterltfilter-specgt), which limits at Git level the max size of objects, and always leads to a valid repository.

HarikrishnanBalagopal:
Any progress on this?

pjbgf:
Closing as per [this comment](https://github.com/go-git/go-git/pull/783#issuecomment-1590056059). Happy to review ideas and proposals that enforce a limit on the storage, but that would be better suited at the `go-git/go-billy` repository.

PR #789, opened 2023-06-07
plumbing: blame, Complete rewrite. Fixes #603

### LINKED ISSUE #603: git blame panic: contents and commits have different length

AriehSchneier:


blame.go
```diff
@@ -30,66 +30,36 @@ func Blame(c *object.Commit, path string) (*BlameResult, error) {
 	// The file to blame is identified by the input arguments:
 	// commit and path. commit is a Commit object obtained from a Repository. Path
 	// represents a path to a specific file contained into the repository.
-	//
```
pjbgf, line 33: It would be good to keep (an updated) high-level description of the implementation.

blame_test.go
```diff
@@ -306,42 +306,45 @@ var blameTests = [...]blameTest{
...
-		repeat("1e3d328a2cabda5d0aaddc5dec65271343e0dc37", 3),
-		repeat("a24001f6938d425d0e7504bdf5d27fc866a85c3d", 12),
-		repeat("1e14f94bcf82694fdc7e2dcbbfdbbed58db0f4d9", 10),
-		repeat("a24001f6938d425d0e7504bdf5d27fc866a85c3d", 69),
-		repeat("b5d999e2986e190d81767cd3cfeda0260f9f6fb8", 7),
-		repeat("a24001f6938d425d0e7504bdf5d27fc866a85c3d", 4),
-	)},
+	/*
+		// Failed
+		{"https://github.com/spinnaker/spinnaker.git", "f39d86f59a0781f130e8de6b2115329c1fbe9545", "pylib/spinnaker/validate_configuration.py", concat(
```
pjbgf, line 311: Are you planning to get this test case to work as well?

references_test.go
```diff
@@ -18,358 +14,6 @@ type ReferencesSuite struct {
...
-			"ed887f6547d7cd2b2d741184a06f97a0a704152b",
-			"d4553dac205023fa77652308af1a2d1cf52138fb",
-			"a596972a661d9a7deca8abd18b52ce1a39516e89",
-			"66ac94f0b4442707fb6f695fbed91d62b3bd9d4a",
-			"079e42e7c979541b6fab7343838f7b9fd4a360cd",
-		}},
-	*/
-}
-
-func (s *ReferencesSuite) TestObjectNotFoundError(c *C) {
```
pjbgf, line 303: Some of the changes within references feel orthogonal to the PR purpose. Do you mind sharing your thought process on this please?

pjbgf (commented):
@AriehSchneier Nice work! I noticed that there are a few optimisation TODO's. What's the performance impact of the changes?

blame_test.go
AriehSchneier: > @AriehSchneier Nice work! I noticed that there are a few optimisation TODO's. What's the performance impact of the changes?

@pjbgf Thats a good question, its hard to know without running tests. It would be good if someone could create a couple of benchmarks that we could run against to compare.

AriehSchneier:
> @AriehSchneier Nice work! I noticed that there are a few optimisation TODO's. What's the performance impact of the changes?

@pjbgf Thats a good question, its hard to know without running tests. It would be good if someone could create a couple of benchmarks that we could run against to compare.

blame_test.go
AriehSchneier: I have not had a chance to have a look at the `failing` tests yet, that will require me to clone the tests and read the source manually. The thing to note is just because 2 implementations of blame give different results does not necessarily mean 1 of them is wrong. It can make a huge difference what algorithms and decisions you make. For example the diff algorithm will make an enormous difference to the results.

I am also trying to decide if we should go the more expensive route for finding changes across multiple branches. In the current implementation if the same change is made on 2 branches it just 'randomly' picks 1 of them (chooses the tree of the parent with the older timestamp - but that doesn't mean it will find the older change). Should we calculate the change across both branches and choose the commit with the older timestamp (I believe this would be _more correct_ but more expensive)?

blame.go
AriehSchneier: Will do (I was planning to)

references_test.go
AriehSchneier: See my comment here: https://github.com/go-git/go-git/pull/780#issuecomment-1580612255

In short, the function `references` was only used by the old implementation of blame, I believe the results that function give are not useful (if not dangerous if used wrong like it was before). It would only really be useful for a file that had no branches in its history. As it is no longer used I removed it (happy to leave it there if you prefer that).

blame_test.go
AriehSchneier: @pjbgf 
I investigated all the failing tests, all the ones still marked as failed now have their reason (different diff tool giving different line number results).

I also changed the code so that it follows both branches and discovered that 'git blame' does not use the earliest commit when a change is added twice, rather it picks the version that was merged into the current branch first - so that is what I changed it to.

references_test.go
pjbgf: You are right, originally `references.go` was [revlist/revlist.go](https://github.com/go-git/go-git/commit/d643cea1e8a6d618b2eddfdbed086c7bdf208658#diff-df6d59375198a3beb203fa89eb592cf8ba5e21ff4b62a6d0b4e61505d51cd87f) but as the code developed it got renamed and a lot of the functionality became blame-centric.

pjbgf:
I noticed that `go-git`'s version of `git blame` returns Author email instead of Author name. I think it would be worth changing that as part of the rewrite:
image

For backwards compatibility, I would add a new AuthorName (or something similar). Then we reassess the API at `v6.0`.

pjbgf:
Please an example on how to use `Blame`:

```go
package main

import (
	"fmt"
	"os"

	"github.com/go-git/go-git/v5"
	. "github.com/go-git/go-git/v5/_examples"
)

// Basic example of how to blame a repository.
func main() {
	CheckArgs("<url>", "<file_to_blame>")
	url := os.Args[1]
	path := os.Args[2]

	tmp, err := os.MkdirTemp("", "go-git-blame-*")
	CheckIfError(err)

	defer os.RemoveAll(tmp)

	// Clone the given repository.
	Info("git clone %s %s", url, tmp)
	r, err := git.PlainClone(
		tmp,
		false,
		&git.CloneOptions{
			URL:  url,
			Tags: git.NoTags,
		},
	)
	CheckIfError(err)

	// Retrieve the branch's HEAD, to then get the HEAD commit.
	ref, err := r.Head()
	CheckIfError(err)

	c, err := r.CommitObject(ref.Hash())
	CheckIfError(err)

	Info("git blame %s", path)

	// Blame the given file/path.
	br, err := git.Blame(c, path)
	CheckIfError(err)

	for i, l := range br.Lines {
		fmt.Printf("%s (%s %s %d) %s\n", l.Hash.String()[:7], l.Author, l.Date.Format("2006-01-02 15:04:05 -0700"), i+1, l.Text)
	}
}
```

pjbgf:
> @pjbgf Thats a good question, its hard to know without running tests. It would be good if someone could create a couple of benchmarks that we could run against to compare.

The previous `Blame` was not working properly, so could not bench it on a large file. The results below are for a file created on single commit:

```
         │   /tmp/old   │                /tmp/new                │
         │    sec/op    │    sec/op      vs base                 │
Blame-16   1.853m ± 10%   103.262m ± 2%  +5473.84% (p=0.002 n=6)

         │   /tmp/old   │                /tmp/new                │
         │     B/op     │     B/op       vs base                 │
Blame-16   1.115Mi ± 0%   77.680Mi ± 0%  +6868.48% (p=0.002 n=6)

         │  /tmp/old   │               /tmp/new                │
         │  allocs/op  │  allocs/op    vs base                 │
Blame-16   11.06k ± 0%   327.18k ± 0%  +2859.02% (p=0.002 n=6)
```

Code used:
```go
func BenchmarkBlame(b *testing.B) {
	b.StopTimer()

	r, err := PlainClone(b.TempDir(), false, &CloneOptions{
		URL:          "https://github.com/go-git/go-git.git",
		Tags:         NoTags,
		SingleBranch: true,
	})
	if err != nil {
		b.Error(err)
	}

	ref, err := r.Head()
	if err != nil {
		b.Error(err)
	}

	c, err := r.CommitObject(ref.Hash())
	if err != nil {
		b.Error(err)
	}

	b.StartTimer()
	for i := 0; i < b.N; i++ {
		_, err = Blame(c, "SECURITY.md")
		if err != nil {
			b.Error(err)
		}
	}
}
```

blame.go
```diff
@@ -29,67 +32,85 @@ type BlameResult struct {
...
+		c,
+		path,
+		contents,
+		needsMap,
+		0,
+		false,
+		0,
+	})
+	for {
+		items := make([]*queueItem, 0)
```
pjbgf, line 84: We may be able to reuse the backing array so it does not need to be created every time. 
So this line could be moved to before the `for`, and here you replace it with:
`items = items[:0]`

blame.go
```diff
@@ -152,96 +162,287 @@ type blame struct {
...
+						need++
+						if need >= len(curItem.NeedsMap) {
+							break out
+						}
+					}
+				case hunks[h].Type == diffmatchpatch.DiffDelete:
+					prevl += hLines
+					continue out
+				default:
+					panic("unreachable")
```
pjbgf, line 335: Let's return an error instead of using `panic`.

pjbgf:
If you are keen on bounce some ideas on this, feel free to reach out via [Slack](https://join.slack.com/t/go-git/shared_invite/zt-1tk7vwrsw-ywCUKRCo_uS_QbKMQ2ukpA).

blame.go
```diff
@@ -152,96 +162,287 @@ type blame struct {
...
-// build graph of a file from its revision history
-func (b *blame) fillGraphAndData() error {
-	//TODO: not all commits are needed, only the current rev and the prev
-	b.graph = make([][]*object.Commit, len(b.revs))
-	b.data = make([]string, len(b.revs)) // file contents in all the revisions
-	// for every revision of the file, starting with the first
-	// one...
-	for i, rev := range b.revs {
+func (b *blame) addBlames(curItems []*queueItem) (bool, error) {
+	curItem := curItems[0]
```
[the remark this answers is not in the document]
AriehSchneier: It is impossible for len(curItems) to be 0 otherwise the calling code would have returned the error `invalid state: no items left on the blame queue`

AriehSchneier:
@pjbgf all comments addressed, I didn't squash yet so you could see the diff. Once we are ready to merge I will squash.

pjbgf:
I ran the blame benchmark from current implementation vs your PR, and got this results:

```
         │  /tmp/old   │              /tmp/new              │
         │   sec/op    │   sec/op     vs base               │
Blame-16   442.8m ± 3%   573.3m ± 6%  +29.49% (p=0.002 n=6)

         │   /tmp/old   │              /tmp/new               │
         │     B/op     │     B/op      vs base               │
Blame-16   294.3Mi ± 0%   373.3Mi ± 0%  +26.83% (p=0.002 n=6)

         │  /tmp/old   │              /tmp/new              │
Blame-16   2.689M ± 0%   3.183M ± 0%  +18.37% (p=0.002 n=6)
```

The test targeted `plumbing/object/commit.go` from your PR. This looks much better.
Although it still is slower than the current, it is not by much and the changes fixes previous inaccuracies. So it LGTM. If there aren't more changes, please squash the commits.

PR #798, opened 2023-06-22
replace os.IsErrNotExist with errors.Is

Diaphteiros:
## What does this PR do?

It replaces all instances of `os.IsErrNotExist(err)` with `errors.Is(err, os.ErrNotExist)`.

## Why?

According to the doc comment on `os.IsErrNotExist`, this is the recommended way to check for specific errors.

In my case, I'm trying to inject a custom filesystem implementation into the default storer. I built a small wrapper which makes the filesystem implementation compatible to the `billy.Filesystem` interface.
However, for non-existing files, the filesystem implementation sometimes returns errors which are not recognized by `os.IsErrNotExist`, which makes the default storer implementation unusable with the filesystem implementation I'd like to use. I did some tests and found out that using `errors.Is` instead of `os.IsErrNotExist` solves the problem.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -172,7 +172,7 @@ func LoadConfig(scope Scope) (*Config, error) {
 	for _, file := range files {
 		f, err := osfs.Default.Open(file)
 		if err != nil {
-			if os.IsNotExist(err) {
+			if errors.Is(err, os.ErrNotExist) {
 				continue
 			}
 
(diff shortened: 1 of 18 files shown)


Diaphteiros:
I ran `make test` to verify this and got a single error. But I got the same error when running it on master, so it is not related to my changes.

pjbgf:
@Diaphteiros the errors are not related to your changes. Please rebase and reword the commit in line with the [contributing guide](https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md#format-of-the-commit-message).

Issue #315, opened 2021-05-19
memory leak with clone


axetroy:
When I cloned a repo about 1G in size, it consumed my 1G of memory and generated a lot of stacks

```go
// 拉取代码
func (p *Puller) Pull() (string, error) {
	p.logger.Printf("Pulling %s\n", p.repo)

	options := git.CloneOptions{
		URL:               p.repo,
		Progress:          p.writer,
		SingleBranch:      true,
		ReferenceName:     plumbing.ReferenceName("refs/heads/dev"),
		Depth:             1,
		RecurseSubmodules: git.DefaultSubmoduleRecursionDepth,
	}

	if p.auth != nil {
		options.Auth = *p.auth
	}

	tempDir := os.TempDir()

	rootPath := filepath.Join(tempDir, p.taskID)

	p.logger.Printf("拉取项目到目录 %s\n", rootPath)

	// clone project
	fs := osfs.New(rootPath)

	gitDir := filepath.Join(rootPath, ".git")

	storage := filesystem.NewStorage(osfs.New(gitDir), cache.NewObjectLRU(cache.MiByte*50))

	_, err := git.CloneContext(p.ctx, storage, fs, &options)

	if err != nil {
		return "", errors.WithStack(err)
	}

	return rootPath, nil
}

```

kbiao:
这个问题有啥进展么....我还发现 clone depth 设置为 1 的也会整个拉下来, 和 git clone --depth
 1 的效果不一样
---
Any progress on this issue .... I also found that clone depths set to 1 also pull down the whole thing, unlike ```
git clone --depth 1
``` 
it has a different effect

**Comment by axetroy** [comment]
Here is for info

```
Showing nodes accounting for 407.33MB, 99.07% of 411.16MB total
Dropped 24 nodes (cum <= 2.06MB)
Showing top 10 nodes out of 14
      flat  flat%   sum%        cum   cum%
  325.48MB 79.16% 79.16%   325.48MB 79.16%  bytes.makeSlice
   81.84MB 19.91% 99.07%   407.33MB 99.07%  github.com/go-git/go-git/v5/plumbing/format/packfile.(*Parser).get
         0     0% 99.07%       64MB 15.57%  bytes.(*Buffer).ReadFrom
         0     0% 99.07%   261.48MB 63.60%  bytes.(*Buffer).Write
         0     0% 99.07%   325.48MB 79.16%  bytes.(*Buffer).grow
[...]

Issue #251, opened 2021-02-02
clone large project , fatal error:out of memory


zhou27:
os : windows 

code:
![image](https://user-images.githubusercontent.com/24773754/106580627-1609f200-657d-11eb-9521-eccaf3c849e9.png)
git file:
![image](https://user-images.githubusercontent.com/24773754/106580469-e9ee7100-657c-11eb-9375-de9a4fc11317.png)
`.runtime: out of memory: cannot allocate 1073741824-byte block (1434288128 in use)
fatal error: out of memory

runtime stack:
runtime.throw(0x115def3, 0xd)
        C:/soft/go/src/runtime/panic.go:1116 +0x64
runtime.largeAlloc(0x3ffffe00, 0x23d0101, 0x1)
        C:/soft/go/src/runtime/malloc.go:1179 +0x15d
runtime.mallocgc.func1()
        C:/soft/go/src/runtime/malloc.go:1071 +0x39
runtime.systemstack(0x0)
        C:/soft/go/src/runtime/asm_386.s:391 +0x53
runtime.mstart()
        C:/soft/go/src/runtime/proc.go:1116

goroutine 56 [running]:
runtime.systemstack_switch()
        C:/soft/go/src/runtime/asm_386.s:352 fp=0x1284dc64 sp=0x1284dc60 pc=0xb0f3a0
runtime.mallocgc(0x3ffffe00, 0x10a6ca0, 0xf55a9b01, 0xc66cda)
        C:/soft/go/src/runtime/malloc.go:1070 +0x698 fp=0x1284dcb4 sp=0x1284dc64 pc=0xabbad8
runtime.makeslice(0x10a6ca0, 0x3ffffe00, 0x3ffffe00, 0x7e00)
        C:/soft/go/src/runtime/slice.go:98 +0x4f fp=0x1284dcc8 sp=0x1284dcb4 pc=0xaf6eff
bytes.makeSlice(0x3ffffe00, 0x0, 0x0, 0x0)
        C:/soft/go/src/bytes/buffer.go:229 +0x64 fp=0x1284dce4 sp=0x1284dcc8 pc=0xb94e14
bytes.(*Buffer).grow(0x12b33f60, 0x200, 0x7e00)
        C:/soft/go/src/bytes/buffer.go:142 +0x10c fp=0x1284dd0c sp=0x1284dce4 pc=0xb948bc
[...]

PR #799, opened 2023-06-27
plumbing: Optimise memory consumption for filesystem storage

pjbgf:
Previously, as part of building the index representation, the `resolveObject`
func would create an interim `plumbing.MemoryObject`, which would then be
saved into storage via `storage.SetEncodedObject`. This meant that objects
would be unnecessarily loaded into memory, to then be saved into disk.

The changes streamlines this process by:
- Introducing the `LazyObjectWriter` interface which enables the write
  operation to take place directly against the filesystem-based storage.
- Leverage multi-writers to process the input data once, while targeting
  multiple writers (e.g. hasher and storage).

An additional change relates to the caching of object info children within
`Parser.get`. The cache is now skipped when a seekable filesystem is being
used.

The impact of the changes can be observed when using seekable filesystem
storages, especially when cloning large repositories.

The stats below were captured by adapting the `BenchmarkPlainClone` test
to clone https://github.com/torvalds/linux.git:

```
              │  /tmp/old   │             /tmp/new              │
              │   sec/op    │   sec/op    vs base               │
PlainClone-16   41.68 ± 17%   48.04 ± 9%  +15.27% (p=0.015 n=6)

              │   /tmp/old    │               /tmp/new               │
              │     B/op      │     B/op       vs base               │
PlainClone-16   1127.8Mi ± 7%   256.7Mi ± 50%  -77.23% (p=0.002 n=6)

              │  /tmp/old   │              /tmp/new              │
PlainClone-16   3.125M ± 0%   3.800M ± 0%  +21.60% (p=0.002 n=6)
```

Notice that on average the memory consumption per operation is over `75%`
smaller. The time per operation was increased by `15%`, which may actual be less
on long running applications, due to the decreased GC pressure and the
garbage collection costs.

Relates to: https://github.com/go-git/go-git/issues/251 https://github.com/go-git/go-git/issues/315

--- 

`pprof` ref using previous stats as base:
image
image

### DIFF
--- a/plumbing/format/packfile/parser.go
+++ b/plumbing/format/packfile/parser.go
@@ -3,6 +3,7 @@ package packfile
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"io"
 
 	"github.com/go-git/go-git/v5/plumbing"
@@ -174,13 +175,25 @@ func (p *Parser) init() error {
 	return nil
 }
 
+type objectHeaderWriter func(typ plumbing.ObjectType, sz int64) error
+
+type lazyObjectWriter interface {
+	// LazyWriter enables an object to be lazily written.
+	// It returns:
+	// - w: a writer to receive the object's content.
+	// - lwh: a func to write the object header.
+	// - err: any error from the initial writer creation process.
+	//
+	// Note that if the object header is not written BEFORE the writer
+	// is used, this will result in an invalid object.
+	LazyWriter() (w io.WriteCloser, lwh objectHeaderWriter, err error)
+}
+
 func (p *Parser) indexObjects() error {
 	buf := sync.GetBytesBuffer()
 	defer sync.PutBytesBuffer(buf)
 
 	for i := uint32(0); i < p.count; i++ {
-		buf.Reset()
-
 		oh, err := p.scanner.NextObjectHeader()
 		if err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


pjbgf:
CI is currently broken, once #802 is merged this needs rebasing.

pjbgf:
Rebased and fixed issue with `ParserSuite.TestThinPack`.

plumbing/format/packfile/parser.go
```diff
@@ -378,6 +424,9 @@ func (p *Parser) get(o *objectInfo, buf *bytes.Buffer) (err error) {
 		}
 	}
 
+	// If the scanner is seekable, caching this data into
+	// memory by offset seems wasteful.
+	// !p.scanner.IsSeekable &&
```
hiddeco, line 429: Is this a `TODO`?

pjbgf: @hiddeco well spotted, thanks for that. In between performance tests I removed that as the further improvements on memory consumption had an impact on execution time. Added the `// TODO` and some additional comments for future reference. PTAL

pjbgf:
CI is broken, fix will be handled on a separate PR.

matankila:
@pjbgf please fix your branch so we all enjoy your performance boost :)

pjbgf:
PR is now rebased.

plumbing/format/packfile/patch_delta.go
```diff
@@ -265,6 +266,107 @@ func patchDelta(dst *bytes.Buffer, src, delta []byte) error {
...
+		buf := *bufp
+		cmd, err := deltaBuf.ReadByte()
+		if err == io.EOF {
+			return 0, plumbing.ZeroHash, ErrInvalidDelta
+		}
+		if err != nil {
+			return 0, plumbing.ZeroHash, err
+		}
+
+		if isCopyFromSrc(cmd) {
```
hiddeco, line 325: Seems like a `switch` may be more fitting here as there is non-conditional `else` behavior which could be captured with a `default`.

hiddeco (approved):
Besides the tiny comment, this looks good to me. Great work @pjbgf :bow: :1st_place_medal:

plumbing/format/packfile/patch_delta.go
pjbgf: Good shout. I will add this change on a follow-up PR. :+1:

PR #809, opened 2023-07-09
*: Handle paths starting with ~Username

AriehSchneier:


pjbgf (approved):
@AriehSchneier Thank you for refactoring this. :bow:

PR #812, opened 2023-07-15
storage: filesystem/dotgit, add support for tmp_objdir prefix

L11R:
more background: since dec 2021 git team changed incoming dir naming template from "incoming-*" to "tmp_objdir-incomming-*": https://github.com/git/git/commit/b3cecf49eac00d62e361bf6e6e81392f5a2fb571

now go-git works properly if it is using in pre-receive hooks

### DIFF
--- a/storage/filesystem/dotgit/dotgit.go
+++ b/storage/filesystem/dotgit/dotgit.go
@@ -582,7 +582,9 @@ func (d *DotGit) hasIncomingObjects() bool {
 		directoryContents, err := d.fs.ReadDir(objectsPath)
 		if err == nil {
 			for _, file := range directoryContents {
-				if strings.HasPrefix(file.Name(), "incoming-") && file.IsDir() {
+				if file.IsDir() && (strings.HasPrefix(file.Name(), "tmp_objdir-incoming-") ||
+					// Before Git 2.35 incoming commits directory had another prefix
+					strings.HasPrefix(file.Name(), "incoming-")) {
 					d.incomingDirName = file.Name()
 				}
 			}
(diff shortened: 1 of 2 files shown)


storage/filesystem/dotgit/dotgit.go
```diff
@@ -582,7 +582,7 @@ func (d *DotGit) hasIncomingObjects() bool {
 		directoryContents, err := d.fs.ReadDir(objectsPath)
 		if err == nil {
 			for _, file := range directoryContents {
-				if strings.HasPrefix(file.Name(), "incoming-") && file.IsDir() {
+				if strings.HasPrefix(file.Name(), "tmp_objdir-incoming-") && file.IsDir() {
```
pjbgf, line 585: To be backwards compatible, we should handle the new prefix while keeping the previous.
Add a comment calling out what version (e.g. `v2.35.0`) of `git` made the change, so we can remove the previous prefix if we ever drop support below that version.

L11R:
@pjbgf i guess this should enough.

L11R:
@pjbgf done!

pjbgf (approved):
@L11R thank you for working on this :bow:.

pjbgf:
The test failures are orthogonal to the changes introduced by this PR:

image

L11R:
@pjbgf thanks for the quick merge! Could we prepare fix release in near future? Understand that pre-receive hooks are not very popular, but still.

pjbgf:
@L11R realistically I am not sure exactly when the next release will be cut. I recommend using the pseudo version (e.g. `github.com/go-git/go-git/v5 v5.7.1-0.20230720184025-0f46b7a16fe1`) from master in the mean time:

```sh
go get github.com/go-git/go-git/v5@master
```

L11R:
Well, that was fast :)

PR #820, opened 2023-07-26
plumbing: transport, handle IPv6 while parsing endpoint. Fixes #740

### LINKED ISSUE #740: Improve IPv6 handling for host names
IPv6 host names are not being properly handled, which results in unhelpful error messages depending on the format used. Examples:

- `git@[::ffff:8c52:7903]:go-git/go-git.git` results in `dial tcp: address [:22: missing ']' in address`
- `git@::ffff:8c52:7903:go-git/go-git.git` results in `dial tcp: lookup git@: no such host`

This issue could be potentially tackled in two ways:
- (short-term) Adding error message highlighting that IPv6 is not supported for host names.
- (long-term) Add full support for IPv6 as host names.

ninedraft:
Fix host formatting for IPv6 adresses

### DIFF
--- a/plumbing/transport/common.go
+++ b/plumbing/transport/common.go
@@ -227,11 +227,17 @@ func parseURL(endpoint string) (*Endpoint, error) {
 		pass, _ = u.User.Password()
 	}
 
+	host := u.Hostname()
+	if strings.Contains(host, ":") {
+		// IPv6 address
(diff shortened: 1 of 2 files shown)


pjbgf (approved):
@ninedraft thank you for working on this. :bow:

Issue #181, opened 2020-10-12
Status() is slow with a large number of untracked files


raviqqe:
`(*Worktree) Status()` is slow when there are a large number of untracked files even if they are ignored by `.gitignore`. It is much slower than `git status`.

I imported this issue from https://github.com/src-d/go-git/issues/844.

akshaybabloo:
This is slow for even four files

j-sattler:
I have the same issue at the moment. Running `worktree.Status()` within a large worktree is quite slow. I tracked this down to the function  `func (w *Worktree) diffStagingWithWorktree(reverse bool) (merkletrie.Changes, error)` where a call to `merkletrie.DiffTree()` is calculating the list of changes between two merkletries. I have not tracked this further down, but looks like the issue is somewhere within this calculation.

steffakasid:
I experience the same. In fact I basically would like to do: `git --git-dir=/home/sid/.dof --work-tree=/home/sid status -uno` is there any way to do it? Otherwise I would get all files in home directory listed as untracked (and which also takes really long).

tcolgate:
This has been outstanding for quite a while, but I just wanted to reiterate how painful it is. We have to avoid go-git for checking status because our internal tools became unusable in our nodejs projects thanks to node_modules.
  I've investigated options for fixing, but some guidance on how the maintainers would like the problem approached would be helpful. I'm willing (and capable), of implementing a correct fix, if people can suggest one.

jwalton:
On a relatively small node.js project, a single call to `Status()` will take 7-8 seconds because of `node_modules`.  On the moderate-sized front end for the project I'm working on now, it takes 46 seconds!

tcolgate:
This still renders the Status() method unusably slow on repos with large numbers of ignored files (like nodejs/npm working directories).

codablock:
I started to analyse the underlying issue that causes slowdowns in the project I'm working and figured out that this issue is the underlying issue.

I also delved into go-git to figure out why it's slower than a regular `git status`. The reason is that `git status` does not traverse into directories if the whole directory is untracked, while go-git does. It even goes as far as calculating hashes for all untracked files.

I'm not sure what the best approach is to fix this. My current best guess is to change the `Hasher` interface to have `Hash() []byte, error` for the Hash function, so that we can implement lazy hash calculations. In the case of untracked files, it will then never trigger hash calculation and speed up the process a lot.

[...]

PR #825, opened 2023-08-06
Calculate filesystem node's hash lazily.

candid82:
The main motivation behind this change is to speed up status operation. Currently it's very slow, especially for repositories with lots of ignored files (e.g. node.js repository with node_modules directory). One of the reasons for this slowness is the fact that traversing filesystem involves calculating file hashes for all files, even if those hashes are not needed in the end because the files are in gitignore.

On my machine, for a random repository with sizable (ignored) node_modules directory this  changes bring the execution time for Worktree.Status from ~3.5s to ~1.4s. This is still very slow, but a significant improvement.

A better fix (instead of or in addition to this one) would be to avoid traversing ignored files in the first place. However, such change seem to be more intrusive and will require much deeper understanding of the codebase.

### DIFF
--- a/utils/merkletrie/filesystem/node.go
+++ b/utils/merkletrie/filesystem/node.go
@@ -29,6 +29,8 @@ type node struct {
 	hash     []byte
 	children []noder.Noder
 	isDir    bool
+	mode     os.FileMode
+	size     int64
 }
 
 // NewRootNode returns the root node based on a given billy.Filesystem.
@@ -50,6 +52,9 @@ func NewRootNode(
 //
 // The hash of a directory is always a 24-bytes slice of zero values
 func (n *node) Hash() []byte {
+	if n.hash == nil {
+		n.calculateHash()
+	}
 	return n.hash
 }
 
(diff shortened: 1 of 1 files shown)


pjbgf:
@candid82 thanks for your PR. Do you mind providing a [benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) of the performance improvements please?

Also, re-word the commit message to align with the [contribution guidelines](https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md#format-of-the-commit-message).

> A better fix (instead of or in addition to this one) would be to avoid traversing ignored files in the first place. However, such change seem to be more intrusive and will require much deeper understanding of the codebase.

Happy for this to be a follow-up PR.

candid82:
@pjbgf Below please find benchstat results for one of my local repositories with sizable node_modules directory. Obviously, the results depend on the size of git repo and the number of ignored files. 
```
> benchstat ~/old.txt  ~/new.txt
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
          │ /Users/candid/old.txt │       /Users/candid/new.txt        │
          │        sec/op         │   sec/op    vs base                │
Status-16              3.875 ± 1%   1.392 ± 3%  -64.07% (p=0.000 n=10)
```

The benchmark code:
```
func BenchmarkStatus(b *testing.B) {
	repo, _ := PlainOpen("<path to repo>")
	worktree, _ := repo.Worktree()
	for i := 0; i < b.N; i++ {
		s, _ := worktree.Status()
		if len(s) > 0 {
			b.Errorf("error!")
		}
	}
}
```

I also adjusted the commit message.

pjbgf:
Thank you for providing the benchstat and rewording the commit. 

Doing another review I noticed that this change could introduce an issue by which once the `Hash()` is called and the hash is calculated, changes to the underlying `fs` will not trigger a new hash recalculation. By looking for the places where this is used, I see no red flags that this would be a problem. Please add some comments in the `Hash()` func to that effect.

Otherwise, LGTM.

codablock:
@candid82 @pjbgf Any updates on this PR? If there is lack of time on @candid82's side, I'm also happy to provide the requested comments/docs after this PR is merged.

Merging this PR would alleviate a lot of pain in regard to https://github.com/go-git/go-git/issues/181. On my side, it would help me avoid a temporary workaround that would exec-out into `git status`, so I'm very happy to instead help here :)

candid82:
@codablock I moved on to other things and currently don't have the time to properly evaluate the concern brought up by @pjbgf in the previous comment. Please feel free to pick this work up.

codablock:
@pjbgf As requested in Slack, I have tested this PR locally (by replacing `github.com/go-git/go-git/v5` with my local version inside go.mod) and can confirm that `Status()` performance gets improved a lot. I used the notorious `node_modules` to test things out and Status() runtime has been reduced to below a second from previously 5-10 seconds.

If you merge this PR, I will follow up with a PR that adds the docs/comments.

pjbgf:
@codablock thanks for the testing. @candid82 thank you for your original contribution.

codablock:
Here is the PR with the requested comment update: https://github.com/go-git/go-git/pull/992

PR #835, opened 2023-09-04
plumbing: Do not swallow http message coming from VCS providers

matejrisek:
For diagnostics reasons we want to surface error messages coming from VCS providers. That's why we introduce the reason field to Err struct in http package. This field can be used by an end user of the library in order to better understand failures.

Context:
Recently ADO started responding with 5xx family of errors in certain cases (indicating an issue on their end) and since the library swallows errors there wasn't a way to understand what's going on.

### DIFF
--- a/plumbing/transport/http/common.go
+++ b/plumbing/transport/http/common.go
@@ -406,12 +406,26 @@ func (a *TokenAuth) String() string {
 // Err is a dedicated error to return errors based on status code
 type Err struct {
 	Response *http.Response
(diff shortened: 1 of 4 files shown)


matejrisek:
> please reword the commit based on the [contributing guide](https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md#format-of-the-commit-message).

@pjbgf sorry forgot about that - added the context

pjbgf (approved):

PR #837, opened 2023-09-11
Bump dependencies and Go to 1.19

pjbgf:
Bump Go to 1.19 and the dependencies below:

- github.com/ProtonMail/go-crypto v0.0.0-20230828082145-3c4c8a2d2371
- github.com/elazarl/goproxy v0.0.0-20230808193330-2592e75ae04a
- github.com/go-git/go-billy/v5 v5.4.2-0.20230911161450-5c1dfec7a041
- golang.org/x/crypto v0.13.0
- golang.org/x/net v0.15.0
- golang.org/x/sys v0.12.0
- golang.org/x/text v0.13.0


.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.19.x, 1.20.x]
+        go-version: [1.20.x, 1.21.x]
```
hiddeco, line 11: Isn't it a tad weird that we support the last 3 stable Go versions, but only test 2?

pjbgf: Agreed - just expanded it to the three versions we support. PTAL

hiddeco (approved):
Thanks @pjbgf ⭐️

PR #849, opened 2023-09-27
*: Add bot to close stale issues

pjbgf:
Due to the limited amount of active maintainers, adding a stale bot to concentrate the available maintainership on non stale PRs and issues.

### DIFF
--- a/.github/workflows/stale-issues-bot.yaml
+++ b/.github/workflows/stale-issues-bot.yaml
@@ -1,0 +1,33 @@
+name: "stale issues bot"
+on:
+  schedule:
+  - cron: "0 7 * * *"
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale-bot:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v8
+      with:
+        operations-per-run: 30
+        days-before-stale: 90
+        days-before-close: 30
+        stale-issue-label: stale
+        stale-pr-label: stale
+        exempt-issue-labels: no-autoclose
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: |
+          To help us keep things tidy and focus on the active tasks, we've introduced a stale bot to spot
+          issues/PRs that haven't had any activity in a while.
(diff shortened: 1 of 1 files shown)


.github/workflows/stale-issues-bot.yaml
```diff
@@ -0,0 +1,23 @@
...
+    - uses: actions/stale@v8
+      with:
+        operations-per-run: 30
+        days-before-stale: 90
+        days-before-close: 30
+        stale-issue-label: stale
+        stale-pr-label: stale
+        exempt-issue-labels: no-autoclose
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'The Stale Bot detected no activity in the past 90 days. This issue will be automatically closed if no further activity occurs in the next 30 days.'
```
hiddeco, line 23: I would be a bit more friendly to people here, and explain why we (have to) take such measures.

hiddeco (approved):

jfontan:
I am not active right now and maybe is not my role to say how to manage the project but just my two cents.

I understand the amount of work to manage the project and the huge amount of opened issues there are right now. Still I see these automatic methods of closing issues a bit rude. That is true that an opened issue for more than one year may be already fixed or is not a problem at all. Maybe adding a stale label could be used to filter out issues without activity.

This can even be used as a list for a task for newcomers to check that the problem still exists.

pjbgf:
@jfontan I agree that automatically closing issues can be perceived as suboptimal at times. The same goes for leaving unanswered issues opened for years.
Some issues are so old that the user left the previous job and is no longer using `go-git`, which is not a great experience for the users/contributors. 

> Maybe adding a stale label could be used to filter out issues without activity.

Agreed. The idea here is to help (through the new notifications that the bot will generate) a review of all the issues and PRs. In time we should be able to go through all issues that are duplicated, already fixed or unanswered. Which hopefully will make the repository look a bit healthier.

Either way, no issues are being deleted. All of them will be tagged and will be available for future reference - and reopening if needed.

Let's give this a try, and see if it works. If it does not, we can always disable the bot and try something else.

Issue #229, opened 2020-12-19
Git SHA256 support


noerw:
[SHA256 support comes in git-2.29](https://git-scm.com/docs/hash-function-transition/2.29.0). Is this library prepared for that?

pjbgf:
Closing this in favour of #706 which aims to detail support status/pending items.

PR #853, opened 2023-09-30
plumbing: commitgraph, allow SHA256 commit-graphs

zeripath:
Since the build-tag sha256 was introduced the commit graph code should be switched to use hash.Size and only use a graph if it has the correct hash version for the version of go-git that is built.

### DIFF
--- a/plumbing/format/commitgraph/file.go
+++ b/plumbing/format/commitgraph/file.go
@@ -2,12 +2,14 @@ package commitgraph
 
 import (
 	"bytes"
+	"crypto"
 	encbin "encoding/binary"
 	"errors"
 	"io"
 	"time"
 
 	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/hash"
 	"github.com/go-git/go-git/v5/utils/binary"
 )
 
@@ -35,6 +37,8 @@ var (
 	parentOctopusMask = uint32(0x7fffffff)
 	parentLast        = uint32(0x80000000)
 )
+
+const commitDataSize = 16
 
 type fileIndex struct {
 	reader              io.ReaderAt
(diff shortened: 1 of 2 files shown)


plumbing/format/commitgraph/file.go
```diff
@@ -170,8 +181,8 @@ func (fi *fileIndex) GetCommitDataByIndex(idx int) (*CommitData, error) {
 		return nil, plumbing.ErrObjectNotFound
 	}
 
-	offset := fi.commitDataOffset + int64(idx)*36
-	commitDataReader := io.NewSectionReader(fi.reader, offset, 36)
+	offset := fi.commitDataOffset + int64(idx)*(hash.Size+16)
```
pjbgf, line 184: Let's move `16` to a constant.

pjbgf (changes requested):
@zeripath nice addition. Please reword the commit in line with the [contributing guide](https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md#format-of-the-commit-message).

zeripath:
OK done.

zeripath:
I have a lot more changes to make to the commit graph code - but some of them will need to change the signature of commit data - would that be ok or should I create a v2 package?

For example, the index within a graph file should be uint32 not an int.

plumbing/format/commitgraph/file.go
zeripath: Done

pjbgf:
@zeripath merging this one as-is as it is mostly self-contained. Let's discuss the rest on the other PR.

> would that be ok or should I create a v2 package? For example, the index within a graph file should be uint32 not an int.

I would rather refrain from changing the API if possible. The support for SHA256 is blocked behind a build tag due to the amount of changes to support a different hash size without going through a shotgun surgery. But for `v6.0.0` we could think about the different API changes that it would require - maybe this could follow the same approach. But happy to consider other solutions as well.

For bouncing some ideas, feel free to reach out via [Slack](https://join.slack.com/t/go-git/shared_invite/zt-1tk7vwrsw-ywCUKRCo_uS_QbKMQ2ukpA).

pjbgf (approved):
@zeripath great job! :1st_place_medal:

pjbgf:
Relates to #229.

PR #855, opened 2023-10-01
fuzzing :  fuzz testing support for oss-fuzz integration

0x34d:
`Fuzz` testing support using [go-fuzz](https://go.dev/security/fuzz) .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,3 +42,13 @@ test-coverage:
 
 clean:
 	rm -rf $(GIT_DIST_PATH)
+
+fuzz:
+	@go test -fuzz=FuzzParser				$(PWD)/internal/revision
+	@go test -fuzz=FuzzParseSignedByte		$(PWD)/plumbing/object
+	@go test -fuzz=FuzzDecode				$(PWD)/plumbing/object
+	@go test -fuzz=FuzzNewEndpoint			$(PWD)/plumbing/transport
+	@go test -fuzz=FuzzDecoder				$(PWD)/plumbing/protocol/packp
+	@go test -fuzz=FuzzDecoder				$(PWD)/plumbing/format/config
+	@go test -fuzz=FuzzPatchDelta			$(PWD)/plumbing/format/packfile
+	@go test -fuzz=FuzzDecodeFile			$(PWD)/utils/merkletrie/internal/fsnoder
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -42,3 +42,17 @@ test-coverage:
 
 clean:
 	rm -rf $(GIT_DIST_PATH)
+
+fuzz:
```
pjbgf, line 46: Please add `make fuzz` to be executed on its own GitHub Action workflow on a [schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule).

Makefile
```diff
@@ -42,3 +42,17 @@ test-coverage:
...
+	@go test -fuzz=FuzzDecode				$(PWD)/plumbing/object
+	@go test -fuzz=FuzzNewEndpoint			$(PWD)/plumbing/transport
+	@go test -fuzz=FuzzDecoder				$(PWD)/plumbing/protocol/packp
+	@go test -fuzz=FuzzDecoder				$(PWD)/plumbing/format/config
+	@go test -fuzz=FuzzPatchDelta			$(PWD)/plumbing/format/packfile
+	@go test -fuzz=FuzzDecodeFile			$(PWD)/utils/merkletrie/internal/fsnoder
+
+fuzzbug:
+	go test -fuzz=TestDecoder				$(PWD)/plumbing/protocol/packp
+	go test -fuzz=TestPatchDelta			$(PWD)/plumbing/format/packfile
```
pjbgf, line 58: We can remove these and fix them as part of your follow-up PR.

pjbgf (changes requested):
@0x34d thanks for working on this. :bow:

Please squash the commits and reword based on the [contributing guidelines](https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md#format-of-the-commit-message).

Makefile
0x34d: > Please add make fuzz to be executed on its own GitHub Action workflow on a [schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule).


I have a better option. how about using `oss-fuzz`?

pjbgf: That's a good shout. Please tag me on the PR to `oss-fuzz`.

pjbgf (approved):
@0x34d thanks for working on this. We can enable this for CI on a follow-up PR.

Issue #863, opened 2023-10-06
config does not support duplicate URLs -- last one wins


BryanStenson-okta:
related (?) to #862 

this works:
```go
func (b *URLSuite) TestUnmarshal_DuplicateUrls(c *C) {
	input := []byte(`[core]
	bare = false
[url "ssh://git@github.com/"]
	insteadOf = https://github.com/
[url "ssh://git@github.com/"]
	insteadOf = https://google.com/
`)

	cfg := NewConfig()
	err := cfg.Unmarshal(input)
	c.Assert(err, IsNil)
	url := cfg.URLs["ssh://git@github.com/"]
	c.Assert(url.Name, Equals, "ssh://git@github.com/")

	c.Assert(url.ApplyInsteadOf("https://google.com/foobar"), Equals, "ssh://git@github.com/foobar")
}
```

this fails:
```go
func (b *URLSuite) TestUnmarshal_DuplicateUrls(c *C) {
	input := []byte(`[core]
	bare = false
[url "ssh://git@github.com/"]
	insteadOf = https://github.com/
[url "ssh://git@github.com/"]
	insteadOf = https://google.com/
`)

	cfg := NewConfig()
	err := cfg.Unmarshal(input)
	c.Assert(err, IsNil)
	url := cfg.URLs["ssh://git@github.com/"]
	c.Assert(url.Name, Equals, "ssh://git@github.com/")

	c.Assert(url.ApplyInsteadOf("https://github.com/foobar"), Equals, "ssh://git@github.com/foobar")
	c.Assert(url.ApplyInsteadOf("https://google.com/foobar"), Equals, "ssh://git@github.com/foobar")
}
```

with this error:
```
...
    c.Assert(url.ApplyInsteadOf("https://github.com/foobar"), Equals, "ssh://git@github.com/foobar")
... obtained string = "https://github.com/foobar"
... expected string = "ssh://git@github.com/foobar"
```

I expect the second test case to successfully replace both `https://github.com/foobar` and `https://google.com/foobar` with `ssh://git@github.com/foobar`

pjbgf:
@sudoforge this is merged on the `main` branch, which is the development branch for the upcoming `v6` release. At present we don't have a release date defined, but I'm keen to get an RC out in the coming months.

Closing this as done for `v6`.

PR #864, opened 2023-10-06
support multiple insteadOf keys in configuration

### LINKED ISSUE #862: config.URL only supports one insteadOf directive -- spec indicates multiple should be supported.
it looks like the "last" `insteadOf` key processed is the winner.  this works:

```go
func (b *URLSuite) TestUnmarshal_MultipleInsteadOf(c *C) {
	input := []byte(`[core]
	bare = false
[url "ssh://git@github.com/"]
	insteadOf = https://github.com/
	insteadOf = https://google.com/
`)

	cfg := NewConfig()
	err := cfg.Unmarshal(input)
	c.Assert(err, IsNil)
	url := cfg.URLs["ssh://git@github.com/"]
	c.Assert(url.Name, Equals, "ssh://git@github.com/")

	c.Assert(url.ApplyInsteadOf("https://google.com/foobar"), Equals, "ssh://git@github.com/foobar")
}
```

this fails:
```go
func (b *URLSuite) TestUnmarshal_MultipleInsteadOf(c *C) {
	input := []byte(`[core]
	bare = false
[url "ssh://git@github.com/"]
	insteadOf = https://github.com/
	insteadOf = https://google.com/
`)

	cfg := NewConfig()
	err := cfg.Unmarshal(input)
	c.Assert(err, IsNil)
	url := cfg.URLs["ssh://git@github.com/"]
	c.Assert(url.Name, Equals, "ssh://git@github.com/")

	c.Assert(url.ApplyInsteadOf("https://github.com/foobar"), Equals, "ssh://git@github.com/foobar")
	c.Assert(url.ApplyInsteadOf("https://google.com/foobar"), Equals, "ssh://git@github.com/foobar")
}
```

with this error:

```
...
    c.Assert(url.ApplyInsteadOf("https://github.com/foobar"), Equals, "ssh://git@github.com/foobar")
... obtained string = "https://github.com/foobar"
... expected string = "ssh://git@github.com/foobar"
```

BryanStenson-okta:
fixes #862 and #863


config/url.go
```diff
@@ -17,7 +17,7 @@ type URL struct {
 	Name string
 	// Any URL that starts with this value will be rewritten to start, instead, with <base>.
 	// When more than one insteadOf strings match a given URL, the longest match is used.
-	InsteadOf string
+	InsteadOf []string
```
pjbgf, line 20: Please add to the commit body (and the PR description) that this introduces a breaking change to `config.URL`.

pjbgf (changes requested):
Overall it looks good to me. Please reword the commit in line with the [contributing guidelines](https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md#format-of-the-commit-message).

config/url.go
stensonb: Should i make this an `interface` (with corresponding type casting) to try to preserve compatibility here?

I know version #s are free, but it seems a little silly to bump the major for only this...

pjbgf:
@mcuadros would you be OK with the breaking backwards compatibility of `config.URL`?

config/url.go
pjbgf: @stensonb I agree that the bump may not be worth the change.

@BryanStenson-okta I would be onboard with adding an additional `InsteadOfSlices []string` field. We would have to support both until `v6`, when we would drop the new field and change the type of the existing. What do you think?

The new field would have a deprecation note as we have in [other places around the code](https://github.com/go-git/go-git/blob/6252084d6fd3173aceeecd6765722ae844ddb266/worktree_status.go#L273).

shanduur:
Any update on this?

pjbgf:
This needs to rebase and change its target to the `v6-exp` branch.

config/url.go
stensonb: Thinking of adding either `InsteadOfSlice []string` (not plural) or `InsteadOfs []string`.

stensonb:
I can revisit this.  I'll rebase against `v6-exp`.

Should I target the final desired API in my PR and work on a "back port" to get this usable in `v5`, OR should I include the deprecated API (and supporting logic) in my PR?

pjbgf:
@stensonb thanks for that. Please the desired API against the `main` branch, which is [now](https://github.com/go-git/go-git/blob/main/CONTRIBUTING.md#branches) the development branch for `v6`.

Once that is merged, let's see how we can backport that as a "bug fix" to v5.

stensonb:
please close this PR -- let's work on the other one (https://github.com/go-git/go-git/pull/1410) , if you don't mind.

PR #866, opened 2023-10-06
Improve handling of remote errors

makkes:
Instead of simply returning the first line that the remote returned, go-git now actively searches all of stderr for lines that may contain a more actionable error message and returns that.

This is helpful in cases where the first line doesn't really indicate the actual issue, e.g. in the output of the GitLab server (over SSH):

```
$ ssh git@gitlab.com "git-receive-pack 'foo/bar'"
remote: 
remote: ========================================================================
remote: 
remote: ERROR: The project you were looking for could not be found or you don't have permission to view it.

remote: 
remote: ========================================================================
remote: 
```

In addition, this change adds a case to map the GitLab-specific error message to an ErrRepositoryNotFound error.

### DIFF
--- a/plumbing/transport/internal/common/common.go
+++ b/plumbing/transport/internal/common/common.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"regexp"
 	"strings"
 	"time"
 
@@ -28,6 +29,10 @@ const (
 
 var (
 	ErrTimeoutExceeded = errors.New("timeout exceeded")
+	// stdErrSkipPattern is used for skipping lines from a command's stderr output.
+	// Any line matching this pattern will be skipped from further
+	// processing and not be returned to calling code.
+	stdErrSkipPattern = regexp.MustCompile("^remote:( =*){0,1}$")
 )
 
 // Commander creates Command instances. This is the main entry point for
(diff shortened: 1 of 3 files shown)


plumbing/transport/internal/common/common.go
```diff
@@ -149,10 +149,29 @@ func (c *client) listenFirstError(r io.Reader) chan string {
...
+					// store the first line as a fallback in case
+					// we can't find a line indicating the actual
+					// error
+					firstLine = line
+				}
+				idx := strings.Index(line, "ERROR: ")
+				if idx != -1 {
+					errLine <- line[idx:]
+					break
+				}
```
pjbgf, line 166: We could simplify this by having a regex of line patterns we want to ignore (e.g. `^remote: (=)*(\r\n|\r|\n){0,1}$`), that way we don't need to hold anything in memory while processing the other lines.

pjbgf:
Please also reword the commit message to align with the [contributing guides](https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md).

plumbing/transport/internal/common/common.go
makkes: Even if we used the skip approach using a regex rather than the match approach as implemented here we would have to hold the first line in memory, anyway so that we can fall back to returning that which might still be better than having nothing to return in case all lines match the regex to be skipped. wdyt?

pjbgf: If the regex is very specific, it should discard only the unhelpful lines - which we would never want to return anyway. In that case, we could return the first valid line, or fallback to a generic error message.

makkes: Yeah I agree that returning `remote: ` doesn't really help. I was just trying to be cognizant of backwards-compatability but I suppose no business logic should actually depend on the text of an error. 😊 I'll adapt the PR accordingly.

makkes: I pushed the changes as agreed. ptal. 🙏🏻

plumbing/transport/internal/common/common.go
```diff
@@ -146,13 +147,21 @@ func (c *client) listenFirstError(r io.Reader) chan string {
 		return nil
 	}
 
+	skipPattern := regexp.MustCompile("^remote: (=)*(\r\n|\n|\r){0,1}$")
```
pjbgf, line 150: Please move this outside the func, so it is compiled only once instead of at every call.

PR #867, opened 2023-10-07
build: Add github workflow to check commit message format

pjbgf:
Adds automation to confirm that the commit messages align with the contributing guidelines.

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -1,0 +1,30 @@
+name: 'PR Validation'
+
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
+
+permissions:
+  contents: read
+
+jobs:
+  check-commit-message:
+    name: Check Commit Messages
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check Package Prefix
+        uses: gsactions/commit-message-checker@v2
+        with:
+          pattern: '^(\*|plumbing|utils|config|_examples|internal|storage|cli|build): .+'
+          error: |
+            Commit message(s) does not align with contribution acceptance criteria.
+
+            Refer to https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md#format-of-the-commit-message for more information.
+          excludeDescription: 'true'
+          excludeTitle: 'true'
+          checkAllCommitMessages: 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/commit.yml
```diff
@@ -0,0 +1,27 @@
...
+      - edited
+      - reopened
+      - synchronize
+
+permissions:
+  contents: read
+
+jobs:
+  check-commit-message:
+    name: Check Commit Messages
```
hiddeco, line 16: Can we think of something smarter here so that it doesn't result in `Check Commit Messages / Check Commit Messages` in the GitHub UI due to `.name` matching `.jobs[].name`?

pjbgf: Changed the redundant terms to `PR Validation > Check Commit Messages` instead. PTAL

hiddeco:
Was trying to see an example of it failing (to see if we could use a multi-line error string, Markdown and/or a link to the `CONTRIBUTING.md` document), which brought me to https://github.com/GsActions/commit-message-checker/issues/95#issuecomment-1635744898 making me feel slightly worried about the experience people will have with this.

pjbgf:
@hiddeco here's an example of a [failure](https://github.com/pjbgf/go-git/actions/runs/6440691021/job/17489752240). Then on the same PR I force pushed it to [pass](https://github.com/pjbgf/go-git/actions/runs/6440695365/job/17489761151). However, I did not try to `--amend` the commit.

Let me try to reproduce amending the commit.

hiddeco:
In this case, I would at least make the message contain the https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md URL as to my knowledge the GitHub Action UI should make this clickable automagically. If multi-lines work without any issue, maybe something like:

```yaml
error: |
  Commit message(s) does not align with CONTRIBUTING.md

  See https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md for more information.
```

pjbgf:
Nice suggestions - all changes done. As for the test of amending the commit messages, it worked fine:
- https://github.com/pjbgf/go-git/actions/runs/6462001191/job/17542819491?pr=22
- https://github.com/pjbgf/go-git/actions/runs/6462006565/job/17542835696?pr=22

I think the issue you linked may be missing `edited` as part of the `pull_request` types.

hiddeco (approved):

Issue #856, opened 2023-10-01
Bugs and found by fuzzing 


0x34d:
PR: https://github.com/go-git/go-git/pull/855

### 1 Panic 
file : plumbing/protocol/packp/uppackresp_test.go

```go
func TestDecoder(t *testing.T) {

	req := NewUploadPackRequest()
	res := NewUploadPackResponse(req)
	defer res.Close()

	res.Decode(io.NopCloser(bytes.NewReader([]byte("000600"))))
}
```

Run: `go test -fuzz=TestDecoder				$(pwd)/plumbing/protocol/packp`

Backtraces:

```log
go test -fuzz=TestDecoder				/home/0x34d/project/go-git/plumbing/protocol/packp
OK: 164 passed
--- FAIL: TestDecoder (0.00s)
panic: runtime error: slice bounds out of range [:3] with capacity 2 [recovered]
	panic: runtime error: slice bounds out of range [:3] with capacity 2

goroutine 192 [running]:
testing.tRunner.func1.2({0x7c0f00, 0xc0001e8720})
	/usr/lib/golang/src/testing/testing.go:1526 +0x24e
testing.tRunner.func1()
	/usr/lib/golang/src/testing/testing.go:1529 +0x39f
panic({0x7c0f00, 0xc0001e8720})
	/usr/lib/golang/src/runtime/panic.go:884 +0x213
github.com/go-git/go-git/v5/plumbing/protocol/packp.(*ServerResponse).decodeLine(0xc00022b600?, {0xc00028507e, 0x2, 0x2})
	/home/0x34d/project/go-git/plumbing/protocol/packp/srvresp.go:104 +0x2d1
github.com/go-git/go-git/v5/plumbing/protocol/packp.(*ServerResponse).Decode(0x40df2a?, 0xc00025d740, 0x0)
	/home/0x34d/project/go-git/plumbing/protocol/packp/srvresp.go:29 +0xd1
github.com/go-git/go-git/v5/plumbing/protocol/packp.(*UploadPackResponse).Decode(0xc00025d6e0, {0x8394d0?, 0xc0002af4d0})
	/home/0x34d/project/go-git/plumbing/protocol/packp/uppackresp.go:63 +0x337
github.com/go-git/go-git/v5/plumbing/protocol/packp.TestDecoder(0xc0002553b0?)
	/home/0x34d/project/go-git/plumbing/protocol/packp/uppackresp_test.go:150 +0x2fe
[...]

PR #868, opened 2023-10-07
plumbing: protocol/packp, Add validation for decodeLine

pjbgf:
Relates to #856.

### DIFF
--- a/plumbing/protocol/packp/srvresp.go
+++ b/plumbing/protocol/packp/srvresp.go
@@ -101,12 +101,14 @@ func (r *ServerResponse) decodeLine(line []byte) error {
 		return fmt.Errorf("unexpected flush")
 	}
 
-	if bytes.Equal(line[0:3], ack) {
-		return r.decodeACKLine(line)
-	}
+	if len(line) >= 3 {
+		if bytes.Equal(line[0:3], ack) {
+			return r.decodeACKLine(line)
+		}
 
-	if bytes.Equal(line[0:3], nak) {
-		return nil
+		if bytes.Equal(line[0:3], nak) {
+			return nil
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


plumbing/protocol/packp/srvresp.go
```diff
@@ -101,6 +101,11 @@ func (r *ServerResponse) decodeLine(line []byte) error {
 		return fmt.Errorf("unexpected flush")
 	}
 
+	err := fmt.Errorf("unexpected content %q", string(line))
+	if len(line) < 3 {
+		return err
```
hiddeco, line 106:
```suggestion
	if l := len(line); l < 3 {
		return fmt.Errorf("unexpected content %q", string(line))
```

maybe?

zeripath: That would break line 117... 

Now the current change will result in the unnecessary stringification of `line` which depending on the length of line this may or may not be a good idea.

It's probably better to do:

```go
func (r *ServerResponse) decodeLine(line []byte) error {
	if len(line) == 0 {
		return fmt.Errorf("unexpected flush")
	}

	if len(line) >= 3 {
		if bytes.Equal(line[0:3], ack) {
			return r.decodeACKLine(line)
		}
		if bytes.Equal(line[0:3], nak) {
			return nil
		}
	}
	return fmt.Errorf("unexpected content %q", string(line))
}
```

Or duplicate the `fmt.Errorf` statement, return a slightly different error e.g. `fmt.Errorf("too short content: %q", string(line))`, or test the len twice.

pjbgf: nice shout, changed for the latter approach. PTAL

hiddeco (approved):

PR #869, opened 2023-10-08
plumbing: commitgraph, Add generation v2 support

zeripath:
This PR adds in support for generation v2 support and a couple of new walkers to match --date-order etc options on log.

### DIFF
--- a/plumbing/format/commitgraph/v2/encoder.go
+++ b/plumbing/format/commitgraph/v2/encoder.go
@@ -3,6 +3,7 @@ package v2
 import (
 	"crypto"
 	"io"
+	"math"
 
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/hash"
@@ -28,13 +29,21 @@ func (e *Encoder) Encode(idx Index) error {
 	hashes := idx.Hashes()
 
 	// Sort the inout and prepare helper structures we'll need for encoding
-	hashToIndex, fanout, extraEdgesCount := e.prepare(idx, hashes)
+	hashToIndex, fanout, extraEdgesCount, generationV2OverflowCount := e.prepare(idx, hashes)
 
 	chunkSignatures := [][]byte{OIDFanoutChunk.Signature(), OIDLookupChunk.Signature(), CommitDataChunk.Signature()}
-	chunkSizes := []uint64{4 * 256, uint64(len(hashes)) * hash.Size, uint64(len(hashes)) * (hash.Size + commitDataSize)}
+	chunkSizes := []uint64{szUint32 * lenFanout, uint64(len(hashes)) * hash.Size, uint64(len(hashes)) * (hash.Size + szCommitData)}
 	if extraEdgesCount > 0 {
 		chunkSignatures = append(chunkSignatures, ExtraEdgeListChunk.Signature())
-		chunkSizes = append(chunkSizes, uint64(extraEdgesCount)*4)
+		chunkSizes = append(chunkSizes, uint64(extraEdgesCount)*szUint32)
+	}
+	if idx.HasGenerationV2() {
+		chunkSignatures = append(chunkSignatures, GenerationDataChunk.Signature())
+		chunkSizes = append(chunkSizes, uint64(len(hashes))*szUint32)
+		if generationV2OverflowCount > 0 {
+			chunkSignatures = append(chunkSignatures, GenerationDataOverflowChunk.Signature())
+			chunkSizes = append(chunkSizes, uint64(generationV2OverflowCount)*szUint64)
+		}
 	}
 
 	if err := e.encodeFileHeader(len(chunkSignatures)); err != nil {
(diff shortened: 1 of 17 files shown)


pjbgf:
@zeripath The tests for Go 1.19 are complaing of: `rightCommit.CommitTime().Compare undefined (type time.Time has no field or method Compare)`. That's because that func only arrived on [Go 1.20](https://tip.golang.org/doc/go1.20#time). 

As we support the last 3 stable versions, the implementation will need to use an alternative method to compare.

zeripath:
Ah fair enough.  I'm currently working out a bug in the chain stuff and adding a testcase too.

zeripath:
OK I finally have this sorted now.

plumbing/object/commitgraph/commitnode_walker_date_order.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/emirpasic/gods/trees/binaryheap"
+)
+
+// NewCommitNodeIterDateOrder returns a CommitNodeIter that walks the commit history,
+// starting at the given commit and visiting its parents in Committer Time and Generation order,
+// but with the  constraint that no parent is emitted before its children are emitted.
+//
+// This matches `git log --date-order`
+func NewCommitNodeIterDateOrder(c CommitNode,
+	seenExternal map[plumbing.Hash]bool,
```
pjbgf, line 15: If all items in `seenExternal` are `true`, this could also be `map[plumbing.Hash]struct{}`.

plumbing/object/commitgraph/commitnode_walker_helper.go
```diff
@@ -0,0 +1,156 @@
...
+	l.l = append(l.l, c)
+}
+
+// Pop pops the most recently added CommitNode from the stack
+func (l *commitNodeLifo) Pop() (CommitNode, bool) {
+	if len(l.l) == 0 {
+		return nil, false
+	}
+	c := l.l[len(l.l)-1]
+	l.l = l.l[:len(l.l)-1]
```
pjbgf, line 35: I suspect this could cause a "slice memory leak".

plumbing/object/commitgraph/commitnode_walker_topo_order.go
```diff
@@ -0,0 +1,159 @@
...
+		if generationV2 {
+			if pc.GenerationV2() < minimumLevel {
+				minimumLevel = pc.GenerationV2()
+			}
+		} else {
+			generationV2 = false
+			if pc.Generation() < minimumLevel {
+				minimumLevel = pc.Generation()
+			}
+		}
```
pjbgf, line 81: We could simplify this by ditching the `else` and removing the redudant line `77`.

```suggestion
		if generationV2 {
			if pc.GenerationV2() < minimumLevel {
				minimumLevel = pc.GenerationV2()
			}
			continue
		}
		
		if pc.Generation() < minimumLevel {
			minimumLevel = pc.Generation()		
		}
```

plumbing/format/commitgraph/v2/encoder.go
```diff
@@ -28,14 +29,22 @@ func (e *Encoder) Encode(idx Index) error {
...
 	if extraEdgesCount > 0 {
 		chunkSignatures = append(chunkSignatures, ExtraEdgeListChunk.Signature())
 		chunkSizes = append(chunkSizes, uint64(extraEdgesCount)*4)
 	}
+	if idx.HasGenerationV2() {
+		chunkSignatures = append(chunkSignatures, GenerationDataChunk.Signature())
+		chunkSizes = append(chunkSizes, uint64(len(hashes))*4)
+		if generationV2OverflowCount > 0 {
+			chunkSignatures = append(chunkSignatures, GenerationDataOverflowChunk.Signature())
+			chunkSizes = append(chunkSizes, uint64(generationV2OverflowCount)*8)
```
pjbgf, line 45: Please replace `4` and `8` with constants.

pjbgf (changes requested):

plumbing/object/commitgraph/commitnode_walker_date_order.go
zeripath: This was chosen to match the signature for NewCommitNodeIterCTime

plumbing/object/commitgraph/commitnode_walker_helper.go
zeripath: Yes, The slice that is used will not reduce in capacity but the slice only exists for the time that the `commitNodeLifo` exists.

This is only used in `NewCommitNodeIterTopoOrder` as the stack for the nodes to that are ready to be printed. Therefore the slice should not get too large and would tend to fluctuate between 1 & 2 nodes up to the max number of active branches for a commit - which would tend to be a similar across the history.

We could make a more complex Lifo, one that say, has a permanent minimum capacity of 8 and if the length falls below a quarter of the current capacity it would reduce to half capacity if preferred

pjbgf: Let's keep as is for simplicity - we can revisit it later if worst comes to worst.

pjbgf (approved):
@zeripath really good job! :1st_place_medal:

Issue #874, opened 2023-10-12
Fast forward only fetch not handled for custom ref


adityasaky:
In [gittuf](https://github.com/gittuf/gittuf), we store additional metadata in custom refs. We noticed that when we constructed a refspec without a leading `+` (eg: `refs/gittuf/reference-state-log:refs/gittuf/reference-state-log`) the destination still gets overwritten after a call to `remote.FetchContext` when the local and remote versions of the ref have diverged.

I'm not 100% sure this is the reason but I wonder if this is the issue: https://github.com/go-git/go-git/blob/master/remote.go#L1203. In this line, our refspecs never trigger the `IsBranch` condition because our ref isn't a typical branch in `refs/heads`. This means that `forceNeeded` doesn't get set, and the destination ref in the local repository gets overwritten.

If there's some clarity on why that condition exists, I'm happy to work on a patch to enable our use. Is the idea to dismiss tags? If so, can that check be explicitly for `refs/tags`?

adityasaky:
Update: I confirmed we're hitting that issue because of the `IsBranch` check. I vendored the library and removed that specific check on that line, and the corresponding test passed.

adityasaky:
I just opened #875 to flip that check from `IsBranch` to `!IsTag`. Happy to iterate there if the solution makes sense!

adityasaky:
It's possible a similar issue exists here: https://github.com/go-git/go-git/blob/master/repository.go#L1090-L1098

adityasaky:
> It's possible a similar issue exists here: https://github.com/go-git/go-git/blob/master/repository.go#L1090-L1098

I think it's harder to say here and we can't use the same fix where we switch to a tag check because that's not enough to determine it's in detached mode. I think this issue may be okay to close? If we run into an issue in that code block, happy to track it separately then. :smile:

PR #875, opened 2023-10-12
remote: Flip clause for fast-forward only check

adityasaky:
In remote.updateLocalReferenceStorage, this commit updates a check to see if the reference is for a branch (in refs/heads) to checking the reference is not a tag. This change ensures that the subsequent fast-forward only handling clauses apply to references that are not standard branches stored in refs/heads.

I'm still missing tests etc here. If this fix makes sense, happy to work towards that (though I might need some pointers again).

#874

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -1198,9 +1198,9 @@ func (r *Remote) updateLocalReferenceStorage(
 			old, _ := storer.ResolveReference(r.s, localName)
 			new := plumbing.NewHashReference(localName, ref.Hash())
 
-			// If the ref exists locally as a branch and force is not specified,
-			// only update if the new ref is an ancestor of the old
-			if old != nil && old.Name().IsBranch() && !force && !spec.IsForceUpdate() {
+			// If the ref exists locally as a non-tag and force is not
+			// specified, only update if the new ref is an ancestor of the old
+			if old != nil && !old.Name().IsTag() && !force && !spec.IsForceUpdate() {
 				ff, err := isFastForward(r.s, old.Hash(), new.Hash())
 				if err != nil {
 					return updated, err
--- a/remote_test.go
+++ b/remote_test.go
@@ -4,16 +4,20 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
+	"testing"
 	"time"
 
+	"github.com/go-git/go-billy/v5/memfs"
 	"github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/cache"
+	"github.com/go-git/go-git/v5/plumbing/object"
 	"github.com/go-git/go-git/v5/plumbing/protocol/packp"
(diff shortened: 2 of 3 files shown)


adityasaky:
I tried a couple times to get the commit message right but looks like I'm still missing something. :(

pjbgf:
> I tried a couple times to get the commit message right but looks like I'm still missing something. :(

That was actually my bad, I forgot to add `git` as a valid prefix. Do you mind [adding that](https://github.com/go-git/go-git/blob/master/.github/workflows/pr-validation.yml#L22) as a separate commit please?

remote.go
```diff
@@ -1200,7 +1200,7 @@ func (r *Remote) updateLocalReferenceStorage(
 
 			// If the ref exists locally as a branch and force is not specified,
 			// only update if the new ref is an ancestor of the old
-			if old != nil && old.Name().IsBranch() && !force && !spec.IsForceUpdate() {
+			if old != nil && !old.Name().IsTag() && !force && !spec.IsForceUpdate() {
```
pjbgf, line 1203: Please amend the comments above to align with the changes.

pjbgf:
> I'm still missing tests etc here. If this fix makes sense, happy to work towards that (though I might need some pointers again).

As for tests, here's an example around fetching: https://github.com/go-git/go-git/blob/master/remote_test.go#L409.

adityasaky:
@pjbgf I actually just added some tests for this but I couldn't work with the existing framework :cry:

But I think the case shows the change in behavior, please let me know if it makes sense.

adityasaky:
> That was actually my bad, I forgot to add git as a valid prefix. Do you mind [adding that](https://github.com/go-git/go-git/blob/master/.github/workflows/pr-validation.yml#L22) as a separate commit please?

Done! That'll still fail because I didn't want to add my tests to my main commit until we figured out how to integrate them into the existing framework.

remote.go
adityasaky: Done!

adityasaky:
Re: https://github.com/go-git/go-git/issues/874#issuecomment-1762320375

I _think_ I hit this when I tried to use Clone with ReferenceName set to the custom ref in the test. At the time, I wasn't setting the master branch at all, but right now the test sets a master branch, clones the repo, then fetches the custom ref for the first time separately.

adityasaky:
I squashed my tests commit to hopefully make CI go green. I still want to highlight that the test setup deviates a fair bit from the default test framework in the repository.

remote_test.go
```diff
@@ -1555,3 +1559,140 @@ func (s *RemoteSuite) TestFetchAfterShallowClone(c *C) {
...
+		},
+	}); err != nil {
+		t.Fatal(err)
+	}
+
+	// 4. Make divergent changes
+	remoteCommitID := writeCommitToRef(t, remoteRepo, customRef, emptyTreeID)
+	// Sleep so that consecutive calls to writeCommitToRef have different
+	// time.Now() values
+	time.Sleep(time.Second)
```
pjbgf, line 1600: Instead of sleeping, we could send time as a parameter for `writeCommitToRef`.

```suggestion
```

adityasaky: Done!

pjbgf (approved):
Thanks for working on this @adityasaky! :bow:

adityasaky:
thank you!

PR #876, opened 2023-10-12
Added initial support for git wire protocol v2 

blmayer:
Hi. In this PR I implemented some parts of the v2 protocol. I have created some tests, but the overall work needs improvements. As this is my first time coding I think some feedback here is cool, since I'm not sure if I coded this in the best place.

The v2 protocol uses commands to negotiate packfiles, I added the initial capability advertisement, ls-refs and fetch commands. Inside each command capabilities change how it is done, I added only the necessary to make it work.

I used this repo to test: https://github.com/blmayer/gwi

So, I'd like some help so I can wrap this up and merge.

Thanks!


plumbing/protocol/packp/comreq.go
```diff
@@ -0,0 +1,171 @@
...
+	case capability.Fetch.String():
+		// ofsdelta
+		var delta uint
+		if res.Args.Supports(capability.OFSDelta) {
+			delta = 50
+		}
+
+		// if done was sent skip acks section
+		if len(res.Haves) > 0 {
+			if res.Args.Supports("done") {
```
pjbgf, line 147: Let's create a const for `done`.

plumbing/transport/server/server.go
```diff
@@ -79,6 +80,7 @@ func (h *handler) NewReceivePackSession(s storer.Storer) (transport.ReceivePackS
 }
 
 type session struct {
+	version  int
```
pjbgf, line 83: Let's create a type for version which defaults to `v1` and gets populated based on the protocol negotiation.

pjbgf (commented):
@blmayer great job progressing the `v2` support. :clap:

I will take a look over the weekend on the failing tests - but my initial impression is that the session needs to be aware of the protocol version and at times have different flows, more on this once I get to the bottom of the issues.

plumbing/protocol/packp/advcaps.go
```diff
@@ -0,0 +1,95 @@
...
+	pe.Flush()
+	pe.EncodeString("version 2\n")
+
+	for _, c := range a.Capabilities.All() {
+		vals := a.Capabilities.Get(c)
+		if len(vals) > 0 {
+			pe.EncodeString(c.String() + "=" + strings.Join(vals, " ") + "\n")
+		} else {
+			pe.EncodeString(c.String() + "\n")
+		}
```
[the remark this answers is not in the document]
blmayer: Not sure if this suggestion works, since each invocation of `EncodeString` will create the line size prefix (part of the pktline protocol). And on v2 all that payload must be on the same line.

blmayer:
Thanks for reviewing this. I'll work on your suggestions on the following days.

aymanbagabas (commented):
Great job here! This is going to be a big update 🙂
It might be a good idea also to support Protocol V1 which is simply V0 plus version line before advertising references. 

See: https://github.com/git/git/blob/master/protocol.h#L25, https://github.com/git/git/blob/master/builtin/upload-pack.c#L60-L68, https://stackoverflow.com/questions/66743099/whats-the-difference-between-git-protocol-v1-and-v0-v2

pjbgf:
> It might be a good idea also to support Protocol V1 which is simply V0 plus version line before advertising references

Let's treat v1 as its own thing, once v2 is implemented. But I agree that by the looks of [v1 upstream](https://github.com/search?q=repo%3Agit%2Fgit+protocol_v1&type=code), it should be fairly straight forward to implement it.

plumbing/protocol/packp/advcaps.go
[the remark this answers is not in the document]
pjbgf: You are absolutely right, updated it to handle that as a var of its own.

plumbing/protocol/packp/advcaps.go
```diff
@@ -0,0 +1,95 @@
...
+	return pe.Flush()
+}
+
+func (a *AdvCaps) Decode(r io.Reader) error {
+	s := pktline.NewScanner(r)
+
+	// decode # SP service=<service> LF
+	s.Scan()
+	f := string(s.Bytes())
+	if i := strings.Index(f, "service="); i < 0 {
```
pjbgf, line 58:
```suggestion
	i := strings.Index(f, "service="); 
	if i < 0 {
```

plumbing/protocol/packp/comreq.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+	}
+
+	return nil
+}
+
+// CommandResponse values represent the information transmitted on a
+// command response message used in wire protocol v2. Values from this type
+// are not zero-value safe, use the New function instead.
+type CommandResponse struct {
```
aymanbagabas, line 113: Perhaps we could make `CommandResponse` an interface with a private method, and then define different command responses to implement this interface. From what i can tell, each v2 command has a different response structure

PR #880, opened 2023-10-14
plumbing: commitgraph, add function to build graphs

zeripath:
Make builder function for creating commitgraphs which could be written with the functionality in encoder.go.

### DIFF
--- a/plumbing/object/commitgraph/commitnode_graph_builder.go
+++ b/plumbing/object/commitgraph/commitnode_graph_builder.go
@@ -1,0 +1,278 @@
+package commitgraph
+
+import (
+	"errors"
+	"math"
+
+	"github.com/go-git/go-git/v5/plumbing"
+	commitgraph "github.com/go-git/go-git/v5/plumbing/format/commitgraph/v2"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/plumbing/storer"
+)
+
+type CreateCommitNodeGraphOptions struct {
+	// Commits holds the commits to be iterated upon. If nil, then all commits in the storage are used as potential heads, otherwise only the commits in the iter are used.
+	Commits object.CommitIter
+	// Append defines whether all commits in the old index should be appended to the new index.
+	Append bool
+	// Chain defines whether the new index should be chained to the old index. Can only be used if Append is false and the old index has generation v2.
+	Chain bool
+}
+
+// CreateCommitGraph takes a provided storage and possible old Index and creates a new commit graph.
+// To create a pack specific graph pass in a storer that is restricted to the packfile.
+func CreateCommitGraph(s storer.EncodedObjectStorer, old commitgraph.Index, opts CreateCommitNodeGraphOptions) (commitgraph.Index, error) {
+	commitIndex := NewGraphCommitNodeIndex(old, s)
+
+	var parent commitgraph.Index
+	if opts.Chain && old != nil && old.HasGenerationV2() {
+		parent = old
+	}
+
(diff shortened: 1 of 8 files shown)


plumbing/object/commitgraph/commitnode_graph_builder.go
```diff
@@ -0,0 +1,231 @@
...
+	"github.com/go-git/go-git/v5/plumbing"
+	commitgraph "github.com/go-git/go-git/v5/plumbing/format/commitgraph/v2"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/plumbing/storer"
+)
+
+type CreateCommitNodeGraphOptions struct {
+	Commits object.CommitIter // If nil then all commits in the storage are used as potential heads, otherwise only the commits in the iter are used
+	Append  bool              // Append all commits in the old index to the new index
+	Chain   bool              // Chain the new index to the old index (only valid if Append is false and the old index has generation v2)
```
pjbgf, line 16:
```suggestion
	// Commits holds the commits to be iterated upon. If nil, then all commits in the storage are used as potential heads, otherwise only the commits in the iter are used.
	Commits object.CommitIter
	// Append defines whether all commits in the old index should be appended to the new index.
	Append  bool
        // Chain defines whether the new index should be chained to the old index. Can only be used if Append is false and the old index has generation v2.
	Chain   bool
```

plumbing/object/commitgraph/commitnode_graph_builder.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// CreateCommitGraph takes a provided storage and possble old Index and creates a new commit graph.
```
pjbgf, line 19:
```suggestion
// CreateCommitGraph takes a provided storage and possible old Index and creates a new commit graph.
```

plumbing/object/commitgraph/commitnode_graph_builder.go
```diff
@@ -0,0 +1,231 @@
...
+// To create a pack specific graph pass in a storer that is restricted to the packfile.
+func CreateCommitGraph(s storer.EncodedObjectStorer, old commitgraph.Index, opts CreateCommitNodeGraphOptions) (commitgraph.Index, error) {
+	// TODO: Implement splits
```
pjbgf, line 22: Please expand on the command on why we need to do that.

plumbing/object/commitgraph/commitnode_graph_builder.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	if isValid(commitNode) {
+		// commitData can be copied in to the graph
+		c.copyCommitNode(commit.Hash, commitNode)
+		return nil
+	}
+
+	// Not in the graph yet
+	// OK so to add a node we need to check its parents
+	// If all of the parents are in the graph then we can add the node.
```
pjbgf, line 176:
```suggestion
	// The node is not in the graph yet. But to add it we first need to check its parents.
	// Only when all of the parents are in the graph that we can add the node.
```

plumbing/object/commitgraph/commitnode_graph_builder.go
zeripath: That's very odd - I'd removed this line! I think I've pushed up the wrong branch!

plumbing/object/commitgraph/commitnode_graph_builder.go
zeripath: Sorry about this I'd done quite a lot of rewriting of the comments - I think things should be a lot clearer in the rewrite!

Issue #844, opened 2023-09-22
Apply `.gitconfig` insteadOf rules to `Repository.clone()`


boggyhole:
First of all: Thanks for this project! It's been a very good experience so far!

I took me some time to understand why my url `insteadOf` rules are not applied to higher level `Clone()`/`PlainClone()` commands. As far as I understood `insteadOf` rules are only applied to existing repositories on `.git/config` remotes. On my journey through the project I didn't find any reason to not apply global url `insteadOf` rules (i.e. defined in `.gitconfig`) to higher level `Clone()`/`PlainClone()` functions. I am using something like this:

```
cloneOpts := &git.CloneOptions{
	URL: repoURL,
}

conf, err := config.LoadConfig(config.GlobalScope)
if err != nil {
	return err
}

// URL replacements would need to be done manually here

storer := memory.NewStorage()
err = storer.SetConfig(conf)
if err != nil {
	return err
}

fs := memfs.New()
repo, err := git.Clone(storer, fs, cloneOpts)
if err != nil {
	return errors.Wrapf(err, "clone repository '%s'", repoURL)
}
```

This code won't apply any `insteadOf` rules from my global config. Am I overlooking something or could url replacement be added to something like: `func (r *Repository) CreateRemote(c *config.RemoteConfig) (*Remote, error)`?

Thanks for suggestions and hints.

jazanne:
It seems like I may be running into this issue too

I have a script that uses go-git and is run from github actions. However, even verifying that the `actions/checkout` step uses token with correct permissions and persists them to git config, i get this message while trying to fetch remote branches i get a `authentication required` error.

robertgzr:
Looks like `Repository.CreateRemote` simply overwrites any configuration previously applied to the `Storer`:
https://github.com/go-git/go-git/blob/1e327a99f5f401cf4e938758fe04bb0e4cf7f31c/repository.go#L652-L670

Br1an67:
I've submitted a fix for this. The issue was that URL insteadOf rules were only being applied when loading config from existing repositories, not during clone operations. I've added support for applying global URL rewrite rules to the clone URL before creating the remote.

PR: #844

PR: #1888

PR #882, opened 2023-10-16
apply insteadof on Remote.{fetch,pull,list} functions -- breaking change(?)

BryanStenson-okta:
Fixes https://github.com/go-git/go-git/issues/844

This applies `insteadOf` rules to `Remote.{fetch,pull,list}` functions.  This may be seen as a breaking change, as the current behavior _does not_ honor `insteadOf` values.  I believe _not_ applying `insteadOf` values is actually a bug (#844)...so I'm not sure if this is a "breaking change", or a "bugfix".

I'm open to a discussion on whether this should be gated via feature flag.  For example, adding a `CloneOptions.ApplyInsteadOf bool` -- defaulting to `false` (to preserve current behavior).

Thoughts?

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -351,7 +351,7 @@ func (c *Config) unmarshalRemotes() error {
 
 	// Apply insteadOf url rules
 	for _, r := range c.Remotes {
-		r.applyURLRules(c.URLs)
+		r.ApplyURLRules(c.URLs)
 	}
 
 	return nil
@@ -583,6 +583,7 @@ type RemoteConfig struct {
 
 	// insteadOfRulesApplied have urls been modified
 	insteadOfRulesApplied bool
+
 	// originalURLs are the urls before applying insteadOf rules
 	originalURLs []string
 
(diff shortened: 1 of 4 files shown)


boggyhole:
I am not super familiar with the codebase, yet. This PR solves the `clone()` issue where `url.insteadOf` rules weren't applied. Are there other use-cases where it would make sense to get them applied (thinking of submodules for example: `func (s *Submodule) Repository()`? So my actual question is: would it make sense to move `r.ApplyURLRules()` to a deeper level? I.e. `func (r *Repository) CreateRemote(c *config.RemoteConfig) (*Remote, error)`?

BryanStenson-okta:
> ... would it make sense to move `r.ApplyURLRules()` to a deeper level? I.e. `func (r *Repository) CreateRemote(c *config.RemoteConfig) (*Remote, error)`?

yup, good point!  i'll update that shortly.

BryanStenson-okta:
thinking about this more, i think this applying the `insteadof` substitution logic should happen on the `Remote` -- so that all fetch, push, etc operations are using substituted urls.

BryanStenson-okta:
I've updated this PR to call `ensureURLRulesApplied()` on each `Remote.{fetch,pull,list}` function, to ensure the URLs will always be rewritten, if necessary.

pjbgf:
> thinking about this more, i think this applying the `insteadof` substitution logic should happen on the `Remote` -- so that all fetch, push, etc operations are using substituted urls.

Good shout! One interesting workflow is when URLs are being set directly at the operation options, which I believe would always overwrite the ones from the Remote. I don't think we need to change the code for this, but I would expand on their comments to highlight that InsteadOf rules does not apply to such fields:

https://github.com/go-git/go-git/blob/6252084d6fd3173aceeecd6765722ae844ddb266/options.go#L178
https://github.com/go-git/go-git/blob/6252084d6fd3173aceeecd6765722ae844ddb266/options.go#L40

We would need tests to that effect.

pjbgf:
> This may be seen as a breaking change, as the current behavior does not honor `insteadOf` values.

The main challenge here is that as a library, not always you would want to be influenced by system wide configuration. However, I agree with your suggested approach below which would support the new behaviour while keeping backwards compatibility:

> adding a `CloneOptions.ApplyInsteadOf` bool -- defaulting to `false` (to preserve current behavior).

boggyhole:
> One interesting workflow is when URLs are being set directly at the operation options, which I believe would always overwrite the ones from the Remote. I don't think we need to change the code for this, but I would expand on their comments to highlight that InsteadOf rules does not apply to such fields:

@pjbgf So would you say that's desired behavior? For my understanding it wouldn't (as I would expect `insteadOf` rules to be applied to every operation)

> The main challenge here is that as a library, not always you would want to be influenced by system wide configuration. 

To my understanding it's not the case? From what I've seen one always needs to manually load system wide configuration and apply it to the `Storer`? Or is there automatic `.gitconfig` loading for some scenarios?

```
gitconfig, err := config.LoadConfig(config.GlobalScope)
if err != nil {
	return nil, err
}

if err := storer.SetConfig(gitconfig); err != nil {
	return nil, err
}
```

Issue #373, opened 2021-09-08
submodule.path & submodule.branch is empty


sigurdblueface:
I've added two submodules:
```
 git submodule add https://mygitlab/prometheus monitoring
 git submodule add https://mygitlab/prometheus-rules
```
.gitmodules file:
```
[submodule "monitoring"]
	path = monitoring
	url = https://mygitlab/prometheus
[submodule "prometheus-rules"]
	path = prometheus-rules
	url = https://mygitlab/prometheus-rules
```
go test code snippet:
```
func FindSubmodules() error {
	repo, err := git.PlainOpen(".")
	if err != nil {
		return err
	}
	cfg, err := repo.Config()
	if err != nil {
		return err
	}
	for key, submodule := range cfg.Submodules{
		log.Printf("key: %s, name: %s, path: %s, url: %s, branch: %s", key, submodule.Name, submodule.Path, submodule.URL, submodule.Branch)
	}
	return nil
}

func TestFindSubmodules(t *testing.T) {
	err := FindSubmodules()
	if err != nil {
		t.Error(err)
	}
}
```
output of the above code is 
```
=== RUN   TestFindSubmodules
2021/09/08 13:02:16 key: prometheus-rules, name: prometheus-rules, path: , url: https://mygitlab/prometheus-rules, branch: 
2021/09/08 13:02:16 key: monitoring, name: monitoring, path: , url: https://mygitlab/prometheus, branch: 
```

Please clarify if it's expected.

andrewpollock:
I'm seeing something related, where the repo's submodule status is different after code has examined it, than before.

Steps to reproduce:

```
$ git clone https://github.com/charlesneimog/pd-server.git
Cloning into 'pd-server'...
remote: Enumerating objects: 849, done.
remote: Counting objects: 100% (10/10), done.
remote: Compressing objects: 100% (8/8), done.
remote: Total 849 (delta 2), reused 4 (delta 2), pack-reused 839
Receiving objects: 100% (849/849), 2.12 MiB | 2.56 MiB/s, done.
Resolving deltas: 100% (371/371), done.
$ cd pd-server
$ git submodule update --init --recursive
Submodule 'pd-lib-builder' (https://github.com/pure-data/pd-lib-builder) registered for path 'pd-lib-builder'
Submodule 'src/cpp-httplib' (https://github.com/yhirose/cpp-httplib) registered for path 'src/cpp-httplib'
[...]

PR #894, opened 2023-10-27
storage: filesystem, handle submodule names with encoding

andrewpollock:
Whilst trying to get to the bottom of unexpected mutative behaviour from calling `submodule.Status()` it was determined that submodule names containing characters that Git URL encodes (such as forward slashes on Linux) were not being handled correctly, resulting in `submodule.Repository()` calling `Init()` with an incorrect submodule name.

Further, Go's URL encoding opts for uppercase hexadecimal characters, whereas Git appears to go for lowercase, so some hacky handling of this is required.

We haven't been able to locate canonical encoding/decoding behaviour in the Git source, so it is currently unknown how comprehensive this is. We have opted for the obvious suspects, being forward and backslashes.

As discussed in #373

### DIFF
--- a/storage/filesystem/module.go
+++ b/storage/filesystem/module.go
@@ -1,6 +1,9 @@
 package filesystem
 
 import (
+	"net/url"
+	"strings"
+
 	"github.com/go-git/go-git/v5/plumbing/cache"
 	"github.com/go-git/go-git/v5/storage"
 	"github.com/go-git/go-git/v5/storage/filesystem/dotgit"
@@ -11,7 +14,12 @@ type ModuleStorage struct {
 }
 
 func (s *ModuleStorage) Module(name string) (storage.Storer, error) {
-	fs, err := s.dir.Module(name)
+	// Submodules can have names that Git URL encodes in the filesystem.
+	encodedName := url.PathEscape(name)
+	// Go's URL encoding uses uppercase, Git seems to use lowercase.
+	lowercasedName := strings.ReplaceAll(encodedName, "%2F", "%2f")   // Replace forward slashes.
+	lowercasedName = strings.ReplaceAll(lowercasedName, "%5C", "%5c") // Replace back slashes.
+	fs, err := s.dir.Module(lowercasedName)
 	if err != nil {
 		return nil, err
 	}


pjbgf:
@andrewpollock please add some tests to confirm the change in behaviour and to avoid regression in the future.

andrewpollock:
> @andrewpollock please add some tests to confirm the change in behaviour and to avoid regression in the future.

@pjbgf I'll need some help, I find the testing framework and the internals in general tricky to grok, so I'm not quite sure where to start when there's no preexisting tests to extend...

pjbgf:
@andrewpollock I found [some comments](https://github.com/git/git/blob/master/submodule.c#L2519-L2538) that seem to be related to this issue in another git library, which makes me think that our solution may not be compatible with other libraries. However, we only aim for `go-git` to be compatible with upstream `git` so as long as that is the case we should be OK.

However, I also could not find how upstream is managing this scenario.

andrewpollock:
> @andrewpollock I found [some comments](https://github.com/git/git/blob/master/submodule.c#L2519-L2538) that seem to be related to this issue in another git library, which makes me think that our solution may not be compatible with other libraries. However, we only aim for `go-git` to be compatible with upstream `git` so as long as that is the case we should be OK.
> 
> However, I also could not find how upstream is managing this scenario.

Yes, we spent some time trawling through https://github.com/git/git/blob/master/submodule.c when looking for other possible URL encodings that would require lowercasing. I have an open enquiry with some colleagues who work on Git for pointers.

andrewpollock:
> However, I also could not find how upstream is managing this scenario.

https://github.com/git/git/blob/2e8e77cbac8ac17f94eee2087187fa1718e38b14/strbuf.c#L458 is the closest thing I can find with fresh eyes, but I haven't been able to connect that with the submodule code yet to confirm...

pjbgf:
@andrewpollock yep, I stumbled into that as well, but it seems orthogonal [as per commit desc](https://github.com/git/git/commit/b44d0118ac5074215ece7ae2bb86ad0713f70674). And the percent encoding they are using is not lower cased.

andrewpollock:
Ugh, I've since discovered we're using a locally patched Git that introduces this URL escaping behaviour :-(

andrewpollock:
So I think the right course of action here is to close this PR until such time that Google upstreams this urlencoding behaviour...

Issue #731, opened 2023-03-30
feat: implement update-server-info


aymanbagabas:
`git update-server-info` is used to generate `info/refs` and `objects/info/packs` files to be used by the HTTP (dumb) transfer protocol.

PR #896, opened 2023-10-29
Implement upload-server-info

aymanbagabas:
This adds UpdateServerInfo along with a new go-git command to generate info files to help git dumb http serve refs and their objects.

This also sorts the refs by name before returning them. Not sure if this would break things :/

Docs: https://git-scm.com/docs/git-update-server-info
Fixes: https://github.com/go-git/go-git/issues/731

### DIFF
--- a/cli/go-git/update_server_info.go
+++ b/cli/go-git/update_server_info.go
@@ -1,0 +1,34 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing/serverinfo"
+	"github.com/go-git/go-git/v5/storage/filesystem"
+)
+
+// CmdUpdateServerInfo command updates the server info files in the repository.
+// This is used by git http transport (dumb) to generate a list of available
+// refs for the repository. See:
+// https://git-scm.com/docs/git-update-server-info
+type CmdUpdateServerInfo struct {
+	cmd
+}
+
+// Usage returns the usage of the command.
+func (CmdUpdateServerInfo) Usage() string {
+	return fmt.Sprintf("within a git repository run: %s", os.Args[0])
+}
+
+// Execute runs the command.
+func (c *CmdUpdateServerInfo) Execute(args []string) error {
+	r, err := git.PlainOpen(".")
+	if err != nil {
+		return err
+	}
+
+	fs := r.Storer.(*filesystem.Storage).Filesystem()
+	return serverinfo.UpdateServerInfo(r.Storer, fs)
+}
--- a/plumbing/serverinfo/serverinfo.go
+++ b/plumbing/serverinfo/serverinfo.go
@@ -1,0 +1,94 @@
+package serverinfo
+
+import (
+	"fmt"
+
+	"github.com/go-git/go-billy/v5"
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/internal/reference"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/plumbing/storer"
+	"github.com/go-git/go-git/v5/storage"
+)
+
+// UpdateServerInfo updates the server info files in the repository.
+//
+// It generates a list of available refs for the repository.
+// Used by git http transport (dumb), for more information refer to:
+// https://git-scm.com/book/id/v2/Git-Internals-Transfer-Protocols#_the_dumb_protocol
+func UpdateServerInfo(s storage.Storer, fs billy.Filesystem) error {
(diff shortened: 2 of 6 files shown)


server_info_test.go
```diff
@@ -0,0 +1,124 @@
...
+	st := memory.NewStorage()
+	r, err := Init(st, fs)
+	c.Assert(err, IsNil)
+	c.Assert(r, NotNil)
+
+	err = UpdateServerInfo(st, fs)
+	c.Assert(err, IsNil)
+}
+
+func testUpdateServerInfo(c *C, f *fixtures.Fixture) {
```
pjbgf, line 26: Would be nice to add another test that asserts subsequent calls to update the server info work properly after new changes to the repository take place.

storage/filesystem/reference.go
```diff
@@ -28,6 +29,8 @@ func (r *ReferenceStorage) IterReferences() (storer.ReferenceIter, error) {
 		return nil, err
 	}
 
+	reference.Sort(refs)
```
pjbgf, line 32: We may need a flag to toggle this behaviour just in case it introduces a regression.

storage/memory/storage.go
```diff
@@ -283,6 +284,8 @@ func (r ReferenceStorage) IterReferences() (storer.ReferenceIter, error) {
 		refs = append(refs, ref)
 	}
 
+	reference.Sort(refs)
```
pjbgf, line 287: Same here about the flag.

server_info.go
```diff
@@ -0,0 +1,85 @@
+package git
```
pjbgf, line 1: Not sure this is the best package for `server_info` and its tests. Maybe somewhere inside `plumbing` could be a better fit?

aymanbagabas: Moved it to `plumbing/serverinfo`

storage/filesystem/reference.go
aymanbagabas: I've added a new `NewOrderedReferenceSliceIter` instead to create an ordered reference iter

cli/go-git/update_server_info.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing/serverinfo"
+	"github.com/go-git/go-git/v5/storage/filesystem"
+)
+
+type CmdUpdateServerInfo struct {
```
pjbgf, line 12: We should add some comments here to briefly explain what it does.

plumbing/serverinfo/serverinfo.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/plumbing/storer"
+	"github.com/go-git/go-git/v5/storage"
+)
+
+// UpdateServerInfo updates the server info files in the repository.
+//
+// It generates a list of available refs for the repository.
+// Used by git http transport (dumb).
```
pjbgf, line 18:
```suggestion
// Used by git http transport (dumb), for more information refer to:
// https://git-scm.com/book/id/v2/Git-Internals-Transfer-Protocols#_the_dumb_protocol
```

pjbgf:
We should add some comment on the compatibility table to call out we have some support and to what extent:
https://github.com/go-git/go-git/blob/master/COMPATIBILITY.md?plain=1#L208

plumbing/storer/reference.go
```diff
@@ -107,14 +108,27 @@ type ReferenceSliceIter struct {
...
+func newReferenceSliceIter(series []*plumbing.Reference, ordered bool) ReferenceIter {
+	if ordered {
+		reference.Sort(series)
+	}
 	return &ReferenceSliceIter{
 		series: series,
 	}
 }
 
+// NewOrderedReferenceSliceIter returns an ordered reference iterator for the
```
pjbgf, line 120:
```suggestion
// NewOrderedReferenceSliceIter returns a lexicographically ordered reference iterator for the
```

storage/filesystem/reference.go
pjbgf: I think we should introduce an `IterOrderedReferences` on the storage interface so we don't change the existing behaviour for users that rely on `IterReferences` already, and the new `NewOrderedReferenceSliceIter` is called there only.

aymanbagabas: I dropped the `IterOrderedReferences` altogether. I don't like the API on the storage interface, [Git sorts the references by refname by default](https://github.com/git/git/blob/master/refs/refs-internal.h#L320) and has the option to not to. We can keep things simple and sort the refs where we need to.

pjbgf (approved):
@aymanbagabas nice one! Thanks for working on this. :bow:

Issue #897, opened 2023-10-30
Running Local Tests Setup


patgmiller:
I initially was using my own shell for running the tests against development and quickly figured out that isn't the most ideal due to my personal git config setup.

Is there a common approach among the maintainers? Or do you all just choose your own preference?

This was my quick and dirty example

```shell
# from your local cloned go-git repo
docker pull golang
docker container run -v $(pwd):/go/src -it --priviledged golang bash
```

```shell
# in container 
apt install -y libcrypt-openssl* libzlcore-* libzltext-* zlib1g-dev libcurl4-open* libexpat1*

export GIT_VERSON=v2.11.0

make build-dist

export GIT_DIST_PATH=$(pwd)/.git-dist
export GIT_EXEC_PATH=$GIT_DIST_PATH
export PATH=$GIT_EXEC_PATH:$PATH
touch ~/known-hosts
export SSH_KNOWN_HOSTS=/root/known-hosts
git --exec-path
git config --global user.email "gha@example.com"
git config --global user.name "GitHub Actions"
make test
```

which got me far enough,
```shell
# ...
----------------------------------------------------------------------
FAIL: worktree_test.go:649: WorktreeSuite.TestCheckoutIndexOS

worktree_test.go:674:
    c.Assert(idx.Entries[0].UID, Not(Equals), uint32(0))
... obtained uint32 = 0x0
... expected uint32 = 0x0

2023/10/26 01:04:13 comparing mod time: 2023-10-26 01:04:13.497602748 +0000 UTC == 2023-10-26 01:04:13.497602748 +0000 UTC on linux (amd64)
OOPS: 291 passed, 1 FAILED
```

but i'm wondering if there can be a common consensus for running tests locally that maybe i or someone can add to the repo?

_Originally posted by @patgmiller in https://github.com/go-git/go-git/discussions/892_

patgmiller:
check progress  @ [feature/897-run-ci-test-locally](https://github.com/patgmiller/go-git/tree/feature/897-run-ci-test-locally)

pjbgf:
@patgmiller yep, that should be a good starting point.

patgmiller:
@pjbgf currently I can run the `go-test-*` in both Linux and Darwin, however it only passes for Linux. There appears to be some different behavior on Darwin, which is actually really stupid b/c docker desktop on Darwin is running a Linux VM. 

The fail in question for `go` `v1.19` is the following: 

```
----------------------------------------------------------------------
[...]

PR #902, opened 2023-11-03
build: Add Makefile targets to run CI tests locally

patgmiller:
See issue #897 and discussion #892 for background. Basically add ability to run `make test` and `make test-coverage` locally using `docker` containers. 

Added make targets:
- `test-go-all` `test-go1.19` `test-go1.20` `test-go1.21`: These will run the `test-coverage` for related go version. 
- `test-git-all` `test-git-master` `test-git-v2.11`: These will build git distribution and then run `make test`.
- `clean-all` `clean-docker`: additional clean up targets.

@pjbgf @jfontan @mcuadros I'm not sure if just Paulo will review or if you prefer to have everyone take a look, but here it is in all of it's glory. Thanks.


patgmiller:
Fixes: #897

Makefile
```diff
@@ -29,6 +34,50 @@ test:
...
+	export BASH_ENV=$(TEST_ENV)
+	$(BASHCMD) _local/test-go.sh "1.19"
+
+test-go1.20: test-env
+	export BASH_ENV=$(TEST_ENV)
+	$(BASHCMD) _local/test-go.sh "1.20"
+
+test-go1.21: test-env
+	BASH_ENV=$(TEST_ENV)
+	$(BASHCMD) _local/test-go.sh "1.21"
```
pjbgf, line 66: If we have target per Go version, we will need to update this twice a year. What about we have a single Makefile target and grab the Go version via environment var?

Makefile
```diff
@@ -29,6 +34,50 @@ test:
...
+test-go-all: test-go1.19 test-go1.20 test-go1.21
+
+.PHONY: test-git-master test-git-v2.11
+test-git-v2.11:
+	BASH_ENV=$(TEST_ENV)
+	$(BASHCMD) _local/test-git-compatability.sh "v2.11.0"
+
+test-git-master:
+	BASH_ENV=$(TEST_ENV)
+	$(BASHCMD) _local/test-git-compatability.sh "master"
```
pjbgf, line 77: Same here, let's have a single target and source the branch name from an environment var.

Makefile
```diff
@@ -40,8 +89,25 @@ test-coverage:
...
+		rm $(TEST_ENV) $(DOCKER_ENV)
+	fi
+	rm coverage*
+
+# doc sys prune in this case only removes dangling images, but filter it anyway
+GOGIT_IMAGES := $(shell docker image ls --filter reference=go-git -q)
+clean-docker:
+	docker system prune -f --filter "label=$(DOCKER_GOGIT_KEY)=$(DOCKER_GOGIT_LABEL)"
+	docker volume prune -f --filter "label=$(DOCKER_GOGIT_KEY)=$(DOCKER_GOGIT_LABEL)"
+	docker image rm -f $(GOGIT_IMAGES)
```
pjbgf, line 108: For simplicity, we probably could just document this instead of adding an additional target. Although your implementation added filters, this is still a system-wide op.

Makefile
```diff
@@ -40,8 +89,25 @@ test-coverage:
...
+
 clean:
-	rm -rf $(GIT_DIST_PATH)
+	@if [ -d $(GIT_DIST_PATH) ]; then
+		rm -rf $(GIT_DIST_PATH)
+	fi
+	if [ -f $(TEST_ENV) ] || [ -f $(DOCKER_ENV) ]; then
+		rm $(TEST_ENV) $(DOCKER_ENV)
+	fi
+	rm coverage*
```
pjbgf, line 101: We could keep it as it was, but instead have a single dir (e.g. `build/`) in which any ephemeral things are created into. So we can remove a single dir to clear up any build/test artefacts.

_local/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+USER ${LOCUSER}
+
+RUN git config --global user.email "gha@example.com"; \
+    git config --global user.name "GitHub Actions"
+
+RUN mkdir -p  ~/.ssh; \
+    ssh-keyscan -H github.com > ~/.ssh/known_hosts
+
+ENTRYPOINT [ "/entrypoint.sh" ]
```
pjbgf, line 38:
```suggestion
ENTRYPOINT [ "/entrypoint.sh" ]

```

_local/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+        tcl8.6* \
+        zlib1g-dev
+
+# worktree_test.go:677:
+# c.Assert(idx.Entries[0].UID, Not(Equals), uint32(0))
+# User MUST NOT == uid=0(root) gid=0(root) groups=0(root)
+RUN DEBIAN_FRONTEND=noninteractive; \
+    addgroup --gid ${GID} --force-badname ${LOCUSER}; \
+    adduser --uid ${UID} --ingroup ${LOCUSER} \
+        --disabled-password --quiet --force-badname ${LOCUSER}
```
pjbgf, line 26: Instead of having to create an user, we can just run it as nobody:

```
USER 65534:65534
```

_local/Dockerfile
```diff
@@ -0,0 +1,38 @@
+ARG GOTAG
+
+FROM golang:${GOTAG}
+
+ARG LOCUSER
+ARG UID
+ARG GID
```
pjbgf, line 7:
```suggestion
```

_local/scripts/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+if [ "${GIT_EXEC_PATH}" != "" ]; then
+    export PATH="$GIT_EXEC_PATH:$PATH"
+fi
+
+exec "$@"
```
pjbgf, line 7:
```suggestion
exec "$@"

```

_local/platform/darwin.sh
```diff
@@ -0,0 +1,29 @@
...
+    # Also fails with confusing "denied: request access" in addition to error
+    # when build fails, instead of just error.
+    if [ "$(docker image ls -q --filter "reference=golang:$tag")" == "" ]; then
+        docker image pull "golang:$tag"
+    fi
+}
+
+### END Platform Specific Section
+
+VERSION=$(bash --version | grep version)
```
pjbgf, line 17: Not sure why we need to enforce a bash version if we would run all the tests within a container we control. Do you mind clarifying why this is needed please?

_local/scripts/build-git.sh
```diff
@@ -0,0 +1,32 @@
...
+
+GIT_ARCHIVE="${GIT_VERSION}.tar.gz"
+GIT_SRC_ROOT=$(dirname "${GIT_DIST_PATH}")
+GIT_URL=https://github.com/git/git/archive/refs/tags
+if [[ ${GIT_VERSION} =~ master ]]; then
+    GIT_URL=https://github.com/git/git/archive/refs/heads
+fi
+GIT_URL="${GIT_URL}/${GIT_ARCHIVE}"
+
+# git clone issues:
```
pjbgf, line 12: Not sure what you mean by these comments. Are this errors that happen intermittently? Or are they issues that were fixed by the code below? Please clarify.

_local/platform/darwin.sh
```diff
@@ -0,0 +1,29 @@
...
+
+function preDockerCommands() {
+    local tag=$1
+    # Docker Desktop for Darwin is stupid, and doesn't read FROM image
+    # Also fails with confusing "denied: request access" in addition to error
+    # when build fails, instead of just error.
+    if [ "$(docker image ls -q --filter "reference=golang:$tag")" == "" ]; then
+        docker image pull "golang:$tag"
+    fi
+}
```
pjbgf, line 13: If this is the main difference between darwin and linux targets, I would probably do that always instead of having to keep platform specific logic.

pjbgf (changes requested):
@patgmiller Thanks for working on this. Overall this is going towards the right direction of reducing the bar for new contributors to be able to run tests locally. I added some comments on ways we could simplify the changes or make them easier to maintain longer term.

Running this on openSUSE Tumbleweed I got an error message:

```
$ make test-go1.21
...
invalid argument "=" for "-l, --label" flag: invalid label '=': empty name
See 'docker container run --help'.
make: *** [Makefile:65: test-go1.21] Error 125
```

patgmiller:
> Running this on openSUSE Tumbleweed I got an error message:
> 
> ```
> $ make test-go1.21
> ...
> invalid argument "=" for "-l, --label" flag: invalid label '=': empty name
> See 'docker container run --help'.
> make: *** [Makefile:65: test-go1.21] Error 125
> ```

@pjbgf could you post the details of `$> docker info`, or at least docker version details? I should've checked asked about version specification earlier. Thanks

pjbgf:
@patgmiller no worries at all, please squash the commits.

Here's the docker info output:

```
$ docker info
Client:
 Version:    24.0.7-ce
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.11.2
    Path:     /usr/lib/docker/cli-plugins/docker-buildx

Server:
 Server Version: 24.0.7-ce
...
```

patgmiller:
@pjbgf regarding commit message check, i can pick semi related package from options: 

```
git|plumbing|utils|config|_examples|internal|storage|cli|build
```

like `_examples`, `internal`, or `build`; however it could be misleading for whomever reads it later. should i perhaps add `_local` as part of my fix, or don't worry about adding one? thanks

Makefile
patgmiller: So you would rather have one target, which possibly never needs to change and it would be run something like this?

```shell
GO_VERSION=1.21; make test-go
```

Makefile
patgmiller: Okay, will do

patgmiller:
@pjbgf your `make test-go1.21` didn't fail due to docker version but i missed an `export` inside the target, so the `BASH_ENV` wasn't included. Would you pull and try it again when you have a minute, should work for you next time.

_local/platform/darwin.sh
patgmiller: The scripts are running the containers, so this version of bash is mainly for Darwin. The built-in bash shell distributed with Darwin is an older 3.x one, and having written a lot of scripts for managing Mac users envs in IT related activities, I've found it simpler to at least upgrade to bash 4 or 5. But if you'd rather, I'll remove the requirement.

_local/Dockerfile
patgmiller: @pjbgf It looks like the `git config` will fail, I tested with your suggestion ...

```shell
GO_VERSION=1.19 gmake test-go
export BASH_ENV=/Users/patrick.miller/src/github.com/patgmiller/go-git/.git-dist/.env
/usr/bin/env bash _local/test-go.sh 1.19
Running checks for: darwin
[+] Building 0.2s (8/9)                                                                                                                                            docker:desktop-linux
 => [internal] load .dockerignore                                                                                                                                                  0.0s
 => => transferring context: 57B                                                                                                                                                   0.0s
 => [internal] load build definition from Dockerfile                                                                                                                               0.0s
 => => transferring dockerfile: 963B                                                                                                                                               0.0s
 => [internal] load metadata for docker.io/library/golang:1.19-bullseye                                                                                                            0.0s
 => [1/5] FROM docker.io/library/golang:1.19-bullseye                                                                                                                              0.0s
 => [internal] load build context                                                                                                                                                  0.0s
 => => transferring context: 156B                                                                                                                                                  0.0s
 => CACHED [2/5] RUN apt update &&     apt install -y autoconf         gettext         libcrypt-openssl*         libcurl4-open*         libexpat1*         libzlcore-*         li  0.0s
 => [3/5] COPY --chown=root:root _local/* /                                                                                                                                        0.0s
 => ERROR [4/5] RUN git config --global user.email "gha@example.com";     git config --global user.name "GitHub Actions"                                                           0.1s
------
 > [4/5] RUN git config --global user.email "gha@example.com";     git config --global user.name "GitHub Actions":
0.127 error: could not lock config file /nonexistent/.gitconfig: No such file or directory
0.128 error: could not lock config file /nonexistent/.gitconfig: No such file or directory
------
Dockerfile:33
--------------------
  32 |
  33 | >>> RUN git config --global user.email "gha@example.com"; \
  34 | >>>     git config --global user.name "GitHub Actions"
  35 |
--------------------
ERROR: failed to solve: process "/bin/sh -c git config --global user.email \"gha@example.com\";     git config --global user.name \"GitHub Actions\"" did not complete successfully: exit code: 255
gmake: *** [Makefile:62: test-go] Error 1
```

 and there by the test i added it for as well as the `ssh-keyscan`, not sure this later part is an issue but i recall the user config being related to a number of tests. 

```shell
# worktree_test.go:677:
# c.Assert(idx.Entries[0].UID, Not(Equals), uint32(0))
# User MUST NOT == uid=0(root) gid=0(root) groups=0(root)
```

There was another reason I added the user, it was because some `.tmp` are created during tests and if the owner id doesn't match the `clean` has to be run with `sudo`.  

Any other thoughts or alternative ideas?

_local/scripts/build-git.sh
patgmiller: The comments below `git clone issues:` were fixed by the code which follows it.

pjbgf:
> like _examples, internal, or build; however it could be misleading for whomever reads it later. should i perhaps add _local as part of my fix, or don't worry about adding one? thanks

@patgmiller `build` would better cover given that this is just build related.

_local/scripts/build-git.sh
pjbgf: Thank you for clarifying. :+1:

_local/test-git-compatability.sh
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env bash
```
pjbgf, line 1: This file name is mispelled, please rename to `test-git-compatibility.sh`.

_local/test-git-compatability.sh
```diff
@@ -0,0 +1,102 @@
...
+PLATFORM=$(uname -s | tr '[[:upper:]]' '[[:lower:]]')
+GIT_DIST_PATH="/git/src"
+GIT_REPOSITORY="${GIT_REPOSITORY:-http://github.com/git/git.git}"
```
pjbgf, line 4:
```suggestion
GIT_REPOSITORY="${GIT_REPOSITORY:-https://github.com/git/git.git}"
```

.dockerignore
```diff
@@ -0,0 +1,2 @@
+*
+!_local/scripts
```
pjbgf, line 2: We can remove the `.dockerignore` and instead be explicit on the files that need to be copied into container.

Makefile
```diff
@@ -29,6 +36,27 @@ test:
...
+	WORKDIR=$(WORKDIR)
+	EOF
+
+test-go: test-env
+	export BASH_ENV=$(TEST_ENV)
+	$(BASHCMD) _local/test-go.sh $(GO_VERSION)
+
+test-git: test-env
+	export BASH_ENV=$(TEST_ENV)
+	$(BASHCMD) _local/test-git-compatability.sh $(GIT_VERSION)
```
pjbgf, line 58:
```suggestion
	$(BASHCMD) _local/test-git-compatibility.sh $(GIT_VERSION)
```

_local/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# worktree_test.go:677:
+# c.Assert(idx.Entries[0].UID, Not(Equals), uint32(0))
+# User MUST NOT == uid=0(root) gid=0(root) groups=0(root)
+RUN DEBIAN_FRONTEND=noninteractive; \
+    addgroup --gid ${GID} --force-badname ${LOCUSER}; \
+    adduser --uid ${UID} --ingroup ${LOCUSER} \
+        --disabled-password --quiet --force-badname ${LOCUSER}
+
+COPY --chown=root:root _local/* /
```
pjbgf, line 28: As per comment above, we can call out the files here.

_local/scripts/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
pjbgf, line 1: Any reason why the shebang for `build-git.sh` and `entrypoint.sh` are different than the ones for `test-*`?

_local/test-git-compatability.sh
```diff
@@ -0,0 +1,102 @@
...
+cat <<EOF >"$DOCKER_ENV"
+COVERAGE_REPORT="/go/src/build/coverage-${GO_VERSION}.out"
+GIT_VERSION=$gitref
+GIT_DIST_PATH=$GIT_DIST_PATH/$gitref
+GIT_EXEC_PATH=$GIT_DIST_PATH/$gitref
+EOF
+
+echo "Building git-dist: $gitref"
+docker container run \
+     -u root --privileged --rm \
```
pjbgf, line 87: We don't seem to need to run this container as `--privileged`.

```suggestion
     -u root --rm \
```

patgmiller:
> > like _examples, internal, or build; however it could be misleading for whomever reads it later. should i perhaps add _local as part of my fix, or don't worry about adding one? thanks
> 
> @patgmiller `build` would better cover given that this is just build related.

sounds good, that's what I eventually went with.

pjbgf (changes requested):
@patgmiller great work so far. I have tested this and does what it says in the tin. :+1: 

Please rename the dir `_local` to `hack` instead. Squash all changes on a single commit. Then add an additional commit which makes `.github/workflows/git.yml` and `.github/workflows/test.yml` call the new make targets.

.dockerignore
patgmiller: Hi @pjbgf so without the `.dockerignore` all of the files are copied to the build "context" when the build happens, so it slows down the build. Do you still want to remove the file?

patgmiller:
@pjbgf sorry for taking so long to respond, I've been recovering from an accident. Before preceding, I need to ask why the following:

> Then add an additional commit which makes `.github/workflows/git.yml` and `.github/workflows/test.yml` call the new make targets.

I thought the intent was to make the pipeline tests available to local, not to move the github actions to containers. It would require github actions to support docker commands via shell. If i'm not mistaken github has specific [requirements for docker actions](https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions) which don't allow for running the container directly.

For example, their `USER` requirement must be `root`, would require refactoring the docker file and go-git test which requires non-root user.

patgmiller:
moved to https://github.com/go-git/go-git/pull/978

Issue #903, opened 2023-11-04
Guard against concurrent map access


mpldr:
During testing of my program I found this concurrent map-access crash:

```
fatal error: concurrent map read and map write

goroutine 66 [running]:
github.com/go-git/go-git/v5/plumbing/format/idxfile.(*MemoryIndex).FindHash(0xc00015e000, 0x5279e500ac9df344?)
	github.com/go-git/go-git/v5@v5.10.0/plumbing/format/idxfile/idxfile.go:177 +0x5b
github.com/go-git/go-git/v5/storage/filesystem.(*ObjectStorage).decodeObjectAt(0xc0003ea158, 0xc0001143c0, 0x7f5cab109f45?)
	github.com/go-git/go-git/v5@v5.10.0/storage/filesystem/object.go:473 +0x34
github.com/go-git/go-git/v5/storage/filesystem.(*ObjectStorage).getFromPackfile(0xc0003ea158, {0x23, 0xe4, 0x88, 0x32, 0x48, 0xf0, 0x47, 0x2d, 0x80, ...}, ...)
	github.com/go-git/go-git/v5@v5.10.0/storage/filesystem/object.go:466 +0x2e5
github.com/go-git/go-git/v5/storage/filesystem.(*ObjectStorage).EncodedObject(0xc0003ea158, 0x1, {0x23, 0xe4, 0x88, 0x32, 0x48, 0xf0, 0x47, 0x2d, ...})
	github.com/go-git/go-git/v5@v5.10.0/storage/filesystem/object.go:318 +0x51
github.com/go-git/go-git/v5/plumbing/object.GetCommit({0x55c9ad2dad80, 0xc0003ea140}, {0x23, 0xe4, 0x88, 0x32, 0x48, 0xf0, 0x47, 0x2d, ...})
	github.com/go-git/go-git/v5@v5.10.0/plumbing/object/commit.go:71 +0x39
github.com/go-git/go-git/v5.(*Repository).CommitObject(...)
	github.com/go-git/go-git/v5@v5.10.0/repository.go:1418
git.sr.ht/~mpldr/gogit/internal/types.(*Repo).Commit(0xc0003e4180, {0xc0003d62d2?, 0x40?})
	git.sr.ht/~mpldr/gogit/internal/types/repo.go:119 +0x8c
git.sr.ht/~mpldr/gogit/internal/handler.(*Handler).repoHandler.commitHandler.func2(0xc0003d624b?)
[...]

PR #904, opened 2023-11-04
index: fix concurrent map access

mpldr:
This commit fixes a runtime error caused by a concurrent map access.

Add a sync.RWMutex and guard the offsetHash field with it.

Fixes: https://github.com/go-git/go-git/issues/903


pjbgf:
@mpldr thanks for submitting this PR. This touches some of the issues covered by https://github.com/go-git/go-git/issues/773. One of the challenges here is on how to add thread safety while either off-setting the performance hit or providing users with a way to opt-out.

For context, a similar [change](https://github.com/go-git/go-git/commit/0a7b552ae2d70f95e8cad6a8864e8505a496a158) exists in the [read-concurrency](https://github.com/go-git/go-git/compare/master...read-concurrency) branch which hasn't been proposed as PR yet due to the reasons above.

PR #913, opened 2023-11-07
cli: Refactor for simplicity and dependency removal

pjbgf:
Remove third party dependency (`github.com/jessevdk/go-flags`) for flags handling to rely on the standard library instead.


pjbgf:
@aymanbagabas do you mind taking a took at these changes please? Pinging you mostly as you were the most recent contributor to change `cli/*`.  :wink:

PR #916, opened 2023-11-08
git: add tracer package

aymanbagabas:
This adds a generic tracing package to log messages to output if the target is set.

### DIFF
--- a/internal/trace/trace.go
+++ b/internal/trace/trace.go
@@ -1,0 +1,55 @@
+package trace
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"sync/atomic"
+)
+
+var (
+	// logger is the logger to use for tracing.
+	logger = newLogger()
+
+	// current is the targets that are enabled for tracing.
+	current atomic.Int32
+)
+
+func newLogger() *log.Logger {
+	return log.New(os.Stderr, "", log.Ltime|log.Lmicroseconds|log.Lshortfile)
+}
+
+// Target is a tracing target.
+type Target int32
+
+const (
+	// General traces general operations.
+	General Target = 1 << iota
+
+	// Packet traces git packets.
+	Packet
+)
+
+// SetTarget sets the tracing targets.
+func SetTarget(target Target) {
+	current.Store(int32(target))
+}
+
+// SetLogger sets the logger to use for tracing.
+func SetLogger(l *log.Logger) {
+	logger = l
+}
+
+// Print prints the given message only if the target is enabled.
+func (t Target) Print(args ...interface{}) {
+	if int32(t)&current.Load() != 0 {
+		logger.Output(2, fmt.Sprint(args...)) // nolint: errcheck
+	}
+}
+
+// Printf prints the given message only if the target is enabled.
+func (t Target) Printf(format string, args ...interface{}) {
+	if int32(t)&current.Load() != 0 {
+		logger.Output(2, fmt.Sprintf(format, args...)) // nolint: errcheck
+	}
+}
--- a/internal/trace/trace_test.go
+++ b/internal/trace/trace_test.go
@@ -1,0 +1,95 @@
+package trace
+
+import (
+	"bytes"
+	"io"
+	"log"
+	"testing"
+)
+
+func TestMain(m *testing.M) {
+	defer SetLogger(newLogger())
+	if code := m.Run(); code != 0 {
+		panic(code)
+	}
+}
+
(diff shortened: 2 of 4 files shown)


pjbgf (approved):
@aymanbagabas nice one, thanks for working on this. :bow:

aymanbagabas:
Added test assertions and benchmarks

```
~/Source/go-git/go-git/utils/trace trace*
› go test -check.b -check.bmem -check.btime 5s
PASS: trace_test.go:63: TraceSuite.BenchmarkDisabledTarget      10000000000              1.96 ns/op       0 B/op           0 allocs/op
PASS: trace_test.go:69: TraceSuite.BenchmarkEnabledTarget       50000000               173 ns/op         14 B/op           1 allocs/op
OK: 2 passed
PASS
ok      github.com/go-git/go-git/v5/utils/trace 28.791s
```

utils/trace/trace_test.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (s *TraceSuite) BenchmarkDisabledTarget(c *C) {
+	for i := 0; i < c.N; i++ {
+		General.Print("test")
+	}
+}
+
+func (s *TraceSuite) BenchmarkEnabledTarget(c *C) {
+	SetTarget(General)
```
pjbgf, line 70: For the purposes of the benchmark, I think the logger could write to `io.Discard` as we only want to understand the overall overhead.

```suggestion
	SetLogger(log.New(io.Discard, "", 0))
	SetTarget(General)
```

This does not seem to affect `check.v1` (as it probably wraps/ignore stderr), with standard benchmark, the call costs seems to be:
~1900 ns/op (default stderr)
~40 ns/op (io.Discard)
~300 ns/op (file on disk)

utils/trace/trace.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// Print prints the given message if tracing is enabled.
+func (t Target) Print(args ...interface{}) {
+	if int32(t)&current.Load() != 0 {
+		logger.Output(2, fmt.Sprint(args...)) // nolint: errcheck
+	}
+}
+
+// Printf prints the given message if tracing is enabled.
```
pjbgf, line 50:
```suggestion
// Printf prints the given message only if the target is enabled.
```

utils/trace/trace.go
```diff
@@ -0,0 +1,55 @@
...
+func SetTarget(target Target) {
+	current.Store(int32(target))
+}
+
+// SetLogger sets the logger to use for tracing.
+func SetLogger(l *log.Logger) {
+	logger = l
+}
+
+// Print prints the given message if tracing is enabled.
```
pjbgf, line 43:
```suggestion
// Print prints the given message only if the target is enabled.
```

utils/trace/trace.go
```diff
@@ -0,0 +1,55 @@
...
+	// current is the targets that are enabled for tracing.
+	current atomic.Int32
+)
+
+func newLogger() *log.Logger {
+	return log.New(os.Stderr, "", log.Ltime|log.Lmicroseconds|log.Lshortfile)
+}
+
+// Target is a tracing target.
+type Target int
```
pjbgf, line 23:
```suggestion
type Target int32
```

utils/trace/trace_test.go
```diff
@@ -0,0 +1,74 @@
+package trace
+
+import (
+	"bytes"
+	"log"
+	"testing"
+
+	. "gopkg.in/check.v1"
```
pjbgf, line 8: For new code, we should move away from `check.v1` as that dependency is no longer maintained.

pjbgf (changes requested):
Thank you for the benchmarks. Initially the costs for pktline seemed expensive, but upon looking into it more carefully it became clear that the `pktline` encoding is quite "cheap", therefore any additional call within that process would make a difference.

Below goes a few nits, otherwise LGTM

utils/trace/trace_test.go
aymanbagabas: switched to stdlib testing

aymanbagabas:
> Thank you for the benchmarks. Initially the costs for pktline seemed expensive, but upon looking into it more carefully it became clear that the `pktline` encoding is quite "cheap", therefore any additional call within that process would make a difference.

Should we remove the printf calls then?

pjbgf (approved):
@aymanbagabas great job! Thanks for working on this. :trophy:

Issue #899, opened 2023-10-31
Support SHA-1 **and** SHA-256 at the same time


6543:
the current implementation makes it hard to have go-git working if you need to support both scenarios.

see https://github.com/go-gitea/gitea/pull/23894

AdamMajer:
It's related to this timeline -- https://github.com/go-git/go-git/issues/706

pjbgf:
Apart from the implementation side of things, one of the main issues for supporting this is breaking backwards compatibility with the current API. So I am marking this to the v6 release.

However, I am not entirely sure when exactly that would land. PRs for the SHA256 support are more than welcome.

pjbgf:
#1527 introduces the concurrent support for SHA1 and SHA256 objects. This was the blocking piece that should enable expanding the SHA256 support across go-git features.

Closing this in favour of #706, which tracks the current status of the SHA256 implementation and supported features.

Issue #806, opened 2023-07-06
Audit the usage of SetUpTest in test suites


AriehSchneier:
We use gocheck for running tests via suites, gocheck allows for setup/teardown for the entire suite and for each test (see documentation here: https://labix.org/gocheck).

It appears that we are using SetUpTest in a number of suites to set up multiple test repos, but then only use 1 of them in each of the tests. For example `MatcherSuite` sets up 8 `billy.Filesystem`s, but then each test only uses 1 (or 2 in one case) (also note we seem to have to `MatcherSuite`s that do similar setup. Another example is `ReceivePackSuite` (again we have multiple instances of) which sets up 3 endpoints for every test but usually only uses 1 of them.

We need to audit all usages of SetUpTest, to see if the code could either be moved to SetUpSuite, or changed to callback functions (so that each test can just call the setup code it needs for that specific test).

pjbgf:
`gocheck` does not seem to be maintained anymore, an option could also be moving on from it.

ricci2511:
> `gocheck` does not seem to be maintained anymore, an option could also be moving on from it.

I believe [testify](https://github.com/stretchr/testify) would be a good alternative. It seamlessly integrates with the stdlib, and therefore it's fully supported by the Go extension in VSCode which can be really useful for VSCode users like me and probably many others since it allows you to conveniently debug and run tests individually directly from within the editor.

pjbgf:
@ricci2511 that sounds good, thanks for volunteering. I think there are two ways to start on this:
1. PoC how to replace `gocheck` in cases where it defines a suite which is being reused across different tests (e.g. `storage/test` is reused across `storage/*` same goes to `/plumbing/transport/test`).
2. Replace `gocheck` on straight forward tests (e.g. `/config/*_test.go`).

The first one seems like a good start as that will define how we would structure those reusable test cases.

ricci2511:
@pjbgf Check out this [commit](https://github.com/ricci2511/go-git/commit/25aff1c936fcc4ba5301b6447df8107b1ecd523f) I made on my fork. 

All storage related tests (unless I have missed something) have been moved to testify as PoC. The reuse of the `BaseStorageSuite` across `storage/*` seems to also work as expected, all tests in the suite are running and passing.

pjbgf:
@ricci2511 that looks good. As we roll-out the gocheck replacement changes, it would be good to start tackling the inefficiencies @AriehSchneier pointed out on this issue.

[...]

PR #920, opened 2023-11-10
exp: plumbling/hasher, Add `ObjectHasher` and `ImmutableHash`

pjbgf:
The introduction of both new types are a stepping stone to enable SHA256 support concurrently - without the need for a build tag.

ImmutableHash provides a way to represent varied sized hashes (for SHA1 or SHA256) with a single type, while keeping similar performance of the existing `plumbing.Hash`.

The names were picked in order to provide a clearer distinction on what they do. `ImmutableHash` is the result of a hash operation and can't be changed once calculated. `ObjectHasher`, adds the Git object header before a normal hash operation that `hash.Hash` or `plumbing/hash.Hash` do.

The performance results shows that SHA1 operations could be slower, however for SHA256 it can be over 3 times faster:

```
~Hasher/hasher-sha1-100B-16         	            2202226	     574.7 ns/op	 174.00 MB/s     272 B/op     7 allocs/op
~Hasher/objecthash-sha1-100B-16     	            1511851	     772.6 ns/op	 129.43 MB/s     272 B/op     9 allocs/op
~Hasher/objecthash-sha256-100B-16   	            5057584	     247.4 ns/op	 404.21 MB/s      96 B/op     7 allocs/op
~Hasher/hasher-sha1-5000B-16        	              96476	     12523 ns/op	 399.25 MB/s    5536 B/op     7 allocs/op
~Hasher/objecthash-sha1-5000B-16    	             105376	     10983 ns/op	 455.27 MB/s     272 B/op     9 allocs/op
~Hasher/objecthash-sha256-5000B-16  	             420741	      2828 ns/op	1767.77 MB/s      96 B/op     7 allocs/op
~HashFromHex/hasher-parse-sha1-valid-16         	23243548     64.65 ns/op	 618.69 MB/s      48 B/op     1 allocs/op
~HashFromHex/objecthash-fromhex-sha1-valid-16   	18120699     79.62 ns/op	 502.36 MB/s      72 B/op     2 allocs/op
~HashFromHex/objecthash-fromhex-sha256-valid-16 	 8969871    124.2 ns/op	     515.22 MB/s      96 B/op     2 allocs/op
```

Relates to #229 #899 

--- 
The other commits (together with the one described above) lay the foundations for a follow-up PR which introduces a `exp/plumbing/format/idxfile` which supports both SHA1 and SHA256.

### DIFF
--- a/exp/plumbing/hasher/hasher.go
+++ b/exp/plumbing/hasher/hasher.go
@@ -1,0 +1,135 @@
+package hasher
+
+import (
+	"crypto"
+	"fmt"
+	"strconv"
+	"sync"
+
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/hash"
+
+	format "github.com/go-git/go-git/v5/plumbing/format/config"
+)
+
+// ObjectHasher computes hashes for Git objects. A few differences
+// it has when compared to plumbing.Hasher:
+//
+//   - ObjectType awareness: produces either SHA1 or SHA256 hashes
+//     depending on the format needed.
+//   - Thread-safety.
+//   - API restricts ability of generating invalid hashes.
+type ObjectHasher interface {
+	// Size returns the length of resulting hash.
+	Size() int
+	// Compute calculates the hash of a Git object. The process involves
+	// first writing the object header, which contains the object type
+	// and content size, followed by the content itself.
+	Compute(ot plumbing.ObjectType, d []byte) (ImmutableHash, error)
+}
+
+// FromObjectFormat returns the correct ObjectHasher for the given
+// ObjectFormat.
+//
+// If the format is not recognised, an ErrInvalidObjectFormat error
+// is returned.
+func FromObjectFormat(f format.ObjectFormat) (ObjectHasher, error) {
+	switch f {
+	case format.SHA1:
+		return newHasherSHA1(), nil
+	case format.SHA256:
+		return newHasherSHA256(), nil
+	default:
+		return nil, format.ErrInvalidObjectFormat
+	}
+}
+
+// FromHash returns the correct ObjectHasher for the given
+// Hash.
+//
+// If the hash type is not recognised, an ErrUnsupportedHashFunction
+// error is returned.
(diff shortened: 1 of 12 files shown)


exp/plumbing/hasher/hasher.go
```diff
@@ -0,0 +1,135 @@
...
+	return h.hasher.Size()
+}
+
+func newHasherSHA256() *objectHasherSHA256 {
+	return &objectHasherSHA256{
+		hasher: hash.New(crypto.SHA256),
+	}
+}
+
+type objectHasherSHA256 struct {
```
aymanbagabas, line 101: both `objectHasherSHA1` and `objectHasherSHA256` share the same logic, can we combine both in one type?

go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	github.com/pjbgf/sha1cd v0.3.0
 	github.com/sergi/go-diff v1.1.0
 	github.com/skeema/knownhosts v1.2.1
+	github.com/stretchr/testify v1.8.4
```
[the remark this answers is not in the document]
pjbgf: I agree we should not have two testing libraries. However, `check.v1` is no longer maintained so we are better off removing it across the board. There is an [issue](https://github.com/go-git/go-git/issues/806#issuecomment-1626348136) already talking about it.

On go-git-fixtures, I [already marked](https://github.com/go-git/go-git-fixtures/commit/af32aeef74e78e5fcdd5a10700c0e2c9e290cfa2) the `check.v1` API as deprecated and on the new code for SHA256 the idea is to rely on either std Go or testify.

exp/plumbing/hasher/hasher.go
pjbgf: Having a single base type was my original design, which gave me a few options:
a) have two different sized arrays, to be used based on the hash at hand.
b) rely on a slice and size it based on the hash at hand.
c) use generics and rely on type constraints.

From the options above, the smallest impact in both performance and API changes was this. But I am not convinced this is it yet - hence why feedback on this is so important.

The original `plumbing.Hash` idea of using a fixed sized array as the base Hash type was really good in so many different ways - including performance. So whatever comes to replace it will need to be quite good.

PR #921, opened 2023-11-10
Refactor server logic

aymanbagabas:
Use go-git backend to serve the file transport instead of relying on the
git binary to do so.

Replace old ioutil.Pipe with std io.Pipe.

Add a clone command

Trace pktline encoder/scanner calls


cli/go-git/go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/go-git/v5/go-git/cli/go-git
 
 go 1.19
 
+replace github.com/go-git/go-git/v5 => ../../
+
```
pjbgf, line 6:
```suggestion

```

plumbing/transport/file/client.go
```diff
@@ -2,132 +2,134 @@
...
 
 import (
-	"bufio"
 	"errors"
 	"io"
 	"os"
-	"path/filepath"
-	"strings"
 
+	"github.com/go-git/go-git/v5/internal/repository"
```
pjbgf, line 9: As this change would be done within a minor release, let's also keep the previous behaviour (marked as deprecated) within `plumbing/transport/file` (say in `client_git.go`), so that users can revert back to via InstallProtocol - in case our implementation introduces some regression.

remote.go
```diff
@@ -1387,7 +1387,7 @@ func pushHashes(
 	allDelete bool,
 ) (*packp.ReportStatus, error) {
 
-	rd, wr := ioutil.Pipe()
+	rd, wr := io.Pipe()
```
pjbgf, line 1390: The existing pipe implementation seems to differ from the std library by having non blocking writes and for blocking reads until the writer is closed. How can we ensure this won't cause regressions? I think some of the errors we are getting may be related to that:

```
remote_test.go:1484:
    c.Assert(err, IsNil)
... value *fs.PathError = &fs.PathError{Op:"write", Path:"|1", Err:0x20} ("write |1: broken pipe")
```

aymanbagabas: Not really, the [bufpipe](https://github.com/acomagu/bufpipe) library is only used for [`js` builds](https://github.com/go-git/go-git/blob/master/utils/ioutil/pipe_js.go), all other builds use the std library which doesn't make sense :/

I'm going to put this change in a separate PR

aymanbagabas:
Superseded by https://github.com/go-git/go-git/pull/938

PR #922, opened 2023-11-13
utils: remove ioutil.Pipe and use std library io.Pipe

aymanbagabas:
ioutil.Pipe literally calls io.Pipe


pjbgf (approved):
@aymanbagabas great job, thanks for working on this. :bow:

Issue #915, opened 2023-11-08
Add support for git notes


JakobDev:
It would be really nice, if support for [git notes](https://git-scm.com/docs/git-notes) would be added. See [this Blog post](https://initialcommit.com/blog/git-notes) for an explanation.

You should add the functions `GetNote()`, `SetNote(text)` and `RemoveNote()` to a Commit.

Git notes are technically just files in the `notes/commits` ref. Try it yourself:
```bash
git init
git add --all
git commit --allow-empty -m "Test"
git notes add -m "This is a Note"
git checkout notes/commits
```
You should now see a file named after your commit which contains the Note.

JakobDev:
I created a proof of concept. It has many problems, but it works in general.

<details>
<summary>Code</summary>

```go
package main

import (
	"errors"
	"fmt"
	"io"
	"os"
	"path/filepath"
	"strings"
	"time"

	"github.com/go-git/go-git/v5"
	"github.com/go-git/go-git/v5/plumbing"
	"github.com/go-git/go-git/v5/plumbing/format/index"
	"github.com/go-git/go-git/v5/plumbing/object"
)

func CheckError(text string, err error) {
	if err != nil {
		fmt.Printf("%s: %s\n", text, err.Error())
		os.Exit(1)
	}
}

func GetNote(repo *git.Repository, repoPath string, commit *object.Commit) (string, error) {
	notes, err := repo.Notes()
	if err != nil {
		return "", err
	}

	noteRef, err := notes.Next()
	if err != nil {
		if errors.Is(err, io.EOF) {
			return "", nil
		}
		return "", err
	}

	noteCommit, err := repo.CommitObject(noteRef.Hash())
	if err != nil {
		return "", err
	}

	tree, err := noteCommit.Tree()
	if err != nil {
		return "", err
	}

	fi, err := tree.File(commit.Hash.String())
	if err != nil {
		if errors.Is(err, object.ErrFileNotFound) {
			return "", nil
		}
		return "", err
	}

	reader, err := fi.Blob.Reader()
	if err != nil {
		return "", err
	}

	by, err := io.ReadAll(reader)
	if err != nil {
		return "", err
	}

	return strings.TrimSpace(string(by)), nil
}

func editNote(repo *git.Repository, repoPath string, commit *object.Commit, remove bool, text string) error {
	notes, err := repo.Notes()
	if err != nil {
		return err
	}

	hasNotesRef := true
	noteRef, err := notes.Next()
	if err != nil {
[...]

PR #928, opened 2023-11-16
Add support for reading git notes

JakobDev:
Part of #915

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -1720,6 +1720,52 @@ func (r *Repository) resolveHashPrefix(hashStr string) []plumbing.Hash {
 	return hashes
 }
 
+// GetNote retuns the note for the given commit. returns a empty string, if the commit has no note.
+func (r *Repository) GetNote(commit *object.Commit) (string, error) {
+	notes, err := r.Notes()
+	if err != nil {
+		return "", err
+	}
+
+	noteRef, err := notes.Next()
+	if err != nil {
+		if errors.Is(err, io.EOF) {
+			return "", nil
+		}
+		return "", err
+	}
+
+	noteCommit, err := r.CommitObject(noteRef.Hash())
+	if err != nil {
+		return "", err
+	}
+
+	tree, err := noteCommit.Tree()
+	if err != nil {
+		return "", err
+	}
+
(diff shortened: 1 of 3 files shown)


repository_test.go
```diff
@@ -3333,3 +3333,20 @@ func BenchmarkPlainClone(b *testing.B) {
...
+	err := r.clone(context.Background(), &CloneOptions{URL: s.GetBasicLocalRepositoryURL()})
+	c.Assert(err, IsNil)
+
+	head, err := r.Head()
+	c.Assert(err, IsNil)
+
+	commit, err := r.CommitObject(head.Hash())
+	c.Assert(err, IsNil)
+
+	note, err := r.GetNote(commit)
```
pjbgf, line 3348: We need a test case that also includes notes in.

You can create a sample repository with 1-2 commits with some notes. Then you submit its pack as a PR to https://github.com/go-git/go-git-fixtures.

That would enable you to refer to that pack via `ByTag` as we have in other parts of the code base.

repository.go
```diff
@@ -1720,6 +1720,52 @@ func (r *Repository) resolveHashPrefix(hashStr string) []plumbing.Hash {
...
+		return "", err
+	}
+
+	noteRef, err := notes.Next()
+	if err != nil {
+		if errors.Is(err, io.EOF) {
+			return "", nil
+		}
+		return "", err
+	}
```
pjbgf, line 1736: Not sure I am missing something, but this looks like we are iterating through the first note only?

JakobDev: No, this is the first ref, not the first Note.  Notes are always in the `notes/commits` ref, so we won't have more than one ref.

pjbgf: Thanks for clarifying.

repository_test.go
pjbgf: Now that we got your [changes](https://github.com/go-git/go-git-fixtures/commit/e089218a43607121fbf7600d940ce7b44ce2f2e8) merged in the other repo, we just need to update the go-git-fixtures version to the latest.

pjbgf:
Please reword the commit to `git: Add support for reading git notes`, so that the PR Validation passes.

PR #936, opened 2023-11-17
Respect pktline error-line errors

aymanbagabas:
Pkt-line defines an error-line format to send errors from/to the client or server. This is mainly used with the git transport since it doesn't have a dedicated stderr channel but can also be used with both ssh and file transports.

This also moves the git-proto-request pktline type into a struct.

### DIFF
--- a/plumbing/format/pktline/error.go
+++ b/plumbing/format/pktline/error.go
@@ -1,0 +1,51 @@
+package pktline
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"strings"
+)
+
+var (
+	// ErrInvalidErrorLine is returned by Decode when the packet line is not an
+	// error line.
+	ErrInvalidErrorLine = errors.New("expected an error-line")
+
+	errPrefix = []byte("ERR ")
+)
+
+// ErrorLine is a packet line that contains an error message.
+// Once this packet is sent by client or server, the data transfer process is
+// terminated.
+// See https://git-scm.com/docs/pack-protocol#_pkt_line_format
+type ErrorLine struct {
+	Text string
+}
+
+// Error implements the error interface.
+func (e *ErrorLine) Error() string {
+	return e.Text
+}
+
+// Encode encodes the ErrorLine into a packet line.
+func (e *ErrorLine) Encode(w io.Writer) error {
+	p := NewEncoder(w)
+	return p.Encodef("%s%s\n", string(errPrefix), e.Text)
+}
+
+// Decode decodes a packet line into an ErrorLine.
+func (e *ErrorLine) Decode(r io.Reader) error {
+	s := NewScanner(r)
+	if !s.Scan() {
+		return s.Err()
+	}
+
+	line := s.Bytes()
+	if !bytes.HasPrefix(line, errPrefix) {
+		return ErrInvalidErrorLine
+	}
+
+	e.Text = strings.TrimSpace(string(line[4:]))
+	return nil
+}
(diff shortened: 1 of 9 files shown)


plumbing/format/pktline/error.go
```diff
@@ -0,0 +1,50 @@
...
+	"errors"
+	"io"
+)
+
+var (
+	// ErrInvalidErrorLine is returned by Decode when the packet line is not an
+	// error line.
+	ErrInvalidErrorLine = errors.New("expected an error-line")
+
+	err = []byte("ERR ")
```
pjbgf, line 14:
```suggestion
	errPrefix = []byte("ERR ")
```

plumbing/format/pktline/error.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Error implements the error interface.
+func (e *ErrorLine) Error() string {
+	return e.Text
+}
+
+// Encode encodes the ErrorLine into a packet line.
+func (e *ErrorLine) Encode(w io.Writer) error {
+	p := NewEncoder(w)
+	return p.Encodef("ERR %s", e.Text)
```
pjbgf, line 33:
```suggestion
	return p.Encodef("%s%s", string(errPrefix), e.Text)
```

plumbing/format/pktline/error.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Decode decodes a packet line into an ErrorLine.
+func (e *ErrorLine) Decode(r io.Reader) error {
+	s := NewScanner(r)
+	if !s.Scan() {
+		return s.Err()
+	}
+
+	line := s.Bytes()
+	if len(line) < 4 || !bytes.Equal(line[:4], err) {
```
pjbgf, line 44:
```suggestion
	if !bytes.HasPrefix(line, errPrefix) {
```

plumbing/format/pktline/scanner.go
```diff
@@ -69,6 +69,17 @@ func (s *Scanner) Scan() bool {
 	s.payload = s.payload[:l]
 	trace.Packet.Printf("packet: < %04x %s", l, s.payload)
 
+	if len(s.payload) > 4 &&
```
pjbgf, line 72: We probably could use `bytes.HasPrefix` here as well.

plumbing/protocol/packp/gitproto.go
```diff
@@ -0,0 +1,116 @@
...
+
+	return nil
+}
+
+// Encode encodes the request into the writer.
+func (g *GitProtoRequest) Encode(w io.Writer) error {
+	if err := g.validate(); err != nil {
+		return err
+	}
+
```
pjbgf, line 50: Check for `w` is `nil` - as this is part of the public API we should.

plumbing/protocol/packp/gitproto.go
```diff
@@ -0,0 +1,116 @@
...
+		return fmt.Errorf("%w: missing null terminator", ErrInvalidGitProtoRequest)
+	}
+
+	parts := strings.SplitN(line, " ", 2)
+	if len(parts) != 2 {
+		return fmt.Errorf("%w: short request", ErrInvalidGitProtoRequest)
+	}
+
+	g.RequestCommand = parts[0]
+	params := strings.Split(parts[1], "\x00")
```
pjbgf, line 97: Let's move `"\x00`" to a constant.

pjbgf (approved):

PR #937, opened 2023-11-19
Replace short field names with more descriptive ones.

matejrisek:
The decision to change the name of these fields came from reading the code further down in the scope and not being clear what `c.c` means.

This also aligns it with the convention from other structs in the same file like e.g. the `session` struct:
```
type session struct {
	auth     AuthMethod
	client   *http.Client
	endpoint *transport.Endpoint
	advRefs  *packp.AdvRefs
}
```


pjbgf (approved):
Thanks @matejrisek! :bow:

PR #939, opened 2023-11-22
git: stop iterating at oldest shallow when pulling. Fixes #305

### LINKED ISSUE #305: `object not found` when trying to pull a repository cloned with `Depth: 1`

dhoizner:
The `CommitPreorderIter` created to check whether two given commits are fast-forwardable was trying to iterate beyond the first commit of a shallow clone, this was causing `object not found` errors when trying to pull down new changes from the remote on a shallow clone.

This pull solves that by passing in an optional parameter specifying the earliest shallow commit and instructing the iterator to stop before trying to visit that commit's parents.

I wasn't sure where/what the best way of testing this was, it looks like the `Pull` tests in `worktree_test.go` use fixtures from [go-git/go-git-fixtures](https://github.com/go-git/go-git-fixtures), but I could use some guidance on the best way to implement a test for this kind of two-stage operation.


dhoizner:
i think the check failure here might be flaky tests, i had to run the test suite a couple of times locally before everything passed (with `-count=1` to ensure that cached results weren't being reused)

pjbgf:
@dhoizner based on [your comment](https://github.com/go-git/go-git/issues/305#issuecomment-1822042880), I am assuming the test for this could very similar to https://github.com/go-git/go-git/blob/fecea417bfc18648757a1bde30ca384548b55197/remote_test.go#L1497 with slight tweaks. 

I am not sure we need a new fixture for this, but if you think we do, here's [how to create new ones](https://github.com/go-git/go-git/issues/305#issuecomment-1822042880).

dhoizner:
> @dhoizner based on [your comment](https://github.com/go-git/go-git/issues/305#issuecomment-1822042880), I am assuming the test for this could very similar to https://github.com/go-git/go-git/blob/fecea417bfc18648757a1bde30ca384548b55197/remote_test.go#L1497 with slight tweaks. 
> 
> 
> 
> I am not sure we need a new fixture for this, but if you think we do, here's [how to create new ones](https://github.com/go-git/go-git/issues/305#issuecomment-1822042880).

Sweet, I was thinking that something like the fetch after shallow test would work also but wasn't sure if we'd want the tests to use fixtures like the rest of the worktree tests. 

I'll get that done in the next few days.

dhoizner:
@pjbgf added a test and a comment about a failure mode that will continue to exist, as well as updating the pull description explaining that failure mode. I think this is still worth merging in as is, with an issue to track the other scenario, but am open to suggestions

pjbgf (approved):
@dhoizner thanks for working on this. :1st_place_medal:

Tang8330:
Just wanted to say thank you for this fix @dhoizner !

ffhan:
@dhoizner thank you for this fix!

One thing to note, though (minimal example to reproduce below):

> clone with depth 1
push two new commits to the remote ref
pull with depth 1
If instead of pulling with depth 1 in the above scenario, the user were to pull with a depth of 2 or an unspecified depth, the pull would succeed.

If `Depth` for pulling is not specified this does not work. It is necessary to set it to a value above 1 to make this work.

```go
package main

import (
	"bufio"
	"context"
	"fmt"
	"github.com/go-git/go-git/v5"
	"github.com/go-git/go-git/v5/plumbing/transport/ssh"
	ssh2 "golang.org/x/crypto/ssh"
	"os"
	"os/signal"
)

func main() {
	if stat, err := os.Stat("./repos/testing"); err == nil && stat.IsDir() {
		if err := os.RemoveAll("./repos/testing"); err != nil {
			panic(err)
		}
	}
	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
	defer stop()

	keys, err := ssh.NewPublicKeysFromFile("git", "/Users/<user>/.ssh/id_rsa", "<password>")
	if err != nil {
		panic(err)
	}
	keys.HostKeyCallback = ssh2.InsecureIgnoreHostKey()

	repo, err := git.PlainCloneContext(ctx, "./repos/testing", false, &git.CloneOptions{
		URL:           "git.w2k.jumptrading.com:<user>/testing.git",
		SingleBranch:  true,
		ReferenceName: "refs/heads/main",
		Auth:          keys,
		Depth:         1,
	})
	if err != nil {
		panic(err)
	}
	fmt.Println("cloned")

	reader := bufio.NewReader(os.Stdin)
	for {
		fmt.Printf("newline to pull, exit to exit")
		// in the meantime commit and push to the repo so there's smth to pull
		line, _, err := reader.ReadLine()
		if err != nil {
			panic(err)
		}
		if string(line) == "exit" {
			return
		}
		tree, err := repo.Worktree()
		if err != nil {
			panic(err)
		}
		fmt.Println("pulling")
		if err = tree.PullContext(ctx, &git.PullOptions{
			Auth: keys,
			//Depth:         100000,
			SingleBranch:  true,
			ReferenceName: "refs/heads/main",
			Force:         true,
		}); err != nil {
			panic(err)
		}
	}
}
```

Runtime (with depth not set):
```
$ go run ./minimal
cloned
newline to pull, exit to exit
pulling
panic: object not found
```

runtime (with `Depth:         10000` in `PullContext`):
```
$ go run ./minimal
cloned
newline to pull, exit to exit
pulling
newline to pull, exit to exit^Cexit
```

In both examples between cloning and pulling I pushed a commit (tested this with multiple commits as well, same behaviour).

This should either be documented or fixed - I couldn't really figure out why this happens, nor why setting a depth to a big number fixes it. Are there any potential drawbacks to setting `Depth` when pulling? Especially for big repositories with a lot of commits.

dhoizner:
hi @ffhan thanks for trying it out! unfortunately, I wasn't able to reproduce with your example :/ 

Screenshot 2023-12-08 at 10 40 02 AM

Code and output are in this [gist](https://gist.github.com/dhoizner/c30782a4d93e634d84be4cc10fee38ed)

Can you make sure that you're referencing a version of `go-git` that includes this pull?

```go get github.com/go-git/go-git/v5@master``` will pull the current HEAD from master

PR #941, opened 2023-11-24
plumbing: object, enable renames in getFileStatsFromFilePatches

djmoch:
Diff has handled renames by default since 2020. This change sets Name for the renamed file in a manner similar to diffstat.


pjbgf:
@djmoch please add to avoid future regression.

djmoch:
@pjbgf I believe you were asking for tests. I've added that in aebf868d2a557384539cf91476d21901ab378643.

pjbgf (approved):
@djmoch thanks for working on this. :bow:

Issue #929, opened 2023-11-16
Repository.Push panics when it tries to push a tag with an empty name


peterzandbergen:
When I Clone a remote repo in memory with no working area, tag the head with an empty tag name and then push it back to origin, Push panics.

Steps to reproduce:
1. Clone a repo from e.g. github in memory
2. tag the head with an empty tag name
3. Push to origin, this panixs

```
// 1
repo, _ := git.CloneContext(ctx, memory.NewStorage(), nil, &co)

// 2
repo.CreateTag("", hash, opts)

// 
repo.PushContext(ctx, opts) // This generates a panic.
```

A test with my program using your great package (serious now)
```
# This command completes successfully.
 peza  c5361530-pop-os  ~  DevProjects  ggit  ERROR  $  go run ./cmd/ggit/     --repo-url https://gitlab.com/peterzandbergen/ggit-test.git      --loglevel debug     --branch main     --user peterzandbergen     --password *************     --name "Peter Zandbergen"     --email peter.zandbergen@myhops.nl     --message "Hallo daar"     --tag "new-tag-1"
time=2023-11-16T10:17:12.377+01:00 level=DEBUG msg="create logger" application=ggit loglevel=DEBUG logformat=text
time=2023-11-16T10:17:12.377+01:00 level=DEBUG msg="execute started" application=ggit
time=2023-11-16T10:17:12.377+01:00 level=DEBUG msg="using basic auth" application=ggit
time=2023-11-16T10:17:12.377+01:00 level=DEBUG msg="authenticator created" application=ggit
time=2023-11-16T10:17:13.005+01:00 level=DEBUG msg="repo cloned" application=ggit
time=2023-11-16T10:17:13.005+01:00 level=DEBUG msg="creating annotated tagger" application=ggit taggerName="Peter Zandbergen" taggerEmail=peter.zandbergen@myhops.nl
time=2023-11-16T10:17:13.005+01:00 level=DEBUG msg="creating annotated tagger" application=ggit taggerName="Peter Zandbergen" taggerEmail=peter.zandbergen@myhops.nl
time=2023-11-16T10:17:13.005+01:00 level=DEBUG msg="annotated tag" application=ggit
time=2023-11-16T10:17:13.005+01:00 level=DEBUG msg="head tagged" application=ggit
time=2023-11-16T10:17:13.005+01:00 level=DEBUG msg="created tag ref spec" application=ggit tagRefSpec=refs/tags/*:refs/tags/*
[...]

PR #949, opened 2023-11-28
git: error on empty tag name (#929)

aymanbagabas:
An empty tagname is an invalid tag.

Fixes: https://github.com/go-git/go-git/issues/929

Issue #927, opened 2023-11-15
PlainCloneContext fails on Windows for quirky ref names


piotrkowalczuk:
I've faced an issue where I could not clone the repository where one of the branches had `>` in its name.

```
open C:\Users\...\2023\11\15\github.com\black-desk\cgtproxy.git\.git\refs\remotes\origin\blackdesk\chore\update-config>master: The filename, directory name, or volume label syntax is incorrect.
```

pjbgf:
@onee-only thank you for looking into this. That makes sense, so closing it.

PR #950, opened 2023-11-28
git: validate reference names (#929)

aymanbagabas:
Check reference names format before creating branches/tags/remotes.

This should probably be in a lower level somewhere in `plumbing`. Validating the names under `plumbing.NewReference*` is not possible since these functions don't return errors.

Fixes: https://github.com/go-git/go-git/issues/929

See https://git-scm.com/docs/git-check-ref-format

### DIFF
--- a/plumbing/reference.go
+++ b/plumbing/reference.go
@@ -3,6 +3,7 @@ package plumbing
 import (
 	"errors"
 	"fmt"
+	"regexp"
 	"strings"
 )
 
@@ -29,6 +30,9 @@ var RefRevParseRules = []string{
 
 var (
 	ErrReferenceNotFound = errors.New("reference not found")
+
+	// ErrInvalidReferenceName is returned when a reference name is invalid.
+	ErrInvalidReferenceName = errors.New("invalid reference name")
 )
 
 // ReferenceType reference type's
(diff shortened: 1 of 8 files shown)


aymanbagabas:
Supersedes https://github.com/go-git/go-git/pull/949

worktree_test.go
```diff
@@ -785,6 +785,30 @@ func (s *WorktreeSuite) TestCheckoutCreateMissingBranch(c *C) {
...
+	}
+
+	for _, name := range []plumbing.ReferenceName{
+		"foo",
+		"-",
+		"-foo",
+		"refs/heads//",
+		"refs/heads/..",
+		"refs/heads/a..b",
+		"refs/heads/.",
```
pjbgf, line 801: Let's add an example with `>` as per https://github.com/go-git/go-git/issues/927

plumbing/reference.go
```diff
@@ -124,6 +128,85 @@ func (r ReferenceName) Short() string {
...
+	if r == HEAD {
+		return nil
+	}
+
+	// rule 7
+	if strings.HasSuffix(s, ".") {
+		return ErrInvalidReferenceName
+	}
+
+	// rule 2
```
pjbgf, line 177: Please add a comment that currently we don't support waiving this rule (e.g. `--allow-onelevel`).

plumbing/reference.go
```diff
@@ -124,6 +128,85 @@ func (r ReferenceName) Short() string {
...
+	for _, part := range parts {
+		// rule 6
+		if len(part) == 0 {
+			return ErrInvalidReferenceName
+		}
+
+		if strings.HasPrefix(part, ".") || // rule 1
+			strings.Contains(part, "..") || // rule 3
+			ctrlSeqs.MatchString(part) || // rule 4
+			strings.ContainsAny(part, "~^:?*[ \t\n") || // rule 4 & 5
```
pjbgf, line 194: Add a comment stating waving via `--refspec-pattern` is not supported.

plumbing/reference.go
```diff
@@ -124,6 +128,85 @@ func (r ReferenceName) Short() string {
...
+	// rule 2
+	parts := strings.Split(s, "/")
+	if len(parts) < 2 {
+		return ErrInvalidReferenceName
+	}
+
+	isBranch := r.IsBranch()
+	isTag := r.IsTag()
+	for _, part := range parts {
+		// rule 6
```
pjbgf, line 186: Same for `--normalize`.

pjbgf (approved):

pjbgf:
Great job @aymanbagabas! :trophy:

PR #953, opened 2023-11-30
storage: filesystem, Add option to set a specific FS for alternates

pjbgf:
Introduces the option to set a FS for alternates, enabling more flexible cross FS sharing of alternates. If none is set, falls back to the current FS used for the object storage.

The changes only process a given path once, and if an alternates dir is not valid, exits with error - aligning behaviour with upstream.

This breaks the previous behaviour, and affected code should use `AlternateFS` to resolve the problem, as per changes on `TestAlternatesRepo`.


storage/filesystem/storage.go
```diff
@@ -37,6 +37,10 @@ type Options struct {
 	// LargeObjectThreshold maximum object size (in bytes) that will be read in to memory.
 	// If left unset or set to 0 there is no limit
 	LargeObjectThreshold int64
+	// AlternatesFS provides the billy filesystem to be used for Git Alternates.
+	// If none is provided, fallbacks to using the underlying instance used for
```
hiddeco, line 41:
```suggestion
	// If none is provided, it falls back to using the underlying instance used for
```

I think?

hiddeco (approved):

zpavlinovic:
@mcuadros Are commits here fixing the vulnerability [GHSA-mw99-9chc-xw7r](https://github.com/advisories/GHSA-mw99-9chc-xw7r) you published?

wxiaoguang:
It seems that AlternateFS doesn't work with git's `clone --shared`. For example:

* There is a repo `/path1/....../gitea-repositories/user2/repo-pull-request-target.git`

* Then "clone --shared" to `/path2/....../temp.git`. 

* Now there is `../../../../../../../../gitea-repositories/user2/repo-pull-request-target.git/objects` in `temp.git/objects/info/alternatives`.

Then, whatever I used to construct AlternateFS, I couldn't make DotGit work, it is unable to handle the relative path? The only working approach is: `AlternativeFS: osfs.New("/")`

Issue #959, opened 2023-12-09
Unable to checkout by tag commit id.


alee-r7:
So I'm trying to checkout by the tag commit id. Basically I get the tag, its hash then do a checkout. Like this.

```
err = w.Checkout(&git.CheckoutOptions{
		Hash: commitHash,
		Force: true,
	})
```

I've verified the working directory has the commit, even checked objects folder by unpacking the repo I'm working with. I'm also able to checkout using the command line.

I tracked down the issue to this here. Here is the order

1. Checkout will call Reset which calls ResetSparsely.
2. ResetSparsely will then call getTreeFromCommitHash.  https://github.com/go-git/go-git/blob/5d08d3bd94c65a3b6c25c6fba6907d12b0dac4ca/worktree.go#L307
3. getTreeFromCommitHash will call r.CommitObject which then calls object.GetCommit.
https://github.com/go-git/go-git/blob/master/repository.go#L1424 
4. GetCommit will then call s.EncodedObject.  https://github.com/go-git/go-git/blob/master/plumbing/object/commit.go#L70
5. This is where it fails.  It seems based off the hash, the object that is retrieved from the pack or objects folder is a TabObject. However, it will fail on the last "IF" check because it expects a commit object. Due to the two types be different it returns an error.

```
func (s *ObjectStorage) EncodedObject(t plumbing.ObjectType, h plumbing.Hash) (plumbing.EncodedObject, error) {
	var obj plumbing.EncodedObject
	var err error

	if s.index != nil {
		obj, err = s.getFromPackfile(h, false)
		if err == plumbing.ErrObjectNotFound {
			obj, err = s.getFromUnpacked(h)
		}
	} else {
		obj, err = s.getFromUnpacked(h)
		if err == plumbing.ErrObjectNotFound {
			obj, err = s.getFromPackfile(h, false)
		}
	}

	// If the error is still object not found, check if it's a shared object
	// repository.
	if err == plumbing.ErrObjectNotFound {
		dotgits, e := s.dir.Alternates()
		if e == nil {
			// Create a new object storage with the DotGit(s) and check for the
			// required hash object. Skip when not found.
			for _, dg := range dotgits {
				o := NewObjectStorage(dg, s.objectCache)
				enobj, enerr := o.EncodedObject(t, h)
				if enerr != nil {
					continue
				}
				return enobj, nil
			}
		}
	}

	if err != nil {
		return nil, err
	}

	if plumbing.AnyObject != t && obj.Type() != t {
[...]

PR #966, opened 2023-12-12
git: worktree checkout tag hash id (#959)

aymanbagabas:
Allow checking out a worktree using a tag hash id.

Fixes: https://github.com/go-git/go-git/issues/959
Supersedes: https://github.com/go-git/go-git/pull/964


worktree.go
```diff
@@ -227,20 +227,21 @@ func (w *Worktree) createBranch(opts *CheckoutOptions) error {
...
+			return b.Hash(), nil
+		}
 
-	if !b.Name().IsTag() {
-		return b.Hash(), nil
+		hash = b.Hash()
 	}
 
-	o, err := w.r.Object(plumbing.AnyObject, b.Hash())
+	o, err := w.r.Object(plumbing.AnyObject, hash)
```
pjbgf, line 244: Keeping `plumbing.AnyObject` here enables an annotated tag to point to another. I am not sure this (nested tags) is supported upstream.

aymanbagabas: You can checkout an annotated tag commit hash upstream `git checkout <tag hash>`

AriehSchneier:
How often do people reference tags by their hash? Wouldn't it be simpler just to resolve the tag to commit before calling checkout?

aymanbagabas:
> How often do people reference tags by their hash? Wouldn't it be simpler just to resolve the tag to commit before calling checkout?

This simply follows the behavior of Git upstream. How often do people reference tags by their hashes, probably not much. Still, I think this is worth implementing since it affects the porcelain, the front facing, API which many people expect to behave the same as upstream Git.

pjbgf (approved):
Thanks @aymanbagabas! :1st_place_medal:

PR #967, opened 2023-12-13
plumbing: object, check legitimacy in (*Tree).Encode

niukuo:
if there is '\x00' in filename, Encode() will produce a malformed object.
also add some checks for entry order

### DIFF
--- a/plumbing/object/tree.go
+++ b/plumbing/object/tree.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"path"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/go-git/go-git/v5/plumbing"
@@ -27,6 +28,7 @@ var (
 	ErrFileNotFound      = errors.New("file not found")
 	ErrDirectoryNotFound = errors.New("directory not found")
 	ErrEntryNotFound     = errors.New("entry not found")
+	ErrEntriesNotSorted  = errors.New("entries in tree are not sorted")
 )
 
 // Tree is basically like a directory - it references a bunch of other trees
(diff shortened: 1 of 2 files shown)


niukuo:
here's a simple script
```
git init
touch a.b
touch a0b
mkdir a
touch a/b
git add a/b
git add a.b
git add a0b
git commit -m init
git cat-file -p HEAD^{tree}
```
results in 
```
100644 blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391	a.b
040000 tree 4277b6e69d25e5efa77c455340557b384a4c018a	a
100644 blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391	a0b
```
indicating that git will add a trailing slash for tree entry in sorting

zeripath:
Yup when encoding trees these entries should be sorted, with directories having an implicit '/' added at the end.

pjbgf:
@niukuo please add some tests to confirm the new behaviour. For example, the new errors we are returned should be part of it.

plumbing/object/tree.go
```diff
@@ -279,7 +302,15 @@ func (t *Tree) Encode(o plumbing.EncodedObject) (err error) {
 	}
 
 	defer ioutil.CheckClose(w, &err)
+
+	if !sort.IsSorted(TreeEntrySorter(t.Entries)) {
+		return errors.New("entries in tree is not sorted")
```
pjbgf, line 307: Let's make this a named error, moving it together with the other `Err...` vars at the top of the file.

plumbing/object/tree.go
```diff
@@ -279,7 +302,15 @@ func (t *Tree) Encode(o plumbing.EncodedObject) (err error) {
...
+	}
+
 	for _, entry := range t.Entries {
+		if strings.IndexByte(entry.Name, 0) != -1 {
+			return fmt.Errorf("malformed filename '%s'", entry.Name)
```
pjbgf, line 312:
```suggestion
			return fmt.Errorf("malformed filename %q", entry.Name)
```

niukuo:
> @niukuo please add some tests to confirm the new behaviour. For example, the new errors we are returned should be part of it.

done

pjbgf (approved):
Thanks @niukuo! :bow:

PR #968, opened 2023-12-13
Decode push options in update request

thegrumpylion:


pjbgf (changes requested):
A few nits, otherwise LGTM.

PR #969, opened 2023-12-14
plumbing: check setAuth error. Fixes #185

### LINKED ISSUE #185: Unlogged error in ssh/common.go
The file ssh/common.go contains the following code

```go
func (r *runner) Command(cmd string, ep *transport.Endpoint, auth transport.AuthMethod) (common.Command, error) {
	c := &command{command: cmd, endpoint: ep, config: r.config}
	if auth != nil {
		c.setAuth(auth)
	}

	if err := c.connect(); err != nil {
		return nil, err
	}
	return c, nil
}
```

The `setAuth` function returns an error, but any error returned is not logged or returned to the client.  This can make it difficult to customize the auth function of the library since users will be unaware their custom auth is not used.

nodivbyzero:
Connected to #185


pjbgf (approved):
Thanks @nodivbyzero! :1st_place_medal:

PR #992, opened 2024-01-10
utils: update comment in node.go's Hash()

codablock:
This reflects the lazy hash calculation and the implications of this.


pjbgf (approved):
Thanks @codablock! :bow:

Issue #993, opened 2024-01-11
Add option to skip git status check when adding a file


moranCohen26:
I noticed that in large repository Status() takes too long time (5 sec in my case). In scale this is something that would be nice to improve to get a better performance, more so if you already know that which files you modified and want to add to stage.
My suggestion is to add "skipStatus" option to AddWithOptions to gain a better performances. 
I will apply only if Path option is given and All option is not given. Anyway from my tests there should be no impact if you are trying to add a non modified file, behavior will be the same like you are not using the "skipStatus" flag.

moranCohen26:
Please see my [PR](https://github.com/go-git/go-git/pull/994) as suggested solution for this

moranCohen26:
@pjbgf will appreciate your response, if anything missing in my issue or PR please let me know :pray:

pjbgf:
Closing as per #994.

PR #994, opened 2024-01-11
Worktree.AddWithOptions: add skipStatus option when providing a specific path

moranCohen26:
Worktree.AddWithOptions: add skipStatus option when providing a specific path to add.
This will skip the git status check which might takes long time in a large repositories. 
This should Address https://github.com/go-git/go-git/issues/993

### DIFF
--- a/options.go
+++ b/options.go
@@ -474,6 +474,11 @@ type AddOptions struct {
 	// Glob adds all paths, matching pattern, to the index. If pattern matches a
 	// directory path, all directory contents are added to the index recursively.
 	Glob string
+	// SkipStatus adds the path with no status check. This option is relevant only
+	// when the `Path` option is specified and does not apply when the `All` option is used.
+	// Notice that when passing an ignored path it will be added anyway.
+	// When true it can speed up adding files to the worktree in very large repositories.
+	SkipStatus bool
 }
 
 // Validate validates the fields and sets the default values.
--- a/worktree_status.go
+++ b/worktree_status.go
@@ -271,7 +271,7 @@ func diffTreeIsEquals(a, b noder.Hasher) bool {
 // no error is returned. When path is a file, the blob.Hash is returned.
 func (w *Worktree) Add(path string) (plumbing.Hash, error) {
 	// TODO(mcuadros): deprecate in favor of AddWithOption in v6.
-	return w.doAdd(path, make([]gitignore.Pattern, 0))
+	return w.doAdd(path, make([]gitignore.Pattern, 0), false)
 }
 
 func (w *Worktree) doAddDirectory(idx *index.Index, s Status, directory string, ignorePattern []gitignore.Pattern) (added bool, err error) {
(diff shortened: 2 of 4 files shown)


moranCohen26:
thanks @pjbgf,  fixed and revert the behavior change in case of lstat error. Understand that error might be "file does not exist" in case of removed file, so continue to skip the error check. Checked that the 2 tests are passing now. Will appreciate if you can re-run the workflows again

options.go
```diff
@@ -474,6 +474,9 @@ type AddOptions struct {
 	// Glob adds all paths, matching pattern, to the index. If pattern matches a
 	// directory path, all directory contents are added to the index recursively.
 	Glob string
+	// SkipStatus adds the path with no status check. This option is relevant only
+	// when the `Path` option is specified and does not apply when the `All` option is used.
```
pjbgf, line 478:
```suggestion
	// SkipStatus adds the path with no status check. This option is relevant only
	// when the `Path` option is specified and does not apply when the `All` option is used.
	// When true it can speed up adding files to the worktree in very large repositories.
```

pjbgf (commented):
What happens with paths that are meant to be ignored? Please add tests to cover that scenario. We will also need to add that to the comments of the new option.

pjbgf:
For future reference, with this changes, the time for adding a small file on a shallow clone of `linux` decreased from `6.2s` to `0.1s`.

worktree_status.go
```diff
@@ -321,24 +321,19 @@ func (w *Worktree) AddWithOptions(opts *AddOptions) error {
...
+		_, err := w.doAdd(".", w.Excludes, false)
 		return err
 	}
 
 	if opts.Glob != "" {
 		return w.AddGlob(opts.Glob)
 	}
 
-	_, err := w.Add(opts.Path)
+	_, err := w.doAdd(opts.Path, make([]gitignore.Pattern, 0), opts.SkipStatus)
```
moranCohen26, line 332: @pjbgf before my change AddWithOptions for a specific file passed empty ignore pattern slice so a file meant to be ignored will be added. I preferred not to change/break this behavior although I'm not sure it's the right option. 
Do you think we should keep it as is?

moranCohen26:
> For future reference, with this changes, the time for adding a small file on a shallow clone of `linux` decreased from `6.2s` to `0.1s`.

See my comment please

worktree_status.go
moranCohen26: Or probably you mean file ignored by .gitignore and not excluded?

moranCohen26: ok, got you. We are not calling diffStagingWithWorktree which exclude the files specified in .gitingore because we are not calling Status(). 
So when skipStatus=true  if passing a path of file that meant to be ignored it will be added and if skipStatus=false it will not added. I will add a test and comment as you asked

pjbgf: Thank you for following this through.

pjbgf (approved):
@moranCohen26 thanks for working on this. :bow:

PR #995, opened 2024-01-11
Add option approximating `git clean -x` flag.

msuozzo:
This introduces an `All` option which, when enabled, removes files even when they are excluded by gitignore. This is equivalent to `git clean` with the `-x` flag but without the ability to customize the excluded patterns as provided by the `-e` flag.

### DIFF
--- a/options.go
+++ b/options.go
@@ -685,7 +685,10 @@ const (
 
 // CleanOptions describes how a clean should be performed.
 type CleanOptions struct {
+	// Dir recurses into nested directories.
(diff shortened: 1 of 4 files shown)


msuozzo:
Any chance this could get reviewed?

msuozzo:
Friendly ping

msuozzo:
bump

pjbgf (approved):
@msuozzo thanks for working on this. :bow:

PR #1009, opened 2024-01-24
git: remote, fix default remote name

dshemin:
Hi!

I had used this package in my project and found this annoying issue. I have a `Remote` with specific name and have to provide it again in `PushOptions`.

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -101,7 +101,7 @@ func (r *Remote) Push(o *PushOptions) error {
 // operation is complete, an error is returned. The context only affects the
 // transport operations.
 func (r *Remote) PushContext(ctx context.Context, o *PushOptions) (err error) {
-	if err := o.Validate(); err != nil {
+	if err := o.Validate(r.c.Name); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


remote.go
```diff
@@ -101,7 +101,7 @@ func (r *Remote) Push(o *PushOptions) error {
 // operation is complete, an error is returned. The context only affects the
 // transport operations.
 func (r *Remote) PushContext(ctx context.Context, o *PushOptions) (err error) {
-	if err := o.Validate(); err != nil {
+	if err := o.Validate(r.c.Name); err != nil {
```
zeripath, line 104: Instead of changing PushOptions.Validate why didn't you go for the simpler solution of doing:

```suggestion
	// If the PushOptions don't have a remote name set - just default them to this remote for ease of use
	if o.RemoteName == "" {
		o.RemoteName = r.c.Name
	}
	if err := o.Validate(); err != nil {
```

Then this PR would reduce considerably in size.

---

One issue with both of these approaches is that the PushOptions are passed in as a pointer so any changes here permanently change the object meaning that if the PushOptions are reused then the result will be different depending on who used them first.

This isn't ideal and perhaps indicates that the Valdate option shouldn't really be changing the struct or it should be returning a duplicate with the default values set, but given that it does it might be that this PR may be inadvisable.

dshemin: I've got you point. I will change.

dshemin: Another option: don't use `RemoteName` from PushOptions at all. Or create `RemotePushOption` struct without `RemoteName` field.

dshemin: Can anyone help me with failed tests?

PR #1016, opened 2024-02-01
add stubs to allow compilation under GOOS=tamago

abarisani:
The [TamaGo bare metal framework](https://github.com/usbarmory/tamago) would benefit from being able to integrate go-git on bare metal firmware.

This PR, in combination with [this go-billy PR](https://github.com/go-git/go-billy/pull/43) adds the required stubs to allow compilation under GOOS=tamago.

Thanks

### DIFF
--- a/worktree_tamago.go
+++ b/worktree_tamago.go
@@ -1,0 +1,26 @@
+// +build tamago
+
+package git
+
+import (
+	"syscall"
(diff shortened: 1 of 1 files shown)


pjbgf:
Please add some tests to ensure we won't get regressions once it is merged.

Issue #830, opened 2023-08-26
How to refer insteadof  in ~/.gitconfig？


yangyile1990:
my gitlab is in my ubuntu computer which ip is 192.168.88.88

my code is in my macbook. so these command are in macbook:
```
cat ~/.gitconfig 

[url "http://192.168.88.88:31080/"]
    insteadof = https://gitlab.yyle.com/
```

my code of go package is:
```
cat go.mod

module gitlab.yyle.com/golang/uvmanage.git

go 1.20
```

my project config git config is:
```
cat .git/config 
[core]
	repositoryformatversion = 0
	filemode = true
	bare = false
	logallrefupdates = true
	ignorecase = true
	precomposeunicode = true

[remote "origin"]
        url = https://gitlab.yyle.com/golang/uvmanage.git
        fetch = +refs/heads/*:refs/remotes/origin/*

[branch "main"]
        remote = origin
        merge = refs/heads/main
```

in this case, how to push my code with 
```
func (r *Repository) PushContext(ctx context.Context, o *PushOptions) error
```

or how to push with:
```
func (r *Remote) PushContext(ctx context.Context, o *PushOptions) (err error)
```

as you know.

the git remote is https://gitlab.yyle.com/golang/uvmanage.git

then use insteadof to get [url "http://192.168.88.88:31080/"]

then push.

but I don't know how to do this action by go-git pkg.

please give me an example.
1. loads the .gitconfig to get insteadof map
2. set the insteadof to somewhere.(I don't know where to set)
3. push the code.

this is the gpt answer but it is wrong:
```
func main() {

	// 打开本地仓库
	repo, err := git.PlainOpen("./repo")
	if err != nil {
		log.Fatal(err)
	}

	// 获取远程仓库
	remote, err := repo.Remote("origin")
	if err != nil {
		log.Fatal(err)
	}

	// 设置insteadof规则
	remote.Config().InsteadOf("https://github.com/", "git://github.com/")

	// 设置身份验证
	auth := &http.BasicAuth{
		Username: "username",
		Password: "password",
	}

	// push代码
	err = remote.PushContext(context.Background(), &git.PushOptions{
		Auth: auth,
	})
	if err != nil {
		log.Fatal(err)
	}

}
```

so why not give me a InsteadOf to make it easy to use?

I can see:
```
func (c *RemoteConfig) applyURLRules(urlRules map[string]*URL)
```
while I have no access to it.

how to call it?

davidalpert:
[...]

Issue #269, opened 2021-03-18
Init() does not honour the `init.defaultBranch` config setting.


jmalloc:
When initialising a new repository the default branch name is always `master`, even if the user has configured a different default in their global configuration file.

I am not sure if this library intends to universally mirror Git's exact behaviour or not - would a change that caused `git.Init()` to use this configuration value be feasible and welcome?

/cc @ezzatron

hairyhenderson:
Related, but I think a slightly different issue - #249

davidalpert:
This also may relate to #395 `support global config (i.e. $HOME/.gitconfig)`

kyoh86:
This is still important for me!

ezzatron:
This is still a pain point.

Issue #239, opened 2021-01-09
'git.PlainOpenWithOptions' does not consider worktree option from .git/config


jsattler:
When initializing an empty repository and changing the worktree option to some other directory, the function 'git.PlainOpenWithOptions' does not consider the specified worktree option under .git/config.

I expected that the worktree option under .git/config is considered when opening the Repository.
I could not find any option or configuration to open the repository with the configured worktree under .git/config.

nscnd:
Hey there,
I'm facing the same issue using `git.PlainOpenWithOptions` on a worktree fails to open the repo.
In that case, no `.git` dir but a `.git` file with `gitdir: /x/y/z `.
I think that it is expected that when running on a worktree we could still open the repo without knowing in advance this is not a fully fledge git repo.
Shouldn't we add an option to handle this case as there is an option to find the 'parent' git repo ?
Or maybe there already is a way to do this I didn't see ?

PR #1019, opened 2024-02-02
git: support partial opt-in to more correct loading of git config via repository.ConfigScoped()

davidalpert:
This PR is an attempt to offer a non-breaking backwards-compatible opt-in approach to an improved handling of loading git configuration according to the documented behavior of the git config subcommand (e.g. see the docs for the latest version as of this PR [v2.43.0](https://git-scm.com/docs/git-config/2.43.0#_description))

As discussed in #395 it was determined by maintainers that, while the go-git v5 implementation diverges from what clients
would expect, fixing it constitutes a breaking change for any client who has coded to the existing v5 config behavior.

Here we introduce new options to the `config.Scope` enumeration:
- V6DefaultScope
- V6SystemScope
- V6GlobalScope
- V6LocalScope

Using one of these options with `config.LoadConfig(scope Scope)` or `repository.ConfigScoped(scope config.Scope)` will load config according to the documented git behavior as described below:

| `config.Scope` value | `config.LoadConfig(..)` | `repository.ConfigScoped(..)` |
| -- | -- | -- |
| `config.V6DefaultScope` | loads system config, merges in global config, and merges in local config from the configured filesystem | same as `config.LoadConfig(config.V6DefaultScope)` then merges in local in-memory config from the repository's config storer |
| `config.V6SystemScope` | equivalent to `git config --system` | equivalent to `git config --system` |
| `config.V6GlobalScope` | equivalent to `git config --global` | equivalent to `git config --global` |
| `config.V6LocalScope` | equivalent to `git config --local` (from os) | equivalent to `git config --local` (from os + from config storer) |

NOTES: 
- This PR does not handle the `--worktree` nor `--file <filename>`  flags as both options require a more extensive refactoring or  reimplementation of the config system to access the repository configuration while resolving config values in the `--worktree` case and accept and pass the `<filename>` parameter to the config module in the `--file` case. Support for these options can be reconsidered in `go-git` v6 planning.
- I recommend that the entire config subsystem be revisited for `go-git` v6 as use of v5 `Repository.Config()` (without the ability to opt-in to more correct configuration loading behavior) is used extensively throughout the `git` package.


config/config.go
```diff
@@ -213,6 +309,48 @@ func Paths(scope Scope) ([]string, error) {
...
+	if !scope.IsV6PreviewScope() {
+		return nil, fmt.Errorf("this method only accepts V6 config scopes")
+	}
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return nil, err
+	}
+
+	systemWideConfigFiles := []string{"/etc/gitconfig"}
+	xdg := os.Getenv("XDG_CONFIG_HOME")
```
pjbgf, line 323: Please make the var names constants.

repository.go
```diff
@@ -563,7 +563,14 @@ func (r *Repository) SetConfig(cfg *config.Config) error {
 // are returned merged in one config value.
 func (r *Repository) ConfigScoped(scope config.Scope) (*config.Config, error) {
 	// TODO(mcuadros): v6, add this as ConfigOptions.Scoped
+	if scope.IsV6PreviewScope() {
+		return r.configScopedV6(scope)
+	}
+
+	return r.configScopedV5(scope)
```
pjbgf, line 570: I would bring the `configScopedV5` and `configScopedV6` logic into `config/config.go` and have a simple call to it here. Potentially following the same approach we discussed above:


```suggestion
	return scope.ConfigScoped()
```

config/config.go
```diff
@@ -35,18 +37,58 @@ var (
...
+	V6DefaultScope
+
+	// V6SystemScope is equivalent to running config commands with the --system flag
+	V6SystemScope
+
+	// V6GlobalScope is equivalent to running config commands with the --global flag
+	V6GlobalScope
+
+	// V6LocalScope is equivalent to running config commands with the --local flag
+	V6LocalScope
```
pjbgf, line 73: What are your thoughts on adding `V6` as a suffix instead of a prefix?

Also, what is the transition/migration strategy for the V6 release? Would we keep the `V6` suffixed Scopes and remove the previous ones?

It would be good to add a TODO for v6 on the outstanding steps.

config/config.go
```diff
@@ -35,18 +37,58 @@ var (
 	ErrRemoteConfigNotFound  = errors.New("remote config not found")
 	ErrRemoteConfigEmptyURL  = errors.New("remote config: empty URL")
 	ErrRemoteConfigEmptyName = errors.New("remote config: empty name")
+
+	// ActiveFS provides a configurable entry point to the billy.Filesystem in active use for reading global and
+	// system level configuration; override in a test to use a fake filesystem then reset to config.DefaultFS() to
+	// restore the default behavior.
+	ActiveFS = DefaultFS()
```
pjbgf, line 44: If this is for testing purposes only, I would make it private instead.

From a public API perspective, we probably want to get `Config` to work more like the core functionality, whereby such configuration is provided via `...WithOptions` when doing New/Load configs.

pjbgf (commented):
@davidalpert thank you for following up on this. Below are some ideas and nits:

pjbgf:
Please update `COMPATIBILITY.md` as part of this PR.

davidalpert:
Thank you @pjbgf I very much appreciate the feedback and will work on those updates today.

Would you prefer additional commits to this branch or rebased to fold this feedback into one or two overall commits?

config/config.go
davidalpert: Moving V6 to a suffix is a fine choice.

As for the transition/migration strategy, I would leave that to you and the maintaining team.

My recommendation from a client/consumer perspective would be to

1. have go-git v5 continue to default to the existing behavior with v6 as an opt-in choice
2. with the release of go-git v6 announce this as a breaking change, flip the default behavior to v6, and either remove or leave the v5 behavior as an opt-in choice for legacy compatibility. in this case, I suggest announcing that it will be maintained for backward compatibility until v7 (or some other point in the future) but is deprecated and no longer supported as the recommended support path is to update to the v6 behavior.

config/config.go
```diff
@@ -35,18 +37,58 @@ var (
...
 )
 
+// DefaultFS provides a billy.Filesystem abstraction over the os filesystem (via osfs.OS) scoped to the
+// root directory / in order to enable access to global and system config files via absolute paths
+func DefaultFS() billy.Filesystem {
+	return osfs.New("/")
```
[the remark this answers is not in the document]
davidalpert: great catch; will fix.

also going to replace
```
 var DefaultLoader = NewFilesystemLoader(osfs.New("")) 
```
with
```
 var DefaultLoader = NewFilesystemLoader(config.DefaultFS) 
```
to reduce that duplication and ensure consistency.

repository.go
davidalpert: This refactor may be trickier than the others, as it hits on one of the complications with the current approach.

The `Repository` has a `Storer` that acts as an abstraction over repository storage; this embeds a `config.ConfigStorer` which seems to be used to keep a copy of local config in memory.

git config commands generally act on the config in one of the stored locations (local, global, system, worktree, or file) and then complete with the config flushed to disk (or its `Storer` equivalent). subsequent commands then act on the resulting merged config by default unless one of those scope arguments is added.

In other words, go-git v5 acts on local configuration alone by default, while git config acts on a merged system/global/local config by default. This is the root cause of issues like #269 and #239 and impacts issues like #830

I will have to reflect on the implications of how to push this Repository-scoped function into the Config package.

davidalpert: Turns out that this is easier than I thought as the existing interface structure supports this refactor quite nicely.

Because `storage.Storer` embeds `config.ConfigStorer` and these methods depend only on the `config.ConfigStorer` part of `storage.Storer` these methods can move into the `config` package quite easily.

config/config.go
```diff
@@ -35,18 +37,58 @@ var (
...
+func (s Scope) IsV6PreviewScope() bool {
+	switch s {
+	case V6DefaultScope:
+		return true
+	case V6SystemScope:
+		return true
+	case V6GlobalScope:
+		return true
+	case V6LocalScope:
+		return true
```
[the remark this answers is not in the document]
davidalpert: TIL the syntax to collapse those `case` statements as this didn't work 😅 
```
    case V6DefaultScope:
    case V6SystemScope:
    case V6GlobalScope:
    case V6LocalScope:
      return true
```

config/config.go
davidalpert: While looking at this I see that a `Repository` stores a `billy.Filesystem` rooted at the worktree of the repo, which does not allow access to files outside the repo.

The core of #395 is that by default `git` operations require files outside the worktree as the default behavior is to merge system and global config into local config before executing any operation.

I can look into adding a `...WithOptions` parameter to pass in the root file system as well as the worktree fs though this may get confusing to consumers, passing in two different filesystems, especially as one is most often a subset of the other.

config/config.go
```diff
@@ -213,6 +329,97 @@ func Paths(scope Scope) ([]string, error) {
...
+			return nil, err
+		}
+	}
+
+	local, err := storer.Config()
+	if err != nil {
+		return nil, err
+	}
+
+	_ = mergo.Merge(global, system)
```
pjbgf, line 394:
```suggestion
	// The merge errors are ignored to keep them in line with existing behaviour.
	_ = mergo.Merge(global, system)
```

config/config.go
```diff
@@ -157,12 +246,11 @@ func ReadConfig(r io.Reader) (*Config, error) {
 	return cfg, nil
 }
 
-// LoadConfig loads a config file from a given scope. The returned Config,
-// contains exclusively information from the given scope. If it couldn't find a
-// config file to the given scope, an empty one is returned.
-func LoadConfig(scope Scope) (*Config, error) {
+func loadConfigV5(scope Scope, ofns ...WithOptions) (*Config, error) {
```
pjbgf, line 249: Do we no longer have `func LoadConfig(scope Scope) (*Config, error)`? This could break existing users.

config/config.go
```diff
@@ -37,16 +39,103 @@ var (
...
+	}
+
+	for _, fn := range ofns {
+		fn(&o)
+	}
+
+	return o
+}
+
+type WithOptions = func(*Options)
```
pjbgf, line 75:
```suggestion
type Option func(*Options)
```

config/config.go
```diff
@@ -37,16 +39,103 @@ var (
...
+		}
+
+		return nil, err
+	}
+
+	defer f.Close()
+	return ReadConfig(f)
+}
+
+func NewOptions(ofns ...WithOptions) Options {
```
pjbgf, line 63:
```suggestion
func NewOptions(ofns ...Option) Options {
```

config/config.go
```diff
@@ -37,16 +39,103 @@ var (
...
+	for _, fn := range ofns {
+		fn(&o)
+	}
+
+	return o
+}
+
+type WithOptions = func(*Options)
+
+func WithFS(fs billy.Filesystem) func(*Options) {
```
pjbgf, line 77: You could just refer to the type you defined:

```suggestion
func WithFS(fs billy.Filesystem) Option {
```

config/config.go
```diff
@@ -37,16 +39,103 @@ var (
...
+	return func(o *Options) {
+		o.fs = fs
+	}
+}
+
+// DefaultFS provides a billy.Filesystem abstraction over the os filesystem (via osfs.OS) scoped to the
+// root directory / in order to enable access to global and system config files via absolute paths
```
pjbgf, line 84:
```suggestion
// root directory / in order to enable access to global and system config files via absolute paths.
```

config/config.go
```diff
@@ -37,16 +39,103 @@ var (
...
+	// DefaultScopeV6 is equivalent to running config commands without any explicit flags
+	DefaultScopeV6
+
+	// SystemScopeV6 is equivalent to running config commands with the --system flag
+	SystemScopeV6
+
+	// GlobalScopeV6 is equivalent to running config commands with the --global flag
+	GlobalScopeV6
+
+	// LocalScopeV6 is equivalent to running config commands with the --local flag
```
pjbgf, line 108:
```suggestion
	// DefaultScopeV6 is equivalent to running config commands without any explicit flags.
	DefaultScopeV6

	// SystemScopeV6 is equivalent to running config commands with the --system flag.
	SystemScopeV6

	// GlobalScopeV6 is equivalent to running config commands with the --global flag.
	GlobalScopeV6

	// LocalScopeV6 is equivalent to running config commands with the --local flag.
```

repository.go
```diff
@@ -546,49 +548,25 @@ func cleanUpDir(path string, all bool) error {
 
 // Config return the repository config. In a filesystem backed repository this
 // means read the `.git/config`.
-func (r *Repository) Config() (*config.Config, error) {
-	return r.Storer.Config()
+func (r *Repository) Config(o ...config.WithOptions) (*config.Config, error) {
```
pjbgf, line 551: I generally find variadic functions quite neat, and probably something we should look into using more on V6. However, it is important to note that this can break backwards compatibility on this specific case.

For example, if we were to go ahead with this change I believe we would break [argocd-autopilot](https://github.com/argoproj-labs/argocd-autopilot/blob/55eb5621fbd4650604a151b3d0efc2592f7dcfdc/pkg/git/gogit/mocks/repository.go#L133) as they define an [interface](https://github.com/argoproj-labs/argocd-autopilot/blob/55eb5621fbd4650604a151b3d0efc2592f7dcfdc/pkg/git/gogit/repo.go#L22) based on our previous signature.

repository.go
```diff
@@ -546,49 +548,25 @@ func cleanUpDir(path string, all bool) error {
...
+// any other `Repository.ConfigScoped`.
 func (r *Repository) SetConfig(cfg *config.Config) error {
 	return r.Storer.SetConfig(cfg)
 }
 
 // ConfigScoped returns the repository config, merged with requested scope and
 // lower. For example if, config.GlobalScope is given the local and global config
 // are returned merged in one config value.
-func (r *Repository) ConfigScoped(scope config.Scope) (*config.Config, error) {
+func (r *Repository) ConfigScoped(scope config.Scope, o ...config.WithOptions) (*config.Config, error) {
```
pjbgf, line 567: Same here on the backwards compatibility. It is always hard to know how folks are using our public API, and changing it on a minor release would be problematic.

config/config.go
```diff
@@ -37,16 +39,103 @@ var (
 	ErrRemoteConfigEmptyName = errors.New("remote config: empty name")
 )
 
+// Options allows configuration of the config package
```
pjbgf, line 42: Maybe?

```suggestion
// Options represents all configurable options for the config package.
```

pjbgf (changes requested):
@davidalpert I really like where this got to, thank you very much for working on this and apologies for the delay getting a second review.

There are a few nits and some points around backwards compatibility. Apart from that it LGTM.

config/config.go
```diff
@@ -37,16 +39,103 @@ var (
...
+type Options struct {
```
pjbgf, line 43: This don't really need to be externally accessible.

```suggestion
type options struct {
```

davidalpert:
> @davidalpert I really like where this got to, thank you very much for working on this and apologies for the delay getting a second review.

@pjbgf I somehow missed this last round of feedback last year. I am glad that you like where this PR has gotten to. I really appreciate the engagement and fine tuning to help me shape this up to land with as little disruption as possible.

I plan rebase this PR and then take a closer look at your last round of feedback and clean it up later this week or this weekend, and will tag you again when I do.

pjbgf:
@davidalpert no worries at all. We are limiting the amount of changes to `master`, as we are working hard on getting ready for `v6`. Therefore, I would recommend on targeting the `v6-exp` branch instead - which then provides less strict rules around breaking backwards compatibility.

pjbgf:
@davidalpert 

> would you prefer I update this PR with a rebase/force-push or close this draft and open a net-new one on the v6-exp branch?

There is no much difference my side, so whatever is easiest for you. If you were to create a new one, just add a link back to this one (e.g. superses X) for context.

> with less concern about breaking changes is the v6-exp branch the place to flip the default behavior to align with v6 git and name the v5 behavior as the opt-in for backwards compatibilty?

Yes, that's sensible. 

> same question with regards to (r *Repository) Config() and (r *Repository) ScopedConfig(); is the v6-exp branch the place to change those signatures to accept variadic config options?

Yes, most definitely.

repository.go
davidalpert: For reference, in my testing a variadic parameter in golang is allowed to be empty, such that the signature above of
```
func (r *Repository) Config(o ...config.Option) ...
```
may be called with or without options (i.e. with an empty or populated slice of options) such that this remains valid syntax
```
r.Config()
```
and should be non-breaking but rather enable opt-in passing of config args without breaking existing `r.Config()` references.

PR #1023, opened 2024-02-05
git: worktree, Fix file reported as `Untracked` while it is committed

### LINKED ISSUE #119: File reported as Untracked while it is commited (i.e. Unmodified)
Assuming my full path to repo is `/Users/panteliskaramolegkos/myrepo`, I am going through the following steps to check the status of a specific file (code below is part of a specific function)

```
	repo, err := git.PlainOpen(fullPathToRepo)
	if err != nil {
		return false, fmt.Errorf("ERROR: Unable to open repository %s\n%s", fullPathToRepo, err)
	}

	workTree, err := repo.Worktree()
	if err != nil {
		return false, fmt.Errorf("ERROR: Unable to open worktree for repository %s\n%s", fullPathToRepo, err)
	}
	workTreeStatus, err := workTree.Status()
	if err != nil {
		return false, fmt.Errorf("ERROR: Unable to retrieve worktree status for repository %s\n%s", fullPathToRepo, err)
	}
	fmt.Printf("%q\n", workTreeStatus.File("relative/path/to/file.yaml").Worktree)
	fmt.Printf("%q\n", workTreeStatus.File("/Users/panteliskaramolegkos/full/path/to/file.yaml").Worktree)
```

In both cases (i.e. using full or relative path, the following gets printed out

```
'?'
'?'
```

According to the [documentation](https://pkg.go.dev/github.com/go-git/go-git/v5@v5.1.0?tab=doc#StatusCode), this corresponds to status `Untracked`

However the file is committed (and pushed to the remote, but I guess this is irrelevant)

rodrigocam:
## Overview

When using the lib I have also faced this problem #119. After looking into the code my understanding is that we were initializing the `status` empty before making the diff operations. This causes an empty status object when no diff is encountered. Which lead us to get `Untracked` status for committed files.

I manage to solve this by adding an initialization method for our `status` that iterates over the `storer` index building an initial status with all the unmodified files included. This way, when we use `status.File` it will find it on the map and will return correctly the `Unmodified` status.

Fix #119

### DIFF
--- a/worktree_status.go
+++ b/worktree_status.go
@@ -47,8 +47,45 @@ func (w *Worktree) Status() (Status, error) {
 	return w.status(hash)
 }
 
+func (w *Worktree) newStatusFromIndex() (Status, error) {
+	idx, err := w.r.Storer.Index()
+	if err != nil {
+		return nil, err
+	}
+
+	idxRoot := mindex.NewRootNode(idx)
+	nodes := []noder.Noder{idxRoot}
+
+	if err != nil {
+		return nil, err
+	}
+
+	status := make(Status)
+
+	for len(nodes) > 0 {
+		var node noder.Noder
+		node, nodes = nodes[0], nodes[1:]
+		if node.IsDir() {
+			children, err := node.Children()
+			if err != nil {
+				return nil, err
+			}
+			nodes = append(nodes, children...)
+			continue
+		}
+		fs := status.File(node.Name())
+		fs.Worktree = Unmodified
+		fs.Staging = Unmodified
+	}
+
+	return status, nil
+}
+
 func (w *Worktree) status(commit plumbing.Hash) (Status, error) {
-	s := make(Status)
+	s, err := w.newStatusFromIndex()
+	if err != nil {
+		return nil, err
+	}
 
 	left, err := w.diffCommitWithStaging(commit, false)
 	if err != nil {
--- a/worktree_status.go
+++ b/worktree_status.go
@@ -29,10 +29,23 @@ var (
 	// ErrGlobNoMatches in an AddGlob if the glob pattern does not match any
 	// files in the worktree.
 	ErrGlobNoMatches = errors.New("glob pattern did not match any files")
+	// ErrUnsupportedStatusStrategy occurs when an invalid StatusStrategy is used
+	// when processing the Worktree status.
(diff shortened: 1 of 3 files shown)


worktree_status.go
```diff
@@ -47,8 +47,45 @@ func (w *Worktree) Status() (Status, error) {
...
+			if err != nil {
+				return nil, err
+			}
+			nodes = append(nodes, children...)
+			continue
+		}
+		fs := status.File(node.Name())
+		fs.Worktree = Unmodified
+		fs.Staging = Unmodified
+	}
```
pjbgf, line 79: This may have a performance impact, as we need to go through all nodes while growing `nodes` as part of that process.

What if instead, we amended `File(path string)` to validate the returning [s[path]](https://github.com/go-git/go-git/blob/master/status.go#L20) and force a default value of `Unmodified` if no value was currently set?

rodrigocam: Sorry, I didn't understand what you proposed. Can you explain it a little more?

pjbgf: The issue here is that going through the entire worktree to create an initial state as unmodified can be quite costly - specially on large repositories. I tested this on a mid-sized repository and noticed that with these changes the `.Status()` call took ~30% longer than before.

My initial suggestion was to amend `Status` so that the `File` func would effectively do the same thing as the pre-init you created. However, this won't really work in some cases, as when there are no changes the `Status` returns an empty list. Besides, the existing logic is incorrect as whatever file we pass on to `status.File(path)` it returns `Untracked` even if that file does not exist across both the staging and the worktree.

In order to fix this, while keeping backwards compatibility and not introducing a performance regression, I think we have a few options:
a) Introduce a new `Worktree.StatusWithOptions()` func which returns a new type that is more flexible and calculates the status of a given file on demand. I suspect this will need to have access to the worktree fs, so that it can handle the scenario I mentioned above, without needing to front-load the entire worktree. It may also need access to the exclusion rules.
b) Introduce a new `Worktree.StatusWithOptions()` func which returns a new type that is more flexible, but simply adds an option so users can opt-in to the front-loading this PR currently introduces.
c) Introduce a new `Worktree.FileStatus()` func which calls `.Status` and ensure the given file is correctly loaded to the Status map.

What are your thoughts on the above?

pjbgf (approved):
@rodrigocam thank you for working on this. I made some of the API changes we discussed before for improved extensibility, otherwise LGTM.

Issue #912, opened 2023-11-07
Create CI tests to confirm all examples build


pjbgf:
As changes are introduced into the main code base, it would be great to have a CI check that validates the code examples within `_examples` to confirm they are not broken by changes introduced from a given PR.

PR #446, opened 2022-01-12
checkout-branch example

dlambda:
This PR adds a `checkout-branch` example and some minor fixes for testing.

It took me a while to figure out how to check-out a branch in my cloned repo, and thought this might save some time and could be useful helping someone else to do so.
Originally, after some digging, I used the method used in this issue: https://github.com/src-d/go-git/issues/777
However, realised it was copying all references, and causes faulty HEAD references in some cases (git "refname HEAD is ambiguous" warning I think, but can't remember exactly now).  Couldn't find any examples of fetching a single branch, hence thought this might be useful.

PR #1030, opened 2024-02-15
build: run example tests during CI workflow

crazybolillo:
A new test has been added in order to verify that all code under
_examples is up to date and has not broken due to API changes.

This test runs as the last step of the test.yml CI workflow.

Closes https://github.com/go-git/go-git/issues/912.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,7 +35,5 @@ jobs:
     - name: Test
       run: make test-coverage
 
-    - name: Examples
-      env:
-        EXAMPLES_DIR: _examples
-      run: go test -test.v -test.run='^TestBuildExamples$'
+    - name: Test Examples
+      run: go test -timeout 30s -v -run '^TestExamples$' github.com/go-git/go-git/v5/_examples --examples
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,3 +34,8 @@ jobs:
 
     - name: Test
       run: make test-coverage
+
+    - name: Examples
+      env:
+        EXAMPLES_DIR: _examples
+      run: go test -test.v -test.run='^TestBuildExamples$'


pjbgf:
@crazybolillo thanks for working on this. But I believe this were made redundant by the changes recently merged from https://github.com/go-git/go-git/pull/446. More specifically https://github.com/go-git/go-git/pull/446/commits/cc10b2a39c46bd22359d88dd5ec94702289d4fbc.

crazybolillo:
That is a nice PR. I have two comments/questions

1. Seems like examples tests are still not being run on the CI? From what I understand, the `test-coverage` target does not run them.
2. Maybe ignored tests can be built but not run? I don't know if this is worth it, but it could catch some issues like the ls example using a deprecated version of commitgraph.

.github/workflows/test.yml
```diff
@@ -34,3 +34,8 @@ jobs:
 
     - name: Test
       run: make test-coverage
+
+    - name: Verify Examples
+      env:
+        EXAMPLES_DIR: _examples
+      run: go test -v -run='^TestBuildExamples$'
```
pjbgf, line 41:
```suggestion
      run: make test
```

.github/workflows/test.yml
```diff
@@ -34,3 +34,8 @@ jobs:
 
     - name: Test
       run: make test-coverage
+
+    - name: Verify Examples
```
pjbgf, line 38: The existing (above) we could change to something like `Test Coverage`.

```suggestion
    - name: Test
```

build_examples_test.go
```diff
@@ -0,0 +1,38 @@
+package git
```
pjbgf, line 1: We don't need this file, as `make test` already does the trick:

```
$ make test
go test  -v _examples/common_test.go _examples/common.go --examples
=== RUN   TestExamples
=== RUN   TestExamples/blame
git clone https://github.com/git-fixtures/basic.git /tmp/go-git-blame-3662081995
git blame CHANGELOG
b8e471f5 (Daniel Ripolles 2015-03-31 13:44:52 +0200 1) Initial changelog
=== RUN   TestExamples/branch
git clone https://github.com/git-fixtures/basic.git /tmp/3841571656
git branch my-branch
git branch -D my-branch
=== RUN   TestExamples/checkout
git clone https://github.com/git-fixtures/basic.git /tmp/3011666005
git show-ref --head HEAD
6ecf0ef2c2dffb796033e5a02219af86ec6584e5
git checkout 35e85108805c84807bc66a02d91535e1e24b38b9
git show-ref --head HEAD
35e85108805c84807bc66a02d91535e1e24b38b9
=== RUN   TestExamples/checkout-branch
git clone https://github.com/git-fixtures/basic.git /tmp/1284450122
git show-ref --head HEAD
6ecf0ef2c2dffb796033e5a02219af86ec6584e5
git checkout branch
local checkout of branch 'branch' failed, will attempt to fetch remote branch of same name.
like `git checkout <branch>` defaulting to `git checkout -b <branch> --track <remote>/<branch>`
checked out branch: branch
git show-ref --head HEAD
...
```

pjbgf (changes requested):
@crazybolillo you are right, the examples were not being executed at CI. I proposed a few changes to this PR that aims to reuse the existing code to achieve the same results, PTAL.

build_examples_test.go
crazybolillo: Yes, I have removed the file

.github/workflows/test.yml
crazybolillo: Since "normal" tests are already being run as part of the coverage, isn't it better to just run the example tests for this new step? It would make the CI faster.

pjbgf: The suggestion for `make test` was because that would also introduce `-race`, which we are not doing at the moment.

crazybolillo: I see, I have changed it to `make test`, however a test is failing. I will look into it to see why.

pjbgf: Hey @crazybolillo, let's keep the previous change you had `go test -v -run='^TestBuildExamples$'`. 

The `make test` has some intermittent issues due to the `-race`. Feel free to propose a separate PR for that.

crazybolillo: Sorry for the late reply. I have gone back to just the build examples. Will see if I can create a separate PR to fix the flakiness from `-race`

pjbgf (approved):

PR #1032, opened 2024-02-21
git: Fix fetching missing commits

AriehSchneier:
This fixes fetching commits that have a reference to them but are not in the pack files.

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/go-git/go-billy/v5/osfs"
+
 	"github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/internal/url"
 	"github.com/go-git/go-git/v5/plumbing"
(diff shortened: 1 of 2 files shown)


remote_test.go
```diff
@@ -1220,17 +1221,20 @@ func (s *RemoteSuite) TestGetHaves(c *C) {
...
 		plumbing.NewReferenceFromStrings(
 			"bar",
-			"fe6cb94756faa81e5ed9240f9191b833db5f40ae",
+			"b8e471f58bcbca63b07bda20e428190409c2db47",
 		),
+		// Doesn't Exist
 		plumbing.NewReferenceFromStrings(
 			"qux",
-			"f7b877701fbf855b44c0a9e86f3fdce2c298b07f",
+			"0000000",
```
pjbgf, line 1237: Having to change this, feels like we are breaking existing/expected behaviour.

pjbgf:
Please add a new test which reproduces the issue you are experiencing. That test should fail without the changes  this PR introduces.

remote_test.go
AriehSchneier: I think the test was wrong, it was passing in a list of references that didn't actually exist

pjbgf: You are right, the previous commits did not exist at [the target](https://github.com/git-fixtures/basic/commits/master/). Thank you for updating and adding comments to that effect.

remote.go
```diff
@@ -849,17 +861,12 @@ func getHavesFromRef(
 		return nil
 	}
 
-	// No need to load the commit if we know the remote already
-	// has this hash.
-	if remoteRefs[h] {
-		haves[h] = true
-		return nil
-	}
```
pjbgf, line 857: Would this not cause a performance regression, as we may be increasing the memory churn?

pjbgf:
@AriehSchneier I finally had time to circle back on this one, apologies for the delay. Please rebase the PR so that I can go ahead and merge it.

AriehSchneier:
> @AriehSchneier I finally had time to circle back on this one, apologies for the delay. Please rebase the PR so that I can go ahead and merge it.

@pjbgf rebased

pjbgf (approved):

PR #1034, opened 2024-02-26
storage, dotgit: translate error code in *DotGit.Object

mallenLF:
Callers of *DotGit.Object expect an os.ErrNotExist error when an object file is not found, but the internal hasObject function might return a plumbing.ErrObjectNotFound error, which needs to be translated.

### DIFF
--- a/storage/filesystem/dotgit/dotgit.go
+++ b/storage/filesystem/dotgit/dotgit.go
@@ -608,6 +608,9 @@ func (d *DotGit) hasIncomingObjects() bool {
 func (d *DotGit) Object(h plumbing.Hash) (billy.File, error) {
 	err := d.hasObject(h)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


storage/filesystem/dotgit/dotgit.go
```diff
@@ -608,6 +608,9 @@ func (d *DotGit) hasIncomingObjects() bool {
 func (d *DotGit) Object(h plumbing.Hash) (billy.File, error) {
 	err := d.hasObject(h)
 	if err != nil {
+		if err == plumbing.ErrObjectNotFound {
```
pjbgf, line 611:
```suggestion
		if errors.Is(err, plumbing.ErrObjectNotFound)
```

pjbgf:
@mallenLF please add some tests that reproduce the issue.

PR #1042, opened 2024-03-06
git: worktree, Don't panic on empty or root path when checking if it is valid

tim775:
I didn't dig into the specific case that was triggering this, but we did have a panic in our production system.


pjbgf (approved):
LGTM! Thanks @tim775! :bow:

PR #1073, opened 2024-04-08
fix(up): divide the encoding of UploadPackReponse into two steps

luckjay:
This PR is to divide the encoding of UploadPackReponse into two steps. So the user can use two different writers for pkt-lines and packfiles when integrating with `side-band` enabled clients, for example, [isomorphic-git](https://isomorphic-git.org/).

For example, the pseudo-code is like the following:
```
var w io.writer
if err = response.EncodePacketLines(w); err != nil {
	return error
}

muxer := sideband.NewMuxer(sideband.Sideband, w)
if err = response.EncodePackfiles(muxer); err != nil {
	return err
}

return nil
```


luckjay:
@pjbgf could you review this pr, thanks!

PR #1082, opened 2024-04-24
Refactor pktline

aymanbagabas:
Taken from https://github.com/aymanbagabas/go-git/pull/1
- [x] Refactor pktline https://github.com/aymanbagabas/go-git/pull/1
  - [x] Clean up and DRY implementation
  - [x] Replace Encoder/Scanner with static methods
  - [x] Support Peek
  - [x] Improve performance


plumbing/protocol/packp/updreq_decode.go
```diff
@@ -83,14 +83,16 @@ func (req *ReferenceUpdateRequest) Decode(r io.Reader) error {
...
 
-	d := &updReqDecoder{r: rc, s: pktline.NewScanner(r)}
+	d := &updReqDecoder{r: rc, s: r}
 	return d.Decode(req)
 }
 
 type updReqDecoder struct {
 	r   io.ReadCloser
-	s   *pktline.Scanner
+	s   io.Reader
```
pjbgf, line 92: The field `s` sounds confusing now that this is no longer a scanner.

```suggestion
	pr   io.Reader
```

plumbing/transport/registry.go
```diff
@@ -0,0 +1,39 @@
...
+	"sync"
+)
+
+// registry are the protocols supported by default.
+var (
+	registry = map[string]Transport{}
+	mtx      sync.Mutex
+)
+
+// Register adds or modifies an existing protocol.
```
pjbgf, line 14: Where the API changed substantially, I think we should either add comments to clarify that, or find another place to amass all the API changes, so it is easier for users to migrate to V6.

```suggestion
// Register adds or modifies an existing protocol.
// Equivalent to client.InstallProtocol in go-git before V6.
```

plumbing/transport/registry.go
```diff
@@ -0,0 +1,39 @@
...
+func Register(protocol string, c Transport) {
+	mtx.Lock()
+	defer mtx.Unlock()
+	registry[protocol] = c
```
pjbgf, line 18: When the code is this simple it is not worth the additional costs of a `defer`.

```suggestion
	registry[protocol] = c
	mtx.Unlock()
```

plumbing/transport/registry.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// Unregister removes a protocol from the list of supported protocols.
+func Unregister(scheme string) {
+	mtx.Lock()
+	defer mtx.Unlock()
+	delete(registry, scheme)
```
pjbgf, line 25:
```suggestion
	delete(registry, scheme)
	defer mtx.Unlock()
```

pjbgf (commented):
Small nits, otherwise LGTM.

pjbgf (approved):

PR #1095, opened 2024-05-07
plumbing: protocol/packp/sideband, Expose Flush() to make muxer usable.

ajnavarro:
Without printing the Flush command, the muxer cannot be used to communicate with a git client. Exposed Flush() method from the pkt-line encoder to be able to call it through the Muxer.


pjbgf (approved):
LGTM

PR #1100, opened 2024-05-24
plumbing: transport/http, Wrap http errors to return reason. Fixes #1097

### LINKED ISSUE #1097: plumbing/transport/http: error handling swallows useful debugging information
I was recently investigating an issue that manifested as Git operations failing to complete while API access continued to function when accessing a popular Git provider. The issue was caused by SSO enforcement on the credentials being used: the provider was configured to require an SSO check every 24h for Git operations.

Unfortunately, the only messaging I was getting out of this library was "authorization failed" (coming from https://github.com/go-git/go-git/blob/master/plumbing/transport/common.go#L35).

Digging into this a bit, I found the `transport/http` [error handling](https://github.com/go-git/go-git/blob/master/plumbing/transport/http/common.go#L414-L441), which discards the response body when the HTTP code maps to a known error type. This hides the message from the Git provider which is akin to "account configured for SSO" or some such.

I think a better way to handle these errors would be to wrap the common error kind using `fmt.Errorf("%w: %s", err, reason)`. This is compatible with the current error output if folks are using [`errors.Is(err, target) bool`](https://pkg.go.dev/errors#Is) to check error equality (which folks should be doing).

With regards to being able to pass the reason string back to the caller: I don't expect that any Git provider would generate output for these errors that would be problematic to pass back to a consumer of the library.

eg:
```
// instead of this
	switch r.StatusCode {
	case http.StatusUnauthorized:
		return transport.ErrAuthenticationRequired
	case http.StatusForbidden:
		return transport.ErrAuthorizationFailed
	case http.StatusNotFound:
		return transport.ErrRepositoryNotFound
	}

// do this
	switch r.StatusCode {
	case http.StatusUnauthorized:
		return fmt.Errorf("%w: %s", transport.ErrAuthenticationRequired, reason)
	case http.StatusForbidden:
		return fmt.Errorf("%w: %s", transport.ErrAuthorizationFailed, reason)
	case http.StatusNotFound:
		return fmt.Errorf("%w: %s", transport.ErrRepositoryNotFound, reason)
	}
```

ggambetti:
This attempts to address https://github.com/go-git/go-git/issues/1097.

A few notes:
 * The underlying testing library doesn't have an `errors.Is` check, so I went and implemented one in `internal/test`. Not sure that's the right spot to put it.
 * The `internal/test.ErrorIs` can be removed if https://github.com/go-check/check/issues/139 gets fixed.
 * I kept the `ErrorMatches` string check in the places I found it.
 * I didn't provide tests for `internal/test.ErrorIs`, would you like them?

### DIFF
--- a/internal/test/checkers.go
+++ b/internal/test/checkers.go
@@ -1,0 +1,43 @@
+package test
+
+import (
+	"errors"
+	"fmt"
+
+	check "gopkg.in/check.v1"
+)
+
+// This check.Checker implementation exists because there's no implementation
+// in the library that compares errors using `errors.Is`. If / when the check
+// library fixes https://github.com/go-check/check/issues/139, this code can
+// likely be removed and replaced with the library implementation.
+//
+// Added in Go 1.13 [https://go.dev/blog/go1.13-errors] `errors.Is` is the
+// best mechanism to use to compare errors that might be wrapped in other
+// errors.
+type errorIsChecker struct {
+	*check.CheckerInfo
+}
+
+var ErrorIs check.Checker = errorIsChecker{
+	&check.CheckerInfo{
+		Name:   "ErrorIs",
+		Params: []string{"obtained", "expected"},
+	},
+}
+
+func (e errorIsChecker) Check(params []interface{}, names []string) (bool, string) {
+	obtained, ok := params[0].(error)
+	if !ok {
+		return false, "obtained is not an error"
+	}
+	expected, ok := params[1].(error)
+	if !ok {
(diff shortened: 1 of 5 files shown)


ggambetti:
@pjbgf

pjbgf:
@ggambetti thanks for working on this. LGTM, just need to reword the commit message (in line with the new PR title), and PR Validation should pass.

ggambetti:
@pjbgf Thanks, updated. :)

pjbgf (approved):

PR #1108, opened 2024-06-06
Add Filter to FetchOptions

vzxv:
Allow a Filter to be specified in FetchOptions.

Positively set the Filter capability when constructing an UploadRequest, and include the Filter when constructing an UploadPackRequest.


pjbgf:
This was superseded by #1375.

Issue #37, opened 2020-04-23
"reference has changed concurrently" when fetch and checkout.


miltalex:
When I try to do a git fetch and then a git checkout to a specific revision I get the following error. 

`error: reference has changed concurrently`

It seems that after 2 retries the whole fetch checkout operation works fine.

There was an old similar issue [here](https://github.com/src-d/go-git/issues/1230).

mcuadros:
Are you using go routines?

miltalex:
No. I just do a git fetch in the beggining.
```
err = repo.Fetch(&git.FetchOptions{
	RemoteName: "origin",
	Auth:       r.Auth,
})
```
Then I resolve the reference that user provided, by checking `referenceTag` and `referenceBranch`, and then I do a checkout and a pull.
```
err = w.Checkout(&git.CheckoutOptions{
	Hash:  *revision,
	Force: true,
})

err = w.Pull(&git.PullOptions{
	RemoteName:    "origin",
	ReferenceName: plumbing.ReferenceName(r.ReferenceName),
	Auth:          r.Auth,
}) 
```

A summary of the code.

mcuadros:
Can you create a script that reproduces the bug, to be able to take a look?

miltalex:
Unfortunately not I can share the code. I noticed that this will happen if something changed on the remote when we call the function, but I wasn't able to reproduce it.

cumminsd:
I also am getting the issue occasionally. It seems to be breaking references:

`error: cannot lock ref 'refs/remotes/origin/EVO-26728': unable to resolve reference 'refs/remotes/origin/EVO-26728': reference broken`

It does appear to be when someone else changes the repository. Here is the function I am calling to update the repository. Note that the commented out section was used to check the repository before calling update, but I entered an issue for that. (If there is an empty folder in the repository, worktree.Status() returns that the tree is dirty, if when it isn't.)

`
// Update does a pull on the specified repository
func Update(folder string) ([]*object.Commit, error) {
	repo, err := git.PlainOpen(folder)
	if err != nil {
		return nil, fmt.Errorf("Error opening repository: %s", err)
	}

	auth := githttp.BasicAuth{Username: "xxxxx", Password: "xxxxx"}

	oldHead, err := repo.Head()
	if err != nil {
		return nil, fmt.Errorf("Error finding HEAD: %s", err)
	}

	oldCommit, err := repo.CommitObject(oldHead.Hash())
	if err != nil {
[...]

PR #1116, opened 2024-06-18
Fix reference updated concurrently error for the filesystem storer

Javier-varez:
If a reference exists in the packed refs file but not in the actual refs folder, we need to fall back to reading the pack refs file when storing the reference.

This happens because by the time we are storing the reference, we open the file with write permissions and read from it. At that point, if the file is empty, we still need to read the valid reference. Otherwise, it will read 0, and assume that the references do not match (because the old value was read from the packed file).

Relates to https://github.com/go-git/go-git/issues/37

### DIFF
--- a/storage/filesystem/dotgit/dotgit.go
+++ b/storage/filesystem/dotgit/dotgit.go
@@ -72,6 +72,9 @@ var (
 	// ErrIsDir is returned when a reference file is attempting to be read,
 	// but the path specified is a directory.
 	ErrIsDir = errors.New("reference path is a directory")
+	// ErrEmptyRefFile is returned when a reference file is attempted to be read,
+	// but the file is empty
+	ErrEmptyRefFile = errors.New("ref file is empty")
 )
 
 // Options holds configuration for the storage.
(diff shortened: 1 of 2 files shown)


storage/filesystem/dotgit/dotgit.go
```diff
@@ -655,24 +655,47 @@ func (d *DotGit) ObjectDelete(h plumbing.Hash) error {
...
+// checkReferenceAndTruncate reads the reference from the given file, or the `pack-refs` file if
+// the file was empty. Then it checks that the old reference matches the stored reference and
+// truncates the file.
 func (d *DotGit) checkReferenceAndTruncate(f billy.File, old *plumbing.Reference) error {
 	if old == nil {
 		return nil
 	}
+
 	ref, err := d.readReferenceFrom(f, old.Name().String())
+	if err == emptyRefFileError {
```
pjbgf, line 689:
```suggestion
	if errors.Is(err, ErrEmptyRefFile) {
```

storage/filesystem/dotgit/dotgit.go
```diff
@@ -655,24 +655,47 @@ func (d *DotGit) ObjectDelete(h plumbing.Hash) error {
 	return err1
 }
 
+type refError string
+
+func (r refError) Error() string {
+	return string(r)
+}
+
+var emptyRefFileError = refError("The file is empty")
```
pjbgf, line 664: Please move this to the top together with the other errors.

```suggestion
var ErrEmptyRefFile = errors.New("ref file is empty")
```

storage/filesystem/dotgit/dotgit.go
```diff
@@ -655,24 +655,47 @@ func (d *DotGit) ObjectDelete(h plumbing.Hash) error {
...
+
 func (d *DotGit) readReferenceFrom(rd io.Reader, name string) (ref *plumbing.Reference, err error) {
 	b, err := io.ReadAll(rd)
 	if err != nil {
 		return nil, err
 	}
 
+	if len(b) == 0 {
+		return nil, emptyRefFileError
```
pjbgf, line 673:
```suggestion
		return nil, ErrEmptyRefFile
```

pjbgf (changes requested):
@Javier-varez nice stuff! Thanks for looking into this issue. Apart from the nits below this LGTM.

Javier-varez:
> @Javier-varez nice stuff! Thanks for looking into this issue. Apart from the nits below this LGTM.

Thank you for reviewing the PR. I just addressed the comments 👍🏻

pjbgf (approved):
@Javier-varez thanks for working on this. :bow:

PR #1138, opened 2024-07-12
storage: filesystem, Fix object cache not working due to uninitialised objects being put into cache

SatelliteMind:
I noticed the `ObjectStorage.getFromUnpacked` function has an object cache, but it is never actually used to resolve any objects. I found that when a new object is being put into cache its hash (which is used as the cache key) is always zero, which makes future loads of that same object always miss when trying the cache. See the debugger screenshots in the expandable sections.

<details>
  <summary>Hash of the queried object at the check cache point</summary>

image

</details>

<details>
  <summary>Hash of the loaded object when it's being put into cache (it's zero)</summary>

image

The hash is later used as cache key:
```Go
func (c *ObjectLRU) Put(obj plumbing.EncodedObject) {
	...
	objSize := FileSize(obj.Size())
	key := obj.Hash()
	...
```

</details>

<details>
  <summary>Hash of the loaded object when it's being returned from the function (now it matches the queried hash)</summary>

image

</details>

It seems to me this is a bug, so I opened this PR with the fix. However, this is my first contribution to this project, so I would appreciate a thorough review.

### DIFF
--- a/storage/filesystem/object.go
+++ b/storage/filesystem/object.go
@@ -431,12 +431,12 @@ func (s *ObjectStorage) getFromUnpacked(h plumbing.Hash) (obj plumbing.EncodedOb
 
 	defer ioutil.CheckClose(w, &err)
 
-	s.objectCache.Put(obj)
-
 	bufp := copyBufferPool.Get().(*[]byte)
 	buf := *bufp
 	_, err = io.CopyBuffer(w, r, buf)
 	copyBufferPool.Put(bufp)
+
+	s.objectCache.Put(obj)
 
 	return obj, err
 }
--- a/storage/filesystem/object_test.go
+++ b/storage/filesystem/object_test.go
@@ -547,3 +547,64 @@ func BenchmarkGetObjectFromPackfile(b *testing.B) {
 		})
 	}
 }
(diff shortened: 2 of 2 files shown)


pjbgf (approved):
@SatelliteMind thanks for working on this. :bow: 

I can confirm that the cache is now working, which translates on improved performance for `EncodedObject` calls (already cached objects). Here's the benchstat results:
```
                         │ /tmp/before  │             /tmp/after              │
                         │    sec/op    │   sec/op     vs base                │
GetObjectFromUnpacked-16   13.663µ ± 3%   5.501µ ± 3%  -59.74% (p=0.000 n=20)

                         │ /tmp/before │             /tmp/after             │
                         │    B/op     │    B/op     vs base                │
GetObjectFromUnpacked-16   5923.0 ± 0%   751.0 ± 0%  -87.32% (p=0.000 n=20)

                         │ /tmp/before │             /tmp/after             │
                         │  allocs/op  │ allocs/op   vs base                │
GetObjectFromUnpacked-16    42.00 ± 0%   14.00 ± 0%  -66.67% (p=0.000 n=20)
```

PR #1144, opened 2024-07-15
Implement IsAdded, IsModified, IsDeleted method for status

### LINKED ISSUE #1140: status not able to track any comitted file
Please see this function:
```
func run() error {
      // Please see that this is just an example for checking status of a file, hence it has hard-coded strings 
	r, err := git.PlainOpenWithOptions("../bsf-ci", &git.PlainOpenOptions{DetectDotGit: true})
	if err != nil {
		return err
	}
	// Get the working tree
	w, err := r.Worktree()
	if err != nil {
		return err
	}
	status, err:= w.Status()
	if err!=nil{
		return err
	}
	fl:= status.File("go.mod")
	println(fl.Staging) // Always printing 63 that means untracked
	println(status.IsUntracked("go.mod")) // always giving false
	fmt.Printf("Staging: %c\n", fl.Staging)
	return nil
}
```
The problem is that it is always giving 63/? for `go.mod`. But once `.git` or `.github` are removed and repository is re-initiated, it starts tracking the file.  

The error is, go git is returning file as untracked, if it is committed 

The bsf-ci repository is: https://github.com/Manik2708/bsf-ci/
The original bsf repository: https://github.com/buildsafedev/bsf

mowazzem:


mowazzem:
Wrong Issue reference. 🙇

PR #1157, opened 2024-07-24
plumbing: transport/ssh, Add support for SSH @cert-authority.

Javier-varez:
The recently released skeema/knownhosts v1.3.0 introduced a `HostKeyDB` type that extends the `HostKeyCallback` functionality to support @cert-authority algorithms in its key `HostKeyAlgorithms` discovery functionality.

`known_hosts` files may contain lines with @cert-authority markers to represent a certificate instead of a key. If a git remote uses cert authorities as the preferred host identification mechanism, the functionality added in skeema/knownhosts v1.3.0 is needed so that go-git can interact with this remote and verify it.

See https://github.com/skeema/knownhosts/pull/9 for details.

### DIFF
--- a/plumbing/transport/ssh/auth_method.go
+++ b/plumbing/transport/ssh/auth_method.go
@@ -230,11 +230,11 @@ func (a *PublicKeysCallback) ClientConfig() (*ssh.ClientConfig, error) {
 //	~/.ssh/known_hosts
 //	/etc/ssh/ssh_known_hosts
 func NewKnownHostsCallback(files ...string) (ssh.HostKeyCallback, error) {
-	kh, err := newKnownHosts(files...)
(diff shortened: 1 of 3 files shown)


pjbgf:
@Javier-varez thanks for looking into this. Please add some tests to confirm the new behaviour, so that we can avoid future regressions.

Javier-varez:
Sure! I added a couple of tests for the new functionality:
- A test that checks the supported key algorithms when the `known_hosts` file contains a regular RSA key.
- And another test that checks the supported certificate algorithms when the `known_hosts` file contains a certificate authority marker.

pjbgf (approved):
LGTM

@Javier-varez thanks for working on this. :bow:

Issue #1003, opened 2024-01-16
worktree.Add does not add a modified file correctly while git add works as expected


lambou123:
Development environment details:

- Ubuntu 23.10
- git 2.40.1
- go-git v5.11.0

cat ~/.gitconfig
```
[merge]
   ff = false
```

I have a modified file (echo "new line" >> readme.txt).

If I run this code, I got a modified file in worktree and index at the same time with the same content.
If I run git add file, this will be added correctly :
```
r,err := git.PlainOpen(dir)
w, err := r.Worktree()
w.Add(file)
```
  
`git diff` shows no changes:
```
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   readme.md

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   readme.md

```

If I run `git add file`, the output is like that:
```
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   readme.md

```

nicholasSUSE:
I have the same problem, 
I thought it was because my file being modified was huge, but it seems it is not the case.  
I am working on a PR for this. 

@pjbgf

nicholasSUSE:
@lambou123 

I have just made a test, please check what happens with yout. 

```
	if err := workTree.AddWithOptions(&git.AddOptions{Path: "."}); err != nil {
		return err
	}
```

This will work except for files that are huge. 
In a test with a .txt small file, it worked.

pjbgf:
Similar issues (file changes not being added) can be related to what osfs abstraction being used. As some of the file modes are not passed through the file system abstraction when using [osfs.WithChroot()](https://github.com/go-git/go-billy/blob/master/osfs/os_chroot.go#L14-L23), which is the default when doing `Plain*` operations.

However, given that the issue at hand here is about changing the actual file, the proposed PR by @nicholasSUSE seems to be the way forward.

PR #1159, opened 2024-08-01
plumbing: filemode, Remove check for setting size of .git/index file 

nicholasSUSE:
#### Issue:

worktree.Add does not add a modified file correctly, while git add works as expected #1003

#### Problem:

There was a check at `doUpdateFileToIndex` that would consider any file with executable permissions as an executable.
This is wrong and should be fixed. 

Despite this, even if a file is an executable, it should be able to be staged; this is the default behavior of `git CLI`. 

#### Solution:

Add an extra check for the executable file that allows the size of `*index.Entry` to be set.

### DIFF
--- a/worktree_status.go
+++ b/worktree_status.go
@@ -559,9 +559,11 @@ func (w *Worktree) doUpdateFileToIndex(e *index.Entry, filename string, h plumbi
 		return err
 	}
 
-	if e.Mode.IsRegular() {
-		e.Size = uint32(info.Size())
-	}
+	// The entry size must always reflect the current state, otherwise
+	// it will cause go-git's Worktree.Status() to divert from "git status".
+	// The size of a symlink is the length of the path to the target.
+	// The size of Regular and Executable files is the size of the files.
+	e.Size = uint32(info.Size())
 
 	fillSystemInfo(e, info.Sys())
 	return nil
--- a/worktree_status_test.go
+++ b/worktree_status_test.go
@@ -1,0 +1,89 @@
+package git
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/go-git/go-billy/v5/osfs"
+	"github.com/go-git/go-git/v5/plumbing/cache"
+	"github.com/go-git/go-git/v5/storage/filesystem"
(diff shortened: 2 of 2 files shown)


worktree_status.go
```diff
@@ -543,9 +543,9 @@ func (w *Worktree) doUpdateFileToIndex(e *index.Entry, filename string, h plumbi
 		return err
 	}
 
-	if e.Mode.IsRegular() {
-		e.Size = uint32(info.Size())
-	}
+	// The size of a symlink is the length of the path to the target
+	// The size of Regular and Executable files is the size of the files
+	e.Size = uint32(info.Size())
```
pjbgf, line 548:
```suggestion
	// The entry size must always reflect the current state, otherwise
	// it will cause go-git's Worktree.Status() to divert from "git status".
	// The size of a symlink is the length of the path to the target.
	// The size of Regular and Executable files is the size of the files.
	e.Size = uint32(info.Size())
```

pjbgf (changes requested):
Thanks for working on this @nicholasSUSE. I managed to reproduce the issue that this fix is addressing. Please add some tests as per off-line chat.

Also, please re-word the commit with the new PR's title.

nicholasSUSE:
@pjbgf, thank you for the help in solving this issue. 

I created a new test file; I don't know if that is what you expected. 
Please review it.

worktree_status_test.go
```diff
@@ -0,0 +1,89 @@
...
+	err = f.Close()
+	require.NoError(t, err)
+
+	_, err = wt.Add(file)
+	assert.NoError(t, err)
+
+	// go-git's Status diverges from "git status", so this check does not
+	// fail, even when the issue is present. As at this point "git status"
+	// reports the unstaged file was modified while "git diff" would return
+	// empy, as the files as the same but the index has the incorrect file
```
pjbgf, line 74:
```suggestion
	// empty, as the files are the same but the index has the incorrect file
```

pjbgf (commented):
Tiny nit, otherwise LGTM.

pjbgf (approved):
Great stuff @nicholasSUSE! :trophy: Thanks for working on this. :bow:

PR #1169, opened 2024-08-20
plumbing: signature, support the same x509 signature formats as git

yoavamit:
Commits and tags signed with x509 certificates may be encoded with the `SIGNED MESSAGE` PEM label. This behavior is consistent in [git](https://github.com/git/git/blob/master/gpg-interface.c\#L63) and [gpgsm](https://github.com/gpg/gnupg/blob/master/sm/sign.c\#L650) which is commonly used to produce these types of signatures.


yoavamit:
Side-note: even though this feature is not very common, I decided to keep supporting the `-----BEGIN CERTIFICATE-----` format as well to prevent introducing breaking changes.

If it's preferred, I'd be happy to remove the ability to successfully parse the incorrect PEM label to keep `go-git` in parity with `git` behavior.

pjbgf (approved):
LGTM! Thanks for working on this @yoavamit! :bow: 

Would you mind proposing another PR with this commit and another removing `-----BEGIN CERTIFICATE-----`? Please target it to the `v6-exp` branch.

yoavamit:
> LGTM! Thanks for working on this @yoavamit! 🙇
> 
> Would you mind proposing another PR with this commit and another removing `-----BEGIN CERTIFICATE-----`? Please target it to the `v6-exp` branch.

Happy to do it :)

I created another PR [here](https://github.com/go-git/go-git/pull/1171).
Please let me know if there's anything missing there.

PR #1170, opened 2024-08-21
fix: allow discovery of non bare repos in fsLoader

jakobmoellerdev:
This is a simple fix to the fs loader that allows loading of non-bare repositories on the filesystem. Previously only bare repositories could be found due to the config check, now it also checks for the default .git directory.

Also kills the TestLoadIgnoreHost case since it was duplicated with TestLoad.
Doesn't use the git dir constant due to an import conflict

### DIFF
--- a/plumbing/transport/server/loader.go
+++ b/plumbing/transport/server/loader.go
@@ -40,8 +40,16 @@ func (l *fsLoader) Load(ep *transport.Endpoint) (storer.Storer, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


jakobmoellerdev:
@pjbgf would you mind retriggering, forgot the commit suffix. Thanks!

pjbgf (approved):
Thanks for working on this @jakobmoellerdev! :bow: 

Would you mind also forward porting this to `v6-exp` please?

PR #1174, opened 2024-08-28
Fixed a bug that caused panic when passing nil pointer to some methods. Fixes #1051

### LINKED ISSUE #1051: (*Worktree).Commit will panic if `CommitOptions` parameter is nil.
I found that `(*Worktree).Commit` method is taking pointer as parameter, and derenferencing it without checking if it is nil.
If someone passes nil there, It should panic for sure.

I think if we need users to pass non-nil pointer, we should document it.
Or we could just replace it with pointer to empty struct if it is nil.

I thought about changing the parameter to non-pointer struct, but it seems like it is breaking change and we will lose consistency in codebase.

Example code:
```go
package main

import (
	"github.com/go-git/go-billy/v5/util"
	"github.com/go-git/go-git/v5"
)

func main() {
	fs := memfs.New()
	storage := memory.NewStorage()

	r, _ := git.Init(storage, fs)
	w, _ := r.Worktree()

	util.WriteFile(fs, "foo", []byte("foo"), 0755)
	w.Add("foo")

	_, err := w.Commit("this will panic", nil)
	if err != nil {
		panic(err)
	}
}
```

eliotee:
Fixed a bug that caused panic when passing nil pointer to some methods.
 Fixes #1051

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -160,6 +160,10 @@ func (w *Worktree) updateSubmodules(o *SubmoduleUpdateOptions) error {
 
 // Checkout switch branches or restore working tree files.
 func (w *Worktree) Checkout(opts *CheckoutOptions) error {
+	if opts == nil {
+		opts = &CheckoutOptions{}
+	}
+
 	if err := opts.Validate(); err != nil {
 		return err
 	}
@@ -279,6 +283,10 @@ func (w *Worktree) setHEADToBranch(branch plumbing.ReferenceName, commit plumbin
 }
 
 func (w *Worktree) ResetSparsely(opts *ResetOptions, dirs []string) error {
+	if opts == nil {
+		opts = &ResetOptions{}
+	}
+
 	if err := opts.Validate(w.r); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


worktree.go
```diff
@@ -997,6 +1009,10 @@ func (gr GrepResult) String() string {
 
 // Grep performs grep on a repository.
 func (r *Repository) Grep(opts *GrepOptions) ([]GrepResult, error) {
+	if opts == nil {
+		opts = &GrepOptions{}
```
pjbgf, line 1013: We should exit with an error here, as per [this](https://github.com/go-git/go-git/issues/1051#issuecomment-2443482527).

worktree.go
```diff
@@ -160,6 +160,10 @@ func (w *Worktree) updateSubmodules(o *SubmoduleUpdateOptions) error {
 
 // Checkout switch branches or restore working tree files.
 func (w *Worktree) Checkout(opts *CheckoutOptions) error {
+	if opts == nil {
+		opts = &CheckoutOptions{}
```
pjbgf, line 164: Same here, we should error instead.

pjbgf:
@eliotee thanks for proposing this PR. Please note that the commit message needs to align with the [contributing guidelines](https://github.com/go-git/go-git/blob/master/CONTRIBUTING.md#format-of-the-commit-message).

On the funcs that will return an error, please add a test to confirm that behaviour.

PR #1177, opened 2024-09-03
Fix reset files in subfolders

linglo:
Fix a bug where `Worktree.Reset()` and `Worktree.Restore()` with a `Files` option will skip over files that are in a subfolders.

PR #1185, opened 2024-09-07
Fix problem with when using Reset with ResetOptions.Files that are in sub directories

ben-tbotlabs:
resetWorktree was using the name instead of the full path when pulling path out of the change. This resulted with everything working for file that exist in to root, but not anything in sub directories


ben-tbotlabs:
@pjbgf I am not sure why the Test / version-matrix (1.20.x, macos-latest) hasn't started, but I seriously doubt it has anything to do with this change

pjbgf:
Superseded by #1177.

Issue #489, opened 2022-03-17
Authentication Error: Authentication Required


Manjinder096:
Trying to clone a repo using go-git, below is the code that I wrote. I'm using BasicAuth from provided "http" package in go-git. 

```
r, err := git.PlainClone(directory, false, &git.CloneOptions{
		Auth: &http.BasicAuth{
			Username: username,
			Password: password,
		},
		URL:      url,
		Progress: os.Stdout,
	})
```

Also tried using in-memory clone but it also gives same error :

```
r, err := git.Clone(storer, fs, &git.CloneOptions{
		URL:  repository,
		Auth: &http.BasicAuth{
			Username: username,
			Password: password,
		},
	})
```

the error simply says _"authentication required"_ . This is the same method all official examples are also following. Not sure if I am missing anything here ?

nathanblair:
I can't say for certain exactly what is causing this but it seems the username and password are required to be valid, though for public repos I haven't confirmed if they auth itself is actually required.

In my case, trying to do a BasicAuth clone from GitHub, it could have something to do with GitHub's git-upload-pack endpoint? It seems to require correct authentication even for a public repository; though running `git fetch` with a URL using dummy credentials at the CLI works fine.

lukehinds:
did you figure this out @Manjinder096 

I am getting the same issue using a PAT token.

mcepl:
Can I repeat here how much I dislike this stupid bot?

https://drewdevault.com/2021/10/26/stalebot.html

pjbgf:
@mcepl do you mind elaborating on the issue you are experiencing? Is this against a specific Git server?

mcepl:
I am mostly interested in the fixing of the linked git-bug issue.

mcepl:
Just prepared a reproducer for my colleague:

This is my local git repo https://mcepl.fedorapeople.org/tmp/wikipediafs-repo.tar.xz, using https://github.com/MichaelMure/git-bug/releases/download/v0.8.0/git-bug_linux_amd64, and then when I run:

```
tumbleweed-pkg~/r/wikipediafs (master)$ /tmp/git-bug_linux_amd64 push sourcehut
Error: authorization failed
tumbleweed-pkg~/r/wikipediafs (master)$ 
```

[...]

PR #1191, opened 2024-09-25
Consume push URLs when they are provided

mcepl:
I am not sure how far this goes with finally fixing #489, but it certainly shouldn’t make things worse.

We now consider also `pushurl` values in the repository config


pjbgf (approved):

pjbgf:
@mcepl thank you for working on this. :bow:

PR #1194, opened 2024-09-28
*: use gocheck's MkDir instead of TempDir for tests. Fixes #807

### LINKED ISSUE #807: Audit the creation and deletion of temp folders in test suites
We use gocheck for running tests via suites, gocheck allows for setup/teardown for the entire suite and for each test (see documentation here: https://labix.org/gocheck).

We also have a number of suites/tests which setup temporary directories and then delete them afterwards. gocheck however already has this feature built in: `c.MkDir()` which will create a temporary folder that will automatically be removed at the end. For example there is quite a lot of usage of `TemporalDir()` we also have multiple `TemporalFilesystem()`s, can also search for `util.TempDir` and `os.MkdirTemp`.

Audit all cases to use the built in temporary directory which reduces the code as well as ensures we clean up correctly at the end.

uragirii:
This PR removes all the usage of `TempDir` and replaces it with gocheck's `MkDir`. It removes the requirement of cleanup functions. 

Note that I haven't removed them from examples directory as they are not tests. 

Fixes: #807

### DIFF
--- a/common_test.go
+++ b/common_test.go
@@ -136,21 +136,6 @@ func (s *BaseSuite) GetLocalRepositoryURL(f *fixtures.Fixture) string {
 	return f.DotGit().Root()
 }
 
-func (s *BaseSuite) TemporalDir() (path string, clean func()) {
-	fs := osfs.New(os.TempDir())
-	relPath, err := util.TempDir(fs, "", "")
-	if err != nil {
-		panic(err)
-	}
-
-	path = fs.Join(fs.Root(), relPath)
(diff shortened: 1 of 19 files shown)


pjbgf (approved):

pjbgf:
@uragirii LGTM. thanks for working on this. :bow: 

Would you mind forward porting this by proposing the same PR against the `v6-exp` branch please?

apoorv-kansal-mxm:
@pjbgf Sure!

PR #1227, opened 2024-11-14
plumbing: object, limit logs by trailing hash 

d-fal:
I believe it would be a nice filter parameter if we can limit the logs with the target hash along with time filter. 

what I like to see in this lib is something similar to ancestry path in git cli
```bash
git rev-list --ancestry-path 7b4a07a..ecf5891
```

then we should be able to select a path on the git tree by setting from/to hash 

```

                     __D     
                   /          
         ____B ___E___H
       /          \__F
     A             
      \___ C

```
if I wanna select `HEB` we set `from: H, To: B`
wdyt?

### DIFF
--- a/plumbing/object/commit_walker_test.go
+++ b/plumbing/object/commit_walker_test.go
@@ -111,8 +111,8 @@ func (s *CommitWalkerSuite) TestCommitLimitIterByTime(c *C) {
 	for i, commit := range commits {
 		c.Assert(commit.Hash.String(), Equals, expected[i])
 	}
-
 }
+
 func (s *CommitWalkerSuite) TestCommitPreIteratorWithSeenExternal(c *C) {
 	commit := s.commit(c, plumbing.NewHash(s.Fixture.Head))
 
--- a/plumbing/object/commit_walker_test.go
+++ b/plumbing/object/commit_walker_test.go
@@ -1,6 +1,8 @@
 package object
 
 import (
+	"time"
+
 	"github.com/go-git/go-git/v5/plumbing"
 
 	. "gopkg.in/check.v1"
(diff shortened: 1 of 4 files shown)


pjbgf:
@d-fal please add some tests. We are avoiding new features on `master`, please target `v6-exp` instead.

d-fal:
@pjbgf , thanks for considering this as a feature.
tests added,

plumbing/object/commit_walker_test.go
```diff
@@ -60,6 +62,57 @@ func (s *CommitWalkerSuite) TestCommitPreIteratorWithIgnore(c *C) {
...
+	}).ForEach(func(c *Commit) error {
+		commits = append(commits, c)
+		return nil
+	})
+
+	for i, commit := range commits {
+		c.Assert(commit.Hash.String(), Equals, expected[i])
+	}
+
+}
```
pjbgf, line 90:
```suggestion
}

```

plumbing/object/commit_walker_test.go
```diff
@@ -60,6 +62,57 @@ func (s *CommitWalkerSuite) TestCommitPreIteratorWithIgnore(c *C) {
...
```
pjbgf, line 115:
```suggestion
}

```

plumbing/object/commit_walker_test.go
```diff
@@ -60,6 +62,57 @@ func (s *CommitWalkerSuite) TestCommitPreIteratorWithIgnore(c *C) {
 	}
 }
 
+func (s *CommitWalkerSuite) TestCommitLimitIterByTrailingHash(c *C) {
+
```
pjbgf, line 66:
```suggestion
```

pjbgf (approved):
@d-fal thanks for working on this. :bow: Can you please just adjust the formatting issues below before we merge this?

pjbgf (approved):

PR #1230, opened 2024-11-23
plumbing: object, don't traverse past shallow commits in bfs commit walk

Ikke:
When trying to determine the merge base between two commits in a shallow clone, bfsCommitIterator tries to traverse the entire history until it reaches the shallow point and tries to look up the parent commits, which do not exist in the shallow clone.

Handle this by obtaining a list of shallow commits, if available.

This fixes an issue with commit.MergeBase, that git proper handles without problems.


Ikke:
There are currently no tests at all for `bfsCommitIterator`. If tests are expected, please advise how this should be tested.

plumbing/object/commit_walker_bfs.go
```diff
@@ -64,6 +64,21 @@ func (w *bfsCommitIterator) Next() (*Commit, error) {
...
 
+		if storer, ok := c.s.(storer.ShallowStorer); ok {
+			shallowCommits, err := storer.Shallow()
+			if err != nil {
+				return nil, err
+			}
+
+			for _, shallowCommit := range shallowCommits {
+				if c.Hash == shallowCommit {
+					// The parents for a shallow commit are not present in the currrent repo
```
pjbgf, line 75:
```suggestion
					// The parents for a shallow commit are not present in the current repo
```

pjbgf:
@Ikke [Here](https://github.com/go-git/go-git/blob/5f852f9e37dd3b4b496c034c4a939e0c525d3fd0/plumbing/object/commit_walker_test.go#L184-L230) are some of the tests that cover some of the BSF iterator. It may be easier to just expand on them.

plumbing/object/commit_walker_bfs.go
```diff
@@ -64,6 +64,21 @@ func (w *bfsCommitIterator) Next() (*Commit, error) {
 
 		w.seen[c.Hash] = true
 
+		if storer, ok := c.s.(storer.ShallowStorer); ok {
+			shallowCommits, err := storer.Shallow()
```
AriehSchneier, line 68: Is this something we would only want to do once and reuse within the `for` loop?

Ikke: I was looking at that. The only issue is that `c.s` is only available in the loop context, so unless there is another way to obtain the storer, we would need to add logic to check if we already obtained the shallow commits.

Ikke: And it turns out that once it reaches that part of the loop, it will always return from the function, so it will only execute once anyway.

AriehSchneier: Ahh, fair enough, I wonder if we should change the logic fractionally to close the for loop before that point, or at least remove the `var c` outside the loop as its useless as is.

plumbing/object/commit_walker_bfs.go
```diff
@@ -64,6 +64,21 @@ func (w *bfsCommitIterator) Next() (*Commit, error) {
...
+			if err != nil {
+				return nil, err
+			}
+
+			for _, shallowCommit := range shallowCommits {
+				if c.Hash == shallowCommit {
+					// The parents for a shallow commit are not present in the current repo
+					// so we should not traverse down the parents.
+					return c, nil
```
AriehSchneier, line 77: Should this be `continue` rather than `return`?

Ikke:
> @Ikke [Here](https://github.com/go-git/go-git/blob/5f852f9e37dd3b4b496c034c4a939e0c525d3fd0/plumbing/object/commit_walker_test.go#L184-L230) are the tests that cover some of the BSF iterator. It may be easier to just expand on them.

FYI, still trying to figure out how to expand the tests to test this change, but not yet following how the fixture is setup and how to setup an instance with shallow commits.

Issue #53, opened 2020-05-03
Pull removes untracked files locally


emaincourt:
Hi,

Thank you for the great work you've been doing with `go-git`. I've used it to start a new project and it works really well.

As a new user I guess I might be doing something wrong, or miss a concept, but [here](https://github.com/emaincourt/dot/blob/master/pkg/sync/git/pull.go#L20) I have a very basic `Pull` operation on a repository which looks like this:
```
if err := w.Pull(&gogit.PullOptions{}); err != nil {
	if err == gogit.NoErrAlreadyUpToDate {
		return nil
	}
	return err
}
```

The folder into which the repository lives also contains untracked files. If I run this pull operation with `go-git`, all the untracked files are being removed from the directory. On the other hand if run a manual `git pull` in my terminal from the same folder, untracked files get preserved as expected.

Could you please help me ?

Thanks in advance

ggrv0408:
I too have the same issue. I am commenting here hoping this gets bumped up.

pete-woods:
I am also suffering this issue. It seems to me like when this library tries to calculate the changes it needs to make to the filesystem, it is including untracked files in the current state (coming from here: https://github.com/go-git/go-git/blob/master/worktree_status.go#L128 in the code).

pete-woods:
It's still a pretty serious bug

pjbgf:
@pete-woods would you be able to share an example on how to reproduce the issue please?

pete-woods:
If I remember correctly, simply doing a "git reset" in a directory with untracked files is enough. Instead of only reverting tracked files, any untracked files are also deleted. This, of course, is not what's supposed to happen.

PR #1235, opened 2024-11-30
Don't delete untracked files when resetting worktree.

kbirk:
This PR is just a direct copy of https://github.com/BillXiang/go-git/commit/8d52a2a431befbeede79854202fd8337d48518be 

I recently ran into this issue: https://github.com/go-git/go-git/issues/53 where running the following code would remove any untracked files from my repo, even if they were in the `.gitignore`

```
worktree.Reset(&git.ResetOptions{
    Mode: git.HardReset,
})
```

This commit fixed the issue for me.

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -469,8 +469,11 @@ func (w *Worktree) resetWorktree(t *object.Tree, files []string) error {
 			}
 		}
 
-		if err := w.checkoutChange(ch, t, b); err != nil {
-			return err
+		status, err := w.Status()
+		if err == nil && status.File(nameFromAction(&ch)).Worktree != Untracked {
+			if err := w.checkoutChange(ch, t, b); err != nil {
+				return err
+			}
 		}
 	}
 
--- a/worktree.go
+++ b/worktree.go
@@ -446,6 +446,11 @@ func (w *Worktree) resetWorktree(t *object.Tree, files []string) error {
 	}
 	b := newIndexBuilder(idx)
 
+	status, err := w.Status()
+	if err != nil {
+		return err
+	}
+
 	for _, ch := range changes {
 		if err := w.validChange(ch); err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


worktree.go
```diff
@@ -469,8 +469,11 @@ func (w *Worktree) resetWorktree(t *object.Tree, files []string) error {
 			}
 		}
 
-		if err := w.checkoutChange(ch, t, b); err != nil {
-			return err
+		status, err := w.Status()
```
pjbgf, line 472: Calling status can be an [expensive](#181) operation, so we should avoid calling it in a loop. Please move `L472` above the `for` statement.

The error handling would have to be updated accordingly.

kbirk: Done! And it is noticeably faster :)

k-anshul:
Hey @pjbgf 

We are also finding these changes useful. Could you please take another look at this PR?

Thanks

pjbgf:
@k-anshul @kbirk thanks for looking into this. This change is making the existing tests to fail. To make it easier to get a review, I'd recommend trying to get the tests fixed so that it is easier to understand the impact of the change. A new test should be added to avoid future regressions.

k-anshul:
> @k-anshul @kbirk thanks for looking into this. This change is making the existing tests to fail. To make it easier to get a review, I'd recommend trying to get the tests fixed so that it is easier to understand the impact of the change. A new test should be added to avoid future regressions.

Hey @pjbgf 

I do not have permissions to make changes in this PR but I raised a new [PR](https://github.com/rilldata/go-git/pull/1) here.
The current functionality is already being verified in an existing [test](https://github.com/go-git/go-git/blob/93ed29d8f65c1f879d438a5370c212ffedfe5030/worktree_test.go#L1121) but the behaviour being checked is wrong so I modified it to check for the correct behaviour.

Could you please take another look at the new PR ?

pjbgf:
Closing this as #1540 was already merged.

Issue #1236, opened 2024-12-05
checkout failed with 'submodule not found' if submodule has a custom name


yankeguo:
If submodule is added with `--name` option, the submodule name is different with submodule path.

In this case, `go-git` checkout will failed with `submodule not found`, since `checkoutChange` function assumes submodule's name is same with it's path

https://github.com/go-git/go-git/blob/63343bf5f918ea5384ae63bfd22bb36689fa0151/worktree.go#L607

PR #1237, opened 2024-12-05
fix: resolve submodule path to name on checkout, #1236

yankeguo:


worktree.go
```diff
@@ -663,14 +663,20 @@ func (w *Worktree) checkoutChangeSubmodule(name string,
 ) error {
 	switch a {
 	case merkletrie.Modify:
-		sub, err := w.Submodule(name)
+		subs, err := w.Submodules()
```
AriehSchneier, line 666: Should you try `w.Submodule(name)` first before checking the Paths?

worktree.go
```diff
@@ -663,14 +663,20 @@ func (w *Worktree) checkoutChangeSubmodule(name string,
...
 		}
-
-		if !sub.initialized {
-			return nil
+		for _, sub := range subs {
+			if sub.Config().Path == name {
+				if !sub.initialized {
+					return nil
+				}
+				goto found
```
AriehSchneier, line 675: Why `goto` rather than just `return` here?

yankeguo: I'm sorry that the exhaustion from debugging until late at night made me write such ugly code. I've fixed it.

worktree.go
yankeguo: I'm thinking that in this case, the "name" variable should actually be the path. So, in order to avoid the situation where the name of the submodule happens to match the path while the path itself fails to match properly, perhaps the "Submodule(name)" method shouldn't be attempted here.

AriehSchneier: If that is indeed the case (I haven't looked at the code or the calling code) then the parameter name is wrong, you should rename it.

yankeguo: The related methods are all using "name" as the variable name. I'm not sure whether it's appropriate to modify it. It seems that "name" represents the file name (and path), and there is ambiguity when dealing with submodules.

pjbgf: The existing code looks up for the submodule by name: [m.Config().Name == name](https://github.com/go-git/go-git/blob/7bf90dec637d9741c45be67402b59aeac3d3eb70/worktree.go#L850). This current change does not seem backwards compatible, as although in some cases name and path may be the same, I am not sure that is a hard rule.

I'd keep the existing check (i.e. `w.Submodule(name)`), and if the submodule was not found, then try to get it by path (e.g. by calling a new func `w.submoduleByPath(path)`).

PR #1238, opened 2024-12-05
fix: Concurrent map read/write in Tree methods

motoki317:
The "tree path cache" internal field in Tree struct is preventing Tree methods to be called from multiple goroutines concurrently - producing "concurrent map read/write" panics.
This PR disables this internal cache and allows the same Tree object methods to be called concurrently.

IMHO, this type of cache is a bad practice in Go and it is better to utilize multiple goroutines to make processing things faster.
Of course, unless the performance gain by this cache was measured thoroughly and this cache definitely improves performance, but I don't think this is the case.

Thanks!

### DIFF
--- a/plumbing/object/tree.go
+++ b/plumbing/object/tree.go
@@ -39,7 +39,6 @@ type Tree struct {
 
 	s storer.EncodedObjectStorer
 	m map[string]*TreeEntry
-	t map[string]*Tree // tree path cache
 }
 
 // GetTree gets a tree from an object storer and decodes it.
@@ -128,27 +127,9 @@ func (t *Tree) TreeEntryFile(e *TreeEntry) (*File, error) {
 
 // FindEntry search a TreeEntry in this tree or any subtree.
 func (t *Tree) FindEntry(path string) (*TreeEntry, error) {
-	if t.t == nil {
-		t.t = make(map[string]*Tree)
-	}
-
 	pathParts := strings.Split(path, "/")
 	startingTree := t
 	pathCurrent := ""
-
-	// search for the longest path in the tree path cache
-	for i := len(pathParts) - 1; i > 1; i-- {
-		path := filepath.Join(pathParts[:i]...)
-
-		tree, ok := t.t[path]
-		if ok {
-			startingTree = tree
-			pathParts = pathParts[i:]
-			pathCurrent = path
-
-			break
-		}
-	}
 
 	var tree *Tree
 	var err error
@@ -158,7 +139,6 @@ func (t *Tree) FindEntry(path string) (*TreeEntry, error) {
 		}
 
 		pathCurrent = filepath.Join(pathCurrent, pathParts[0])
-		t.t[pathCurrent] = tree
 	}
 
 	return tree.entry(pathParts[0])
@@ -182,10 +162,6 @@ func (t *Tree) dir(baseName string) (*Tree, error) {
 }
 
 func (t *Tree) entry(baseName string) (*TreeEntry, error) {
-	if t.m == nil {
-		t.buildMap()
-	}
-
 	entry, ok := t.m[baseName]
 	if !ok {
 		return nil, ErrEntryNotFound
@@ -268,6 +244,8 @@ func (t *Tree) Decode(o plumbing.EncodedObject) (err error) {
 			Name: baseName,
 		})
 	}
+
+	t.buildMap()
 
 	return nil
 }
--- a/plumbing/object/tree_test.go
+++ b/plumbing/object/tree_test.go
@@ -264,6 +264,7 @@ func (s *TreeSuite) TestTreeDecodeEncodeIdempotent(c *C) {
 		err = newTree.Decode(obj)
 		c.Assert(err, IsNil)
 		tree.Hash = obj.Hash()
+		tree.buildMap()
 		c.Assert(newTree, DeepEquals, tree)
 	}
 }


motoki317:
@pjbgf Hello! Excuse me for tagging you, but is there any chance this change could be reviewed?

pjbgf:
@motoki317 thank you for your PR. Changes like this are better off targeting the `v6-exp` branch which will be our next major release.

Would you be able to provide a benchmark test showing the performance improvements?

motoki317:
@pjbgf Thanks for taking a look at this.

I don't think this change is a breaking one, but should I still target the v6 branch? I think it'd be better to import this into v5 (and maybe to v6 as well) if v6 is going to take a bit longer.
(In fact, I'm using my forked go-git by using `replace` in go.mod in my project. See PR description)

For benchmark, I could try adding one, but I'm not sure how to go about benchmarking this.
Ultimately, the performance gain (or loss) by this change is really dependent on the structure of real projects; a "normal" project with not many (around tens) files per directory and not that deep structure would, I think, benefit from this change. I'm not sure if any project/usecase would suffer from this change, but I don't think there's many of such projects and the performance impact should be unnoticeable.

motoki317:
At least in my usecase, this change resulted in at least a few times faster processing time (a few hundreds ms to a few seconds gain), in most of real projects (I'm mining a lot of OSS repositories), using parallel calls to these methods.

pjbgf:
@motoki317 the `v5` release is frozen, changes to it are now mostly bug fixes or bumps for dependencies. Changes such as this will need to go against `main`, for the `v6` release.

> At least in my usecase, this change resulted in at least a few times faster processing time (a few hundreds ms to a few seconds gain), in most of real projects (I'm mining a lot of OSS repositories), using parallel calls to these methods.

It is common for performance improvements to have side-effects, sometimes the slower processing times causes higher allocations, or memory usage, which leads to slower long-running applications due to GC pressure. 

For example [this](https://github.com/go-git/go-git/pull/799/commits/1c361adbc1f4b0e3d0743d11f187fb0b3ac4cb4d) commit made  clones 15% slower, however it reduced the memory cost by 77%. By running [Go benchmarks](https://pkg.go.dev/testing#hdr-Benchmarks) and pprof we can understand what are the trade-offs before making such decisions.

motoki317:
@pjbgf Thank you for the response.

I see, I have updated the base branch to `main` if you say so.

As for the benchmark, I added a (rough-cut) benchmark code above, but I seem to be getting `file: already closed` or `zlib: invalid header` errors like this, if I try to call FindEntry() concurrently:
```
    repository_test.go:3622: error reading utils/merkletrie/change_test.go: read /tmp/BenchmarkTree_FindEntry_Parallel4239159960/001/objects/pack/pack-af8e45f7de79ee1ca5d7f555bf5868880425ba96.pack: file already closed
    repository_test.go:3622: error reading utils/merkletrie/difftree.go: zlib: invalid header
    repository_test.go:3622: error reading plumbing/format/packfile/types.go: zlib: invalid header
    repository_test.go:3622: error reading plumbing/format/commitgraph/v2/commitgraph.go: zlib: invalid header
    repository_test.go:3622: error reading plumbing/object.go: zlib: invalid header
    repository_test.go:3622: error reading utils/merkletrie/filesystem/node.go: zlib: invalid header
    repository_test.go:3622: error reading plumbing/protocol/packp/capability/capability.go: zlib: invalid header
    repository_test.go:3622: error reading plumbing/protocol/packp/ulreq_encode.go: zlib: invalid header
    repository_test.go:3622: error reading plumbing/protocol/packp/ulreq_test.go: zlib: invalid header
    repository_test.go:3622: error reading plumbing/format/config/common_test.go: zlib: invalid header
```
This doesn't occur in my tool's usecase, so I may be missing something here, or it may be just that go-git's implementation is not really thread-safe at many places other than this.

Nevertheless, when I ignore these errors, I obtained the following results. I run benchmarks on both main and in my branch, and both on `github.com/go-git/go-git` and `github.com/torvalds/linux` (as an example of a larger repository).
```
v6 main
$ go test -run=^$ -test.bench ^BenchmarkTree_FindEntry$ -benchmem
goos: linux
goarch: amd64
cpu: 12th Gen Intel(R) Core(TM) i9-12900K
BenchmarkTree_FindEntry-24            99          11549066 ns/op        27529918 B/op      46009 allocs/op
PASS
ok      github.com/go-git/go-git/v5     2.403s

v6 fix/race-in-tree
$ go test -run=^$ -test.bench ^BenchmarkTree_FindEntry -benchmem
goos: linux
goarch: amd64
cpu: 12th Gen Intel(R) Core(TM) i9-12900K
BenchmarkTree_FindEntry-24                    20          56532524 ns/op        146480136 B/op    264872 allocs/op
BenchmarkTree_FindEntry_Parallel-24           50          20483664 ns/op        128628483 B/op    192093 allocs/op
PASS
ok      github.com/go-git/go-git/v5     5.109s

v6 main (github.com/torvalds/linux)
$ go test -run=^$ -test.bench ^BenchmarkTree_FindEntry$ -benchmem
goos: linux
goarch: amd64
cpu: 12th Gen Intel(R) Core(TM) i9-12900K
BenchmarkTree_FindEntry-24             2         551947129 ns/op        1216659400 B/op  2608151 allocs/op
PASS
ok      github.com/go-git/go-git/v5     3.247s

v6 fix/race-in-tree (github.com/torvalds/linux)
$ go test -run=^$ -test.bench ^BenchmarkTree_FindEntry -benchmem
goos: linux
goarch: amd64
cpu: 12th Gen Intel(R) Core(TM) i9-12900K
BenchmarkTree_FindEntry-24                     1        13438525538 ns/op       36206744264 B/op        129991164 allocs/op
BenchmarkTree_FindEntry_Parallel-24            1        2665111342 ns/op        33130471120 B/op        111777776 allocs/op
PASS
ok      github.com/go-git/go-git/v5     17.653s
```
which is telling me that this change is making the code slower, at least for this benchmark code.

However, I have checked twice now that this change indeed improves performance in my usecase, by checking performance before and after my tool's commit [here](https://github.com/salab/iccheck/commit/7afd704505b799af7aea810daa2103c25b4619de).
In a specific configuration, my tool's runtime improved from 16s to 14s.
I guess the performance improvement in my usecase comes from the fact that I need to "preload" all files sequentially due to the fact that go-git's methods are not thread-safe, so the "preloading" affects the tool's runtime. On the other hand, if go-git's methods become thread-safe (after applying this PR's patch), it no longer needs to be "preloaded" and I can call it from any goroutine anytime, so parallelism improves and the overall runtime improves.

So, actually I am not sure what to do now. I'll probably keep using this patched go-git in my tool, and I may improve the "preload" implementation so it doesn't become a bottleneck in my tool and report it back here (when I have bandwidth to do so).
Nevertheless, I think it's making it hard for go-git to be used in Golang codes if its methods are not thread-safe, like in my example, so IMO this patch is still worth considering. There's probably many other non-thread-safe code to be fixed if you're going to make methods thread-safe though.

Issue #234, opened 2021-01-05
Need an example of creating a git server using go-git


zh-jn:
Hi all, is there have an example of creating a git server using this repository? thanks!

MatisseB:
Hello, 
also have a hard time to use this lib to create a git server. 
Trying to find the way with this [sample](https://play.golang.org/p/j2AKA1dJb4A) using old repo.

If someone have any snippets, even of a simple http upload/receive-pack server, it would be much appreciated.

Thanks!

lkingland:
This might help: https://github.com/movsb/sgits

saul-data:
I would also like to see an example of http server implementation.

seankhliao:
For future reference, an demo server for just git-upload-pack with go-git/v5: https://github.com/seankhliao/gitreposerver

prologic:
I'm trying to do this over ssh and run into this error: `create receive-pack response: reference delta not found`

code sample:

```#!go
func handleReceivePack(dir string, ch ssh.Channel) error {
	ctx := context.Background()

	ep, err := transport.NewEndpoint("/")
	if err != nil {
		return fmt.Errorf("create transport endpoint: %w", err)
	}
	bfs := osfs.New(dir)
	ld := server.NewFilesystemLoader(bfs)
	svr := server.NewServer(ld)
	sess, err := svr.NewReceivePackSession(ep, nil)
	if err != nil {
		return fmt.Errorf("create receive-pack session: %w", err)
	}

	ar, err := sess.AdvertisedReferencesContext(ctx)
	if err != nil {
		return fmt.Errorf("get advertised references: %w", err)
	}
	err = ar.Encode(ch)
	if err != nil {
		return fmt.Errorf("encode advertised references: %w", err)
	}

	rur := packp.NewReferenceUpdateRequest()
	err = rur.Decode(ch)
	if err != nil {
		return fmt.Errorf("decode reference-update request: %w", err)
	}

	res, err := sess.ReceivePack(ctx, rur)
	if err != nil {
		return fmt.Errorf("create receive-pack response: %w", err)
	}
	err = res.Encode(ch)
	if err != nil {
		return fmt.Errorf("encode receive-pack response: %w", err)
	}

	return nil
}
```

This is based on [gitreposerver](https://github.com/seankhliao/gitreposerver) by @seankhliao -- Pretty sure I'm doing something wrong here 🤔

andskur:
[...]

PR #1249, opened 2024-12-16
fix: support empty pushes

### LINKED ISSUE #1014: Cannot handle empty ReferenceUpdateRequest
Hello,
I've been trying to implement a Git server using this library. I have encountered an issue when pushing to it. When pushing first time with some commits present in local repo that are not in remote repo, everything is fine. When I try to push again immediately afterwards, push fails. The same action works fine for GitHub, GitLab and other git services.

I have used example from [here](https://github.com/go-git/go-git/issues/234#issuecomment-1364724467).

I have done some debugging and pinned it down to `updReqDecoder`'s [`Decode`](https://github.com/go-git/go-git/blob/03a57f8f5179a990083e8b9c2c9e40854ae402ea/plumbing/protocol/packp/updreq_decode.go#L96) method. From my investigation, it turns out that `git` client responds with just flush-pkt line when there are no refs to be updated - in other words, nothing to push by client. However, `ReferenceUpdateRequest` and its decoder, is coded in such a way that it always expects some refs being sent. That causes it to fail with `capabilities delimiter not found` - it comes from [here](https://github.com/go-git/go-git/blob/03a57f8f5179a990083e8b9c2c9e40854ae402ea/plumbing/protocol/packp/updreq_decode.go#L173).

I've looked at `ReferenceUpdateRequest` and it is not straightforward for me how to fix it - maybe someone else have better idea? Nonetheless, I can craft a PR if you are willing to accept it.

Thanks for all the hard work on `go-git`!

iainjreid:
This PR aims to fix #1014 by returning a new `ErrFlushPacketRecieved` error when an empty `recieve-pack` request is sent by an up-to-date git client. This error can then be gracefully handled in upstream code as required.

I've also included a fix to the reader closing errors reported in #234, but this might be over eager, because the original code appears to have been deliberately structured to only use the `io.NopCloser` utility in a specific case?

I can split up the commit into the respective fixes with the ticket numbers if required. Looking forward to any and all feedback.

### DIFF
--- a/plumbing/protocol/packp/updreq_decode.go
+++ b/plumbing/protocol/packp/updreq_decode.go
@@ -19,6 +19,7 @@ var (
 
 var (
 	ErrEmpty                        = errors.New("empty update-request message")
+	ErrFlushPacketRecieved          = errors.New("flush packet recieved")
 	errNoCommands                   = errors.New("unexpected EOF before any command")
 	errMissingCapabilitiesDelimiter = errors.New("capabilities delimiter not found")
 )
@@ -76,14 +77,7 @@ func errMalformedCommand(err error) error {
 
 // Decode reads the next update-request message form the reader and wr
 func (req *ReferenceUpdateRequest) Decode(r io.Reader) error {
-	var rc io.ReadCloser
-	var ok bool
-	rc, ok = r.(io.ReadCloser)
-	if !ok {
-		rc = io.NopCloser(r)
-	}
-
-	d := &updReqDecoder{r: rc, s: pktline.NewScanner(r)}
+	d := &updReqDecoder{r: io.NopCloser(r), s: pktline.NewScanner(r)}
 	return d.Decode(req)
 }
 
@@ -168,6 +162,9 @@ func (d *updReqDecoder) decodeCommands() error {
 
 func (d *updReqDecoder) decodeCommandAndCapabilities() error {
 	b := d.s.Bytes()
+	if bytes.Equal(b, pktline.Flush) {
+		return ErrFlushPacketRecieved
+	}
 	i := bytes.IndexByte(b, 0)
 	if i == -1 {
 		return errMissingCapabilitiesDelimiter


plumbing/protocol/packp/updreq_decode.go
```diff
@@ -168,6 +162,9 @@ func (d *updReqDecoder) decodeCommands() error {
 
 func (d *updReqDecoder) decodeCommandAndCapabilities() error {
 	b := d.s.Bytes()
+	if bytes.Equal(b, pktline.Flush) {
+		return ErrFlushPacketRecieved
+	}
```
pjbgf, line 167: This would also represent an empty `update-request` message, and due to that I am thinking we are better off reusing the existing `ErrEmpty`.

plumbing/protocol/packp/updreq_decode.go
```diff
@@ -76,14 +77,7 @@ func errMalformedCommand(err error) error {
...
 func (req *ReferenceUpdateRequest) Decode(r io.Reader) error {
-	var rc io.ReadCloser
-	var ok bool
-	rc, ok = r.(io.ReadCloser)
-	if !ok {
-		rc = io.NopCloser(r)
-	}
-
-	d := &updReqDecoder{r: rc, s: pktline.NewScanner(r)}
+	d := &updReqDecoder{r: io.NopCloser(r), s: pktline.NewScanner(r)}
```
pjbgf, line 80: This is trickier, as in the existing behaviour the callee is responsible for closing the reader `r`, with this change this swifts to the caller. For breaking changes such as this, the commit/PR would have to target `v6-exp` instead.

pjbgf (changes requested):
Added some comments. For the empty push change can you please add some tests to confirm the new behaviour?

plumbing/protocol/packp/updreq_decode.go
scabala: I think this can be revert, right? So the fix for up-to-date git client could be fixed.

scabala:
As for test: simple operation of `git pull && git pull` is able to trigger this bug. Dunno how it could be implemented in tests but at least it's a starting point

PR #1296, opened 2024-12-28
plumbing: format/pktline, gocheck to testify migration. Fixes #1295

### LINKED ISSUE #1295: plumbing/format/pktline: gocheck to testify migration

BeChris:


BeChris:
Hello,
Strangely, TestDecodeOversizePktLines() test fails (it succeed before migrating to testify).
I expected it to check for an error returned by sc.Scan() but it checks there are no error.
The name of the test looks like there might be an error to check.
Strange that it worked with gocheck.
Was this test executed ?

pjbgf (changes requested):
Found these two issues, but will need to take a better look at the scanner.

aymanbagabas:
> Hello, Strangely, TestDecodeOversizePktLines() test fails (it succeed before migrating to testify). I expected it to check for an error returned by sc.Scan() but it checks there are no error. The name of the test looks like there might be an error to check. Strange that it worked with gocheck. Was this test executed ?

Indeed, weirdly, the tests passed before this migration. Another reason to move away from gocheck.

While looking at the scanner, I found a bug in the `sc.Bytes()` method. Fixed in https://github.com/go-git/go-git/pull/1301. After rebasing this onto the fix, we need this patch for the tests to work.

```diff
diff --git a/plumbing/format/pktline/scanner_test.go b/plumbing/format/pktline/scanner_test.go
index b01bd2855630..598edfddc8b3 100644
--- a/plumbing/format/pktline/scanner_test.go
+++ b/plumbing/format/pktline/scanner_test.go
@@ -21,7 +21,7 @@ func TestSuiteScanner(t *testing.T) {
 
 func (s *SuiteScanner) TestInvalid() {
 	for _, test := range [...]string{
-		"0001", "0002", "0003", "0004",
+		"0003",
 		"0001asdfsadf", "0004foo",
 		"fff5", "ffff",
 		"gorka",
@@ -31,7 +31,8 @@ func (s *SuiteScanner) TestInvalid() {
 	} {
 		r := strings.NewReader(test)
 		sc := pktline.NewScanner(r)
-		_ = sc.Scan()
+		for sc.Scan() {
+		}
 		s.ErrorContains(sc.Err(), pktline.ErrInvalidPktLen.Error(),
 			fmt.Sprintf("data = %q", test))
 	}
@@ -47,7 +48,7 @@ func (s *SuiteScanner) TestDecodeOversizePktLines() {
 		r := strings.NewReader(test)
 		sc := pktline.NewScanner(r)
 		_ = sc.Scan()
-		s.NoError(sc.Err())
+		s.ErrorIs(sc.Err(), pktline.ErrInvalidPktLen)
 	}
 }
 
@@ -61,11 +62,11 @@ func (s *SuiteScanner) TestValidPktSizes() {
 		r := strings.NewReader(test)
 		sc := pktline.NewScanner(r)
 		hasPayload := sc.Scan()
-		obtained := sc.Bytes()
+		obtained := fmt.Sprintf("%04x%s", sc.Len(), sc.Bytes())
 
 		s.True(hasPayload)
 		s.NoError(sc.Err())
-		s.Equal([]byte(test), obtained)
+		s.Equal(strings.ToLower(test), obtained)
 	}
 }
 
@@ -205,7 +206,7 @@ func (s *SuiteScanner) TestReadSomeSections() {
 	sectionCounter := 0
 	lineCounter := 0
 	for sc.Scan() {
-		if len(sc.Bytes()) == 0 {
+		if sc.Len() == pktline.Flush {
 			sectionCounter++
 		}
 		lineCounter++
```

- TestInvalid: All the removed cases are now valid cases. `0001` is a delim packet with no payload. The same goes for `0002` and `0004` with the latter being an empty string packet.
- TestDecodeOversizePktLines: There was a bug in the v5 implementation of `pktline.Scanner` where it accepts pktlines over the expected limit, i.e. `65520` or `0xFFF0` see https://github.com/go-git/go-git/blob/master/plumbing/format/pktline/scanner.go#L112 which should error.
- TestValidPktSizes: This is a format error. The test cases expect a full pktline (length + payload) while `sc.Bytes()` only return the payload.
- TestReadSomeSections: Nothing wrong with this one, except that we want to make it clear we're expecting a flush packet. The `sc.Bytes()` function will return a `nil` value when encountering a special packet. In Go, `nil`s have a length of zero.

pjbgf:
@aymanbagabas thank you very much for taking a look on this. I have now merged #1301.

@BeChris please rebase this PR.

BeChris:
@pjbgf rebase done.

pjbgf (approved):

PR #1111, opened 2024-06-08
Refactor transport and drop git binary dependency

aymanbagabas:
This PR refactors the transport sub-package and implements server-side git operations like `UploadPack` and `ReceivePack` along with delegating client-side operations to `SendPack` and `FetchPack` following the canonical git implementation.

- [x] Use shared transport logic across all transports `file`, `git`, `ssh` and `http` using `NeogotiatePack`, `FetchPack`, and `SendPack`.
- [x] Support HTTP(s) dumb transport (in a follow-up https://github.com/aymanbagabas/go-git/tree/transport-http-dumb)
- [x] Supports git protocol V1
- [x] Support server-side `no-progress`, `sideband`, and `sideband64k` capabilities.
- [ ] Add shallow pack support to `packfile.Encoder` (needed to get the `file` transport to work with the pure server implementation)
- [ ] Fix tests (currently only fails because the above is missing :/)

Future follow-ups:
- Git daemon implementation (`git-daemon` support)
- HTTP middleware (`git-http-backend` support)
- go-git standalone cli

**Needs:** https://github.com/go-git/go-git/pull/1082

New transport api: https://gist.github.com/aymanbagabas/1b5ccc265ee62dd574c62d378b4b57bb

Supersedes: https://github.com/go-git/go-git/pull/757
Supersedes: https://github.com/go-git/go-git/pull/938
Fixes: https://github.com/go-git/go-git/issues/883
Fixes: https://github.com/go-git/go-git/issues/838

PR #1299, opened 2024-12-29
utils: add ioutil.ContextReader and ioutil.ContextWriter and drop ctx…

aymanbagabas:
…io package

This adds `ioutil.ContextReader` and `ioutil.ContextWriter` to the `ioutil` package and drops the `ctxio` package. It also introduces a `CloserFunc` type to implement the `io.Closer` interface with a function.

This extracts changes from #1111

### DIFF
--- a/utils/ioutil/context.go
+++ b/utils/ioutil/context.go
@@ -1,0 +1,117 @@
+package ioutil
+
+import (
+	"context"
+	"io"
+	"slices"
+)
+
+type ioret struct {
+	err error
+	n   int
+}
+
+type Writer interface {
+	io.Writer
+}
+
+type ctxWriter struct {
+	w   io.Writer
+	ctx context.Context
+}
+
+// NewContextWriter wraps a writer to make it respect the given Context.
+// If there is a blocking write, the returned Writer will return
+// whenever the context is cancelled (the return values are n=0
+// and err=ctx.Err().)
+//
+// Note that this wrapper DOES NOT ACTUALLY cancel the underlying
+// write, as there is no way to do that with the standard Go io
+// interface. So the read and write _will_ happen or hang. Use
+// this sparingly, make sure to cancel the read or write as necessary
+// (e.g. closing a connection whose context is up, etc.)
+//
+// Furthermore, in order to protect your memory from being read
+// _after_ you've cancelled the context, this io.Writer will
+// first make a **copy** of the buffer.
+func NewContextWriter(ctx context.Context, w io.Writer) *ctxWriter {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	return &ctxWriter{ctx: ctx, w: w}
+}
+
+func (w *ctxWriter) Write(buf []byte) (int, error) {
+	buf2 := slices.Clone(buf)
+
+	c := make(chan ioret, 1)
+
+	go func() {
+		n, err := w.w.Write(buf2)
+		c <- ioret{err, n}
+		close(c)
+	}()
+
+	select {
+	case r := <-c:
+		return r.n, r.err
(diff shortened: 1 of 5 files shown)


utils/ioutil/context.go
```diff
@@ -0,0 +1,117 @@
...
+//
+// Note well: this wrapper DOES NOT ACTUALLY cancel the underlying
+// write-- there is no way to do that with the standard go io
+// interface. So the read and write _will_ happen or hang. So, use
+// this sparingly, make sure to cancel the read or write as necesary
+// (e.g. closing a connection whose context is up, etc.)
+//
+// Furthermore, in order to protect your memory from being read
+// _after_ you've cancelled the context, this io.Writer will
+// first make a **copy** of the buffer.
```
pjbgf, line 35:
```suggestion
// NewContextWriter wraps a writer to make it respect the given Context.
// If there is a blocking write, the returned Writer will return
// whenever the context is cancelled (the return values are n=0
// and err=ctx.Err().)
//
// Note that this wrapper DOES NOT ACTUALLY cancel the underlying
// write, as there is no way to do that with the standard Go io
// interface. So the read and write _will_ happen or hang. Use
// this sparingly, make sure to cancel the read or write as necessary
// (e.g. closing a connection whose context is up, etc.)
//
// Furthermore, in order to protect your memory from being read
// _after_ you've cancelled the context, this io.Writer will
// first make a **copy** of the buffer.
```

utils/ioutil/context.go
```diff
@@ -0,0 +1,117 @@
...
+// Note well: this wrapper DOES NOT ACTUALLY cancel the underlying
+// write-- there is no way to do that with the standard go io
+// interface. So the read and write _will_ happen or hang. So, use
+// this sparingly, make sure to cancel the read or write as necesary
+// (e.g. closing a connection whose context is up, etc.)
+//
+// Furthermore, in order to protect your memory from being read
+// _before_ you've cancelled the context, this io.Reader will
+// allocate a buffer of the same size, and **copy** into the client's
+// if the read succeeds in time.
```
pjbgf, line 87:
```suggestion
// NewContextReader wraps a reader to make it respect the given Context.
// If there is a blocking read, the returned Reader will return
// whenever the context is cancelled (the return values are n=0
// and err=ctx.Err().)
//
// Note that this wrapper DOES NOT ACTUALLY cancel the underlying
// read, as there is no way to do that with the standard Go io
// interface. The read and write _will_ happen or hang. Use
// this sparingly, make sure to cancel the read or write as necessary
// (e.g. closing a connection whose context is up, etc.)
//
// Furthermore, in order to protect your memory from being read
// _before_ you've cancelled the context, this io.Reader will
// allocate a buffer of the same size, and **copy** into the client's
// if the read succeeds in time.
```

utils/ioutil/context.go
```diff
@@ -0,0 +1,117 @@
...
+func NewContextWriter(ctx context.Context, w io.Writer) *ctxWriter {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	return &ctxWriter{ctx: ctx, w: w}
+}
+
+func (w *ctxWriter) Write(buf []byte) (int, error) {
+	buf2 := make([]byte, len(buf))
+	copy(buf2, buf)
```
pjbgf, line 45: Can't we use [Clone](https://pkg.go.dev/golang.org/x/exp/slices#Clone) instead?
```suggestion
	buf2 := slices.Clone(buf)
```

pjbgf (changes requested):

pjbgf (approved):
LGTM

pjbgf:
Let's keep an eye to see whether this resolves #1323.

Issue #910, opened 2023-11-07
API and Behaviour Changes for next major release


pjbgf:
Potential API changes for the next major release. This changes will be merged into the development branch [main](https://github.com/go-git/go-git/tree/main), before they are released on the upcoming `v6.0.0` release.

#### Changes:
- [ ] SHA256 Support: The support for SHA256 and SHA1 in parallel may require API changes on `plumbing.Hash` or where it is currently used.
  - xref: https://github.com/go-git/go-git/issues/706 https://github.com/go-git/go-git/issues/899
- [ ] Git config management (TBC)
  - xref: https://github.com/go-git/go-git/issues/395
- [ ] `worktree_status.go`
    - https://github.com/go-git/go-git/blob/cf196ea5ba6c8b16712a9868cc2d4cca4f287ae4/worktree_status.go#L273
    - https://github.com/go-git/go-git/blob/cf196ea5ba6c8b16712a9868cc2d4cca4f287ae4/worktree_status.go#L372
- [ ] `difftree.go`
    - https://github.com/go-git/go-git/blob/cf196ea5ba6c8b16712a9868cc2d4cca4f287ae4/plumbing/object/difftree.go#L63C1-L63C1
- [ ] `options.go`
    - https://github.com/go-git/go-git/blob/cf196ea5ba6c8b16712a9868cc2d4cca4f287ae4/options.go#L711
- [ ] `repository.go`
    - https://github.com/go-git/go-git/blob/cf196ea5ba6c8b16712a9868cc2d4cca4f287ae4/repository.go#L561
    - https://github.com/go-git/go-git/blob/0a1c5ab118998a22ef88d8e6c42970197d77413c/repository.go#L426
    - https://github.com/go-git/go-git/blob/0a1c5ab118998a22ef88d8e6c42970197d77413c/repository.go#L439
- [ ] `server.go`
    - https://github.com/go-git/go-git/blob/cf196ea5ba6c8b16712a9868cc2d4cca4f287ae4/plumbing/transport/server/server.go#L92
- [x] `commitgraph`
  - https://github.com/go-git/go-git/pull/854#discussion_r1342188779
  - https://github.com/go-git/go-git/blob/cf196ea5ba6c8b16712a9868cc2d4cca4f287ae4/plumbing/format/commitgraph/memory.go#L10
[...]

PR #1300, opened 2024-12-29
plumbing: move TagMode type to plumbing

aymanbagabas:
This type will be later used in the transport package so it makes sense for it to be shared among modules and defined under plumbing like other
  types such as ReferenceName, etc.

This extracts changes from https://github.com/go-git/go-git/pull/1111


options.go
```diff
@@ -174,21 +172,6 @@ func (o *PullOptions) Validate() error {
...
-	// TagFollowing any tag that points into the histories being fetched is also
-	// fetched. TagFollowing requires a server with `include-tag` capability
-	// in order to fetch the annotated tags objects.
-	TagFollowing
-	// AllTags fetch all tags from the remote (i.e., fetch remote tags
-	// refs/tags/* into local tags with the same name)
-	AllTags
-	// NoTags fetch no tags from the remote at all
-	NoTags
-)
```
pjbgf, line 190: As we are still early days for the V6 release, it may be beneficial to keep a type redirect so that users consuming `v6-exp` (soon to become `main`), don't have to change their code.

```golang
// TODO: delete for V6
type TagMode = plumbing.TagMode

const (
	InvalidTagMode = plumbing.InvalidTagMode
	TagFollowing   = plumbing.TagFollowing
	AllTags        = plumbing.AllTags
	NoTags         = plumbing.NoTags
)
```

aymanbagabas: Fair enough. Updated the patch

pjbgf (approved):
LGTM

pjbgf:
Tagging https://github.com/go-git/go-git/issues/910 to track the removal of the type redirection.

PR #1301, opened 2024-12-29
plumbing: pktline, fix scanner EOF and special packet handling

aymanbagabas:
This fixes the scanner to handle EOF correctly, return nil for special packets such as [Flush], [Delim], and [ResponseEnd], and to return the payload of the line instead of the whole line.

Fixes: https://github.com/go-git/go-git/pull/1082


pjbgf (approved):
LGTM

Issue #492, opened 2022-03-17
Cloning cgit repositories over HTTP gives `error: unexpected EOF`


agorgl:
Trying to clone a repository in a cgit host over HTTP gives:
```
error: unexpected EOF
```
Host is configured with `enable-http-clone=1` in `cgitrc` to allow cloning over HTTP.
Problem is reproducible when trying to clone any git repo in a cgit instance using the basic `clone` example in the `go-git` repo here.

From a rough debugging session error seems to originate from here:
https://github.com/go-git/go-git/blob/c785af3f4559ebac52c42f12d17cb118aac383ad/plumbing/transport/http/common.go#L67

agorgl:
Seems that this is caused when cloning repositories over `dumb` http transport instead of `smart` http transport:
https://git-scm.com/book/en/v2/Git-Internals-Transfer-Protocols

perj:
Was looking for this for a couple of hours now, but this seems to be the issue- Go-git simply assumes that the smart protocol is supported and will generate an unexpected EOF if it gets the dumb response.
Did you find easy to use workaround when just wanting to serve a git repo (dumb protocol served by `git update-server-info` + `python -m http.server`)

perj:
I ended up following the lighttpd guide as shown by `git http-backend --help`

pjbgf:
> Go-git simply assumes that the smart protocol is supported

`go-git` does not support the dumb protocol, as per its [compatibility doc](https://github.com/go-git/go-git/blob/master/COMPATIBILITY.md#transport-schemes).

perj:
> `go-git` does not support the dumb protocol, as per its [compatibility doc](https://github.com/go-git/go-git/blob/master/COMPATIBILITY.md#transport-schemes).

Which is fine of course, but it could be worth giving it a dedicated error message, I think, instead of just unexpected EOF.

pjbgf:
Closing as this is now implemented on `v6` - currently only available at `main`.

PR #1337, opened 2024-12-31
refactor transport: plumbing: transport, refactor and define transport operations (2/5)

aymanbagabas:
This change redefines the transport package to be more modular and
extensible. It redefines the `transport.Transport`, `transport.Session`,
and `transport.Commander` and `transport.Command` interfaces. It also
introduces a new `transport.Connection` that abstracts the logic of
sending and receiving git data to and from a server.

There are 2 main transports. The pack transport which is a full-duplex
command based transport that uses `transport.PackSession`. And the other
is the HTTP transport which is a half-duplex transport that uses
`transport.HTTPSession`. All of file, git, and ssh transports use the
pack transport.

The rest of the abstractions are defined in `transport.NegotiatePack`,
`transport.FetchPack`, `transport.UploadPack`, `transport.SendPack`, and
`tansport.ReceivePack`. All of these static functions are used by the
`transport.Connection` to send and receive data to and from a
`storage.Storer` interface type.

Reference: https://gist.github.com/aymanbagabas/1b5ccc265ee62dd574c62d378b4b57bb

Based on: https://github.com/go-git/go-git/pull/1303

PR #1303, opened 2024-12-29
refactor transport: plumbing: packp, refactor types (1/5)

aymanbagabas:
This change makes packp types follow the git protocol and simplifies the implementation of `Encode` and `Decode` methods to be consistent with the rest of the types.

Reference: https://git-scm.com/docs/gitprotocol-pack
This extracts changes from https://github.com/go-git/go-git/pull/1111

TODO:
- [x] Refactor `transport` https://github.com/go-git/go-git/pull/1337
- [x] Implement `multi_ack` support based on https://github.com/go-git/go-git/pull/1338
- [x] Implement HTTP dumb protocol https://github.com/go-git/go-git/pull/1339
- [ ] Fix tests

This needs https://github.com/go-git/go-git/pull/1299 and https://github.com/go-git/go-git/pull/1300

Supersedes: https://github.com/go-git/go-git/pull/938
Supersedes: https://github.com/go-git/go-git/pull/1111

PR #1339, opened 2024-12-31
refactor transport: plumbing: transport/http, support http dumb protocol (4/5)

aymanbagabas:
This adds support to the HTTP dumb protocol described in
https://git-scm.com/book/en/v2/Git-Internals-Transfer-Protocols

Fixes: https://github.com/go-git/go-git/issues/492

Based on: https://github.com/go-git/go-git/pull/1303 and https://github.com/go-git/go-git/pull/1337

### DIFF
--- a/plumbing/transport/http/dumb.go
+++ b/plumbing/transport/http/dumb.go
@@ -1,0 +1,432 @@
+package http
+
+import (
+	"bufio"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"net/http"
+	"path/filepath"
+	"strings"
+
+	"github.com/go-git/go-billy/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/format/idxfile"
+	"github.com/go-git/go-git/v5/plumbing/format/objfile"
+	"github.com/go-git/go-git/v5/plumbing/format/packfile"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/plumbing/transport"
+	"github.com/go-git/go-git/v5/utils/ioutil"
+)
+
+func (s *HTTPSession) fetchDumb(ctx context.Context, req *transport.FetchRequest) error {
+	if req.Depth != 0 {
+		return fmt.Errorf("dumb http protocol does not support shallow capabilities")
+	}
+
+	fsi, ok := s.st.(interface {
+		Filesystem() billy.Filesystem
+	})
+	if !ok {
+		return fmt.Errorf("dumb http protocol requires a filesystem")
+	}
+
+	repoFs := fsi.Filesystem()
+	r := newFetchWalker(s, ctx, repoFs)
+	if err := r.process(); err != nil {
+		return err
+	}
+
+	if err := r.fetch(); err != nil {
+		return fmt.Errorf("error fetching objects: %w", err)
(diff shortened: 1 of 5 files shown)


plumbing/transport/http/dumb.go
```diff
@@ -0,0 +1,432 @@
...
+	"github.com/go-git/go-git/v5/plumbing/format/objfile"
+	"github.com/go-git/go-git/v5/plumbing/format/packfile"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/plumbing/transport"
+	"github.com/go-git/go-git/v5/utils/ioutil"
+)
+
+func (s *HTTPSession) fetchDumb(ctx context.Context, req *transport.FetchRequest) error {
+	if req.Depth != 0 {
+		return fmt.Errorf("dumb http protocol does not support shallow capabilities")
```
pjbgf, line 26:
```suggestion
		return errors.New("dumb http protocol does not support shallow capabilities")
```

plumbing/transport/http/dumb.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	fsi, ok := s.st.(interface {
+		Filesystem() billy.Filesystem
+	})
+	if !ok {
+		return fmt.Errorf("dumb http protocol requires a filesystem")
```
pjbgf, line 33:
```suggestion
		return errors.New("dumb http protocol requires a filesystem")
```

plumbing/transport/http/dumb.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	if _, err := io.Copy(f, res.Body); err != nil {
+		_ = f.Close()
+		return err
+	}
+
+	if err := f.Close(); err != nil {
+		return err
+	}
```
pjbgf, line 133: As `res.Body` [must](https://github.com/golang/go/blob/858a0e9dfd10ac94a0b9de4429749f0cb99e8cb8/src/net/http/response.go#L61-L64) always be read to completion, maybe something like:

```suggestion
	copy := func(w io.Writer) error {
		if _, err := io.Copy(w, res.Body); err != nil {
			return err
		}

		if err := f.Close(); err != nil {
			return err
		}
	}
	
	f, err := r.fs.TempFile(filepath.Dir(path), filepath.Base(path)+".temp")
	if err != nil {
		copy(io.Discard)	
		return err
	}

	copy(f)	
```

plumbing/transport/http/dumb.go
```diff
@@ -0,0 +1,432 @@
...
+
+	// TODO: support hardlinks and "core.createobject" configuration
+	return r.fs.Rename(f.Name(), path)
+}
+
+// getHead returns the HEAD reference from the server.
+func (r *fetchWalker) getHead() (*plumbing.Reference, error) {
+	url := fmt.Sprintf("%s/HEAD", r.ep.String())
```
pjbgf, line 141: Let's use [JoinPath](https://pkg.go.dev/net/url@go1.21#JoinPath) instead.

plumbing/transport/http/dumb.go
```diff
@@ -0,0 +1,432 @@
...
+	req, err := http.NewRequestWithContext(r.ctx, "GET", url, nil)
```
pjbgf, line 142: Instead of `GET` -> `http.MethodGet`.

pjbgf (approved):

PR #1345, opened 2025-01-02
feat(plumbing/format/pktline): add DelimPkt and ResponseEndPkt support (protocol v2)

bored-engineer:
Git's [v2 wire protocol](https://git-scm.com/docs/protocol-v2) defines a few special packets in addition to the existing `flush-pkt`:
> In protocol v2 these special packets will have the following semantics:
> * `0000` Flush Packet (`flush-pkt`) - indicates the end of a message
> * `0001` Delimiter Packet (`delim-pkt`) - separates sections of a message
> * `0002` Response End Packet (`response-end-pkt`) - indicates the end of a response for stateless connections

This PR adds support for reading and writing the `delim-pkt` and `response-end-pkt` within the existing `plumbing/format/pktline` package.

The behavior is kept the same as `flush-pkt` where `Bytes` returns an empty slice when these are encountered, but now the "type" of that packet can be determined via the `FlushPkt`, `DelimPkt`, `ResponseEndPkt` methods which return true/false.


aymanbagabas:
@bored-engineer Thank you for the PR. V2 wire protocol progress is being tracked in https://github.com/go-git/go-git/tree/v6-exp and https://github.com/go-git/go-git/tree/v6-transport. See https://github.com/go-git/go-git/blob/v6-transport/plumbing/format/pktline/common.go

Issue #1341, opened 2024-12-31
host key mismatch after upgrading module from v5.12.0 to v5.13.0


jgoodall:
SSH git clone fails after update to v5.13.0. Specifically, the issue comes from `ssh.FixedHostKey(hostKey)` where we try to set the host key callback. The following works fine with `v5.12.0`, but fails with `v5.13.0`; the error is `ssh: handshake failed: ssh: host key mismatch`. So I dont think it is actually a host key mismatch, but something that changed in this module (or a dependency).

```
	hostKey, _, _, _, err := ssh.ParseAuthorizedKey([]byte(serverHostKey))
	if err != nil {
		return "", fmt.Errorf("unable to parse host key: %w", err)
	}
	pk, err := ssh2.NewPublicKeys(gitUser, []byte(repoPrivateKey), "")
	if err != nil {
		return "", fmt.Errorf("unable to create git public key: %w", err)
	}
	pk.HostKeyCallback = ssh.FixedHostKey(hostKey)
	deployBranch := gitplum.NewBranchReferenceName(gitBranch)
	r, err := git.PlainClone(destDirPath, false, &git.CloneOptions{
		URL:           repoURL,
		Auth:          pk,
	})
	if err != nil {
		return "", fmt.Errorf("unable to clone git repo [%s] (branch [%s]): %w", repoURL, deployBranch, err)
	}
```

I can use insecure and it works, but obviously I would like to verify the host key:

```
	pk.HostKeyCallback = ssh.InsecureIgnoreHostKey()
```

pjbgf:
@jgoodall can you provide some more insights into this? I tried reproducing it with `v5.13.0` against `github.com` and the clone operation worked fine.

The pubkey I used was: `github.com ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOMqqnkVzrm0SdG6UOoqKLsabgH5C9okWi0dh2l9GKJl`.

jgoodall:
It is a private gitlab server. I only bumped go-git (`go get -u github.com/go-git/go-git/v5`), although that did update some required modules:

```
diff --git a/go.mod b/go.mod
index dc8f2d4..e5e229a 100644
--- a/go.mod
+++ b/go.mod
@@ -12 +12 @@ require (
-	github.com/go-git/go-git/v5 v5.12.0
+	github.com/go-git/go-git/v5 v5.13.0
@@ -32 +32 @@ require (
[...]

Issue #654, opened 2023-01-09
ssh: unable to ignore host key auth callback 


aymanbagabas:
Hi,

The changes introduced in https://github.com/go-git/go-git/commit/c35b8082c863f2106de1c3c95ba9ed21d30f9371 breaks the ability to ignore the client host key check. In our case, we're using `ssh.InsecureIgnoreHostKey()` for integration testing where we don't care about validating the host key. The changes made in the commit deprecate the `HostKeyCallbackHelper` interface and don't provide any workaround.

Implementing a `ssh.AuthMethod` with a custom `HostKeyCallback` won't work either since the associated `ssh.ClientConfig` will be overwritten by [`ssh.SetConfigHostKeyFields()`](https://github.com/go-git/go-git/commit/c35b8082c863f2106de1c3c95ba9ed21d30f9371#diff-4b7927140f2bd6fe4030126982cb323cf2b49656e75e7f235ca1fbf0336aadf4R124-R128) when establishing the connection.

Before these changes were introduced, you could pass a `HostKeyCallbackHelper` to any ssh AuthMethod and it will set the `HostKeyCallback` function to be used to validate the host key.

```go
	pks, _ := ssht.NewPublicKeysFromFile("git", filepath, "")
	pks.HostKeyCallbackHelper = ssht.HostKeyCallbackHelper{
	 	HostKeyCallback: ssh.InsecureIgnoreHostKey(),
	}
	err = r.Push(&git.PushOptions{
		RemoteName: "origin",
		Auth:       pks,
	})
```

I tried wrapping the `PublicKeys` AuthMethod and modifying the `HostKeyCallback`, but it gets [overwritten](https://github.com/go-git/go-git/commit/c35b8082c863f2106de1c3c95ba9ed21d30f9371#diff-4b7927140f2bd6fe4030126982cb323cf2b49656e75e7f235ca1fbf0336aadf4R124-R128) before the connection happens.

```go
type pkAuthMethod struct {
	*ssht.PublicKeys
}

func (p *pkAuthMethod) ClientConfig() (*ssh.ClientConfig, error) {
	cc, err := p.PublicKeys.ClientConfig()
	if err != nil {
		return nil, err
	}
	cc.HostKeyCallback = ssh.InsecureIgnoreHostKey()
	return cc, nil
}

func main() {
	pks, _ := gssh.NewPublicKeysFromFile("git", pkPath, "")
	auth := &pkAuthMethod{pks}
[...]

PR #1346, opened 2025-01-02
Revert "plumbing: transport/ssh, Add support for SSH @cert-authority."

pjbgf:
Reverts go-git/go-git#1157 due to #1341.

### DIFF
--- a/plumbing/transport/ssh/auth_method.go
+++ b/plumbing/transport/ssh/auth_method.go
@@ -230,11 +230,11 @@ func (a *PublicKeysCallback) ClientConfig() (*ssh.ClientConfig, error) {
 //	~/.ssh/known_hosts
 //	/etc/ssh/ssh_known_hosts
 func NewKnownHostsCallback(files ...string) (ssh.HostKeyCallback, error) {
-	db, err := newKnownHostsDb(files...)
-	return db.HostKeyCallback(), err
+	kh, err := newKnownHosts(files...)
+	return ssh.HostKeyCallback(kh), err
 }
 
-func newKnownHostsDb(files ...string) (*knownhosts.HostKeyDB, error) {
+func newKnownHosts(files ...string) (knownhosts.HostKeyCallback, error) {
 	var err error
 
 	if len(files) == 0 {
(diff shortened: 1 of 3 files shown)


aymanbagabas (approved):

evanelias:
Since you've now locked #1341, I will post a final response here, as your last comment there may indicate there was some misunderstanding. My [previous comment in that issue](https://github.com/go-git/go-git/issues/1341#issuecomment-2624989012) wasn't about the *timing* of the revert in any way. Again, doing a quick revert ahead of an urgent security release is 100% understandable and is absolutely the right thing to do. I completely agree with you there.

My comment was about the lack of PR author notification, not the timing. Notification could have occurred after the revert/release. No one ever commented on the original PR (#1157) to indicate there was a problem. Nor did anyone tag the PR author in #1341, nor here in the revert PR.

This is identical to my experience a few years back when my PR #548 apparently caused a regression (#654) - no one ever notified me, so I had no idea whatsoever that there was a problem.

Previously in #1341, you said you wanted PRs from folks who are "keen on fixing arising issues". But contributors cannot possibly fix issues if no one tells them there is an issue in the first place.

Anyway, my takeaway from these interactions is that it appears that the substantial amount of time spent on recent improvements in [skeema/knownhosts](https://github.com/skeema/knownhosts) has been a completely wasted effort: go-git is the largest external direct importer of skeema/knownhosts, but you are no longer using any of the recent knownhosts improvements in go-git v5, and meanwhile in go-git v6 you've made an internal fork of skeema/knownhosts in order to add a single debug printf (but without ever reporting a corresponding problem to skeema/knownhosts).

I am strongly considering permanently abandoning and archiving the public [skeema/knownhosts](https://github.com/skeema/knownhosts) repo to avoid losing any more time on it in the future.

pjbgf:
> No one ever commented on the original PR (https://github.com/go-git/go-git/pull/1157) to indicate there was a problem.

The revert linked the original PR, which should have caused a notification to the author. As the issue was closed, potentially the author ignored or maybe missed it:

Image

As for the previous experience, as that was not as obvious a regression, that was completely missed (at least on my part).

Either way, I agree with you maybe the right course of action would be to create a new issue tagging the author for a follow-up. Point-taken for future regressions. 👍 

The other points I will answer on an open issue.

evanelias:
Thanks. Regarding "The revert linked the original PR, which should have caused a notification to the author" - in my experience, GitHub doesn't ever send a notification email for that type of linking. Perhaps there's a setting for it, but I can't seem to find one to subscribe to emails for that type of activity.

PR #1360, opened 2025-01-05
git: worktree, fix restoring dot slash files. Fixes #1176

### LINKED ISSUE #1176: Can't checkout a file
The title might be a misleading, but I can't seem to figure out how to do this `git checkout -- ./blah ./foo` with the exact behavior in git (i.e. discard unstaged files only). I'd expect it to look something like this:
```go
// Open the repository.
repo, err := git.PlainOpen(path)
if err != nil {
	return err
}

// Get the working directory tree.
w, err := repo.Worktree()
if err != nil {
	return err
}

return w.Checkout(&git.CheckoutOptions{
	Force: true,
	Keep:  true,
        Files: []string{"./blah", "./foo"},
})
```

I can't find anything in the issues or the codebase that even remotely hints at being able to do this.

BeChris:


pjbgf (approved):
@BeChris thanks for working on this. :bow:

Issue #706, opened 2023-03-07
Support for SHA256


pjbgf:
Implement SHA256 support as per https://git-scm.com/docs/hash-function-transition/.

High level changes:
- [ ] Repository Format Extension
- [ ] Loose object index
- [ ] Bidirectional (SHA-1 <-> SHA-256) mapping table
- [ ] Fetching from SHA1 server
- [ ] Signed Commits
- [ ] Signed Tags
- [ ] Mergetag
- [ ] Submodules
- [ ] Shallow clone (unsupported - fail gracefully)

AdamMajer:
Thanks for working on this. I think the transition doesn't yet need to be implemented, since it doesn't really work in git tools anyway. What should be implemented is simply ability to work with SHA256 repositories - I think this is the scope of the work here.

SHA1 <> SHA256 repo interop probably will need some more thought and implementation upstream first.

AdamMajer:
Seems a duplicate of https://github.com/go-git/go-git/issues/229 although more work is done here so maybe close the other issue.

pjbgf:
@AdamMajer thanks for calling out the duplicated issue - just closed it.

I agree with your comment on the transition. Given that `go-git` aims to be aligned with upstream, the items pending for SHA256 will always have a hard blocker in terms of whether or not upstream already merged the given feature/change.

#707 Already added support to create and make commits in SHA256 repositories, but this is still experimental - hence why it is blocked by a build tag.

AdamMajer:
@pjbgf , just to make sure I'm reading the patch you linked correctly, this basically just allows go-git to use only sha256 or sha1, depending on the build tag?  There is no possibility of using both hash types in one program?

pjbgf:
@AdamMajer that's correct. That is a temporary limitation which aims to keep backwards compatibility with the API - as `plumbing.Hash` defines the hash's length. There are also some performance/memory footprint considerations that will need to be considered when we implement that.

Once the bilateral mapping table is implemented, we should remove that limitation.

gecube:
Hi!

I am waiting for changes, as I just experienced that Gitlab.com offers the new SHA-256 algorithm, but unfortunately tools bases on go-git could not work with such a repos...

![image](https://github.com/user-attachments/assets/bb8e7314-cd6c-43ca-9e67-61a26c125c43)

pjbgf:
@gecube thanks for the ping - I wasn't aware of that support from Gitlab. :bow: 
[...]

PR #1362, opened 2025-01-06
[v6] Make hash usage extensible to enable concurrent support of multiple hash types

pjbgf:
Refactor the codebase so that it stops relying on `plumbing.Hash`'s implementation details, and instead rely on exported methods.
This is a stepping stone to enable support of multiple hash types, which will be dealt with as a follow-up PR.

The `hash` package now contains all code regarding hash, with exception to `Hasher` related logic, which is spread between `plumbing` and `hasher`. `plumbing.Hash` now redirects to `hash.SHA1Hash`, which may (or may not) be dropped at a later stage.

The previous sha256 specific tests that relied on build tags were removed as they are now redundant.

Relates to #706 #899.
Supersedes to #920.


plumbing/hash/hash_sha256.go
```diff
@@ -1,15 +1,48 @@
...
+	hash [SHA256Size]byte
+}
+
+func (ih SHA256Hash) Size() int {
+	return len(ih.hash)
+}
+
+func (ih SHA256Hash) IsZero() bool {
+	var empty SHA256Hash
+	return ih == empty
```
aymanbagabas, line 19:
```suggestion

var empty256 SHA256Hash

func (ih SHA256Hash) IsZero() bool {
	return ih == empty256
```

plumbing/hash/hash_sha1.go
```diff
@@ -1,15 +1,58 @@
...
+}
+
+func (ih SHA1Hash) Size() int {
+	return len(ih.hash)
+}
+
+func (ih SHA1Hash) IsZero() bool {
+	var empty SHA1Hash
+	return ih == empty
+}
```
aymanbagabas, line 30:
```suggestion

var empty SHA1Hash

func (ih SHA1Hash) IsZero() bool {
	return ih == empty
}
```

plumbing/hash/new.go
```diff
@@ -0,0 +1,126 @@
...
+//
+// Defaults to SHA1-sized hash if the provided format is not supported.
+func ZeroFromObjectFormat(f format.ObjectFormat) ObjectID {
+	switch f {
+	case format.SHA256:
+		return SHA256Hash{}
+	default:
+		return SHA1Hash{}
+	}
+}
```
aymanbagabas, line 126: We don't use these funcs now. We can add them later when the need arises

plumbing/hash/object.go
```diff
@@ -0,0 +1,29 @@
+package hash
+
+import "io"
+
+// ObjectID represents a calculated hash, which is immutable by nature.
+type ObjectID interface {
```
aymanbagabas, line 6: I'd rename this to `Hash` to keep naming consistent

plumbing/hash/object.go
```diff
@@ -0,0 +1,29 @@
...
+	String() string
+	// Bytes returns a slice of bytes containing the ObjectID.
+	Bytes() []byte
+	// HasPrefix verifies whether the objectID start with a given prefix.
+	HasPrefix([]byte) bool
+}
+
+// LazyObjectID represents an object hash which may not be known at the time
+// the object is created. Or an objectID which changes during its lifetime.
+type LazyObjectID interface {
```
aymanbagabas, line 23: Same, I'd rename this to `LazyHash`

plumbing/hasher/hasher.go
```diff
@@ -0,0 +1,158 @@
...
+//     depending on the format needed.
+//   - Thread-safety.
+//   - API restricts ability of generating invalid hashes.
+type ObjectHasher interface {
+	// Size returns the length of resulting hash.
+	Size() int
+	// Compute calculates the hash of a Git object. The process involves
+	// first writing the object header, which contains the object type
+	// and content size, followed by the content itself.
+	Compute(ot plumbing.ObjectType, d []byte) (ObjectID, error)
```
aymanbagabas, line 34: Why don't you use the types from `hash`?

plumbing/hash/hash_sha1.go
```diff
@@ -1,15 +1,58 @@
...
 const (
 	// CryptoType defines what hash algorithm is being used.
 	CryptoType = crypto.SHA1
 	// Size defines the amount of bytes the hash yields.
-	Size = 20
+	Size = SHA1Size
 	// HexSize defines the strings size of the hash when represented in hexadecimal.
-	HexSize = 40
+	HexSize = SHA1HexSize
 )
```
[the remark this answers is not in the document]
pjbgf: This is still being used, instead of replacing all the refs to SHA1Size, to then have to change them again later, I think we are better off leaving as is and replace as the specific packages get multi-hash support.

plumbing/hash/hash_zero.go
```diff
@@ -0,0 +1,9 @@
+package hash
+
+func Zero() SHA1Hash {
+	return SHA1Hash{}
+}
+
+func ZeroSHA256() SHA256Hash {
+	return SHA256Hash{}
+}
```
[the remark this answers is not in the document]
pjbgf: The main reason for a func was to avoid the awkard position were API users are able to change the meaning of zero, using a single var will revert to that. However, I agree with the change for the time being, but it is likely this code will change again before it being ready for merging back to `v6-exp`.

plumbing/hash/object.go
pjbgf: I am holding back on making this change while `plumbing.Hash` exists, to avoid additional confusion on having multiple "Hash" types. At present `plumbing.Hash` has more than one use which is leading me to think that potentially it could be split into two different types. Either way, let's make that decision together with the final design for the API (as per discord thread) - which will be in a follow-up PR.

pjbgf:
@aymanbagabas changes made and code rebased. PTAL

plumbing/hash/new.go
```diff
@@ -0,0 +1,99 @@
...
+// FromHex parses a hexadecimal string and returns an ObjectID
+// and a boolean confirming whether the operation was successful.
+// The hash (and object format) is inferred from the length of the
+// input.
+//
+// If the operation was not successful, the resulting hash is nil
+// instead of a zeroed hash.
+func FromHex(in string) (ObjectID, bool) {
+	if len(in) < SHA1HexSize ||
+		len(in) > SHA256HexSize {
```
aymanbagabas, line 47:
```suggestion
	if len(in) != SHA1HexSize &&
		len(in) != SHA256HexSize {
```

plumbing/hash/new.go
```diff
@@ -0,0 +1,99 @@
...
+// FromBytes creates an ImmutableHash object based on the value its
+// Sum() should return.
+// The hash type (and object format) is inferred from the length of
+// the input.
+//
+// If the operation was not successful, the resulting hash is nil
+// instead of a zeroed hash.
+func FromBytes(in []byte) (ObjectID, bool) {
+	if len(in) < SHA1Size ||
+		len(in) > SHA256Size {
```
aymanbagabas, line 81:
```suggestion
	if len(in) != SHA1Size &&
		len(in) != SHA256Size {
```

plumbing/hasher/hasher.go
```diff
@@ -0,0 +1,153 @@
...
+		hasher: hash.New(crypto.SHA1),
+		buf:    *bytes.NewBuffer(make([]byte, 32)),
+	}
+}
+
+type objectHasherSHA1 struct {
+	hasher hash.Hash
+	m      sync.Mutex
+	// both fields below are allocation optimisations.
+	b   [20]byte
```
aymanbagabas, line 75:
```suggestion
	b   [hash.SHA1Size]byte
```

plumbing/hasher/hasher.go
```diff
@@ -0,0 +1,153 @@
...
+		hasher: hash.New(crypto.SHA256),
+		buf:    *bytes.NewBuffer(make([]byte, 32)),
+	}
+}
+
+type objectHasherSHA256 struct {
+	hasher hash.Hash
+	m      sync.Mutex
+	// both fields below are allocation optimisations.
+	b   [32]byte
```
aymanbagabas, line 116:
```suggestion
	b   [hash.SHA256Size]byte
```

plumbing/hash/hash_sha1.go
```diff
@@ -1,15 +1,57 @@
...
+
+type SHA1Hash struct {
```
aymanbagabas, line 19: Perhaps we could move the actual implementation in a sub package `plumbing/hash/sha1` with a unified API for both sha1 and sha256

I could also see the use for an object format registry similar to `crypto.RegisterHash`

pjbgf: > I could also see the use for an object format registry similar to `crypto.RegisterHash`

I had some discussions with the folks from [gittuf](https://github.com/gittuf/gittuf) a while back, and that seemed to support some use cases they had. Let's definitely consider this.

Some of these changes I already have locally. I am thinking on `v6-transport` merged first, and then getting them going. Just to avoid all the rebasing.

pjbgf:
Latest feedback adopted and rebased from `main`.

PR #1364, opened 2025-01-06
plumbing: use the correct user agent string. Fixes #883

### LINKED ISSUE #883: Always sends HTTP user-agent of `git/1.0`?
It seems this library always uses the user agent `git/1.0` in HTTP requests, since that is hardcoded here:
https://github.com/go-git/go-git/blob/72ce9961c676b73c9bdd0678b64af7b0a42253f8/plumbing/transport/http/common.go#L27

I see some code to support "in-band"(?) user-agents:
https://github.com/go-git/go-git/blob/72ce9961c676b73c9bdd0678b64af7b0a42253f8/plumbing/protocol/packp/capability/capability.go#L246-L254

But this never seems to influence the HTTP User-Agent header. Am I correct in my assumption? Would it be possible to expose this value in the HTTP header?

uragirii:
Fixes: #883 

Makes sure the transport uses correct `userAgent` instead of hardcoded `git/1.0`

### DIFF
--- a/plumbing/transport/http/common.go
+++ b/plumbing/transport/http/common.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/protocol/packp"
+	"github.com/go-git/go-git/v5/plumbing/protocol/packp/capability"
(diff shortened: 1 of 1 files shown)


pjbgf (approved):
@uragirii thanks for fixing this issue. :bow:

Issue #713, opened 2023-03-15
How to clone remote repo with --filter option?


RoxKilly:
My golang application needs to list all file and folder paths in a subfolder of a git repo (a search path).  I only care about the HEAD commit.  I'm using `Clone()` first, then walking the tree from the HEAD commit.

```
repo, _ = git.Clone(memory.NewStorage(), nil, git.CloneOptions{URL: repoUrl, Depth: 1})
// Get commit object at HEAD
// get tree of commit object
// iterate into tree and append file names if file is within my search path
```

This works but the clone operation is very slow as the repo gets large and probably gives me a ton of stuff I don't need (I'm only interested in a subset of the repo tree).  I noticed that the [git clone docs][2] have a `--filter` option that could serve my purposes.  Is this option available via the go-git library? I don't see anything in the [CloneOptions object][3]

  [2]: https://git-scm.com/docs/git-clone
  [3]: https://github.com/go-git/go-git/blob/master/options.go#L37-L67

AnthonyHerman:
I would also like to see this functionality come to go-git.

jfontan:
You can set the `Depth` to 1 in `CloneOptions` to only download that commit.

jfontan:
My bad. I had not noticed you are already using `Depth: 1`. Is the repository that takes so long public?

pjbgf:
@RoxKilly `go-git` does not support [partial clones ](https://git-scm.com/docs/partial-clone) atm. But I would be happy to review a PR that implements it - so feel free to tag me if you decide to give it a go.

In the meantime, for a more efficient clone, you could change your `CloneOptions` to the following:

```
 &git.CloneOptions{
			URL:           repoUrl,
			SingleBranch:  true,
			Depth:         1,
			Tags: git.NoTags,
		})
```

More info as to why: https://github.com/go-git/go-git/issues/545#issuecomment-1353681676

pjbgf:
Closing this as it seems that this feature is no longer supported upstream - at least not from a path perspective.

Git v2.22.0 [release notes](https://github.com/git/git/blob/master/Documentation/RelNotes/2.22.0.txt):
>  * The filter specification "--filter=sparse:path=<path>" used to  create a lazy/partial clone has been removed. 

The latest version of [git-rev-list](https://git-scm.com/docs/git-rev-list#_object_traversal) also states that:
> Note that the form `--filter=sparse:path=<path>` that wants to read from an arbitrary path on the filesystem has been dropped for security reasons.

[...]

PR #1375, opened 2025-01-13
git: fetch and clone --filter support

EarthCompass:
This PR adds Filter support for git clone and fetch. #713 

I see an open PR #1108 that implemented this for fetch but was left open somehow. This is a more complete version of #1108.

### DIFF
--- a/options.go
+++ b/options.go
@@ -12,6 +12,7 @@ import (
 	"github.com/go-git/go-git/v5/plumbing"
 	formatcfg "github.com/go-git/go-git/v5/plumbing/format/config"
 	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/plumbing/protocol/packp"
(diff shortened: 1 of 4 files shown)


ste93cry:
I came across this pull request while looking for a way to fetch the contents of a single file regardless of repository size. The `blob:none` filter works perfectly for my use case, but I noticed that the changes were merged only into the `v6-exp` branch, which appears to be quite far from a stable release. Would it be possible to backport this pull request and include it in the next stable `5.x` version?

pjbgf:
@ste93cry this is now in [main](https://github.com/go-git/go-git/commit/69f456373e09f01822f46db0486cd2225a128503) for `v6`. We should be cutting an alpha release for it soon.

ste93cry:
It would still be very helpful to have this backported to `v5` as well. I don't expect `v6` to reach production-ready stability in the near future, and relying on an alpha release in production isn't really feasible for me. The change appears to be self‑contained, and from experiments I've done, `v5` already includes all the necessary `packp` internals to support it.

PR #1379, opened 2025-01-14
Improve Status performance with many ignored files

silkeh:
This PR contains two changes that significantly improve the performance of `WorkTree.Status()` (from 491 s to 5 s):

- Skip ignored files when walking through the worktree.
  This signigifantly improves the performance of `Status()`:
  In a repository with 3M ignored files `Status` now takes 5 s instead of 160 s.
- Fix loading of `.gitignore` files in nested ignored directories. 
  These were not matched in the current implementation, as that only checks the ignores of the current directory.
  As a side-effect this change also leads to significantly improved performance: in a repository with 3M ignored files `Status` now takes 160 s instead of 491 s.

I have a backport to `master` ready if desired.

### DIFF
--- a/plumbing/format/gitignore/dir.go
+++ b/plumbing/format/gitignore/dir.go
@@ -51,10 +51,15 @@ func readIgnoreFile(fs billy.Filesystem, path []string, ignoreFile string) (ps [
 // recursively traversing through the directory structure. The result is in
 // the ascending order of priority (last higher).
 func ReadPatterns(fs billy.Filesystem, path []string) (ps []Pattern, err error) {
-	ps, _ = readIgnoreFile(fs, path, infoExcludeFile)
+	return extendPatterns(fs, nil, path)
+}
 
+func extendPatterns(fs billy.Filesystem, start []Pattern, path []string) (ps []Pattern, err error) {
+	dirps, _ := readIgnoreFile(fs, path, infoExcludeFile)
 	subps, _ := readIgnoreFile(fs, path, gitignoreFile)
-	ps = append(ps, subps...)
+	dirps = append(dirps, subps...)
+	ignores := append(start, dirps...)
+	ps = dirps
 
 	var fis []os.FileInfo
 	fis, err = fs.ReadDir(fs.Join(path...))
@@ -64,12 +69,12 @@ func ReadPatterns(fs billy.Filesystem, path []string) (ps []Pattern, err error) 
 
 	for _, fi := range fis {
 		if fi.IsDir() && fi.Name() != gitDir {
-			if NewMatcher(ps).Match(append(path, fi.Name()), true) {
+			if NewMatcher(ignores).Match(append(path, fi.Name()), true) {
 				continue
 			}
 
 			var subps []Pattern
-			subps, err = ReadPatterns(fs, append(path, fi.Name()))
+			subps, err = extendPatterns(fs, append(start, dirps...), append(path, fi.Name()))
 			if err != nil {
 				return
 			}
--- a/plumbing/format/gitignore/dir_test.go
+++ b/plumbing/format/gitignore/dir_test.go
@@ -48,7 +48,9 @@ func (s *MatcherSuite) SetupTest() {
 	s.NoError(err)
 	_, err = f.Write([]byte("ignore.crlf\r\n"))
 	s.NoError(err)
-	_, err = f.Write([]byte("ignore_dir\n"))
+	_, err = f.Write([]byte("/ignore_dir\n"))
+	s.NoError(err)
+	_, err = f.Write([]byte("nested/ignore_dir\n"))
 	s.NoError(err)
 	err = f.Close()
 	s.NoError(err)
@@ -71,6 +73,17 @@ func (s *MatcherSuite) SetupTest() {
 	_, err = fs.Create("ignore_dir/file")
 	s.NoError(err)
 	err = f.Close()
+	s.NoError(err)
+
+	err = fs.MkdirAll("nested/ignore_dir", os.ModePerm)
+	s.NoError(err)
+	f, err = fs.Create("nested/ignore_dir/.gitignore")
+	s.NoError(err)
+	_, err = f.Write([]byte("!file\n"))
+	s.NoError(err)
+	err = f.Close()
+	s.NoError(err)
+	_, err = fs.Create("nested/ignore_dir/file")
 	s.NoError(err)
 
 	err = fs.MkdirAll("another", os.ModePerm)
@@ -284,13 +297,14 @@ func (s *MatcherSuite) SetupTest() {
 
 func (s *MatcherSuite) TestDir_ReadPatterns() {
 	checkPatterns := func(ps []Pattern) {
-		s.Len(ps, 7)
+		s.Len(ps, 8)
 		m := NewMatcher(ps)
 
 		s.True(m.Match([]string{"exclude.crlf"}, true))
 		s.True(m.Match([]string{"ignore.crlf"}, true))
 		s.True(m.Match([]string{"vendor", "gopkg.in"}, true))
 		s.True(m.Match([]string{"ignore_dir", "file"}, false))
+		s.True(m.Match([]string{"nested", "ignore_dir", "file"}, false))
 		s.False(m.Match([]string{"vendor", "github.com"}, true))
 		s.True(m.Match([]string{"multiple", "sub", "ignores", "first", "ignore_dir"}, true))
 		s.True(m.Match([]string{"multiple", "sub", "ignores", "second", "ignore_dir"}, true))
--- a/plumbing/format/gitignore/noder.go
+++ b/plumbing/format/gitignore/noder.go
@@ -1,0 +1,138 @@
+package gitignore
+
+import (
+	"slices"
+
+	"github.com/go-git/go-git/v6/utils/merkletrie/noder"
+)
+
+var _ noder.Noder = (*MatchNoder)(nil)
+
+// MatchNoder is an implementation of [noder.Noder] that only includes nodes based on a pattern.
+type MatchNoder struct {
+	noder.Noder
+
+	matcher  Matcher
+	invert   bool
+	path     []string
+	children []noder.Noder
+}
+
+// IgnoreNoder returns a [MatchNoder] that filters out the given pattern.
+func IgnoreNoder(m Matcher, n noder.Noder) *MatchNoder {
+	var path []string
+	if name := n.Name(); name != "." {
+		path = []string{name}
+	}
+
+	return &MatchNoder{matcher: m, invert: true, Noder: n, path: path}
+}
+
+// Children returns matched children.
+// It implements [noder.Noder].
+func (n *MatchNoder) Children() ([]noder.Noder, error) {
+	if len(n.children) > 0 {
+		return n.children, nil
+	}
+
+	children, err := n.Noder.Children()
+	if err != nil {
+		return nil, err
+	}
+
+	n.children = n.ignoreChildren(children)
+
+	return n.children, nil
+}
+
+func (n *MatchNoder) ignoreChildren(children []noder.Noder) []noder.Noder {
+	found := make([]noder.Noder, 0, len(children))
+	pathBuf := slices.Grow(n.path, len(n.path)+1)
+
+	for _, child := range children {
+		path := append(pathBuf, child.Name())
+		if n.match(path, child.IsDir()) {
+			continue
+		}
+
+		found = append(found, n.newChild(child, path))
+	}
+
+	return found
+}
+
+func (n *MatchNoder) match(path []string, isDir bool) bool {
+	if n.matcher != nil && n.matcher.Match(path, isDir) {
+		return n.invert
+	}
+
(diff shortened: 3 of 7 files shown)


pjbgf:
@silkeh thanks for proposing this PR. Although this may improve the use case of multiple ignored files, it seems to cause a regression when that is not the case. I did a quick benchmark testing a call to `Status()` on a copy of the `go-git` repo, and these changes seem to have a negative impact on both allocation numbers and time per operation:

```
BenchmarkStatus-16 (v5.13.0)      	      48	  21514427 ns/op	19511754 B/op	   69999 allocs/op
BenchmarkStatus-16 (this PR)      	      34	  34658273 ns/op	42274701 B/op	  106194 allocs/op
```

I am struggling for time to pinpoint optimisation areas in the PR, would you be able to optimise it for the standard use case as well?

silkeh:
@pjbgf: Thanks for taking a look! I'll take a look later this week and see what I can improve.

onee-only:
Hello @silkeh, Are you still working on this? This could be a big improvement for https://github.com/go-git/go-git/issues/181.

silkeh:
@onee-only: That was the plan, but I've been busy with other things. I'll try to make some time next week.

silkeh:
I finally had some time to take a look and rebase the PR. The difference compared to `origin/main` (3a68d040) seems minor now. I got the following results:

```
BenchmarkStatus-12(v5.16.2)        433	  27579139 ns/op	18265183 B/op	   43611 allocs/op
BenchmarkStatus-12(origin/main)    324	  37745332 ns/op	 2106671 B/op	   41265 allocs/op
BenchmarkStatus-12(first commit)   321	  37490431 ns/op	 2107606 B/op	   41274 allocs/op
BenchmarkStatus-12(second commit)  307	  39710261 ns/op	 2174177 B/op	   42200 allocs/op
```

v5.16.2 is a lot faster, but this PR in its current form doesn't hurt performance too much. If you test with a repo on disk it actually seems to be a bit faster.

Benchmark I used:

```go
func BenchmarkStatus(b *testing.B) {
	r, err := Clone(memory.NewStorage(), memfs.New(), &CloneOptions{
		URL: "https://github.com/go-git/go-git",
	})
	require.NoError(b, err)

	workTree, err := r.Worktree()
	require.NoError(b, err)

	_, err = workTree.Status()
	require.NoError(b, err)

	b.ReportAllocs()
	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		workTree.Status()
	}
}
```

If I read the profiling correctly most of the performance degradation is introduced in `MatchNoder.NumChildren` because it now has to run through the patterns instead of just returning the info from disk:

image

I did a *tiny* optimization to reduce the allocations a bit, but the rest of the performance improvements would probably have to be done in the `gitignore.matcher`.

plumbing/format/gitignore/dir.go
```diff
@@ -51,10 +51,15 @@ func readIgnoreFile(fs billy.Filesystem, path []string, ignoreFile string) (ps [
 // recursively traversing through the directory structure. The result is in
 // the ascending order of priority (last higher).
 func ReadPatterns(fs billy.Filesystem, path []string) (ps []Pattern, err error) {
```
onee-only, line 53:
```suggestion
func ReadPatterns(fs billy.Filesystem, path string) (ps []Pattern, err error) {
```
Since exported function is now separated, can we make `path` to single string? Without reading it's internal behavior, I first thought it represents specific directories to read ignore patterns, not path parts to be joined.

plumbing/format/gitignore/noder.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+var _ noder.Noder = (*MatchNoder)(nil)
+
+// MatchNoder is an implementation of [noder.Noder] that only includes nodes based on a pattern.
+type MatchNoder struct {
+	noder.Noder
+
+	matcher  Matcher
+	invert   bool
```
onee-only, line 16: Could you please elaborate what `invert` does? It seems that the value of `invert` is always true.

plumbing/format/gitignore/noder.go
```diff
@@ -0,0 +1,138 @@
...
+	path     []string
+	children []noder.Noder
+}
```
onee-only, line 19: 9727b9be4056599d0ec720e2a698e99a537cd3a3 Introduced a `Option` struct in `utils/merkletrie/filesystem`. What do you think of adding `gitignore.Matcher` as an option in there? I believe we can achieve higher cohension with this.

onee-only: Doing that and making `IgnoreNoder` interface could be a good idea for decoupling the implementation from the `DiffTree()`.

plumbing/format/gitignore/noder.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (n *MatchNoder) noderPaths(path noder.Path) []string {
+	parts := make([]string, len(path))
+
+	for i, p := range path {
+		parts[i] = p.Name()
+	}
+
+	return parts
+}
```
onee-only, line 138:
```suggestion
func findChild(node noder.Noder, name string) (noder.Noder, bool) {
	children, _ := node.Children()

	for _, child := range children {
		if child.Name() == name {
			return child, true
		}
	}

	return nil, false
}

func convertToStrings(path noder.Path) []string {
	parts := make([]string, len(path))

	for i, p := range path {
		parts[i] = p.Name()
	}

	return parts
}
```

Since these functions aren't restricted to the use case of `MatchNoder`, we can move the functions out of the struct.

utils/merkletrie/difftree.go
```diff
@@ -448,3 +454,18 @@ func diffDirs(changes *Changes, ii *doubleIter) error {
 
 	return nil
 }
+
+func isIgnoredNode(tree noder.Noder, path noder.Path) bool {
```
onee-only, line 458:
```suggestion

// isIgnoredPath reports if the path is ignored in the tree.
func isIgnoredPath(tree noder.Noder, path noder.Path) bool {
	in, ok := tree.(*gitignore.MatchNoder)

	return ok && in.PathIgnored(path)
}
```
Maybe this is more intuitive?

utils/merkletrie/difftree.go
```diff
@@ -448,3 +454,18 @@ func diffDirs(changes *Changes, ii *doubleIter) error {
...
+	in, ok := tree.(*gitignore.MatchNoder)
+
+	return ok && in.PathIgnored(path)
+}
+
+func ignoredNode(tree noder.Noder, path noder.Path) (noder.Path, bool) {
```
onee-only, line 464:
```suggestion

// findIgnoredNode finds the node that matches the path in the tree. 
// If the node is found and ignored, it returns the node and true.
func findIgnoredNode(tree noder.Noder, path noder.Path) (noder.Path, bool) {
```
Maybe this is more intuitive?

worktree_status.go
```diff
@@ -154,26 +162,29 @@ func (w *Worktree) diffStagingWithWorktree(reverse, excludeIgnoredChanges bool)
 		return nil, err
 	}
 
-	if excludeIgnoredChanges {
-		return w.excludeIgnoredChanges(c), nil
+	if excludeIgnoredChanges && m != nil {
+		return w.excludeIgnoredChanges(m, c), nil
 	}
```
onee-only, line 167: Do we still need to traverse through the changes?
I think it was already done in MatchNoder and DiffTree.

utils/merkletrie/difftree.go
```diff
@@ -298,8 +299,14 @@ func DiffTreeContext(ctx context.Context, fromTree, toTree noder.Noder,
...
-				if err = ret.AddRecursiveDelete(from); err != nil {
-					return nil, err
+				if node, ok := ignoredNode(toTree, from); ok {
+					if err = diffNodesSameName(&ret, ii, ii.from.current, node); err != nil {
+						return nil, err
+					}
+				} else {
+					if err = ret.AddRecursiveDelete(from); err != nil {
+						return nil, err
+					}
```
onee-only, line 309: I think this should also be added to `onlyToRemains` case.
We should support ignored tree being passed as `toTree`, for instance `resetWorktree()` uses reversed diff of index and worktree.

utils/merkletrie/difftree.go
```diff
@@ -350,8 +357,10 @@ func diffNodes(changes *Changes, ii *doubleIter) error {
 	// compare their full paths as strings
 	switch from.Compare(to) {
 	case -1:
-		if err = changes.AddRecursiveDelete(from); err != nil {
-			return err
+		if ok := isIgnoredNode(to[0], from); !ok {
+			if err = changes.AddRecursiveDelete(from); err != nil {
+				return err
+			}
```
onee-only, line 363: Same as the comment above. It should also be added to the opposite.

onee-only (commented):
Hello @silkeh! I'm looking forward to this being merged. I've left my opinions on the change. I hope these help to improve the code.

plumbing/format/gitignore/dir_test.go
```diff
@@ -73,6 +75,17 @@ func (s *MatcherSuite) SetupTest() {
...
+	err = fs.MkdirAll("nested/ignore_dir", os.ModePerm)
+	s.NoError(err)
+	f, err = fs.Create("nested/ignore_dir/.gitignore")
+	s.NoError(err)
+	_, err = f.Write([]byte("!file\n"))
+	s.NoError(err)
+	err = f.Close()
+	s.NoError(err)
+	_, err = fs.Create("nested/ignore_dir/file")
+	s.NoError(err)
```
pjbgf, line 87: Not closing files can lead to failing tests in Windows, so ideally we would consistently close them:

```suggestion
	f, err = fs.Create("nested/ignore_dir/file")
	s.NoError(err)
	err = f.Close()
	s.NoError(err)
```

pjbgf (changes requested):
I Benchmarked main vs this PR for the Linux Kernel repo, which is quite large. The changes makes it 2.5x slower while also churning 3 times more memory:

```
BenchmarkStatus/osfs.BoundOS-16 (main)   	       1	 8356676003 ns/op 	2158831416 B/op	15644121 allocs/op
BenchmarkStatus/osfs.BoundOS-16 (this PR)	       1	20240289471 ns/op	6315067472 B/op	16515943 allocs/op
```

I appreciate that this may help some specific use cases, so if we were to go ahead this would need to be an opt-in, using the options that @onee-only [suggested](https://github.com/go-git/go-git/pull/1379#discussion_r2463264406).

Overall, I think that Status could be optimised by leaning on the caching mechanisms that upstream has such as the index extensions `TREE`, `LINK` and `UNTR`. The support to those are being added in https://github.com/go-git/go-git/pull/1622. Once merged, they will need to be put to use in the Status logic.

onee-only:
> I Benchmarked main vs this PR for the Linux Kernel repo, which is quite large. The changes makes it 2.5x slower while also churning 3 times more memory:

@pjbgf Note that this branch is currently 98 commits behind `main`. There were several PRs that improved performance as far as I remember.

I've merged `main` to this branch locally and the difference doesn't seem that large.

tested with the snippet above:
```
goos: darwin
goarch: arm64
cpu: Apple M2
         │ before.txt  │             after.txt             │
         │   sec/op    │   sec/op     vs base              │
Status-8   11.16m ± 0%   11.83m ± 1%  +6.04% (p=0.002 n=6)

         │  before.txt  │             after.txt              │
         │     B/op     │     B/op      vs base              │
Status-8   2.104Mi ± 0%   2.138Mi ± 0%  +1.64% (p=0.002 n=6)

         │ before.txt  │             after.txt             │
         │  allocs/op  │  allocs/op   vs base              │
Status-8   46.35k ± 0%   46.80k ± 0%  +0.96% (p=0.002 n=6)
```

pjbgf:
@onee-only rebasing this PR and testing it against the Linux Kernel is still 40% slower, but to your point, it is not as bad as using the current state of the PR.


```
goos: linux
goarch: amd64
cpu: AMD Ryzen 7 PRO 8840HS w/ Radeon 780M Graphics
                     │ /tmp/main  │               /tmp/ig               │
                     │   sec/op   │   sec/op     vs base                │
Status/osfs.BoundOS-16   7.846 ± 4%   11.150 ± 5%  +42.12% (p=0.000 n=10)

                     │  /tmp/main   │               /tmp/ig               │
                     │     B/op     │     B/op      vs base               │
Status/osfs.BoundOS-16   805.1Mi ± 0%   809.6Mi ± 0%  +0.56% (p=0.001 n=10)

                     │  /tmp/main  │              /tmp/ig               │
Status/osfs.BoundOS-16   15.60M ± 0%   15.61M ± 0%  +0.05% (p=0.001 n=10)
```

The benchmark is using osfs rather than memfs. We probably should add a `BenchmarkStatus` to the code base so that we all use the same going forwards - ideally running across the built-in storages as they will be impacted in different manners.

onee-only:
Hello @silkeh, If you don't mind, I'd like to take over this PR and help this proposal get merged. Please let me know if you're planning to work on this someday.

silkeh:
@onee-only: This PR is still on my radar, but I've been busy with other priorities. Feel free to take this over if you do have the time to work on it soon!

PR #1409, opened 2025-01-27
Bump go-git module to v6

pjbgf:
Bumps the go-git module to v6, in preparation for the API and behaviour changes that are soon to arrive at `main`.
It also cherry-picks all recent changes from `main`.

This PR includes some of the behaviour/API changes called out in https://github.com/go-git/go-git/issues/910.


aymanbagabas (approved):
We need to update the default agent capability string to `go-git/6.x` probably in a follow up PR

PR #1415, opened 2025-01-30
Add link to comment about the "legal issues"

jcbhmr:
Why? Because some people see "legal issues" and think "uh oh... a dark cloud is looming over this project now." Providing a link to more context to dispel the reflexive negative reaction to the words "legal issues" seems like a good idea. The GitHub comment was the best that I could find. If there's a better summary somewhere then update this link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ It's being actively developed since 2015 and is being used extensively by [Keyba
 Project Status
 --------------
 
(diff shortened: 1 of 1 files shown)


pjbgf (approved):
@jcbhmr fair point, thanks for adding the link to the historical thread here. :bow:

PR #1416, opened 2025-01-31
plumbing: packfile parser process deltaRef before deltaObj in thin pack

manland:
A delta can reference a delta-ref we need to process them before delta.

### DIFF
--- a/plumbing/format/packfile/parser.go
+++ b/plumbing/format/packfile/parser.go
@@ -117,7 +117,6 @@ func (p *Parser) Parse() (plumbing.Hash, error) {
 
 		case ObjectSection:
 			oh := data.Value().(ObjectHeader)
-
 			if oh.Type.IsDelta() {
 				if oh.Type == plumbing.OFSDeltaObject {
 					pendingDeltas = append(pendingDeltas, &oh)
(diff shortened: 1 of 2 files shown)


plumbing/format/packfile/parser.go
```diff
@@ -248,7 +247,7 @@ func (p *Parser) parentReader(parent *ObjectHeader) (io.ReaderAt, error) {
 	// If the parent is not an external ref and we don't have the
 	// content offset, we won't be able to inflate via seeking through
 	// the packfile.
-	if parent.externalRef && parent.ContentOffset == 0 {
+	if !parent.externalRef && parent.ContentOffset == 0 {
```
manland, line 250: @pjbgf I found that we looking around this code... Don't know if the comment ("not an external") or the code ("parent.externalRef") was bad... I guess it was the code but no test fail in both case!?

pjbgf: Good catch. We should expand on tests to avoid a future regression, but will merge this as-is to unblock you.

IMO, the change aligns [with the v5 code](https://github.com/go-git/go-git/blob/releases/v5.x/plumbing/format/packfile/parser.go#L402-L405) around thin packs.

pjbgf (approved):
@manland great work, thanks for working on this. :bow:

PR #1442, opened 2025-02-27
Upstream fork sync to update the fork with upstream changes

grahambrooks:


grahambrooks:
Opened in error - trying to update my fork with changes. The fork was created to support bazel build of an app using go-git. The ProtonMail dependency caused problem with building an app depending on go-git using bazel.

PR #1462, opened 2025-03-15
repository: adopt variadic options pattern for Init and PlainInit

davidalpert:
This PR refactors the options for the Repository `Init` and `PlainInit` methods to accept some variadic style options in the same style as the recently refactored [packfile.Parser](https://github.com/go-git/go-git/blob/main/plumbing/format/packfile/parser.go#L45).

This refactoring for the repository factory functions paves the way for additional options to opt-in to legacy v5 configuration handlings (see #395 and the initial work in #1019)

I am looking for feedback to confirm that this syntax and semantics are going in an aligned direction. Specifically I ran into the similarity between `Init` and `PlainInit` args and the fact that Golang does not support two functions named the same of different types (e.g. `WithDefaultBranch` might become `WithPlainInitDefaultBranch` but that seems to add no value) so I compromised and introduced a `WithInitOptions` helper that allows passing `initOption` helpers into the `plainInitOption`'s inner `initOptions`

A different approach would be to collapse all the init options into a single struct and remove `PlainInit`, replacing it with `Init` and the options that configure a bare repo.

@pjbgf I am looking for some confirmation or direction on this set of changes before I push forward to address the rest of #395

### DIFF
--- a/options.go
+++ b/options.go
@@ -791,15 +791,54 @@ type PlainOpenOptions struct {
 // Validate validates the fields and sets the default values.
 func (o *PlainOpenOptions) Validate() error { return nil }
 
-type PlainInitOptions struct {
-	InitOptions
-	// Determines if the repository will have a worktree (non-bare) or not (bare).
-	Bare         bool
-	ObjectFormat formatcfg.ObjectFormat
+type plainInitOptions struct {
+	initOptions []InitOption
+
+	bare         bool
+	objectFormat formatcfg.ObjectFormat
+}
+
+type PlainInitOption func(*plainInitOptions)
+
+// newPlainInitOptions creates a new instance of plainInitOptions with default settings.
+func newPlainInitOptions() plainInitOptions {
+	return plainInitOptions{
+		initOptions:  make([]InitOption, 0),
+		bare:         false,
+		objectFormat: "",
+	}
+}
+
+// WithInitOptions passes a set of general InitOption functions through to the shared initOptions.
+func WithInitOptions(opts ...InitOption) PlainInitOption {
+	return func(o *plainInitOptions) {
+		o.initOptions = opts
+	}
+}
+
+// WithWorkTree determines if the repository will have a worktree (non-bare) or not (bare).
+func WithWorkTree(hasWorkTree bool) PlainInitOption {
+	return func(o *plainInitOptions) {
+		o.bare = !hasWorkTree
+	}
+}
+
+// WithBare is for backwards compatibility and is the opposite of WithWorkTree; a bare repo has no worktree.
+func WithBare(isBare bool) PlainInitOption {
+	return func(o *plainInitOptions) {
+		o.bare = isBare
+	}
+}
+
+// WithObjectFormat sets the object format when invoking PlainInitWithOptions.
+func WithObjectFormat(of formatcfg.ObjectFormat) PlainInitOption {
+	return func(o *plainInitOptions) {
+		o.objectFormat = of
+	}
 }
 
 // Validate validates the fields and sets the default values.
-func (o *PlainInitOptions) Validate() error { return nil }
+func (o *plainInitOptions) Validate() error { return nil }
 
 var ErrNoRestorePaths = errors.New("you must specify path(s) to restore")
 
--- a/repository.go
+++ b/repository.go
@@ -77,31 +77,45 @@ type Repository struct {
 	wt billy.Filesystem
 }
 
-type InitOptions struct {
-	// The default branch (e.g. "refs/heads/master")
-	DefaultBranch plumbing.ReferenceName
+type initOptions struct {
+	defaultBranch plumbing.ReferenceName
+}
+
+func newInitOptions() initOptions {
+	return initOptions{
+		defaultBranch: plumbing.Master,
+	}
+}
+
+type InitOption func(*initOptions)
+
+// WithDefaultBranch sets the  default branch for the new repo (e.g. "refs/heads/master").
+func WithDefaultBranch(b plumbing.ReferenceName) InitOption {
+	return func(o *initOptions) {
+		o.defaultBranch = b
+	}
 }
 
 // Init creates an empty git repository, based on the given Storer and worktree.
 // The worktree Filesystem is optional, if nil a bare repository is created. If
 // the given storer is not empty ErrRepositoryAlreadyExists is returned
 func Init(s storage.Storer, worktree billy.Filesystem) (*Repository, error) {
-	options := InitOptions{
-		DefaultBranch: plumbing.Master,
-	}
-	return InitWithOptions(s, worktree, options)
+	return InitWithOptions(s, worktree,
+		WithDefaultBranch(plumbing.Master),
+	)
 }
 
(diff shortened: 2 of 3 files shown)


pjbgf (changes requested):
@davidalpert thank you for putting this together. I added a few comments, PTAL.

We may have some additional challenges when expanding the variadic options to `Clone`, to keep the API simple and avoid the name collision on the `With` funcs.

repository.go
```diff
@@ -77,31 +77,45 @@ type Repository struct {
...
-	return InitWithOptions(s, worktree, options)
+	return InitWithOptions(s, worktree,
+		WithDefaultBranch(plumbing.Master),
+	)
 }
 
-func InitWithOptions(s storage.Storer, worktree billy.Filesystem, options InitOptions) (*Repository, error) {
-	if err := initStorer(s); err != nil {
-		return nil, err
+func InitWithOptions(s storage.Storer, worktree billy.Filesystem, opts ...InitOption) (*Repository, error) {
```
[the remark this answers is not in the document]
davidalpert: addressed in [ef4c87e](https://github.com/go-git/go-git/pull/1462/commits/ef4c87e7145c64347f954cdbda4bd6cb393a75b6); if approved I can rebase and squash into a single commit.

options.go
```diff
@@ -791,15 +791,54 @@ type PlainOpenOptions struct {
...
-type PlainInitOptions struct {
-	InitOptions
-	// Determines if the repository will have a worktree (non-bare) or not (bare).
-	Bare         bool
-	ObjectFormat formatcfg.ObjectFormat
+type plainInitOptions struct {
+	initOptions []InitOption
+
+	bare         bool
+	objectFormat formatcfg.ObjectFormat
```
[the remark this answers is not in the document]
davidalpert: addressed in [ef4c87e](https://github.com/go-git/go-git/pull/1462/commits/ef4c87e7145c64347f954cdbda4bd6cb393a75b6); if approved I can rebase and squash into a single commit.

repository.go
```diff
@@ -77,35 +77,70 @@ type Repository struct {
...
+	}
+}
+
+// WithBare is for backwards compatibility and is the opposite of WithWorkTree; a bare repo sets the worktree to nil.
+func WithBare() InitOption {
+	return func(o *initOptions) {
+		o.workTree = nil
+	}
+}
+
```
pjbgf, line 116: I understand the idea behind having something for users to relate to. But it feels largely redundant.

```suggestion
```

repository.go
```diff
@@ -77,35 +77,70 @@ type Repository struct {
...
+		o.defaultBranch = b
+	}
+}
+
+// WithWorkTree sets the internal worktree for the new repo.
+func WithWorkTree(worktree billy.Filesystem) InitOption {
+	return func(o *initOptions) {
+		o.workTree = worktree
+	}
+}
```
pjbgf, line 108:
```suggestion
// WithWorkTree sets the worktree filesystem for the repo. If not used, or a `nil` is
// passed as argument, will result in a bare repository.
func WithWorkTree(worktree billy.Filesystem) InitOption {
	return func(o *initOptions) {
		o.workTree = worktree
	}
}
```

repository.go
```diff
@@ -77,35 +77,70 @@ type Repository struct {
...
+
+// WithBare is for backwards compatibility and is the opposite of WithWorkTree; a bare repo sets the worktree to nil.
+func WithBare() InitOption {
+	return func(o *initOptions) {
+		o.workTree = nil
+	}
+}
+
+// WithObjectFormat sets the object format when invoking PlainInitWithOptions.
```
pjbgf, line 117:
```suggestion
// WithObjectFormat sets the repository's object format.
```

repository.go
```diff
@@ -77,35 +77,70 @@ type Repository struct {
...
+
+func newInitOptions() initOptions {
+	return initOptions{
+		defaultBranch: plumbing.Master,
+		workTree:      nil,
+		objectFormat:  "",
+	}
+}
+
+type InitOption func(*initOptions)
```
pjbgf, line 94: For the purposes of this PR, we can leave this as-is. However, as we expand other parts of the code we may need to amend this so that we can enable the reuse of the same `With<Option>` across different use cases (e.g. `Init`, `Clone`, etc), without having to prefix them.

Let's leave this comment open for future reference.

pjbgf (changes requested):
@davidalpert overall LGTM.

One additional change needed is to replace the `PlainInitWithOptions` from `_examples/sha256`.

Please squash your commits and rebase (and retarget the PR) against [v6-transport](https://github.com/go-git/go-git/tree/v6-transport) branch instead? We already bumped the API version on that branch to `v6` so we would be able to merge this right away.

repository.go
davidalpert: will remote `WithBare(..)` and incorporate this comment update in my rebased PR.

given the updated comments on `WithWorkTree(..)` what is the significance of keeping the `isBare` argument in
```
func PlainInit(path string, isBare bool, options ...InitOption) (*Repository, error) {
   ...
}
```

perhaps it makes sense to remove `isBare bool` there and leave the signature as
```
func PlainInit(path string, options ...InitOption) (*Repository, error) {
   ...
}
```
so that there is one consistent syntax for a repository as both `Init` and `PlainInit` default to a bare repo unless a worktree is passed using `WithWorkTree(...)`

repository.go
davidalpert: I had the same thought here, looking ahead to the possibility of future collisions, especially after initially running into them between `Init` and `PlainInit`.

One possibility that comes to mind is establishing a policy that all `With<Option>` functions need to be in a scoped package which helps guide clients into using the right option set for the package, e.g. `repo.WithWorkTree(...)` rather than `git.WithWorkTree(...)`

repository.go
davidalpert: also, tests are failing for me locally on the `v6-transport` branch before I rebase my changes...

```
❯ make test
running against git version 2.35.1
...
--- FAIL: TestProxyEnvSuite (0.35s)
    --- FAIL: TestProxyEnvSuite/TestCommand (0.34s)
        testing.go:1399: race detected during execution of test
FAIL
FAIL	github.com/go-git/go-git/v6/internal/transport/ssh/test	2.317s
...
```

davidalpert:
- [x] rebased onto `v6-transport`
- [x] squashed into a single commit
- [x] force-pushed to this branch
- [x] retargeted this PR against the `v6-transport` branch

repository.go
davidalpert: The `--bare` option is one that may collide, for example, as both `git init --bare` and `git clone --bare` are supported by the CLI.

davidalpert:
rebased again on top of `upstream/v6-transport`

davidalpert:
@pjbgf feedback has been addressed and this PR is awaiting your review/approval.

davidalpert:
I appreciate the feedback on this one. It will help guide me as I pick up again the Config alignment that got me started.

My original interest in the init methods was to pass in a stub filesystem during tests. interestingly, the `git config` command does not require that it is run inside a repository context so I will still look to expose a way to stub the filesystem used by the config module at the module level.

PR #1463, opened 2025-03-16
plumbing: make objfile closing idempotent

christian-roggia:
We observed unexpected behavior where `objfile.Close()` is called multiple times. This is likely caused by `sync.PutZlibWriter(w.zlib)`, which can lead to race conditions and unintended side effects during fast parallel writes.

This pull request ensures that calling `Close()` multiple times is idempotent, aligning with common patterns in other `io.Writer` implementations. After the first call, subsequent calls will have no effect and return the same error, if any.

Reproduce with fast concurrent writes and the following sample code:
```golang
func Encode(objectType plumbing.ObjectType, data []byte) (plumbing.Hash, []byte, error) {
	var buffer bytes.Buffer

	writer := objfile.NewWriter(&buffer)
        defer writer.Close()

	if err := writer.WriteHeader(objectType, int64(len(data))); err != nil {
		return plumbing.ZeroHash, nil, err
	}

	if n, err := writer.Write(data); err != nil || n != len(data) {
		return plumbing.ZeroHash, nil, err
	}

	if err := writer.Close(); err != nil {
		return plumbing.ZeroHash, nil, err
	}

	return writer.Hash(), buffer.Bytes(), nil
}
```

### DIFF
--- a/plumbing/format/objfile/writer.go
+++ b/plumbing/format/objfile/writer.go
@@ -25,6 +25,8 @@ type Writer struct {
 
 	closed  bool
 	pending int64 // number of unwritten bytes
+
+	closeErr error
 }
 
 // NewWriter returns a new Writer writing to w.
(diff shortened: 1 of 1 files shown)


pjbgf (approved):

pjbgf:
@christian-roggia thanks for working on this. :bow:

PR #1465, opened 2025-03-18
plumbing: transport, Ensure file:/// support in Windows

pjbgf:
This fixes all failing tests in `v6-transport`.


aymanbagabas (commented):
On Windows, we can support the `\\server\path` scheme for a native standard. The `file:` URI scheme should _always_ use `/` slashes even on Windows. We should probably use `url.URL` or similar to encode these endpoints.

See https://en.wikipedia.org/wiki/File_URI_scheme

pjbgf:
> On Windows, we can support the `\\server\path` scheme for a native standard. The `file:` URI scheme should _always_ use `/` slashes even on Windows. We should probably use url.URL or similar to encode these endpoints.

@aymanbagabas thanks for looking into this. During the implementation I tested it and the behaviour aligned with the `git` CLI for local clones.

If we were to actively support remote servers in Windows, we would need to provide an opt-in mechanism, so that `go-git` can't be used to map the network without the application developer's consent.

PR #1469, opened 2025-03-24
fix: use "errors" package for comparaison

PascalBourdier:
Comparison with errors using equality operators fails on wrapped errors

use the golang 1.13 best practices: https://tip.golang.org/doc/go1.13#error_wrapping


plumbing/transport/file/client.go
```diff
@@ -89,13 +89,12 @@ func (r *runner) Command(cmd string, ep *transport.Endpoint, auth transport.Auth
...
 	if err != nil {
-		if e, ok := err.(*execabs.Error); ok && e.Err == execabs.ErrNotFound {
+		var e *execabs.Error
+		if errors.As(err, &e) && errors.Is(e.Err, execabs.ErrNotFound) {
 			cmd, err = prefixExecPath(cmd)
 			if err != nil {
 				return nil, err
 			}
-		} else {
-			return nil, err
```
pjbgf, line 98: Although the `else` is redundant, we still need the `return nil, err`. Please add it back.

pjbgf (changes requested):
@PascalBourdier thank you for proposing this PR. There is only a small change needed, otherwise LGTM. Please rebase your changes.

PascalBourdier:
the default branch has changed so I have to create a new PR: https://github.com/go-git/go-git/pull/1718

PR #1471, opened 2025-03-26
plumbing: fix err return in ResolveReference

PascalBourdier:
currently we can return no error if r is nil


pjbgf (approved):
@PascalBourdier thanks for working on this. :bow:

PR #1475, opened 2025-03-26
plumbing: fix err mngt in packfile processDelta

PascalBourdier:
this func could fail silently and couldn't be catch at higher level


pjbgf (approved):
@PascalBourdier thanks for fixing this. :bow:

Issue #1417, opened 2025-01-31
Re-add support for SSH @cert-authority to v5


pjbgf:
> The recently released skeema/knownhosts v1.3.0 introduced a `HostKeyDB` type that extends the `HostKeyCallback` functionality to support @cert-authority algorithms in its key `HostKeyAlgorithms` discovery functionality.

> `known_hosts` files may contain lines with @cert-authority markers to represent a certificate instead of a key. If a git remote uses cert authorities as the preferred host identification mechanism, the functionality added in skeema/knownhosts v1.3.0 is needed so that go-git can interact with this remote and verify it.

This was previously implemented on https://github.com/go-git/go-git/pull/1157 and reverted on https://github.com/go-git/go-git/pull/1346 due to https://github.com/go-git/go-git/issues/1341.

We can re-add this to v5, provided that we have enough regression testing and affected users confirm the regression is no longer in place.

cc: @Javier-varez @evanelias

Javier-varez:
I will try to find time to look at this issue over the weekend. Thank you @evanelias, I did not get notified about this issue before your mentions, but I would definitely be interested to bring this functionality back into go-git.

Javier-varez:
I finally managed to get some time to investigate issue #1341 so that I can figure out the right way to bring this feature back.

At this point I am relatively confident I have found the root cause of the issue, unfortunately it is more complex than I anticipated, so the fix will also be less straight forward than I was hoping for. I will describe the situation here, as well as a couple of potential alternatives to fix it.

The problem described in #1341 is related to the usage of the [`ssh.FixedHostKey`](https://github.com/golang/crypto/blob/d0a798f774735c176ed0d3500ac986957a02660f/ssh/client.go#L269) `HostKeyCallback` together with the [`PublicKeys`](https://github.com/go-git/go-git/blob/ec133065ea1201a4488bbc793ccd9337b6c06877/plumbing/transport/ssh/auth_method.go#L121) `AuthMethod` of `go-git`. I will examine how this is handled both without the implementation of #1157 and with it.

# Using the `PublicKeys` auth method with a custom `HostKeyCallback` before the introduction of #1157 

The following code overrides the configuration of the `HostKeyCallback` in `PublicKeys`.

```go 
	pk, err := ssh2.NewPublicKeys("git", []byte(privateKey), "")
	if err != nil {
		return "", fmt.Errorf("unable to create git public key: %s", err)
	}
[...]

PR #1482, opened 2025-03-31
plumbing: add cert auth support to `releases/v5.x`

Javier-varez:
Re-adds support for cert authorities using skeema/knownhosts. 
It performs two separate changes: 
- Simplify the `AuthMethod`s provided by go-git so that they populate HostKeyCallback and HostKeyAlgorithms, while still allowing the user to inject a custom HostKeyCallback. If the HostKeyCallback is empty, then the `HostKeyCallback` and the `HostKeyAlgorithms` are set using the knownhosts DB functionality, which now supports cert authorities.
- Add an option so that a user of the default `AuthMethod`s provided by go-git can set a custom `HostKeyAlgorithms` when they use a custom `HostKeyCallback`. This is needed for usages of `ssh.FixedHostKey`.

fixes 1417

### DIFF
--- a/plumbing/transport/ssh/auth_method.go
+++ b/plumbing/transport/ssh/auth_method.go
@@ -230,11 +230,23 @@ func (a *PublicKeysCallback) ClientConfig() (*ssh.ClientConfig, error) {
 //	~/.ssh/known_hosts
 //	/etc/ssh/ssh_known_hosts
 func NewKnownHostsCallback(files ...string) (ssh.HostKeyCallback, error) {
-	kh, err := newKnownHosts(files...)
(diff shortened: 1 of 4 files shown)


pjbgf:
Relates to https://github.com/go-git/go-git/issues/1417.

runxiyu:
This does seem to be a breaking change though

pjbgf:
@Javier-varez These changes have introduced two regressions:
1. `SetHostKeyCallback` was removed.
2. In some cases known_hosts are no longer being considered.

The linked [PR](https://github.com/go-git/go-git/pull/1515) aim to address both regressions. PTAL if you have some time.

Issue #1455, opened 2025-03-12
`SparseCheckoutDirectories` works only for 1st level directory.


exu:
**Describe the bug**

The `SparseCheckoutDirectories` works only for top level directories (here `test`). 
It doesn't work for more than one directory (here `test/k6`). 

**To Reproduce**

Run the test: 

```go
package server

import (
	"io/fs"
	"os"
	"path/filepath"
	"strings"
	"testing"

	"github.com/go-git/go-git/v5"
	"github.com/go-git/go-git/v5/plumbing"
)

func TestGit(t *testing.T) {
	path := "test/k6"

	dir, err := os.MkdirTemp("", "discovery")
	if err != nil {
		t.Fatal("can't create temp dir")
	}

	r, err := git.PlainClone(dir, false, &git.CloneOptions{
		URL:               "https://github.com/kubeshop/testkube.git",
		NoCheckout:        true,
		SingleBranch:      true,
		Depth:             1,
		Tags:              git.NoTags,
		ShallowSubmodules: true,
		ReferenceName:     plumbing.ReferenceName("refs/heads/main"),
	})
	if err != nil {
		t.Fatal("can't clone repository")
	}

	w, err := r.Worktree()
	if err != nil {
		t.Fatal("can't get a worktree")
	}

	options := &git.CheckoutOptions{
		Branch: plumbing.ReferenceName("refs/remotes/origin/main"),
	}
	if path != "" {
		options.SparseCheckoutDirectories = []string{"test/k6"}
	}

	err = w.Checkout(options)
	if err != nil {
		t.Fatal("can't checkout repository")
	}

	pathCheckedOut := false
	filepath.Walk(dir, func(f string, info fs.FileInfo, err error) error {
		t.Logf("- %s", f)
		if err != nil {
			return err
		}

		if strings.Contains(f, "test/k6") {
			pathCheckedOut = true
		}

		return nil
	})

	if !pathCheckedOut {
		t.Fatal("path not checked out")
	}
}
```

**Expected behavior**
Sparse checkout should check out only path which contains passed parameter.

patricsss:
I found the bug. Working on a fix.

patricsss:
https://github.com/go-git/go-git/pull/1484

PR #1484, opened 2025-04-02
utils: fix diff so subpaths work for sparse checkouts, fixes 1455

patricsss:
# Summary

Fix for https://github.com/go-git/go-git/issues/1455

When attempting to check out anything that isn't a root directory the Sparse Checkout options do not work correctly. This is reproducible with the examples in the above issue. 

# Problem 
"virtual" nodes are created in the tree that represent directories. The first time a file is added to the tree it's parts are split and any missing nodes (including any directories) are added to the tree. 

For example, if the first added file is `foo/bar/file.yaml` and this file should be skipped, entries are created for `foo`, `foo/bar`, and `foo/bar/file.yaml`, with all three entries setting `SkipWorktree` to `true`. If `foo/baz/file.yaml` is subsequently added, but should _not_ be skipped, this is the current bug. `foo` already exists in the tree, and had `SkipWorktree` set to `true`. Even though `foo/baz` and `foo/baz/file.yaml` will have skip to false, they will never be included in the diff because their parent `foo` was already marked as skipped. 

# Fix
This change makes it so that any existing parents are re-checked and have their `SkipWorktree` boolean reset when _any_ child of that subtree has `SkipWorktree` set to false. 

The final issue is in the `addRecursive()` method. It doesn't take into consideration the `SkipWorktree` flag via the `Skip()` interface. It now checks this condition and does not add any subtrees which were marked as `Skip() == true`

# Testing
I tested the changes with the reproduction case provided as well as through a simple unit test that attempts to highlight the issue and solution. 

Open to any suggestions.

### DIFF
--- a/utils/merkletrie/index/node_test.go
+++ b/utils/merkletrie/index/node_test.go
@@ -2,7 +2,7 @@ package index
 
 import (
 	"bytes"
-	"path/filepath"
+	"path"
 	"testing"
 
 	"github.com/go-git/go-git/v5/plumbing"
@@ -47,14 +47,14 @@ func (s *NoderSuite) TestDiff() {
 func (s *NoderSuite) TestDiffChange() {
 	indexA := &index.Index{
 		Entries: []*index.Entry{{
-			Name: filepath.Join("bar", "baz", "bar"),
+			Name: path.Join("bar", "baz", "bar"),
 			Hash: plumbing.NewHash("8ab686eafeb1f44702738c8b0f24f2567c36da6d"),
 		}},
 	}
 
 	indexB := &index.Index{
 		Entries: []*index.Entry{{
-			Name: filepath.Join("bar", "baz", "foo"),
+			Name: path.Join("bar", "baz", "foo"),
 			Hash: plumbing.NewHash("8ab686eafeb1f44702738c8b0f24f2567c36da6d"),
 		}},
 	}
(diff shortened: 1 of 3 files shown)


utils/merkletrie/index/node_test.go
```diff
@@ -64,6 +64,43 @@ func (s *NoderSuite) TestDiffChange() {
...
+		a, err := c.Action()
+		s.NoError(err)
+		switch a {
+		case merkletrie.Delete:
+			deleteCount++
+		case merkletrie.Insert:
+			insertCount++
+		}
+	}
+	s.Equal(1, deleteCount)
```
onee-only, line 100: The test isn't passing. 
I think the `deleteCount` should be 0, since it is marked as skipped.

patricsss: Thanks I'll take a look. Weirdly it passed in my local branch before I pushed upstream, but without the checks running it was hard to debug what else might need to be updated. Which tests failed depended on whether I ran them from Windows or Linux so it was hard to identify my own environment misconfiguration vs actual failures. 

Will look into this and push some test updates. Thanks for the quick response!

patricsss: Yeah it passes on windows, but when I run the test in WSL it fails...

Ubuntu24: 

```
go test ./utils/merkletrie/index
--- FAIL: TestNoderSuite (0.00s)
    --- FAIL: TestNoderSuite/TestDiffSkipIssue1455 (0.00s)
        node_test.go:87:
                Error Trace:    /mnt/c/Users/Miri/Projects/go-git/utils/merkletrie/index/node_test.go:87
                Error:          "[<Insert bar/biz/bat>]" should have 2 item(s), but has 1
                Test:           TestNoderSuite/TestDiffSkipIssue1455
        node_test.go:100:
                Error Trace:    /mnt/c/Users/Miri/Projects/go-git/utils/merkletrie/index/node_test.go:100
                Error:          Not equal:
                                expected: 1
                                actual  : 0
                Test:           TestNoderSuite/TestDiffSkipIssue1455
FAIL
FAIL    github.com/go-git/go-git/v5/utils/merkletrie/index      0.004s
FAIL
```

Windows: 

```
=== RUN   TestNoderSuite
=== RUN   TestNoderSuite/TestDiff
--- PASS: TestNoderSuite/TestDiff (0.00s)
=== RUN   TestNoderSuite/TestDiffChange
--- PASS: TestNoderSuite/TestDiffChange (0.00s)
=== RUN   TestNoderSuite/TestDiffDir
--- PASS: TestNoderSuite/TestDiffDir (0.00s)
=== RUN   TestNoderSuite/TestDiffSameRoot
--- PASS: TestNoderSuite/TestDiffSameRoot (0.00s)
=== RUN   TestNoderSuite/TestDiffSkipIssue1455
--- PASS: TestNoderSuite/TestDiffSkipIssue1455 (0.00s)
--- PASS: TestNoderSuite (0.00s)
PASS
```

patricsss: Well as a temporary solution since the only thing we actually care to test is there is exactly one insertion I updated the tests to only check that the insertion count is `1`. 

That said, it's _very_ bizarre that the behavior is different across operating systems. Curious if you know why that might be? If not, no worries.

onee-only: I think the problem is that `filepath.Join()` on the test uses 'os-specific' seperator to join the strings.
I believe we can simply change it into `path.Join()` .

patricsss: Yep, that was the problem. Done.

patricsss:
@onee-only all tests pass. I re-based all the commits into a single one which I believe adheres to the comment format. Please LMK if you need anything else. Otherwise if you can r+ and merge it'd be greatly appreciate! :)

pjbgf (approved):

pjbgf:
@patricsss thank you for working on this. :+1:  @onee-only thank you for the initial reviews. :bow:

PR #1486, opened 2025-04-06
Introduce a no-merges traversal option

MoralCode:
This PR is the result of a fix I made to solve a personal usecase in which i needed to be able to view the git log without seeing the contents of merged commits (since they seem to intersperse themselves into the order of commits).

I eventually found that the `--first-parent` option to `git log` from https://stackoverflow.com/a/39586986 was the perfect solution, however I did not see an existing option in the library that enables me to traverse the log this way. With search results leading me to a few mentions of this flag, none of which proposed an alternative way to achieve this, I decided to build my own.

This PR represents a cleaned-up version of my hacked-together changes.

Since I did not see any existing options objects passed into the existing iterators, I decided to copy paste the one that was closest to my usecase and modify it. I realize that this now introduces the issue where the other traversal methods probably also need their own first-parent versions as well, however that seemed like it was both quite out of scope for my personal usecase and also would likely require discussion to determine if theres a better way to handle this in the first place.

### DIFF
--- a/plumbing/object/commit_walker.go
+++ b/plumbing/object/commit_walker.go
@@ -184,6 +184,74 @@ func (w *commitPostIterator) ForEach(cb func(*Commit) error) error {
 
 func (w *commitPostIterator) Close() {}
 
+type commitPostIteratorNM struct {
+	stack []*Commit
+	seen  map[plumbing.Hash]bool
+}
+
+// NewCommitPostorderIter returns a CommitIter that walks the commit
+// history like WalkCommitHistory but in post-order. This means that after
+// walking a merge commit, the merged commit will be walked before the base
+// it was merged on. This can be useful if you wish to see the history in
+// chronological order. Ignore allows to skip some commits from being iterated.
+func NewCommitPostorderIterNoMerge(c *Commit, ignore []plumbing.Hash) CommitIter {
+	seen := make(map[plumbing.Hash]bool)
+	for _, h := range ignore {
+		seen[h] = true
+	}
+
+	return &commitPostIteratorNM{
+		stack: []*Commit{c},
+		seen:  seen,
+	}
+}
+
+func (w *commitPostIteratorNM) Next() (*Commit, error) {
+	for {
+		if len(w.stack) == 0 {
+			return nil, io.EOF
+		}
+
+		c := w.stack[len(w.stack)-1]
+		w.stack = w.stack[:len(w.stack)-1]
+
+		if w.seen[c.Hash] {
+			continue
+		}
+
+		w.seen[c.Hash] = true
+		// TODO: change this
+		return c, c.Parents().ForEach(func(p *Commit) error {
+			w.stack = append(w.stack, p)
+			return nil
+		})
+	}
+}
+
+func (w *commitPostIteratorNM) ForEach(cb func(*Commit) error) error {
+	for {
+		c, err := w.Next()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return err
+		}
+
+		err = cb(c)
+		if err == storer.ErrStop {
+			break
+		}
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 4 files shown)


plumbing/object/commit_walker.go
```diff
@@ -184,6 +184,76 @@ func (w *commitPostIterator) ForEach(cb func(*Commit) error) error {
...
+			}
+			return nil
+		})
+	}
+}
+
+func (w *commitPostIteratorFirstParent) ForEach(cb func(*Commit) error) error {
+	for {
+		c, err := w.Next()
+		if err == io.EOF {
```
pjbgf, line 236:
```suggestion
		if errors.Is(err, io.EOF) {
```

plumbing/object/commit_walker.go
```diff
@@ -184,6 +184,76 @@ func (w *commitPostIterator) ForEach(cb func(*Commit) error) error {
...
+		return c, c.Parents().ForEach(func(p *Commit) error {
+			if len(c.ParentHashes) > 0 && p.Hash == c.ParentHashes[0] {
+				w.stack = append(w.stack, p)
+			}
+			return nil
+		})
+	}
+}
+
+func (w *commitPostIteratorFirstParent) ForEach(cb func(*Commit) error) error {
```
pjbgf, line 233: This entire `func` has the same logic as `func (w *commitPostIterator) ForEach(cb func(*Commit) error) error`. Please extract the core of it into a helper function and call it from both places instead.

plumbing/object/commit_walker.go
```diff
@@ -184,6 +184,76 @@ func (w *commitPostIterator) ForEach(cb func(*Commit) error) error {
...
+type commitPostIteratorFirstParent struct {
+	stack []*Commit
+	seen  map[plumbing.Hash]bool
+}
+
+// NewCommitPostorderIter returns a CommitIter that walks the commit
+// history like WalkCommitHistory but in post-order. This means that after
+// walking a merge commit, the merged commit will be walked before the base
+// it was merged on. This can be useful if you wish to see the history in
+// chronological order. Ignore allows to skip some commits from being iterated.
```
pjbgf, line 196:
```suggestion
// NewCommitPostorderIterFirstParent returns a CommitIter that walks the commit
// history like WalkCommitHistory but in post-order. 
// 
// This option gives a better overview when viewing the evolution of a particular
// topic branch, because merges into a topic branch tend to be only about
// adjusting to updated upstream from time to time, and this option allows
// you to ignore the individual commits brought in to your history by such
// a merge.
// 
// Ignore allows to skip some commits from being iterated.
```

pjbgf (changes requested):
@MoralCode thank you for your first contribution. :bow: Apart from the tiny nits below, this LGTM.

Once you made the changes below, please squash the commits of this PR.

plumbing/object/commit_walker.go
MoralCode: All of them seem to implement ForEach the same way - should I refactor all of them? It kinda feels a bit out of scope for this PR to touch literally all of the other commit walkers as well

plumbing/object/commit_walker.go
MoralCode: This would require importing errors and (ideally) changing all the other `ForEach` implementations to do this check in the same way. Is it okay for me to make such a refactor without much knowledge of how this may affect other parts of the codebase?

plumbing/object/commit_walker.go
MoralCode: It feels focusing specifically on back-merges from main branches into topic branches may be too specific - This change basically returns only commits that are directly on the branch that is being traversed. This essentially means that, regardless of the selected branch (be that main or a feature branch), when a merge commit is encountered, only the merge commit itself and the base commit (i.e. the last shared commit before the branches diverged from main) will be traversed.

MoralCode:
How do things look now (other than not squashing things)?

pjbgf (approved):
@MoralCode thank you for contributing this. :bow:

PR #1487, opened 2025-04-06
[v6] plumbing: fix, prevent double decoding of commands and flush

aymanbagabas:
This refactors the decoding of commands and capabilities in the update-requests decoder to avoid double decoding the commands and eating the next pktlines in the reader.

The issue was that the decoder doesn't stop once it reaches a flush packet and continues to read the next line. This is incorrect as an update-requests always expects a flush packet after the commands have been sent.

### DIFF
--- a/options.go
+++ b/options.go
@@ -286,7 +286,7 @@ type PushOptions struct {
 	// ForceWithLease allows a force push as long as the remote ref adheres to a "lease"
 	ForceWithLease *ForceWithLease
 	// PushOptions sets options to be transferred to the server during push.
-	Options map[string]string
+	Options []string
 	// Atomic sets option to be an atomic push
 	Atomic bool
 	// ProxyOptions provides info required for connecting to a proxy.
--- a/plumbing/transport/common.go
+++ b/plumbing/transport/common.go
@@ -131,7 +131,7 @@ type PushRequest struct {
 	Progress sideband.Progress
 
 	// Options is a set of push-options to be sent to the server during push.
-	Options map[string]string
+	Options []string
 
 	// Atomic indicates an atomic push.
 	// If the server supports atomic push, it will update the refs in one
--- a/plumbing/protocol/packp/pushopts.go
+++ b/plumbing/protocol/packp/pushopts.go
@@ -1,0 +1,50 @@
+package packp
+
+import (
+	"io"
+
+	"github.com/go-git/go-git/v6/plumbing/format/pktline"
+)
+
+// PushOptions represents a list of update request push-options.
+//
+// See https://git-scm.com/docs/gitprotocol-pack#_reference_update_request_and_packfile_transfer
+type PushOptions struct {
+	Options []string
+}
+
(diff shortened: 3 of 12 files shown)


plumbing/transport/common.go
```diff
@@ -131,7 +131,7 @@ type PushRequest struct {
 	Progress sideband.Progress
 
 	// Options is a set of push-options to be sent to the server during push.
-	Options map[string]string
+	Options []string
```
pjbgf, line 134: As per upstream [docs](https://git-scm.com/docs/git-push#Documentation/git-push.txt-pushpushOption):

> This is a multi-valued variable, and an empty value can be used in a higher priority configuration file... to clear the values inherited from a lower priority configuration files

A map here would make that overwritting logic more straight-foward.

options.go
```diff
@@ -286,7 +286,7 @@ type PushOptions struct {
 	// ForceWithLease allows a force push as long as the remote ref adheres to a "lease"
 	ForceWithLease *ForceWithLease
 	// PushOptions sets options to be transferred to the server during push.
-	Options map[string]string
+	Options []string
```
pjbgf, line 289: Similarly, a map/kv pair here would make it more straight-forward for handling keys that must overwrite options from the config.

pjbgf (commented):
@aymanbagabas thanks for working on this. Do you mind clarifying the motivation on replacing the option map with an array? If looks like keeping the map would make it easier to mirror the upstream behaviour of surfacing those options from the configs and overriding them with any passed on push-option args.

plumbing/transport/common.go
aymanbagabas: This is only applicable when no `git push --push-option` are given.

In that case, options should be read from the repository and global config files. For example, to clear config options described in `/etc/gitconfig`, you would define new options in `.git/config` of the repository and use `push-options = ` to clear the global ones.

See https://git-scm.com/docs/git-push#Documentation/git-push.txt---push-optionltoptiongt

Which means when you use `git push --push-option` you can send duplicates and any string, see https://github.com/git/git/blob/485f5f863615e670fd97ae40af744e14072cfe18/send-pack.c#L375

options.go
aymanbagabas: Duplicates should be allowed. In case of no options are given, we should derive them from config files.

pjbgf: My point was regarding [this](https://git-scm.com/docs/git-push#Documentation/git-push.txt-pushpushOption):

> This is a multi-valued variable, and an empty value can be used in a higher priority configuration file (e.g. .git/config in a repository) to clear the values inherited from a lower priority configuration files (e.g. $HOME/.gitconfig).

However, that is orthogonal to the actual transport-level behaviour, so happy for us to transition the type here and handle the config options separately.

plumbing/transport/common.go
pjbgf: Fair point, thanks for clarifying.

pjbgf (approved):

Issue #1460, opened 2025-03-14
`Worktree.Add()` silently fails for absolute paths


msuozzo:
Repro:

```
repo := must(git.Init(memory.NewStorage(), memfs.New()))
w := must(repo.Worktree())
must(w.Filesystem.Create("/foo"))
must(w.Add("/foo"))  // No error!
fmt.Println(must(w.Status()).String())  // "?? foo" -> /foo is untracked
```

I suspect the fix is to localize the Add() path before updating the index.

pjbgf:
Had a quick look at this, and the issue only occurs when an absolute path is passed to `Worktree.Add`. This is filesystem/storer agnostic. The first two use cases on this tests work fine:

```golang
func TestInconsistentAddStatus(t *testing.T) {
	tests := []struct {
		create string
		add    string
	}{
		{create: "foo", add: "foo"},
		{create: "/foo", add: "foo"},
		{create: "/foo", add: "/foo"},
		{create: "foo", add: "/foo"},
	}

	for i, tc := range tests {
		t.Run(fmt.Sprintf("#%d %q %q", i, tc.create, tc.add), func(t *testing.T) {
			repo, err := Init(memory.NewStorage(), memfs.New())
			require.NoError(t, err)

			w, err := repo.Worktree()
			require.NoError(t, err)

			_, err = w.Filesystem.Create(tc.create)
			require.NoError(t, err)

			h, err := w.Add(tc.add)
			require.NoError(t, err)
			assert.NotEqual(t, plumbing.ZeroHash, h)

			st, err := w.Status()
			require.NoError(t, err)

			got := st.String()
			assert.Equal(t, fmt.Sprintf("A  %s\n", tc.add), got)
		})
	}
}
```

> I suspect the fix is to localize the Add() path before updating the index.

@msuozzo I think the entry name needs to be canonised in [addOrUpdateFileToIndex](https://github.com/go-git/go-git/blob/adcc06bf122076146fefe6514c724fc4d7351229/worktree_status.go#L535).

tmchow:
Implemented in https://github.com/go-git/go-git/pull/1949 - converts absolute paths to worktree-relative before updating the index.

PR #1491, opened 2025-04-07
fix: canonicalize pathnames when calling Worktree.Add()

0xDEC0DE:
Ensure that adding files off the filesystem root gets the filenames right.

Fixes: Issue #1460

### DIFF
--- a/worktree_status.go
+++ b/worktree_status.go
@@ -370,7 +370,11 @@ func (w *Worktree) doAdd(path string, ignorePattern []gitignore.Pattern, skipSta
 		}
 	}
 
+	// Relative paths only; see: https://github.com/go-git/go-git/issues/1460
 	path = filepath.Clean(path)
+	if filepath.IsAbs(path) {
+		path, _ = filepath.Rel(w.Filesystem.Root(), path)
+	}
 
 	if err != nil || !fi.IsDir() {
 		added, h, err = w.doAddFile(idx, s, path, ignorePattern)
--- a/worktree_test.go
+++ b/worktree_test.go
@@ -1529,7 +1529,7 @@ func (s *WorktreeSuite) TestAddUntracked() {
 	err = util.WriteFile(w.Filesystem, "foo", []byte("FOO"), 0755)
 	s.NoError(err)
 
-	hash, err := w.Add("foo")
+	hash, err := w.Add("/foo")
 	s.Equal("d96c7efbfec2814ae0301ad054dc8d9fc416c9b5", hash.String())
 	s.NoError(err)
 


worktree_status.go
```diff
@@ -370,7 +370,11 @@ func (w *Worktree) doAdd(path string, ignorePattern []gitignore.Pattern, skipSta
 		}
 	}
 
+	// Relative paths only; see: https://github.com/go-git/go-git/issues/1460
 	path = filepath.Clean(path)
+	if filepath.IsAbs(path) {
+		path, _ = filepath.Rel(w.Filesystem.Root(), path)
+	}
```
pjbgf, line 377: I believe this is not compatible with [BoundOS](https://github.com/go-git/go-billy/blob/main/osfs/os_bound.go#L48), and unfortunately `go-billy` does not expose an os-aware Rel func.

0xDEC0DE: Correct, it fails spectacularly under BoundOS.  Yikes.

But `Rel` seems to work OK, I'll push another attempt at a fix for your consideration.

worktree_status.go
```diff
@@ -370,7 +370,12 @@ func (w *Worktree) doAdd(path string, ignorePattern []gitignore.Pattern, skipSta
 		}
 	}
 
+	// Relative paths only; see: https://github.com/go-git/go-git/issues/1460
 	path = filepath.Clean(path)
+	if filepath.IsAbs(path) {
+		root := w.Filesystem.Root()
+		path, _ = filepath.Rel(root, filepath.Join(root, ".", path))
+	}
```
0xDEC0DE, line 378: Hmm.  This works on OSX and Linux, and fails on Windows.  I don't have access to any Windows boxes to figure out what's going wrong, can anybody else add any colour here?

worktree_status.go
```diff
@@ -370,7 +370,13 @@ func (w *Worktree) doAdd(path string, ignorePattern []gitignore.Pattern, skipSta
 		}
 	}
 
+	// Relative paths only; see: https://github.com/go-git/go-git/issues/1460
+	// filepath.IsAbs is hilariously bugged on Windows, check by hand
 	path = filepath.Clean(path)
+	if string(path[0]) == string(filepath.Separator) {
+		root := w.Filesystem.Root()
+		path, _ = filepath.Rel(root, filepath.Join(root, ".", path))
+	}
```
pjbgf, line 379: @0xDEC0DE thanks for looking into this. This logic seems better suited in [go-billy](https://github.com/go-git/go-billy/), potentially as an `util` that receives a `Filesystem` and a path, returning a canonical path if the file exists.

We'd need to test this (in `go-billy`) for all built-in FS types.

0xDEC0DE: Not entirely sure I understand, as the use case here is creating a new file; but there's already `util.getUnderlyingAndPath`, which I suppose could use an exported version, would that be sufficient for this?

PR #1493, opened 2025-04-07
plumbing: correct matching of file vs. dir globs in `.gitignore`, Fixes #154

### LINKED ISSUE #154: The pattern "/a*/**" in .gitignore should not match a file "ab" in the project root
We use go-git in a project, and ran across the issue that a modified file already under version control was not committed when performing:

```golang
worktree.Commit("...", &git.CommitOptions{All: true, ...}})
```
I've tracked this down to a pattern contained in the `.gitignore` of the project in the form of "/a*/**". This should match all files contained in a **folder** that startes with "a" or any of it's sub-folders. The pattern matching of go-git also matches a **file** starting with "a" contained in the project root. This is not the case with git itself. I've opened the following PR to provide a test-case:

https://github.com/go-git/go-git/pull/153

0xDEC0DE:
e.g., the pattern `/a*/**` in `.gitignore` should not match a file `ab` in the project root.

Fix a broken test that was enforcing the incorrect behavior.

Fixes: Issue #154

### DIFF
--- a/plumbing/format/gitignore/pattern.go
+++ b/plumbing/format/gitignore/pattern.go
@@ -146,6 +146,10 @@ func (p *pattern) globMatch(path []string, isDir bool) bool {
 			}
 			matched = true
 			path = path[1:]
+			// files matching dir globs, don't match
+			if len(path) == 0 && i < len(p.pattern)-1 {
+				matched = false
+			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


pjbgf (approved):
@0xDEC0DE thanks for working on this. :bow:

PR #1498, opened 2025-04-08
[v6] Add a flush line after sideband

OlshaMB:

PR #1502, opened 2025-04-10
[V6] HTTP middleware

aymanbagabas:
This adds a generic git http middleware to be used with to serve git
repositories over http using the go standard http library.
The handler takes a loader to load the repositories from storage and
an optional error logger.

The API is still WIP and open to change. Feedbacks are welcome!

This needs critical fixes from https://github.com/go-git/go-git/pull/1498, https://github.com/go-git/go-git/pull/1488, and https://github.com/go-git/go-git/pull/1487


plumbing/http/http.go
```diff
@@ -0,0 +1,327 @@
...
+				},
+			)
+		case transport.ReceivePackService:
+			err = transport.ReceivePack(ctx, st, nil, ioutil.WriteNopCloser(w),
+				&transport.ReceivePackOptions{
+					GitProtocol:   version,
+					AdvertiseRefs: true,
+					StatelessRPC:  true,
+				},
+			)
```
OlshaMB, line 278: What if the service isn't either of those

plumbing/http/http.go
```diff
@@ -0,0 +1,327 @@
...
+	if !ok {
+		logf(errorLog, "http writer must implement http.Flusher interface")
+		// The writer must implement the Flusher interface which the default
+		// http.ResponseWriter does.
+		renderStatusError(w, http.StatusBadRequest)
+		return
+	}
+
+	// Write the response to the client in chunks.
+	p := make([]byte, 1024)
```
OlshaMB, line 194: Should we still chunk it when we can return the sideband error?
My answer is no, we should answer from the start

plumbing/http/http.go
```diff
@@ -0,0 +1,327 @@
...
+
+type contextKey string
+
+type service struct {
+	method  string
+	handler http.HandlerFunc
+	svc     transport.Service
+}
+
+var services = map[string]service{
```
OlshaMB, line 31: Maybe move services as a default into config? We should be able to add/remove services.

plumbing/http/http.go
```diff
@@ -0,0 +1,327 @@
...
+	hdrCacheForever(w)
+	sendFile(w, r, "application/x-git-packed-objects")
+}
+
+func getIdxFile(w http.ResponseWriter, r *http.Request) {
+	hdrCacheForever(w)
+	sendFile(w, r, "application/x-git-packed-objects-toc")
+}
+
+func renderStatusError(w http.ResponseWriter, code int) {
```
OlshaMB, line 311: You should make this be passed from config and have this as default value

plumbing/transport/receive_pack.go
```diff
@@ -34,7 +34,7 @@ func ReceivePack(
 	w io.WriteCloser,
 	opts *ReceivePackOptions,
 ) error {
-	if r == nil || w == nil {
+	if w == nil {
 		return fmt.Errorf("nil reader or writer")
```
OlshaMB, line 38:
```suggestion
		return fmt.Errorf("nil writer")
```

OlshaMB (changes requested):
Looks great! It needs ability to inject auth, e.g. check write permissions.

plumbing/http/http.go
aymanbagabas: Good catch, we should return an error

plumbing/http/http.go
aymanbagabas: This is meant to implement the default services and be sort of a replacement for `git-http-backend`. It's just a middleware, other services can be defined as handlers and wrap this one.

plumbing/http/http.go
OlshaMB: ok. Can we maybe do it concurently? (Just a question)

plumbing/http/http.go
OlshaMB: Ok, what about disabling dumb protocol routes?

plumbing/http/http.go
OlshaMB: Ok, agreed

aymanbagabas:
> Looks great! It needs ability to inject auth, e.g. check write permissions.

Auth can be implemented using a custom `transport.Loader` or as a separate middleware layer.

OlshaMB:
> > Looks great! It needs ability to inject auth, e.g. check write permissions.
> 
> Auth can be implemented using a custom `transport.Loader` or as a separate middleware layer.

I don't want to check receive-pack path twice

plumbing/http/http.go
aymanbagabas: I'm not sure, honestly. HTTP1 only handles one stream at a time, but using HTTP2 _may_ benefit from this since it can handle multiple streams at once.

OlshaMB (approved):
LGTM!

OlshaMB:
> > > Looks great! It needs ability to inject auth, e.g. check write permissions.
> > 
> > 
> > Auth can be implemented using a custom `transport.Loader` or as a separate middleware layer.
> 
> I don't want to check receive-pack path twice

We can discuss it later when i use this, i will see how it would work

OlshaMB (approved):

pjbgf (changes requested):
This is looking good @aymanbagabas. Thanks for working on it. Please add some tests around the `http.go` code.

plumbing/http/http.go
```diff
@@ -0,0 +1,334 @@
...
+		logger.Printf(format, v...)
+	}
+}
+
+// Handler returns a new HTTP handler that serves git repositories over HTTP.
+// It uses the given loader to load the repositories from storage.
+func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	h.init()
+	for pattern, service := range services {
+		pat := regexp.MustCompile(pattern)
```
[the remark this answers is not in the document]
aymanbagabas: Good call

plumbing/transport/receive_pack.go
```diff
@@ -64,6 +64,10 @@ func ReceivePack(
 		return nil
 	}
 
+	if r == nil {
+		return fmt.Errorf("nil reader")
+	}
+
```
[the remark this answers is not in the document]
aymanbagabas: The nil check is only needed if we're not advertising references in a stateless session

OlshaMB (approved):

pjbgf (approved):
@aymanbagabas thanks for working on this. :bow:

backend/http/http.go
```diff
@@ -0,0 +1,345 @@
...
+// [transport.UpdateServerInfo] before serving the repository.
+type Handler struct {
+	// Loader is the loader used to load repositories from storage. If nil, the
+	// default loader is used.
+	Loader transport.Loader
+	// ErrorLog is the logger used to log errors. If nil, no errors are logged.
+	ErrorLog *log.Logger
+	// Prefix is a path prefix that will be stripped from the URL path before
+	// matching the service patterns.
+	Prefix string
```
OlshaMB, line 70: What if I want postfix, maybe make a function?

backend/http/http.go
```diff
@@ -0,0 +1,345 @@
...
+// implement the [storer.FilesystemStorer] interface. Keep in mind that
+// repositories that wish to be server using the Dumb-HTTP protocol must update
+// the server info files. This can be done by using
+// [transport.UpdateServerInfo] before serving the repository.
+type Handler struct {
+	// Loader is the loader used to load repositories from storage. If nil, the
+	// default loader is used.
+	Loader transport.Loader
+	// ErrorLog is the logger used to log errors. If nil, no errors are logged.
+	ErrorLog *log.Logger
```
OlshaMB, line 67: Maybe make error logging optional?

backend/http/http.go
```diff
@@ -0,0 +1,345 @@
...
+		reader = r.Body
+	}
+
+	frw := &flushResponseWriter{ResponseWriter: w, log: errorLog}
+
+	switch svc {
+	case transport.UploadPackService:
+		err = transport.UploadPack(ctx, st, reader, frw,
+			&transport.UploadPackOptions{
+				GitProtocol:   version,
```
OlshaMB, line 182: Should we verify the version before passing it? I know it's verified inside, but what if something changes inside?

backend/http/http.go
```diff
@@ -0,0 +1,345 @@
...
+		return
+	}
+	fs := fss.Filesystem()
+	f, err := fs.Open(file)
+	if err != nil {
+		renderStatusError(w, http.StatusNotFound)
+		return
+	}
+
+	defer f.Close() //nolint:errcheck
```
OlshaMB, line 236: Nitpick: defer should be near the line before, but with an empty line after

backend/http/http.go
```diff
@@ -0,0 +1,345 @@
...
+func getPackFile(w http.ResponseWriter, r *http.Request) {
+	hdrCacheForever(w)
+	sendFile(w, r, "application/x-git-packed-objects")
+}
+
+func getIdxFile(w http.ResponseWriter, r *http.Request) {
+	hdrCacheForever(w)
+	sendFile(w, r, "application/x-git-packed-objects-toc")
+}
+
```
OlshaMB, line 328: Maybe make factory function?

backend/http/http.go
```diff
@@ -0,0 +1,345 @@
...
+	w.Header().Set("Pragma", "no-cache")
+	w.Header().Set("Cache-Control", "no-cache, max-age=0, must-revalidate")
+}
+
+func hdrCacheForever(w http.ResponseWriter) {
+	now := time.Now()
+	expires := now.Add(365 * 24 * time.Hour)
+	w.Header().Set("Date", now.Format(http.TimeFormat))
+	w.Header().Set("Expires", expires.Format(http.TimeFormat))
+	w.Header().Set("Cache-Control", "public, max-age=31536000")
```
OlshaMB, line 344: Maybe move out max-age into a variable?

backend/http/writer.go
```diff
@@ -0,0 +1,62 @@
...
+	http.ResponseWriter
+	log *log.Logger
+}
+
+// ReadFrom implements io.ReaderFrom interface.
+func (f *flushResponseWriter) ReadFrom(r io.Reader) (int64, error) {
+	flusher := http.NewResponseController(f.ResponseWriter) // nolint: bodyclose
+
+	var n int64
+	p := make([]byte, chunkSize)
```
OlshaMB, line 30: Maybe make chunk size configurable?

OlshaMB: Make it at least public.

plumbing/storer/storer.go
```diff
@@ -13,3 +15,9 @@ type Initializer interface {
 	// any.
 	Init() error
 }
+
+// FilesystemStorer is a storer that can be used to store objects and references
+// in a filesystem. It is used by the filesystem storage.
+type FilesystemStorer interface {
```
OlshaMB, line 21: Still think it's too generic, but if I need I will fix in later pr.

OlshaMB (commented):
A few netpicks but looks ready to me, the suggested changes if needed can be put in place at a later date

OlshaMB:
Forgot to choose approve.

backend/http/http.go
OlshaMB: Thinking now, it can be resolved in the loader.

aymanbagabas: You can't have postfix because that will interfere with git routes ex. given `POST /my/long/prefix/gitrepo/git-receive-pack`, you can't have anything appended after `git-receive-pack` because that's what the git client expects the endpoint to be.

However, say you want to mount your middleware under `http://mydomain/gitstuff/` then you would use `Prefix: "/gitstuff"` and the middleware will strip that part when it tries to match for the repository name before giving it to the `Loader` to find.

backend/http/http.go
aymanbagabas: It is optional now, when nil, no errors are logged. This is similar to `http.Server{ErrorLog}`

backend/http/http.go
aymanbagabas: I don't think that's necessary, as this middleware is just a frontend to the actual server implementation. If something changes in `UploadPack`/`ReceivePack`, we'd also need to update the logic everywhere these routines are used.

backend/http/http.go
```diff
@@ -0,0 +1,345 @@
...
+
+// Handler is a Git HTTP handler that serves git repositories over HTTP.
+//
+// It supports serving repositories using both the Smart-HTTP and the Dumb-HTTP
+// protocols. When the Dumb-HTTP protocol is used, the repository store must
+// implement the [storer.FilesystemStorer] interface. Keep in mind that
+// repositories that wish to be server using the Dumb-HTTP protocol must update
+// the server info files. This can be done by using
+// [transport.UpdateServerInfo] before serving the repository.
+type Handler struct {
```
pjbgf, line 62: From an API/UX perspective, the public fields can be changed for a created `Handler`. We are better off making all the fields internal, and have them set via the[ variadic options pattern](https://github.com/go-git/go-git/blob/93ed29d8f65c1f879d438a5370c212ffedfe5030/plumbing/format/packfile/parser_test.go#L39-L40) based on a `NewHandler(opts ...)` instead.

backend/http/writer.go
pjbgf: There may be heuritistics known by the backend that could change this to an optimal value to optimise an specific use-case. 

Instead of making it public, let's make it a field of `flushResponseWriter`. If needed, we could then expose it via variadic option of the Handler.

pjbgf (changes requested):

OlshaMB (approved):

Issue #509, opened 2022-04-06
Hostname from ssh config is handled incompatible to how git+openssh does it


codablock:
Consider the following example ssh config:
```
Host example.com
    Hostname ssh.example.com
```

When openssh is used to connect to `example.com` through a non-default port, it will properly honor the non-default port and only replace the hostname with `ssh.example.com`, e.g.:
```sh
$ ssh -p 2222 example.com
```
This will connect to ssh.example.com via port 2222.

Also, using git to clone the url `ssh://git@example.com:2222/my-repo` will properly handle this the expected way.

The same URL however will cause go-git to unexpectedly switch to port 22. This happens in `doGetHostWithPortFromSSHConfig`, which uses `DefaultSSHConfig.Get` to retrieve the user config port. This will however not find a match in the config (as there is no `Port` for the host) and thus return the default value `22` for it, causing the unexpected port override.

To fix this, a change in the dependent https://github.com/kevinburke/ssh_config library is required, as it's currently not possible to check if a value exists, which is needed to decide if the port should be overridden or not. I'll link the issue that I'm going to create in the library here later. EDIT [here it is](https://github.com/kevinburke/ssh_config/issues/47)

Another inconsistency that I found in the same function is that `doGetHostWithPortFromSSHConfig` will ignore `Port` in case no `Hostname` is configured. My understanding (without having it verified) of the ssh config is that you should be able to override the port without overriding the hostname.

PR #1509, opened 2025-04-12
plumbing: transport/ssh allow different AuthMethod basd on user/host

### LINKED ISSUE #218: ssh private key not being picked up
By default git pickups my ssh key from ~/.ssh, but for pull and push operations this fails for go-git.

```
Error: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
```

If I do an `ssh-add` with my key it will start working.

This is on OSX.

tjamet:
When using the plain git with SSH, git reads the SSH config and can consider all the relevant SSH options depending on the hostname.

With go-git, it is impossible to use different configurations like IdentityFiles based on the hostname or other configurations as the information is not provided to the ssh auth config.

This commit provides such information while keeping the backward compatibility with older auth methods which do not.

I expect this will help fix #218 and #509


tjamet:
I have tested this change with the auth method which allowed me to consider both ssh agents *and* individual identity files for a given host.

```golang
import (
	// ssh-client-config is a WIP project to create a working ssh config
	// as configured in the various ssh config files
	sshclientconfig "github.com/tjamet/ssh-client-config"
)

type GitAuth struct {
}

func (a *GitAuth) SetAuth(r *http.Request) {
}

func (a *GitAuth) Name() string {
	return "GitAuth"
}

func (a *GitAuth) String() string {
	return "GitAuth"
}
func (a *GitAuth) ClientConfig(user string, host string) (*ssh.ClientConfig, error) {
	return sshclientconfig.NewSSHClientConfig("").SSHClientConfig(host, func(cc *ssh.ClientConfig) {
		// go git already implements the HostKeyCallback
		// so far, sshclientconfig does not at the moment
		cc.HostKeyCallback = nil
		cc.User = user
	})
}
```

PR #1510, opened 2025-04-12
[v5] plumbing: support mTLS for HTTPS protocol

hiddeco:
This adds support for mTLS for the HTTPS protocol, the same way custom CA bundles were introduced for `v5`. It is specifically targeting `v5`, as I believe in `v6` we may benefit from a full rehaul of how we work with transports (and may benefit from more fine-grain `http.Transport` injection).

### DIFF
--- a/plumbing/transport/http/common.go
+++ b/plumbing/transport/http/common.go
@@ -15,12 +15,13 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/golang/groupcache/lru"
+
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/protocol/packp"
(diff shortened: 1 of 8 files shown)


plumbing/transport/http/common.go
```diff
@@ -245,6 +263,8 @@ func newSession(c *client, ep *transport.Endpoint, auth transport.AuthMethod) (*
 			configureTransport(transport, ep)
```
hiddeco, line 263: Is the error swallowing here on purpose? It seems like this could result in unexpected behavior.

pjbgf: I don't think so, we should at the very least log it. Do you mind handling the errors from the `configureTransport` calls into a separate commit please?

pjbgf (approved):

PR #1512, opened 2025-04-13
Support pushing to an anonymous remote

mhagger:
Git allows you to push to a remote that isn't defined in your gitconfig:

    git push https://example.com/foo.git refs/heads/bar:refs/heads/bar

The effect of this command is to push to the remote repository at the specified URL, without updating any remote-tracking references corresponding to the reference that was pushed.

It would seem that this should be possible with `go-git`, by writing something like

    r, _ := git.PlainOpen(".")
    _ = r.Push(&git.PushOptions{
        RemoteURL: "https://example.com/foo.git",
        RefSpecs: []config.RefSpec{"refs/heads/bar:refs/heads/bar"},
    })

But `go-git` wasn't handling this correctly:

* `PushOptions.Validate()` saw the empty `o.RemoteName` and changed it to `DefaultRemoteName` ("origin").

* If no "origin" remote was defined in the gitconfig, `Repository.Push()` failed entirely with `ErrRemoteNotFound`.

* If: an "origin" _was_ defined in the gitconfig, then after the push, remote-tracking references were incorrectly created for the "origin" remote (e.g., `refs/remotes/origin/foo`), even though the push was probably to a remote that is not related to "origin".

Alternatively, if the `PushOptions` were created using `RemoteName: "anonymous"`, the push still failed with `ErrRemoteNotFound` because a remote called "anonymous" was sought in the gitconfig, and that remote (presumably!) doesn't exist.

This commit fixes all three problems:

* Change the behavior of `PushOptions.Validate()`: if there is a `RemoteURL` but no `RemoteName`, then set the `RemoteName` to "anonymous" rather than setting it to `DefaultRemoteName`. (If neither a `RemoteName` nor a `RemoteURL` is set, then set `RemoteName` to `DefaultRemoteName` as before.)

* Change `Remote.updateRemoteReferenceStorage()` to skip updating any remote-tracking references if the remote is anonymous.

The earlier patches in this series introduce some helpers to make it easier to deal with anonymous remotes in the first place, and somewhat formalizing the concept (which already existed in the form of `Repository.CreateRemoteAnonymous()` but didn't seem to be used much).

Aside: In my opinion, it would be preferable to use a more distinct name than "anonymous" for anonymous remotes, to avoid the (admittedly unlikely) possibility that the user has created an actual remote named "anonymous". For example, an anonymous remote could use the empty string for its name, or at least something more distinct like "_anonymous_" or "(anonymous)" or "*anonymous*". But there is already a precedent in `Repository.CreateRemoteAnonymous()` to use the name "anonymous", so I didn't make that change.

This is my first contribution to go-git. I broke this PR into separate patches because I think that makes them easier to review, but if this project prefers single-commit PRs, I can either squash them together or submit separate PRs, as you prefer. (But the PRs will depend on each other, so there doesn't seem to be much value in the latter.)

In case you're curious, I contributed extensively to `git` several years ago and currently work at GitHub. I haven't used this library before, but am considering it for a small internal project. I'm happy that it exists and am glad to contribute a little bit to it :heart:


pjbgf (changes requested):
Hey @mhagger, thanks for looking into this and for the detailed description - that's much appreciated.

The main change I'd suggest for your PR is to avoid creating the arbitrary Remote name. Instead, we could set the RemoteName to the default in the `PushContext`, whenever both `pushOptions.URL` and `pushOptions.RemoteName` are empty. 
That way, later down the execution path we will always have a Remote name set, unless we are handling an anonymous push. WDYT?

Also, do you mind targeting the `v6-transport` branch instead? We are working transport related work there which will be part of our `v6` release.

repository_test.go
```diff
@@ -1708,6 +1710,40 @@ func (s *RepositorySuite) TestPush() {
 	})
 }
 
+func (s *RepositorySuite) TestPushAnonymous() {
+	url, err := os.MkdirTemp("", "")
+	s.NoError(err)
```
pjbgf, line 1715:
```suggestion
	url := s.T().TempDir()
```

mhagger:
@pjbgf Thanks for your quick reply.

I rebased this branch onto the `v6-transport` branch, changed the target to that branch, and squashed in your suggested change.

I'm not exactly sure what you mean by "avoid creating the arbitrary Remote name". Are you suggesting that anonymous remotes should have the empty string as their name? That's doable, though it would require the following changes:

* It would require a change to `Repository.CreateRemoteAnonymous()`, which currently requires that the name of an anonymous remote be "anonymous", and also to the tests `TestCreateRemoteAnonymous()` and `TestCreateRemoteAnonymousInvalidName()`, which explicitly test for this behavior.
* It would require a change to `Remote.String()`, which currently incorporates the remote's name into its string representation.

If we go that route, then for backwards compatibility, one could imagine having `Repository.CreateRemoteAnonymous()` accept either "" or "anonymous" as the input `RemoteConfig.name`. But since the passed-in `RemoteConfig` is (unfortunately?) built straight into the constructed `Remote`, one would either have to:

1. accept either name as meaning "anonymous" in a `Remote` instance, thus making the code more, rather than less, complicated, and continue to live with the inability to handle a configured remote that happens to be named "anonymous";
2. overwrite "anonymous" with the empty string in the passed-in `RemoteConfig`;
3. copy the `RemoteConfig` but with "anonymous" changed to "", and store the copy (not the original) in the `Remote` instance; or
4. slightly break backwards compatibility by changing `Repository.CreateRemoteAnonymous()` to require anonymous remotes to be created with the empty string as their name.

If that's what you want, please confirm and let me know if you have a preference for one of the above options, then I'll make the change. If you had something else in mind, please clarify.

repository_test.go
mhagger: :heavy_check_mark:

pjbgf:
@mhagger thanks for rebasing this. :bow: 

> Are you suggesting that anonymous remotes should have the empty string as their name?

Yes. I added a few comments below with the changes we may need to do to achieve this.


> It would require a change to `Repository.CreateRemoteAnonymous()`

Given that we are already working on the new V6 API, we can remove this method altogether. Let's align with upstream UX, and handle this internally at the point of use (e.g. at a Push operation for example).

> It would require a change to `Remote.String()`, which currently incorporates the remote's name into its string representation.

Upstream this doesn't really happen, as adhoc remotes are not really listed. I think we can either leave it as empty, or show something that is not a valid `Remote` name, so that it is clear this is not a valid `Remote` e.g. `<unnamed>`.


Please squash the changes into 1-2 commits. For example, one for removing the concept of "anonymous" remotes, and another to handle non-persistent remotes.

repository.go
```diff
@@ -1245,9 +1247,21 @@ func (r *Repository) PushContext(ctx context.Context, o *PushOptions) error {
...
-		return err
+	var remote *Remote
+	if o.RemoteName == config.AnonymousRemoteName {
+		var err error
+		remote, err = r.CreateRemoteAnonymous(
+			config.NewAnonymousRemoteConfig(o.RemoteURL),
+		)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
mhagger: True, but that would mean that the `RemoteConfig` doesn't go through any validation at all. Are you OK with that?

mhagger:
I just re-pushed the branch, rebased on the current v6-transport, with a fixup commit on top addressing most of your feedback. Once we are agreed on the approach I'll squash the PR down to one or two commits as you requested.

The main open question for me is whether/when the `RemoteConfig` should be validated if `NewRemote()` is called. Currently it is validated by `Repository.CreateRemote()` but not by `NewRemote()`, which seems strange to me. Perhaps the validation should be moved to `NewRemote()` (which is called by `Repository.CreateRemote()`) to cover both code paths?

pjbgf:
> Currently it is validated by `Repository.CreateRemote()` but not by `NewRemote()`, which seems strange to me. Perhaps the validation should be moved to `NewRemote()` (which is called by `Repository.CreateRemote()`) to cover both code paths?

@mhagger +1 on your suggestion. Let's move this to `NewRemote()` and update the signature to also return an `error`. Otherwise, LGTM.

PR #1515, opened 2025-04-14
plumbing: transport, Reintroduce SetHostKeyCallback. Fix #1514

### LINKED ISSUE #1514: SetHostKeyCallback removed in minor version release (v5.15)
The `HostKeyCallbackHelper.SetHostKeyCallback()` was removed (or more correctly, replaced with `SetHostKeyCallbackAndAlgorithms()`?)

See diff:  https://github.com/go-git/go-git/compare/v5.14.0...v5.15.0#diff-512f53c734d78ae512802447f2d46d550e849d296a6102ec8515de90f1fa3563L300

Would it be possible to add a (deprecated) alias with the old name, or restore the old implementation if they are significantly different?

FWIW, I am not using this method in my own codebase, but it is [used by Argo CD](https://github.com/argoproj/argo-cd/blob/faf0073ebc1cccdf53766a56a084f9e5052814cd/util/git/ssh.go#L58), which is preventing my project from compiling.

Thanks!

pjbgf:
The PR #1482 renamed `SetHostKeyCallback` to `SetHostKeyCallbackAndAlgorithms`, due to the change in behaviour in that method. For keeping backwards compatibility with v5, a new `SetHostKeyCallback` is being introduced that simply calls back to `SetHostKeyCallbackAndAlgorithms`.

Some tests were introduced to enforce expected behaviour.

### DIFF
--- a/plumbing/transport/ssh/auth_method.go
+++ b/plumbing/transport/ssh/auth_method.go
@@ -231,7 +231,10 @@ func (a *PublicKeysCallback) ClientConfig() (*ssh.ClientConfig, error) {
 //	/etc/ssh/ssh_known_hosts
 func NewKnownHostsCallback(files ...string) (ssh.HostKeyCallback, error) {
 	kh, err := NewKnownHostsDb(files...)
-	return kh.HostKeyCallback(), err
+	if err != nil {
+		return nil, err
+	}
+	return kh.HostKeyCallback(), nil
 }
 
 // NewKnownHostsDb returns knownhosts.HostKeyDB based on a file based on a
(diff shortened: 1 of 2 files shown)


plumbing/transport/ssh/auth_method.go
```diff
@@ -311,13 +314,40 @@ type HostKeyCallbackHelper struct {
 	// HostKeyAlgorithms is a list of supported host key algorithms that will
 	// be used for host key verification.
 	HostKeyAlgorithms []string
+
+	// fallback allows for injecting the fallback call, which is called
+	// when a HostKeyCallback is not set.
+	fallback func(files ...string) (ssh.HostKeyCallback, error)
```
Javier-varez, line 320: Might be worth indicating that this is only meant for testing, but in practice it always calls `NewKnownHostsCallback`

Javier-varez (commented):
Thanks, I did not realize that there are users of `HostKeyCallbackHelper` outside `go-git`. Thank you for addressing it.

hiddeco (approved):

PR #1516, opened 2025-04-14
Allow all types of objects to be pushed using non-wildcard refspecs

mhagger:
For non-wildcard refspecs, `addOrUpdateReferences()` was trying to resolve the refspec source to a commit. But there's no reason that other objects (tags, trees, blobs) shouldn't be allowed to be pushed by OID. So resolve the refspec source to whatever object type it happens to be.


pjbgf (approved):
@mhagger thanks for working on this. :bow:

Issue #1188, opened 2024-09-12
High memory allocations from concurrent clones


tylfin:
We're using `go-git` to clone several repositories in parallel and noticed a memory spike leading to OOM/unresponsive behavior on smaller machines. I went ahead and profiled the application, noticing that the `PlainClone` can accumulate files on the heap during checkout. Is there a way to get these to stream directly from the provider or an option to avoid the `MemoryObject` allocation?

<img width="1366" alt="Screenshot 2024-09-12 at 12 19 33 PM" src="https://github.com/user-attachments/assets/2ee895d0-4cc9-4e23-8fc7-4d9774cc252e">

pjbgf:
@tylfin Are you doing shallow clones or you need the entire depth of the repository? Can you share a code snippet as depending on what you need and how you are using `go-git`, there may be some optimisations available.

As for disk based cloning, as opposed to in memory, there is currently an overhead which I am resolving as part of the packfile scanner/parser rebuild. That will decrease memory allocations and get the time to clone closer together between the two types of storage. This should be made available on the `v6-exp` in the coming weeks.

tylfin:
Hey @pjbgf, thanks for getting back so quickly. Our `CloneOptions` generally look like this:

```
opts := &git.CloneOptions{
	URL:               url,
	RecurseSubmodules: git.SubmoduleRescursivity(recurseSubmodules),
	SingleBranch:      true,
	Depth:             1,
	Tags:              git.AllTags,
	Auth:              auth,
}
```

The repositories are user-specified and can container large files/artifacts. We do need all the tags as we're polling these repositories to build packages for customers.

> This should be made available on the v6-exp in the coming weeks.

That all sounds great, we'll switch to the new packfile when it's available.

grachevko:
Repository size ~450mb

```sh
Enumerating objects: 276370, done.
Counting objects: 100% (185698/185698), done.98)
Compressing objects: 100% (32804/32804), done.804)
Total 276370 (delta 167651), reused 152914 (delta 152884), pack-reused 90672 (from 1)
```

```golang
repo, err := git.PlainCloneContext(ctx, repoDir, false, &git.CloneOptions{
	Auth:              auth,
	URL:               gitRepo.HttpUrl,
	SingleBranch:      true,
	RecurseSubmodules: git.DefaultSubmoduleRecursionDepth,
	ShallowSubmodules: true,
	Progress:          os.Stderr,
})
```

[...]

Issue #1451, opened 2025-03-08
Regression (bisected): poor performance reading packfiles


runxiyu:
## Test program

```go
package main

import (
	"log"
	"os"

	"github.com/go-git/go-git/v5"
)

func main() {
	log.SetFlags(log.LstdFlags | log.Lmicroseconds)
	repo, err := git.PlainOpen(os.Args[1]); if err != nil { panic(err) }
	headRef, err := repo.Head(); if err != nil { panic(err) }
	commit, err := repo.CommitObject(headRef.Hash()); if err != nil { panic(err) }
	tree, err := commit.Tree(); if err != nil { panic(err) }
	log.Println("Iterating tree")
	for _, entry := range tree.Entries {
		size, err := tree.Size(entry.Name); if err != nil { panic(err) }
		log.Println(entry.Name, size)
	}
	log.Println("Done")
}
```

## Test environment

[This commit from Asahi Linux](https://github.com/AsahiLinux/linux/commit/00e802beb831ad451914908402c61fd345a00bbd)

go version go1.23.6 linux/arm64

M1 MacBook Air (2020)

## Expected behavior

The entire process completes in about 0.1s

## Actual behvior

It takes about 1.3s.

## Profiling

![Image](https://github.com/user-attachments/assets/6a83ca9a-b572-4316-818c-f3a411f0daca)

It looks like the main problem is how genOffsetHash uses maps. I'm no expert when it comes to the Go runtime so I'll leave it here.

## First bad commit

```
commit bb5c35196ee55b82f6506d69257de0e59cd8c41d
Date:   Sun Nov 17 00:34:11 2024 +0000

    plumbing: packfile, Refactor Parser and Scanner logic
    The changes focus on increasing thread-safety, simplifying the code base,
    enabling support for sha256 and improving general time and space complexities.

    (snip: some statistics showing fewer allocs)
```

runxiyu:
Note that the problematic commit is not in 5.14.0 yet; it's in the main branch tho

runxiyu:
The 'assign to a map a million times' behaviour could be seen in commits as old as 83649a1c08e0dcc103de54652ed5b9c33d301326 when running `genOffsetHash` e.g. when fetching blob contents from packfiles

![Image](https://github.com/user-attachments/assets/64a718fd-67bf-4cf2-ba0a-b3be04fffea7)

Additionally, memory usage also explodes

[...]

Issue #1331, opened 2024-12-30
v6: PackfileIter with KeepDescriptors


pjbgf:
During the [migration](https://github.com/go-git/go-git/pull/1312) from go-check to testify we noticed that `TestPackfileIterKeepDescriptors` was no longer working. This can be related to https://github.com/go-git/go-git/pull/1228, or may have happened beforehand.

In either case, that test is currently being [skipped](https://github.com/go-git/go-git/pull/1312/commits/dab914087d5b6296aef4370a4aee82d64498ee9a). For `v6` we could either find a more performant approach to reading the packfile, and remove the KeepDescriptors option, or fix `PackfileIter`'s support for it and unskip this test.

PR #1523, opened 2025-04-17
[v6] plumbing: format/packfile, Optimise packfile delta processing

pjbgf:
The memory churn for processing delta objects was quite high, due to the performance optimisation which kept in memory all the inflated contents for each object.

The changes introduced allows for supporting storage implementations (i.e. `storage/filesystem`) to only keep in memory the contents of the parents needed to resolve the contents of the delta being processed. The buffers used for all the contents are reused to avoid unnecessary churn.

Users can opt-out by using `WithHighMemoryUsage()` when creating the Parser, or `HighMemoryMode` in the storage option. 

A new `ioutil.Copy` was introduced to simplify the process of reusing buffers during Copy operations.

Fixes https://github.com/go-git/go-git/issues/1188 https://github.com/go-git/go-git/issues/1331.

---
Benchmarks comparing this PR against the [current](https://github.com/go-git/go-git/commit/5d6af409877bf81b28775aef461d5d2738c9c309) `main` branch using BenchmarkPlainClone against the `kubernetes/kubernetes` repository.

#### Filesystem storage (1.4GB less memory churn)
image
image

#### Memory storage (1GB less memory churn)
image
image

#### Reduced GC pressure
The decreased memory churn results in reduced GC pressure:
image

---
Running the benchmark against `torvalds/linux` with `high` or `low` memory usage, compared with the [current version](https://github.com/go-git/go-git/commit/5d6af409877bf81b28775aef461d5d2738c9c309) in `main`:
```
BenchmarkPlainClone/high-memory-16                 1	21255659542 ns/op	1418664608 B/op	 2227027 allocs/op
BenchmarkPlainClone/low-memory-16          	   1	22576359873 ns/op	1402975104 B/op	 2167243 allocs/op
BenchmarkPlainClone/main-16    	                   1	22154340340 ns/op	4962471016 B/op	 2483031 allocs/op
```

Note that even the version with High memory usage yield over 250k less allocations than the previous version. The difference in memory churn aligns with what the pprof information above (although this is for a different repo).

### DIFF
--- a/plumbing/format/packfile/packfile.go
+++ b/plumbing/format/packfile/packfile.go
@@ -14,6 +14,7 @@ import (
 	"github.com/go-git/go-git/v6/plumbing/format/idxfile"
 	"github.com/go-git/go-git/v6/plumbing/storer"
 	"github.com/go-git/go-git/v6/utils/ioutil"
+	gogitsync "github.com/go-git/go-git/v6/utils/sync"
 )
 
 var (
@@ -324,13 +325,17 @@ func (p *Packfile) getMemoryObject(oh *ObjectHeader) (plumbing.EncodedObject, er
 			return nil, fmt.Errorf("cannot find base object: %w", err)
 		}
 
-		err = p.scanner.inflateContent(oh.ContentOffset, &oh.content)
+		if oh.content == nil {
+			oh.content = gogitsync.GetBytesBuffer()
+		}
+
+		err = p.scanner.inflateContent(oh.ContentOffset, oh.content)
 		if err != nil {
 			return nil, fmt.Errorf("cannot inflate content: %w", err)
 		}
 
 		obj.SetType(parent.Type())
-		err = ApplyDelta(obj, parent, oh.content.Bytes()) //nolint:ineffassign
+		err = ApplyDelta(obj, parent, oh.content) //nolint:ineffassign
 
 	default:
 		err = ErrInvalidObject.AddDetails("type %q", oh.Type)
--- a/plumbing/format/packfile/packfile.go
+++ b/plumbing/format/packfile/packfile.go
@@ -345,19 +345,6 @@ func (p *Packfile) getMemoryObject(oh *ObjectHeader) (plumbing.EncodedObject, er
 	return obj, nil
 }
 
-// isInvalid checks whether an error is an os.PathError with an os.ErrInvalid
-// error inside. It also checks for the windows error, which is different from
-// os.ErrInvalid.
-func isInvalid(err error) bool {
-	pe, ok := err.(*os.PathError)
-	if !ok {
-		return false
-	}
-
-	errstr := pe.Err.Error()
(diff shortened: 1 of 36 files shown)


runxiyu:
This may also fix https://github.com/go-git/go-git/issues/1451. I will test it against my workloads when I have time.

utils/ioutil/sync.go
```diff
@@ -0,0 +1,17 @@
+package ioutil
```
runxiyu, line 1: I think this is very confusingly named, since it's also the name of the deprecated `io/ioutil`

plumbing/format/packfile/parser_test.go
```diff
@@ -135,7 +135,7 @@ func TestResolveExternalRefsInThinPack(t *testing.T) {
 
 	checksum, err := parser.Parse()
 	assert.NoError(t, err)
-	assert.NotEqual(t, plumbing.ZeroHash, checksum)
+	assert.NotEqual(t, checksum, plumbing.ZeroHash)
```
runxiyu, line 138: How would this make a difference?

utils/ioutil/sync.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"io"
+
+	"github.com/go-git/go-git/v6/utils/sync"
+)
+
+// Copy calls io.CopyBuffer and uses a buffer from sync.GetByteSlice,
+// to reduce the complexity when using it while avoiding the allocation
+// of a new buffer per call.
+func Copy(dst io.Writer, src io.Reader) (n int64, err error) {
```
aymanbagabas: Honestly, I would rather make the caller call `io.CopyBuffer` directly or make them implement `io.ReaderFrom` and `io.WriterTo`. Or rename this to something like `CopyBufferPool` to avoid confusion with `io.Copy`

plumbing/format/packfile/packfile.go
```diff
@@ -324,13 +325,17 @@ func (p *Packfile) getMemoryObject(oh *ObjectHeader) (plumbing.EncodedObject, er
 			return nil, fmt.Errorf("cannot find base object: %w", err)
 		}
 
-		err = p.scanner.inflateContent(oh.ContentOffset, &oh.content)
+		if oh.content == nil {
+			oh.content = gogitsync.GetBytesBuffer()
```
aymanbagabas, line 329: Shouldn't we put the buffer back after finishing using it?

aymanbagabas: Never mind, I see that we put it back later but only if lowMemoryMode is set 🤔

pjbgf: Yep, we need them in memory to be able to inflate its descendants. We only remove them from memory once they have been used or sync decides it is time for it to go. Whatever happens first.

utils/ioutil/sync.go
pjbgf: Nice one, I'll rename to CopyBufferPool.

PR #1525, opened 2025-04-19
git: checkout.validate should set to branch only if both branch and hash are empty

GaikwadPratik:
add validation usecase for Force and keep.

This would avoid the error condition where if Hash is provided on CheckoutOptions but NOT branch and validate is called since per documentation:

// Branch to be checked out, if Branch and Hash are empty is set to master.
Branch plumbing.ReferenceName

### DIFF
--- a/options.go
+++ b/options.go
@@ -349,6 +349,7 @@ type SubmoduleUpdateOptions struct {
 var (
 	ErrBranchHashExclusive  = errors.New("Branch and Hash are mutually exclusive")
 	ErrCreateRequiresBranch = errors.New("Branch is mandatory when Create is used")
+	ErrForceKeepExclusive   = errors.New("Force and branch are mutually exclusive")
 )
 
 // CheckoutOptions describes how a checkout operation should be performed.
@@ -375,16 +376,23 @@ type CheckoutOptions struct {
 
 // Validate validates the fields and sets the default values.
 func (o *CheckoutOptions) Validate() error {
+	// if not create, both hash and branch cannot be provided
 	if !o.Create && !o.Hash.IsZero() && o.Branch != "" {
 		return ErrBranchHashExclusive
 	}
 
-	if o.Create && o.Branch == "" {
+	// if create with branch not provided but hash is provided, ask for branch name?
+	if o.Create && o.Branch == "" && !o.Hash.IsZero() {
 		return ErrCreateRequiresBranch
 	}
 
-	if o.Branch == "" {
+	// if create with both branch and hash are not provided, set branch to master
+	if o.Create && o.Branch == "" && o.Hash.IsZero() {
 		o.Branch = plumbing.Master
+	}
+
+	if o.Force && o.Keep {
+		return ErrForceKeepExclusive
 	}
 
 	return nil
--- a/options_test.go
+++ b/options_test.go
@@ -120,6 +120,31 @@ func (s *OptionsSuite) writeGlobalConfig(cfg *config.Config) func() {
 
 	return func() {
 		os.Setenv("XDG_CONFIG_HOME", "")
-
 	}
+}
+
+func (s *OptionsSuite) TestCheckoutOptionsValidate() {
+	checkoutOpts := CheckoutOptions{}
+	err := checkoutOpts.Validate()
+	s.NotNil(err)
+	s.Equal(ErrBranchHashExclusive, err)
+
+	checkoutOpts.Create = true
+	err = checkoutOpts.Validate()
+	s.Nil(err)
+	s.Equal(plumbing.Master, checkoutOpts.Branch)
+
+	checkoutOpts.Branch = ""
+	checkoutOpts.Hash = plumbing.NewHash("ab1b15c6f6487b4db16f10d8ec69bb8bf91dcabd")
+	err = checkoutOpts.Validate()
+	s.NotNil(err)
+	s.Equal(ErrCreateRequiresBranch, err)
+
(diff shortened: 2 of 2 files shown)


pjbgf (approved):
@GaikwadPratik thanks for working on this. :bow:

options_test.go
```diff
@@ -120,6 +120,31 @@ func (s *OptionsSuite) writeGlobalConfig(cfg *config.Config) func() {
...
 		os.Setenv("XDG_CONFIG_HOME", "")
-
 	}
 }
+
+func (s *OptionsSuite) TestCheckoutOptionsValidate() {
+	checkoutOpts := CheckoutOptions{}
+	err := checkoutOpts.Validate()
+	s.NotNil(err)
+	s.Equal(ErrBranchHashExclusive, err)
```
pjbgf, line 130:
```suggestion
	s.ErrorIs(ErrBranchHashExclusive, err)
```

options_test.go
```diff
@@ -120,6 +120,31 @@ func (s *OptionsSuite) writeGlobalConfig(cfg *config.Config) func() {
...
+	checkoutOpts.Create = true
+	err = checkoutOpts.Validate()
+	s.Nil(err)
+	s.Equal(plumbing.Master, checkoutOpts.Branch)
+
+	checkoutOpts.Branch = ""
+	checkoutOpts.Hash = plumbing.NewHash("ab1b15c6f6487b4db16f10d8ec69bb8bf91dcabd")
+	err = checkoutOpts.Validate()
+	s.NotNil(err)
+	s.Equal(ErrCreateRequiresBranch, err)
```
pjbgf, line 141:
```suggestion
	s.ErrorIs(ErrCreateRequiresBranch, err)
```

options_test.go
```diff
@@ -120,6 +120,31 @@ func (s *OptionsSuite) writeGlobalConfig(cfg *config.Config) func() {
...
+
+	checkoutOpts.Branch = "test"
+	checkoutOpts.Force = true
+	checkoutOpts.Keep = true
+
+	err = checkoutOpts.Validate()
+	s.NotNil(err)
+	s.Equal(ErrForceKeepExclusive, err)
```
pjbgf, line 149:
```suggestion
	s.ErrorIs(ErrForceKeepExclusive, err)
```

GaikwadPratik:
Hi @pjbgf , applied the suggestions, can you please trigger another pipeline?

pjbgf:
@GaikwadPratik I generally run `make test` or rely on `go test` for specific tests I'm trying to run. Please note that some tests require `git-daemon` installed, without it you will get some errors.

GaikwadPratik:
> @GaikwadPratik I generally run `make test` or rely on `go test` for specific tests I'm trying to run. Please note that some tests require `git-daemon` installed, without it you will get some errors.

Can we please update the documentation?

Also any idea why pipeline is failing even though my code doesn't touch those tests?

options.go
```diff
@@ -375,18 +376,25 @@ type CheckoutOptions struct {
...
 
-	if o.Create && o.Branch == "" {
+	// if create with branch not provided but hash is provided, ask for branch name?
+	if o.Create && o.Branch == "" && !o.Hash.IsZero() {
 		return ErrCreateRequiresBranch
 	}
 
-	if o.Branch == "" {
+	// if create with both branch and hash are not provided, set branch to master
+	if o.Create && o.Branch == "" && o.Hash.IsZero() {
```
onee-only, line 390: I think the failures are because of this check.
Is o.Create necessary to set the branch to master?

options_test.go
```diff
@@ -120,6 +120,31 @@ func (s *OptionsSuite) writeGlobalConfig(cfg *config.Config) func() {
...
 		os.Setenv("XDG_CONFIG_HOME", "")
-
 	}
 }
+
+func (s *OptionsSuite) TestCheckoutOptionsValidate() {
+	checkoutOpts := CheckoutOptions{}
+	err := checkoutOpts.Validate()
+	s.NotNil(err)
+	s.ErrorIs(ErrBranchHashExclusive, err)
```
onee-only, line 130: I don't think this is valid. Since it has to satisfy `!o.Create && !o.Hash.IsZero() && o.Branch != ""`

onee-only (commented):
FYI, this fails on my codespace too.

options.go
GaikwadPratik: Am I incorrectly reading https://github.com/go-git/go-git/blob/main/options.go#L360?
> // Branch to be checked out, if Branch and Hash are empty is set to `master`.

so we don't need to check for `o.Create`?

onee-only: Yeah I believe that default should be set regardless of `o.Create`. otherwise it will result in `reference not found` error.

GaikwadPratik:
Hi @onee-only ,
> FYI, this fails on my codespace too.

Can you please provide instructions on how you have set up for local testing? I think I am missing something and I would prefer to do local testing instead of pushing the code in git all the time just for testing

options_test.go
GaikwadPratik: You are right... I tried to be cheeky.. Going back to basics and fixing it :)

options.go
GaikwadPratik: Fixed it

GaikwadPratik:
@pjbgf , okay my changes were breaking one of the test in `worktree_test`. Fixed the code conditions and tests for `CheckoutOptions` as well by keeping them simple and to the basics. Ran `make test` locally and they ran successfully.

image

PR #1529, opened 2025-04-25
[v6] backend: add git transport backend

aymanbagabas:
This implements a Git transport backend ~and server~ for Go-Git. ~The backend expects a `net.Conn` and handles raw Git transport protocol requests. It is suitable to be used with a `net.Listener` or any other `net.Conn` source.~ The server implementation has been moved to go-git/cli https://github.com/go-git/cli/pull/13.

~Since Go doesn't have a standard raw TCP server API, we need a new `git.Server` to handle our requests. The API looks as follows.~

I'm still not sure if we should keep a TCP server implementation in `go-git`. Perhaps that code can live somewhere else where we don't need to maintain it here, like in `go-git/cli`. Please see https://github.com/go-git/cli/pull/13

```go
// Backend represents a Git transport server handler that can handle git-upload-pack, git-receive-pack, and git-upload-archive requests over TCP.
type Backend struct {
	// Loader is used to load repositories. It uses [transport.DefaultLoader]
	// when nil.
	Loader transport.Loader
	// UploadPack indicates whether the handler should handle
	// git-upload-pack requests.
	UploadPack bool
	// ReceivePack indicates whether the handler should handle
	// git-receive-pack requests.
	ReceivePack bool
}

func (b *Backend) ServeTCP(ctx context.Context, c io.ReadWriteCloser, req *packp.GitProtoRequest)
```

### DIFF
--- a/backend/git/git.go
+++ b/backend/git/git.go
@@ -2,12 +2,17 @@ package git
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"log"
+	"math/rand"
 	"net"
 	"net/url"
 	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
 
 	"github.com/go-git/go-git/v6/plumbing/format/pktline"
 	"github.com/go-git/go-git/v6/plumbing/protocol/packp"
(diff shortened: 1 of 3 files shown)


backend/git/git.go
```diff
@@ -0,0 +1,476 @@
...
+	svc := transport.Service(req.RequestCommand)
+	if (svc != transport.UploadPackService &&
+		// svc != transport.UploadArchiveService &&
+		svc != transport.ReceivePackService) ||
+		(svc == transport.UploadPackService && !b.UploadPack) ||
+		// (svc == transport.UploadArchiveService && !b.UploadArchive) ||
+		(svc == transport.ReceivePackService && !b.ReceivePack) {
+		renderError(wc, transport.ErrUnsupportedService)
+		return
+	}
```
pjbgf: For simplicity:
```suggestion
	switch {
	case svc == transport.UploadPackService && b.UploadPack,
		svc == transport.ReceivePackService && b.ReceivePack:
	default:
		renderError(wc, transport.ErrUnsupportedService)
		return
	}
```

We can add the archive service once we support it.

backend/git/git.go
```diff
@@ -0,0 +1,476 @@
...
+	url, err := url.JoinPath(fmt.Sprintf("git://%s", host), req.Pathname)
+	if err != nil {
+		renderError(wc, transport.ErrRepositoryNotFound)
+		return
+	}
+
+	ep, err := transport.NewEndpoint(url)
+	if err != nil {
+		// XXX: Should we use a more descriptive error?
+		renderError(wc, transport.ErrRepositoryNotFound)
```
pjbgf:
```suggestion
		renderError(wc, fmt.Errorf("%w: %w", transport.ErrRepositoryNotFound, err)
```

backend/git/git.go
```diff
@@ -0,0 +1,476 @@
...
+	case transport.ReceivePackService:
+		err = transport.ReceivePack(ctx, st,
+			io.NopCloser(r), ioutil.WriteNopCloser(wc),
+			&transport.ReceivePackOptions{
+				GitProtocol: version,
+			})
+	}
+
+	if err != nil {
+		renderError(wc, transport.ErrRepositoryNotFound)
```
pjbgf:
```suggestion
		renderError(wc, fmt.Errorf("%w: %w", transport.ErrRepositoryNotFound, err)
```

backend/git/git.go
```diff
@@ -0,0 +1,476 @@
...
+}
+
+// Close immediately closes the server and all active connections. It returns
+// any error returned from closing the underlying listeners.
+func (s *Server) Close() error {
+	s.inShutdown.Store(true)
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	err := s.closeListenersLocked()
```
pjbgf: To simplify the multiple calls to Lock/Unlock here, what if we did that mutex control within `closeListenersLocked`? That way we can remove that from the other caller as well.

backend/git/git.go
```diff
@@ -0,0 +1,476 @@
...
+
+// ListenAndServe listens on the TCP network address and serves Git
+// protocol requests using the provided handler.
+func (s *Server) ListenAndServe() error {
+	if s.shuttingDown() {
+		return ErrServerClosed
+	}
+	addr := s.Addr
+	if addr == "" {
+		addr = ":9418" // Default Git protocol port
```
pjbgf: Let's move the default to a `const` so we have it set in a single place, and ensure that by default we are only binding for localhost. Users wanting to expose on any (or all) interfaces can always opt-in accordingly.

```suggestion
		addr = "127.0.0.1:9418" // Default Git protocol port
```

backend/git/git.go
```diff
@@ -0,0 +1,476 @@
...
+		unixSec := c.unixSec.Load()
+		if unixSec == 0 {
+			// New connection, skip it.
+			quiescent = false
+			continue
+		}
+		c.Close() //nolint:errcheck
+		delete(s.activeConn, c)
+	}
+	return quiescent
```
pjbgf:
```suggestion
	idle := true
	for c := range s.activeConn {
		unixSec := c.unixSec.Load()
		if unixSec == 0 {
			// New connection, skip it.
			idle = false
			continue
		}
		c.Close() //nolint:errcheck
		delete(s.activeConn, c)
	}
	return idle
```

backend/git/git.go
```diff
@@ -0,0 +1,476 @@
...
+		s.listenerGroup.Add(1)
+	} else {
+		delete(s.listeners, ln)
+		s.listenerGroup.Done()
+	}
+	return true
+}
+
+// closeIdleConns closes all idle connections. It returns whether the server is
+// quiescent.
```
pjbgf:
```suggestion
// closeIdleConns closes all idle connections. It returns true only if no new connection
// was found.
```

backend/git/git.go
aymanbagabas: I should bind to all interfaces, though, and people wanting to bind it to a specific one should explicitly specify that. This is the same behavior as `http.Server` and `git daemon`.

pjbgf (approved):

PR #1540, opened 2025-05-02
git: worktree, Don't delete local untracked files when resetting work…

k-anshul:
…tree. Fixes https://github.com/go-git/go-git/issues/53

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -309,14 +309,15 @@ func (w *Worktree) ResetSparsely(opts *ResetOptions, dirs []string) error {
 		return err
 	}
 
+	var removedFiles []string
 	if opts.Mode == MixedReset || opts.Mode == MergeReset || opts.Mode == HardReset {
-		if err := w.resetIndex(t, dirs, opts.Files); err != nil {
+		if removedFiles, err = w.resetIndex(t, dirs, opts.Files); err != nil {
 			return err
 		}
 	}
 
 	if opts.Mode == MergeReset || opts.Mode == HardReset {
-		if err := w.resetWorktree(t, opts.Files); err != nil {
+		if err := w.resetWorktree(t, removedFiles); err != nil {
 			return err
 		}
 	}
@@ -369,23 +370,24 @@ func (w *Worktree) Reset(opts *ResetOptions) error {
 	return w.ResetSparsely(opts, nil)
 }
 
-func (w *Worktree) resetIndex(t *object.Tree, dirs []string, files []string) error {
+func (w *Worktree) resetIndex(t *object.Tree, dirs []string, files []string) ([]string, error) {
 	idx, err := w.r.Storer.Index()
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	b := newIndexBuilder(idx)
 
 	changes, err := w.diffTreeWithStaging(t, true)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
+	var removedFiles []string
 	for _, ch := range changes {
 		a, err := ch.Action()
 		if err != nil {
-			return err
+			return nil, err
 		}
 
 		var name string
@@ -396,7 +398,7 @@ func (w *Worktree) resetIndex(t *object.Tree, dirs []string, files []string) err
 			name = ch.To.String()
 			e, err = t.FindEntry(name)
 			if err != nil {
-				return err
+				return nil, err
 			}
 		case merkletrie.Delete:
 			name = ch.From.String()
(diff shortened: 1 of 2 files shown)


worktree.go
```diff
@@ -477,8 +485,13 @@ func (w *Worktree) resetWorktree(t *object.Tree, files []string) error {
 			}
 		}
 
-		if err := w.checkoutChange(ch, t, b); err != nil {
-			return err
+		// only checkout an untracked file if it is in the list of files
+		// a reset should leave untracked files alone
+		file := nameFromAction(&ch)
+		if status.File(file).Worktree != Untracked || inFiles(files, file) {
```
onee-only, line 491: Can we use `ch` instead of status? I think something like

```go
action, err := ch.Action()
if err != nil {
	return err
}

if action != merkletrie.Insert {
    // checkout the file
}
```

Would work since "Untracked" would mean that the file appears in worktree but not in staging.

For reference, `Status()` also works this way internally.

https://github.com/go-git/go-git/blob/93ed29d8f65c1f879d438a5370c212ffedfe5030/worktree_status.go#L93-L118

This way we can avoid re-doing computations that we already have done.

onee-only, line 491: Is there any specific reason that we check `inFiles` here? I think it is already done in block above.

k-anshul: Hey @onee-only  

Thanks for the reviews. I just realised that I don't need the check at all and I can directly do checkout. 
The `files` being passed is always going to just include the files which need changes and no other files.

k-anshul: Hey 

You were right actually. I just removed the complete condition.

UnseenBook:
I have been looking for a solution to the same problem. I have tried these changes and they work in most cases, but when checking out to a branch with the same commit, so no files change, it still deletes all untracked files (including files that are in the gitignore).
From brief manual testing I think it's because when no tracked files change the new list of `removedFiles` is empty, so it is not checked. If I understand correctly, the list of files passed to `resetWorktree` are supposed to be the only files that need to be reset in the worktree. (I have been trying to piece together the logic behind all this and find it very difficult to figure out why certain things are done the way they are. Not that there is anything wrong with that, just saying that I might be completely wrong.)
If my assumption is correct then the logic in `resetWorktree` has to change slightly. If the list of files passed to it is empty the for loop can be skipped, because there is no work to be done. I don't know if the whole function can be skipped, because there is also some index writing done, but I don't know if that needs to stay or not.
There might be other solutions, but from my brief testing this solved the issue of deleted untracked files. Tomorrow I will do some more testing.

k-anshul:
Hey @UnseenBook 

Thanks for looking into this and yes it does fail when reseting on to same commit. I am also discovering some more edge cases. I will take a look at it later again.

k-anshul:
> @k-anshul I have created a PR to your branch with changes that I believe are what we need to solve this issue. I realized that a hard reset does need to remove untracked and ignored files. So I made it only skip `resetWorktree` if the removedFiles slice is empty AND when mode is set to MergeReset, so that HardReset still uses the list of files provided in the ResetOptions.
> 
> The tests run successfully with this change. I hope this helps.

Sounds good @UnseenBook. Just merged your changes in my PR. Thanks for looking into this.

UnseenBook:
Thanks @k-anshul !
What do you expect there is left to fully fix this issue? You mentioned you encountered more edge cases, might they be fixed with my changes, or are they very different? I would love to help :)

k-anshul:
> Thanks @k-anshul ! What do you expect there is left to fully fix this issue? You mentioned you encountered more edge cases, might they be fixed with my changes, or are they very different? I would love to help :)

I could not reproduce the issues after your fix :)

jacksoncalvert:
@k-anshul  can you adjust the commit message to get this merged please?

k-anshul:
> @k-anshul can you adjust the commit message to get this merged please?

Done @jacksoncalvert

jacksoncalvert:
@onee-only  we are seeing this problem in multiple instances. can we get this merged please

jacksoncalvert:
@pjbgf Hi Paulo, can we please get this merged in. My team is using this package and this issue is hugely blocking for some of our tasks. 

Cheers

UnseenBook:
If it's desired to have a test for this fix, I created a PR to add one: https://github.com/rilldata/go-git/pull/4

@k-anshul If you merge the PR by using the rebase option, it should not create a merge commit that would fail the commit message check.

pjbgf (approved):
@k-anshul thanks for working on this. :bow:

Ch00k:
Is there any chance this change could be incorporated in one of the upcoming v5.x.x releases?

pjbgf:
@Ch00k feel free to create a [back-port](https://github.com/go-git/go-git/pull/1564/changes/aa09c632f192bba78b2e37526ffee00070269845) targeting the v5 release branch and tagging me for a review.

PR #1555, opened 2025-05-28
git: remote, Add remote ref not found error, Fixes #1480

### LINKED ISSUE #1480: Missing error variable for "couldn't find remote ref"
If you perform a fetch and the refspecs don't match, the error that is returned contains the string "couldn't find remote ref". There is currently no built-in error to match it against so instead this workaround is required:

```go
if !strings.Contains(err.Error(), "couldn't find remote ref")
```

Perhaps this is because the error originates from the HTTP layer rather than the library itself, but it would be nice to capture this internally and pass on a matchable error so that the above snippet can be replaced with:

```go
if err != git.ErrRemoteRefNotFound
```

CLBRITTON2:


CLBRITTON2:
Targeted wrong branch - Close

Issue #1197, opened 2024-10-01
config parsing fails with "Empty subsection name"


srijan-paul:
If the `.git/config` file has an empty subsection name like this:

```toml
[remote ""]
url = "foo"
fetch = "bar"
```

The official git implementation works fine, but go-git fails to parse the config with an error that says "empty subsection name".
Found out while using go-git at work to analyze git repos.

I'd be happy to send a patch to make go-git behave the same as git.

Minimal reproduction:

```go
package main

import "github.com/go-git/go-git/v5"

func must(err error) {
	if err != nil {
		panic(err)
	}
}

func main() {
       // somerepo/.git/config should have an empty toml subsection.
	repo, err := git.PlainOpen("./somerepo")
	must(err)

	remotes, err := repo.Remotes()
	must(err) // panics here

	for _, remote := range remotes {
		println(remote.Config().Name)
	}
```

CLBRITTON2:
I can probably take a look at this

CLBRITTON2:
@pjbgf I don't think it would be incredibly difficult to adapt `go-git/gcfg` to support. 
I didn't dive too deep but it seemed like removing the empty name check/error and it's test case from gcfg allowed tests to pass.
I'd have to look into what go-git tests would need modifying to allow empty subsection names.
That would at least allow empty subsections to exist in a config without throwing errors.
Adapting to match upstream `git` might be a little bit more involved. I believe both config parsing and encoding would have to be modified to capture options that fall under an empty subsection name.
I'll continue looking into this.

PR #1562, opened 2025-06-04
plumbing: format/config, remove empty config subsection name test

CLBRITTON2:
step 1 in fix for issue #1197
point go.mod to fork of gcfg that removes empty subsection name check and error
This will stop go-git from throwing an error when parsing a config file with an empty subsection name.
To match git we'll have to account for options that fall under empty subsection names like URLs for example during config marshal/unmarshal.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -379,6 +379,21 @@ func (c *Config) unmarshalRemotes() error {
 		c.Remotes[r.Name] = r
 	}
 
+	// Check if the main remote section still has options
+	// after unmarshaling named subsections - this indicates an empty subsection name
+	if len(s.Options) > 0 {
+		emptySubsection := &format.Subsection{
+			Name:    "",
+			Options: s.Options,
+		}
+
+		r := &RemoteConfig{}
+		if err := r.unmarshal(emptySubsection); err != nil {
+			return err
+		}
+
+		c.Remotes[r.Name] = r
+	}
 	// Apply insteadOf url rules
 	for _, r := range c.Remotes {
 		r.applyURLRules(c.URLs)
--- a/plumbing/format/config/decoder_test.go
+++ b/plumbing/format/config/decoder_test.go
@@ -47,12 +47,12 @@ func (s *DecoderSuite) TestDecodeFailsWithEmptySectionName() {
 	decodeFails(s, t)
 }
 
-func (s *DecoderSuite) TestDecodeFailsWithEmptySubsectionName() {
+func (s *DecoderSuite) TestDecodeSucceedsWithEmptySubsectionName() {
 	t := `
 	[remote ""]
 	key=value
 	`
-	decodeFails(s, t)
+	decodeSucceeds(s, t)
 }
 
 func (s *DecoderSuite) TestDecodeFailsWithBadSubsectionName() {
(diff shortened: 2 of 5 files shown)


plumbing/format/config/decoder_test.go
```diff
@@ -47,14 +47,6 @@ func (s *DecoderSuite) TestDecodeFailsWithEmptySectionName() {
 	decodeFails(s, t)
 }
 
-func (s *DecoderSuite) TestDecodeFailsWithEmptySubsectionName() {
```
pjbgf, line 50: Should we not keep this test and change the assertion to confirm that it does not fail instead?

CLBRITTON2: Yeah, it does make sense to keep the test in place for regression/version mismatch

repository_test.go
```diff
@@ -407,7 +407,7 @@ func (s *RepositorySuite) TestEmptyCreateBranch() {
 	r, _ := Init(memory.NewStorage())
 	err := r.CreateBranch(&config.Branch{})
 
-	s.NotNil(err)
+	s.Nil(err)
```
pjbgf:
```suggestion
	s.NoError(err)
```

repository_test.go
```diff
@@ -605,7 +605,7 @@ func (s *RepositorySuite) TestCreateBranchInvalid() {
...
```
pjbgf:
```suggestion
	s.Error(err)
```

config/branch_test.go
```diff
@@ -16,17 +16,17 @@ func TestBranchSuite(t *testing.T) {
...
 	}
-	badBranch := Branch{
+	noNameBranch := Branch{
 		Remote: "some_remote",
 		Merge:  "refs/heads/master",
 	}
-	b.Nil(goodBranch.Validate())
-	b.NotNil(badBranch.Validate())
+	b.Nil(namedBranch.Validate())
+	b.Nil(noNameBranch.Validate())
```
pjbgf:
```suggestion
	b.NoError(namedBranch.Validate())
	b.Error(noNameBranch.Validate())
```

config/branch.go
```diff
@@ -54,6 +49,10 @@ func (b *Branch) Validate() error {
 		return errBranchInvalidRebase
 	}
 
+	// Skip validation on empty branch names to avoid violating git ref format rules
+	if b.Name == "" {
+		return nil
+	}
```
pjbgf: Not sure we should change the existing branch validation, name seems to be compulsory here.

plumbing/format/config/decoder_test.go
```diff
@@ -98,10 +98,16 @@ func decodeFails(s *DecoderSuite, text string) {
 	s.NotNil(err)
 }
 
-func FuzzDecoder(f *testing.F) {
+func decodeSucceeds(s *DecoderSuite, text string) {
+	r := bytes.NewReader([]byte(text))
+	d := NewDecoder(r)
+	cfg := &Config{}
+	err := d.Decode(cfg)
+	s.Nil(err)
```
pjbgf: Add a check to confirm the option value is properly loaded. We may need to change it from `key` to something already in place inside `Config`.

config/config.go
```diff
@@ -379,6 +379,21 @@ func (c *Config) unmarshalRemotes() error {
...
+			Options: s.Options,
+		}
+
+		r := &RemoteConfig{}
+		if err := r.unmarshal(emptySubsection); err != nil {
+			return err
+		}
+
+		c.Remotes[r.Name] = r
+	}
```
pjbgf, line 396: Would this not lead to options of a different subsections to end up in the empty remote section? Please add tests to that effect. We probably need at least two test cases, one that the first remote is unnamed and the second isn't, and the next which is revered. They should have different options so that we can verify we are loading the correct values.

Assuming this works as intended, we probably could remove this `if` block and amend the original `for` to:

```golang
	for _, sub := range s.Subsections {
		r := &RemoteConfig{}
		if len(s.Options) > 0 {
			sub.Options = s.Options
		}

		if err := r.unmarshal(sub); err != nil {
			return err
		}
		c.Remotes[r.Name] = r
	}
```

pjbgf (approved):
@CLBRITTON2 really good stuff! Thank you for working on this. :bow:

PR #1576, opened 2025-06-15
status: preloadWithStatus should return the full path, not just file name

### LINKED ISSUE #1384: worktree status with preload option is broken
Currently, when fetching the `Worktree` status with `Preload` strategy, it stores only file names in the resulting `Status` struct without the relative path to repository root.
As a result, querying `IsUntracked` on it returns true all the time except for files at the repository root.

It seems that using `node.Name()` is wrong in `preloadStatus` file [`status.go#142` ](https://github.com/go-git/go-git/blob/master/status.go#L142)

It should use the node.path but it’s not exposed by the interface.

davidrios:
Not sure if this is the correct way, but this fixes #1384

### DIFF
--- a/status.go
+++ b/status.go
@@ -118,6 +118,11 @@ func (s StatusStrategy) new(w *Worktree) (Status, error) {
 	return nil, fmt.Errorf("%w: %+v", ErrUnsupportedStatusStrategy, s)
 }
 
+type NodeWithPath struct {
+	Node noder.Noder
+	Path string
+}
+
 func preloadStatus(w *Worktree) (Status, error) {
 	idx, err := w.r.Storer.Index()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


status.go
```diff
@@ -118,28 +118,44 @@ func (s StatusStrategy) new(w *Worktree) (Status, error) {
...
+					path = node.Path + "/" + child.Name()
+				}
+				nodeWP := NodeWithPath{Node: child, Path: path}
+				childrenWP = append(childrenWP, nodeWP)
+			}
+			nodes = append(nodes, childrenWP...)
 			continue
 		}
-		fs := status.File(node.Name())
+		fs := status.File(node.Path)
```
pjbgf: If the only thing we need is the full node path, I believe we could do that with: 
```suggestion
		fs := status.File(node.String())
```

In that case, we could drop the rest of the changes.

pjbgf (changes requested):
@davidrios please add some tests to confirm the expected behaviour.

status.go
davidrios: Good one

pjbgf:
This PR was targeting the wrong branch. I cherry-picked the changes and updated based on the latest `main`. Closing this in favour of that PR.

Issue #1578, opened 2025-06-16
Extra commit headers are ignored


jrobsonchase:
Git-compatible tools such as [jujutsu](https://github.com/jj-vcs/jj) may encode extra information in the git commit headers. 

Currently, the [Commit.Decode](https://github.com/go-git/go-git/blob/v5.16.2/plumbing/object/commit.go#L186) method skips all headers that it doesn't know about.

This loses information and leads to false negative GPG signature verifications, as git will both preserve unknown headers in the object and keep them in the signature payload. Since `go-git` omits them, it will always believe that commits containing unknown headers have an invalid signature, despite git giving them the all-clear.

pjbgf:
Closed by https://github.com/go-git/go-git/pull/1627.

PR #1579, opened 2025-06-16
Keep raw headers around to preserve ordering and unknown extra headers

jrobsonchase:
WIP fix for #1578.

Preserves extra headers for a very narrow usecase - re-encoding them exactly as they were in the original object. This allows for the successful PGP validation of commits containing such extra headers, such as jj's `change-id`. A test for this specific case is provided, and all existing tests currently pass.

There's still some "quirky" behavior around the presence of the raw headers causing the stuctured fields to be ignored for encoding. This probably needs to be resolved as it can result in some surprising behavior (see incidental test changes where I had to clear the `headersRaw` field to get them to pass). Not sure how exactly to present that to the user though.

### DIFF
--- a/plumbing/object/commit.go
+++ b/plumbing/object/commit.go
@@ -63,7 +63,13 @@ type Commit struct {
 	// Encoding is the encoding of the commit.
 	Encoding MessageEncoding
 
-	s storer.EncodedObjectStorer
+	headersRaw []headerEntry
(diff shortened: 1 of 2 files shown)


baloo:
I missed your PR and issue and made duplicates.

This should be fixed by https://github.com/go-git/go-git/pull/1627 and there is a backport in https://github.com/go-git/go-git/pull/1633

PR #1583, opened 2025-06-17
fix: OnInflatedObjectContent doesn't include object contents

bored-engineer:
In #799 [packfile.Parser](https://pkg.go.dev/github.com/go-git/go-git/v5@v5.16.2/plumbing/format/packfile#Parser) was refactored to reduce memory usage and no longer properly invokes the [packfile.Observer](https://pkg.go.dev/github.com/go-git/go-git/v5@v5.16.2/plumbing/format/packfile#Observer) interface. Specifically, `OnInflatedObjectContent` is invoked but `content` will always be `nil`.

This PR restores the `content` bytes for callers that are directly using [packfile.Parser](https://pkg.go.dev/github.com/go-git/go-git/v5@v5.16.2/plumbing/format/packfile#Parser) outside of the more standard workflow where the values are stored in a storer.

### DIFF
--- a/plumbing/format/packfile/parser.go
+++ b/plumbing/format/packfile/parser.go
@@ -330,7 +330,7 @@ func (p *Parser) resolveDeltas() error {
 			return err
 		}
 
-		if err := p.onInflatedObjectContent(obj.SHA1, obj.Offset, obj.Crc32, nil); err != nil {
+		if err := p.onInflatedObjectContent(obj.SHA1, obj.Offset, obj.Crc32, buf.Bytes()); err != nil {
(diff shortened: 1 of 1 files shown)


pjbgf:
@bored-engineer thanks for opening this PR. We are trying to limit changes to the existing v5, please propose this against the `main` branch instead.

It would make sense to have this an [opt-in](https://github.com/bored-engineer/go-git/blob/main/plumbing/format/packfile/parser_options.go) - due to the performance implications. To do this safely we will have to return a copy of the underlying array.

Issue #1156, opened 2024-07-23
Lost data due to clone logic


syntastical:
Hi there,

I was just working on a project using this library and I made the unfortunate mistake of cloning to my desktop thinking that a folder would be created based on the repo name, like the git cli does.   However, what ended up happening was that all the files on my Desktop were deleted and the git repo was clone directly to the desktop.  I partially blame myself for being more careful, but I wouldn't expect clone to clear a folder out.  I feel like the default logic should not clear an existing directory, and you could add a flag to the clone options to delete if you need it, then error if the directory already exists and isn't empty.  If that's not acceptable, I'd like to request that this deletion logic is boldly documented.

onee-only:
Since #1594 is merged, this issue can be closed.

AriehSchneier:
Closing as per comment above

PR #1594, opened 2025-06-30
git: repository, add empty dir check to clone

CLBRITTON2:
Proposed fix for #1156
Opening as a draft PR to see whether or not the cleanup logic is necessary.
I recycled the cleanup function into an empty dir check to prevent unwanted file deletions on error, but I think an early error would allow the cleanup logic to stay in place without interfering with a user's expectations if we want to keep it.


repository.go
```diff
@@ -491,21 +490,26 @@ func dotGitCommonDirectory(fs billy.Filesystem) (commonDir billy.Filesystem, err
...
-// ErrRepositoryAlreadyExists is returned.
+// ErrCloneDirNotEmpty is returned.
 func PlainClone(path string, o *CloneOptions) (*Repository, error) {
 	return PlainCloneContext(context.Background(), path, o)
 }
 
 // PlainCloneContext a repository into the path with the given options, isBare
 // defines if the new repository will be bare or normal. If the path is not empty
-// ErrRepositoryAlreadyExists is returned.
+// ErrCloneDirNotEmpty is returned.
```
pjbgf: I'd keep the existing error, as that more closely aligns with upstream.

repository.go
```diff
@@ -491,21 +489,26 @@ func dotGitCommonDirectory(fs billy.Filesystem) (commonDir billy.Filesystem, err
 
 // PlainClone a repository into the path with the given options, isBare defines
 // if the new repository will be bare or normal. If the path is not empty
-// ErrRepositoryAlreadyExists is returned.
+// ErrCloneDirNotEmpty is returned.
```
pjbgf:
```suggestion
// ErrTargetDirNotEmpty is returned.
```

repository.go
```diff
@@ -491,21 +489,26 @@ func dotGitCommonDirectory(fs billy.Filesystem) (commonDir billy.Filesystem, err
...
 func PlainClone(path string, o *CloneOptions) (*Repository, error) {
 	return PlainCloneContext(context.Background(), path, o)
 }
 
 // PlainCloneContext a repository into the path with the given options, isBare
 // defines if the new repository will be bare or normal. If the path is not empty
-// ErrRepositoryAlreadyExists is returned.
+// ErrCloneDirNotEmpty is returned.
```
pjbgf:
```suggestion
// ErrTargetDirNotEmpty is returned.
```

pjbgf (commented):
@CLBRITTON2 I'll take a proper look later on this week. But at first, there are some comments that need changing as per below.

Issue #1616, opened 2025-08-07
Double square brackets [[ around ipv6 addresses


harrys522:
I saw from the changes that followed the previous ipv6 related issue https://github.com/go-git/go-git/issues/740 that go-git inserts square brackets if it figures out that the address is ipv6, fine, but I think maybe one of the networking libraries has also done this recently which means that when the library got bumped now it inserts the brackets twice. 

I had to fork this to make progress in my project since it seemed like v5 didn't really support ipv6 over ssh transport (only http) either.

https://github.com/harrys522/go-git/commit/1b787f2236afbcef1f40159ce99edcf1438bcd52

kristiangronas:
Looks like #1706 fixes this, but it's currently only in the main branch

PR #1619, opened 2025-08-08
Ipv6 double square brackets

harrys522:
It seems like a networking library duplicates the formatting logic that was added in https://github.com/go-git/go-git/commit/4ec6b3f4fa9cdfe8f10d0953ac7d398d01a90f17

Without removing this code, I am unable to clone via ssh from an ipv6 address as this logic will add a second set of square brackets around the address [[ipv6_address]]:22 resulting in no port found (not invalid address?)

https://github.com/go-git/go-git/issues/1616


pjbgf:
@harrys522 please add some test to confirm the new behaviour.

PR #1622, opened 2025-08-09
plumbing: fully support TREE, REUC, LINK, UNTR, EOIE, FSMN, IEOT index extensions

christian-roggia:
This pull request introduces full support for the TREE, REUC, LINK, UNTR, FSMN, IEOT, and EOIE index extensions. Partial decoding support for the TREE, EOIE and REUC extensions already existed, but encoding was missing. There are a few other official extensions not yet implemented, which can be added in future updates.

I would appreciate an initial review of these changes as I continue testing and validating support for the new index extensions in our environment.


christian-roggia:
NOTE: The TREE extension decoder has been updated to match the behavior of the C implementation. The decoder should continue reading the entry value until a newline is encountered to ensure the buffer advances correctly to the next entry. Additionally, invalidated TREE entries should be preserved rather than discarded as in the original logic. Preserving these entries retains valuable information and enables re-encoding the index byte-for-byte exactly as intended.

christian-roggia:
NOTE: The REUC decoder has been updated to correctly decode stages in the intended order. Previously, iterating over the map caused a random order since maps are unordered and iteration order is not guaranteed.

plumbing/format/index/encoder.go
```diff
@@ -206,6 +210,254 @@ func (e *Encoder) encodeRawExtension(signature string, data []byte) error {
 	return nil
 }
 
+func (e *Encoder) encodeExtensions(idx *Index) error {
+	// Note: always write EOIE first to mark the boundary.
+	if idx.EndOfIndexEntry != nil {
+		if err := e.encodeEOIE(idx.EndOfIndexEntry); err != nil {
+			return err
+		}
+	}
```
christian-roggia: Reading the documentation again - EOIE is supposed to be the last extension in the index file.

plumbing/ewah/ewah_io.go
```diff
@@ -0,0 +1,60 @@
+package ewah
+
+import (
+	"encoding/binary"
+	"io"
+)
+
+func ReadFrom(r io.Reader) (*Bitmap, error) {
+	var bits uint32
```
pjbgf, line 25: Missing a `nil` check for `r`.

plumbing/ewah/bitmap.go
```diff
@@ -0,0 +1,94 @@
...
+
+const (
+	RLWRunningBits = 32
+	RLWLiteralBits = 64 - 1 - RLWRunningBits
+
+	RLWLargestRunningCount = (1 << RLWRunningBits) - 1
+	RLWLargestLiteralCount = (1 << RLWLiteralBits) - 1
+)
+
+func GetRunBit(rlw uint64) bool {
```
pjbgf:
```suggestion
// RunBit returns whether the run bit in rlw is set.
func RunBit(rlw uint64) bool {
```

plumbing/ewah/bitmap.go
```diff
@@ -0,0 +1,94 @@
...
+	return rlw&1 != 0
+}
+
+func GetRunningLen(rlw uint64) uint64 {
```
pjbgf:
```suggestion
// RunningLen extracts rlw's running length.
func RunningLen(rlw uint64) uint64 {
```

plumbing/ewah/bitmap.go
```diff
@@ -0,0 +1,94 @@
...
+	return uint64((rlw >> 1) & RLWLargestRunningCount)
+}
+
+func GetLiteralWords(rlw uint64) uint64 {
```
pjbgf:
```suggestion
// LiteralWords extracts the number of literal words in rlw.
func LiteralWords(rlw uint64) uint64 {
```

plumbing/ewah/bitmap.go
```diff
@@ -0,0 +1,94 @@
...
+			return (litWord & (1 << (pos % 64))) != 0
+		}
+		pos -= literals * 64
+
+		i++
+	}
+	return false
+}
+
+// ForEach calls fn() for each set bit.
```
pjbgf:
```suggestion
// ForEach calls fn() for each set bit.
// The returning bool from fn defines whether iteration should continue.
```

plumbing/ewah/bitmap.go
```diff
@@ -0,0 +1,94 @@
...
+				}
+				pos++
+			}
+
+			pos += 64
+		}
+	}
+}
+
+func (b *Bitmap) NumBits() uint64 {
```
pjbgf, line 193: What's the core difference between NumBits and Bits? Please document both funcs.

plumbing/ewah/bitmap.go
```diff
@@ -0,0 +1,94 @@
...
+
+func GetRunningLen(rlw uint64) uint64 {
+	return uint64((rlw >> 1) & RLWLargestRunningCount)
+}
+
+func GetLiteralWords(rlw uint64) uint64 {
+	return uint64(rlw >> (1 + RLWRunningBits))
+}
+
+func (b *Bitmap) Get(pos uint64) bool {
```
pjbgf: Wouldn't `At` be a better name here? I'm assuming this checks whether a bit is set at a given position. Is that right?

```suggestion
func (b *Bitmap) At(pos uint64) bool {
```

Please document this func.

plumbing/format/index/encoder.go
```diff
@@ -206,6 +210,324 @@ func (e *Encoder) encodeRawExtension(signature string, data []byte) error {
...
+
+	if idx.ResolveUndo != nil {
+		if err := e.encodeREUC(idx.ResolveUndo); err != nil {
+			return err
+		}
+	}
+
```
pjbgf: Is this not duplicated with L233-L237?

```suggestion
```

pjbgf (changes requested):
@christian-roggia thanks for looking into this. The changes are looking good, although I need to take a closer look around the extensions on a follow-up review.

Please add some tests around the `ewah` code and rebase the PR.

christian-roggia:
Thanks for the review! All comments addressed and rebased onto the latest upstream main.
Ready for another look.

plumbing/format/index/encoder.go
```diff
@@ -232,6 +240,322 @@ func (e *Encoder) encodeRawExtension(signature string, data []byte) error {
...
+		if _, err := buf.WriteString(i.Path); err != nil {
+			return err
+		}
+		if err := buf.WriteByte(0); err != nil {
+			return err
+		}
+
+		for _, stage := range []Stage{AncestorMode, OurMode, TheirMode} {
+			if _, ok := i.Stages[stage]; ok {
+				if _, err := buf.WriteString(strconv.FormatInt(int64(stage), 10)); err != nil {
```
Soph: These three fields are git's octal **file modes** for stages 1-3, not the stage numbers.

[`resolve-undo.c:48-49`](https://github.com/git/git/blob/2c78326f810173a4f3aefd8021f1e07575412481/resolve-undo.c#L48-L49):

```c
for (i = 0; i < 3; i++)
        strbuf_addf(sb, "%o%c", ui->mode[i], 0);   /* "100644\0" */
```

This writes the stage index (`1`, `2`, `3`) in base 10 instead. Two errors are compounding:

- the value is the stage, not the mode
- the base is decimal, not octal, which is masked here only because 1/2/3 are identical in both

git accepts the result without complaint: `1`/`2`/`3` are valid octal and non-zero, so [`resolve_undo_read`](https://github.com/git/git/blob/2c78326f810173a4f3aefd8021f1e07575412481/resolve-undo.c#L84-L93) goes on to read exactly the three hashes that follow and the byte count still lines up. But every recorded conflict comes back with mode `000001`/`000002`/`000003`, so a later `git checkout -m` recreates it with nonsense modes.

The root cause sits upstream of the encoder: `ResolveUndoEntry.Stages` is `map[Stage]plumbing.Hash`, with nowhere to put a mode, and `readStage` parses the octal mode correctly only to use it as a presence flag and discard it. Decoding was already lossy; adding an encoder turns that loss into bad data on write.

Fixing it means carrying a mode per stage on `ResolveUndoEntry`, which is a breaking change to a pre-existing type. That looks fine for v6, but worth confirming before you invest in it.

Note that `TestExtensions_REUC` can't catch this today: it compares `Path` and the stage hashes, and the mode isn't in the model at all.

plumbing/format/index/decoder.go
```diff
@@ -579,10 +619,348 @@ func (d *endOfIndexEntryDecoder) Decode(e *EndOfIndexEntry) error {
...
+
+	return nil
+}
+
+type linkExtensionDecoder struct {
+	r *bufio.Reader
+}
+
+func (d *linkExtensionDecoder) Decode(ext *Link) error {
+	if _, err := ext.ObjectID.ReadFrom(d.r); err != nil {
```
Soph, line 656: `ext.ObjectID` is zero-valued here, so this reads 20 bytes regardless of the repository's object format.

`plumbing.ObjectID` starts at `format.UnsetObjectFormat`, and `ObjectFormat.Size()` falls through to `default: return DefaultObjectFormat.Size()`, which is SHA-1 at 20 bytes. On a SHA-256 repository there are 32 bytes on disk, so this desyncs the stream by 12 and everything after it is garbage.

The three pre-existing extension decoders avoid this by carrying the hash and sizing every read, e.g. in `treeExtensionDecoder`:

```go
e.Hash.ResetBySize(d.h.Size())
```

`linkExtensionDecoder` is declared as `struct { r *bufio.Reader }` with no hash field, so it can't. Same fix as the others: add `h hash.Hash`, and `ResetBySize(d.h.Size())` before the read. The encoder side needs it too, since `Bytes()` also returns `Size()` bytes.

plumbing/format/index/decoder.go
```diff
@@ -579,10 +619,348 @@ func (d *endOfIndexEntryDecoder) Decode(e *EndOfIndexEntry) error {
...
+	}
+
+	flags, err := binary.ReadUint32(d.r)
+	if err != nil {
+		return err
+	}
+
+	ext.DirFlags = flags
+
+	if _, err := ext.InfoExcludeHash.ReadFrom(d.r); err != nil {
```
Soph, line 728: Same SHA-256 problem as `linkExtensionDecoder`: `untrackedCacheDecoder` is `struct { r *bufio.Reader }` with no hash field, so `InfoExcludeHash`, `ExcludesFileHash`, and every element of `Hashes` (line 793) read 20 bytes on a SHA-256 repository where 32 were written.

`fsMonitorDecoder` reads no object IDs, so LINK and UNTR are the only two affected.

plumbing/format/index/encoder.go
```diff
@@ -232,6 +240,322 @@ func (e *Encoder) encodeRawExtension(signature string, data []byte) error {
...
+	if idx.EndOfIndexEntry != nil {
+		if err := e.encodeEOIE(idx.EndOfIndexEntry); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (e *Encoder) encodeEOIE(ext *EndOfIndexEntry) error {
```
Soph: EOIE's offset and hash are derived from the byte layout of the file being written: the offset points at the start of the extension section, and the hash covers the (signature, size) headers of the extensions preceding it. This writes back whatever was decoded, so both go stale as soon as anything changes, and go-git rewrites the index after every `Add`, `Commit`, and `Checkout`.

To be clear about the blast radius: nothing corrupts. git validates the recorded hash against the extension headers it actually finds and bails out to sequential parsing on any mismatch -- [`read-cache.c:3624-3625`](https://github.com/git/git/blob/2c78326f810173a4f3aefd8021f1e07575412481/read-cache.c#L3624-L3625):

```c
if (!hasheq(hash, (const unsigned char *)index, the_repository->hash_algo))
        return 0;
```

That is one of several `return 0` bail-outs in [`read_eoie_extension`](https://github.com/git/git/blob/2c78326f810173a4f3aefd8021f1e07575412481/read-cache.c#L3548-L3632). The practical effect is that go-git writes an optimisation record that silently never applies.

Related: the extension write order in `encodeExtensions` also differs from git's, which changes the hash input as well.

Worth either computing this at write time, or not writing EOIE at all until it can be computed.

plumbing/format/index/encoder.go
```diff
@@ -232,6 +240,322 @@ func (e *Encoder) encodeRawExtension(signature string, data []byte) error {
...
+	if err := binary.WriteUint32(buf, uint32(len(ext.DirtyBitmap))); err != nil {
+		return err
+	}
+	if _, err := buf.Write(ext.DirtyBitmap); err != nil {
+		return err
+	}
+	return e.encodeRawExtension("FSMN", buf.Bytes())
+}
+
+func (e *Encoder) encodeIEOT(ext *EntryOffsetTable) error {
```
Soph, line 640: Same shape as EOIE: these are entry-block offsets into the file being written, replayed from what was decoded rather than recomputed, so they are stale after any change to the entries.

This one has a little more edge to it. Once past the EOIE gate, git seeks straight to `ieot->entries[i].offset` and parses entries from there with no further validation -- [`read-cache.c:2103-2106`](https://github.com/git/git/blob/2c78326f810173a4f3aefd8021f1e07575412481/read-cache.c#L2103-L2106):

```c
for (i = p->ieot_start; i < p->ieot_start + p->ieot_blocks; i++) {
        p->consumed += load_cache_entry_block(p->istate, p->ce_mem_pool,
                p->offset, p->ieot->entries[i].nr, p->mmap, p->ieot->entries[i].offset, NULL);
```

It only gets that far when EOIE validated, since the table is only read when `extension_offset` is non-zero -- [`read-cache.c:2289-2293`](https://github.com/git/git/blob/2c78326f810173a4f3aefd8021f1e07575412481/read-cache.c#L2289-L2293). So in practice a stale EOIE hash ends up protecting a stale IEOT. That safety is incidental rather than designed, though.

plumbing/format/index/encoder.go
```diff
@@ -232,6 +240,322 @@ func (e *Encoder) encodeRawExtension(signature string, data []byte) error {
 	return nil
 }
 
+func (e *Encoder) encodeExtensions(idx *Index) error {
```
Soph, line 305: git writes extensions in a different order -- [`read-cache.c:2973-3063`](https://github.com/git/git/blob/2c78326f810173a4f3aefd8021f1e07575412481/read-cache.c#L2973-L3063):

    IEOT, LINK, TREE, REUC, UNTR, FSMONITOR, sparse, EOIE

versus here:

    TREE, LINK, UNTR, REUC, FSMN, IEOT, EOIE

Parsing is order-independent, so nothing breaks. But it does mean a git-written index won't survive a decode/encode round trip byte-for-byte, which is the goal you described when you changed the TREE decoder to preserve invalidated entries. It also changes the input to the EOIE hash.

plumbing/ewah/bitmap.go
```diff
@@ -0,0 +1,159 @@
...
+
+		runBits := RunningLen(rlw) * 64
+		if pos-seen < runBits {
+			return RunBit(rlw)
+		}
+		seen += runBits
+
+		literals := LiteralWords(rlw)
+		if pos-seen < literals*64 {
+			litWord := b.words[i+(pos-seen)/64]
```
Soph, line 112: `literals` comes from the RLW with no check that `i + literals <= len(b.words)`, so a corrupt or crafted bitmap indexes past the slice and panics rather than returning an error. `ForEach` has the same pattern at `b.words[i+k]`, and `ReadFrom` does no structural validation of the run/literal layout.

Same fuzz-reachable path as the allocation issue in `ewah_io.go`. A bounds check in both loops, or validating the run structure once at read time, would cover both.

plumbing/ewah/bitmap_test.go
```diff
@@ -0,0 +1,182 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/go-git/go-git/v6/plumbing/ewah"
+)
+
+// makeRLW assembles a run-length word from its components.
+func makeRLW(runBit bool, runLen, literalWords uint64) uint64 {
```
Soph, line 15: `makeRLW` and `buildEWAH` build bitmaps from the package's own constants, so this suite is self-consistent by construction: it would pass unchanged if `RLWRunningBits` disagreed with git.

For what it's worth I did check the constants against [`ewah/ewok_rlw.h:24-25`](https://github.com/git/git/blob/2c78326f810173a4f3aefd8021f1e07575412481/ewah/ewok_rlw.h#L24-L25) and they are correct:

```c
#define RLW_RUNNING_BITS (sizeof(eword_t) * 4)                       /* 32 */
#define RLW_LITERAL_BITS (sizeof(eword_t) * 8 - 1 - RLW_RUNNING_BITS) /* 31 */
```

and all three accessors match -- [`rlw_get_run_bit`](https://github.com/git/git/blob/2c78326f810173a4f3aefd8021f1e07575412481/ewah/ewok_rlw.h#L34), [`rlw_get_running_len`](https://github.com/git/git/blob/2c78326f810173a4f3aefd8021f1e07575412481/ewah/ewok_rlw.h#L63), [`rlw_get_literal_words`](https://github.com/git/git/blob/2c78326f810173a4f3aefd8021f1e07575412481/ewah/ewok_rlw.h#L68) -- including the `>> (1 + RLW_RUNNING_BITS)` shift. But these tests aren't what establishes that.

Since the original ask was specifically for ewah coverage: one fixture of real bytes taken from a git-produced index carrying UNTR or FSMN, asserted against known set-bit positions, would pin the format against git rather than against ourselves.

plumbing/format/index/decoder.go
```diff
@@ -579,10 +619,348 @@ func (d *endOfIndexEntryDecoder) Decode(e *EndOfIndexEntry) error {
...
+
+	ext.PerDirIgnoreFile = string(ignoreFile)
+
+	count, err := binary.ReadVariableWidthInt(d.r)
+	if err != nil {
+		return err
+	}
+
+	if count != 0 {
+		ext.Entries = make([]UntrackedCacheEntry, count)
```
Soph: `count` comes straight from a variable-width int in the file, so this is the same unbounded-allocation shape as `ewah.ReadFrom` -- a small crafted extension can request a very large slice. The same fuzz target reaches it.

plumbing/format/index/encoder.go
```diff
@@ -232,6 +236,318 @@ func (e *Encoder) encodeRawExtension(signature string, data []byte) error {
...
+		if err := buf.WriteByte(0); err != nil {
+			return err
+		}
+		if _, err := fmt.Fprintf(buf, "%d %d\n", i.Entries, i.Trees); err != nil {
+			return err
+		}
+		if _, err := buf.Write(i.Hash.Bytes()); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
Soph: checked, this is resolved

plumbing/ewah/bitmap.go
```diff
@@ -0,0 +1,159 @@
...
+			word := b.words[i+k]
+			for j := range uint64(64) {
+				if word&(1<<j) != 0 {
+					if !fn(pos) {
+						return
+					}
+				}
+				pos++
+			}
+		}
```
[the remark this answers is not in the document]
Soph: Imho a false positive

plumbing/format/index/decoder.go
```diff
@@ -579,10 +619,348 @@ func (d *endOfIndexEntryDecoder) Decode(e *EndOfIndexEntry) error {
...
+		for i := range count {
+			entry, err := d.readEntry()
+			if err != nil {
+				return err
+			}
+			ext.Entries[i] = *entry
+		}
```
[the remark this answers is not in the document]
Soph: Imho a false positive

plumbing/format/index/decoder.go
```diff
@@ -579,10 +619,348 @@ func (d *endOfIndexEntryDecoder) Decode(e *EndOfIndexEntry) error {
...
+}
+
+func (d *untrackedCacheDecoder) readEntry() (*UntrackedCacheEntry, error) {
+	e := &UntrackedCacheEntry{}
+
+	entries, err := binary.ReadVariableWidthInt(d.r)
+	if err != nil {
+		return nil, err
+	}
+	e.Entries = make([]string, entries)
```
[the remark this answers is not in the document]
Soph: Imho a false positive

plumbing/format/index/decoder.go
```diff
@@ -579,10 +619,348 @@ func (d *endOfIndexEntryDecoder) Decode(e *EndOfIndexEntry) error {
...
+	}
+	e.Name = string(name)
+
+	for i := range entries {
+		value, err := binary.ReadUntil(d.r, '\x00')
+		if err != nil {
+			return nil, err
+		}
+		e.Entries[i] = string(value)
+	}
```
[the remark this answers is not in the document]
Soph: Imho a false positive

plumbing/format/index/index_test.go
```diff
@@ -116,3 +150,328 @@ func TestIndexGlob(t *testing.T) {
...
+func TestExtensions_UNTR(t *testing.T) {
+	t.Parallel()
+	idx := &Index{
+		Version: 4,
+		UntrackedCache: &UntrackedCache{
+			Environments: []string{"BAR=FOO", "FOO=BAR"},
+
+			InfoExcludeStats: UntrackedCacheStats{
+				CreatedAt:  time.Date(2018, 1, 1, 0, 0, 0, 0, time.Local),
+				ModifiedAt: time.Date(2019, 1, 1, 0, 0, 0, 0, time.Local),
```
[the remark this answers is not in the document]
Soph: Imho a false positive

plumbing/ewah/ewah_io.go
```diff
@@ -0,0 +1,71 @@
...
+
+	var count uint32
+	if err := binary.Read(r, binary.BigEndian, &count); err != nil {
+		return nil, err
+	}
+
+	words := make([]uint64, count)
+	if err := binary.Read(r, binary.BigEndian, &words); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
Soph: Confirming this is still open, and it's worth more attention than the summary suggests.

Two details that sharpen it:

**The allocation happens twice.** `make([]uint64, count)` is one; `binary.Read` allocates its own `[]byte` of the same size internally on the next line. So `count = 0xFFFFFFFF` asks for ~34 GB, twice over, from four bytes of input — all before any payload is read.

**The `io.LimitedReader` doesn't help.** `getExtensionReader` bounds the *read*, not the *allocation*. The `make` has already happened by the time the short read fails, so the extension-length cap gives no protection here.

It's reachable from a crafted index via LINK, UNTR, and FSMN, and `plumbing/format/index/decoder_fuzz_test.go` already runs under ClusterFuzzLite — so this becomes live fuzz surface as soon as this merges.

git bounds the equivalent copy against the remaining length before touching the buffer (`ewah/ewah_io.c:119`):

```c
data_len = st_mult(self->buffer_size, sizeof(eword_t));
if (len < data_len)
        return error("corrupt ewah bitmap: eof in data "
                     "(%"PRIuMAX" bytes short)", ...);
```

Checking `count * 8` against what's actually available before the `make` would close it. `TestReadFromTruncated` covers a short payload but not a large declared count, so a case for that would be worth adding alongside the fix.

Related, same area: `At` and `ForEach` in `bitmap.go` index `b.words[i+...]` using the literal-word count straight from the RLW with no check that it's within `len(b.words)`, so the same class of crafted input panics rather than returning an error.

plumbing/ewah/bitmap.go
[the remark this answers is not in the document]
pjbgf: +1, the build and tests pass. :smiley:

plumbing/format/index/decoder.go
```diff
@@ -554,13 +592,15 @@
...
+	// stage as absent from the conflict.
+	mode, err := strconv.ParseInt(string(ascii), 8, 64)
 	if err != nil {
 		return err
 	}
 
-	if stage != 0 {
-		e.Stages[s] = plumbing.ZeroHash
+	if mode != 0 {
+		e.Stages[s] = ResolveUndoStage{Mode: filemode.FileMode(mode)}
```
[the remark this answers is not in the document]
pjbgf: This is now fixed.

PR #1627, opened 2025-08-15
commit: support extra headers, support jujutsu signed commit

### LINKED ISSUE #1626: EncodeWithoutSignature breaks signatures generated by Jujutsu
The canonical commit representation returned by [`EncodeWithoutSignature`](https://pkg.go.dev/github.com/go-git/go-git/v5/plumbing/object#Commit.EncodeWithoutSignature) will strip out any extra field it doesn't support.

[Jujutsu](https://jj-vcs.github.io) commits would add extra fields like:
```
tree eddf38b125ad5ed8fbd9438b33e0c41662f8739d
parent 7f92e7685ce2cac51a8348d38be665faf1af314d
parent 185bd1d7dbb2464cf744dc33c0993de1965685a6
author Arthur Gautier <...> 1753582512 +0000
committer Arthur Gautier <...> 1754670137 -0700
change-id xkmmqzonpyyrvxvtxsuutmmxrzyyvwon
gpgsig -----BEGIN PGP SIGNATURE-----

 -----END PGP SIGNATURE-----

Foobar
```
(notice the `change-id` which is jujutsu specific).

The canonical representation (the one to be signed) should be:
```
tree eddf38b125ad5ed8fbd9438b33e0c41662f8739d
parent 7f92e7685ce2cac51a8348d38be665faf1af314d
parent 185bd1d7dbb2464cf744dc33c0993de1965685a6
author Arthur Gautier <...> 1753582512 +0000
committer Arthur Gautier <...> 1754670137 -0700
change-id xkmmqzonpyyrvxvtxsuutmmxrzyyvwon

Foobar
```

but instead we get:
```
tree eddf38b125ad5ed8fbd9438b33e0c41662f8739d
parent 7f92e7685ce2cac51a8348d38be665faf1af314d
parent 185bd1d7dbb2464cf744dc33c0993de1965685a6
author Arthur Gautier <...> 1753582512 +0000
committer Arthur Gautier <...> 1754670137 -0700

Foobar
```

Which in effects breaks signatures.

The implementation should instead just filter out the gpgsig field and leave out the other ones.

baloo:
This adds support for extra headers. While git has a set of  ["standard headers"] (`tree`, `parent`, `author`, `committer`, and `encoding`), it will also support [extra headers] when serializing.

The extra headers must come after the standard ones, but they are otherwise freetyped.

[Jujutsu] takes advantage of that to store its own identifier (`change-id`) as an extra header.

Because signatures will cover the hash of the whole commit (standard headers, extra headers and the message. Everything but the signature itself), if we deserialize a commit and then `EncodeWithoutSignature` to get back the "canonical" representation of a commit, if we don't serialize back the extra headers, the hash will no longer match and the signature will fail to verify.

This adds support for parsing and reencoding the extra headers from the original commit and it's expected to fix support for Jujutsu signed commits.

Fixes #1626

["standard headers"]: https://github.com/git/git/blob/724518f3884d8707c5f51428ba98c115818229b8/commit.c#L1450
[extra headers]: https://github.com/git/git/blob/724518f3884d8707c5f51428ba98c115818229b8/commit.c#L1690
[Jujutsu]: https://github.com/jj-vcs/jj


baloo:
Would this be accepted in a backport to `release-v5`?
Edit: hopefully yes: https://github.com/go-git/go-git/pull/1633

baloo:
@pjbgf any chance to get a review here? (Could I ask you'd click the button to run the CI for me? CI runs are restricted for first time contributors)

baloo:
Rebased with fixes in https://github.com/go-git/go-git/pull/1643. 

Turns out I can run actions on my fork: https://github.com/baloo/go-git/pull/1

plumbing/object/commit.go
```diff
@@ -341,6 +345,12 @@ func (c *Commit) encode(o plumbing.EncodedObject, includeSig bool) (err error) {
 		}
 	}
 
+	for _, header := range c.ExtraHeaders {
+		if _, err = fmt.Fprintf(w, "\n%s", header); err != nil {
+			return err
+		}
+	}
+
```
pjbgf: Please add some validation around this logic. I'm not completely sure of all the validation that takes place upstream, but it would be good to explore adding checks for:
- Empty values
- Headers with the same name as core headers (`tree`, `author`, `committer`, `parent`, etc.)
- Header names without values
- Header values without names

Whatever we can confirm upstream blocks, we should also block.

pjbgf (changes requested):
@baloo thanks for looking into this. Overall the changes look good, I just think we are missing some input validation when encoding the extra header into the commits. Can you take a look at that please?

pjbgf:
> Would this be accepted in a backport to `release-v5`?

I'm happy to support the backporting once this PR is merged.

plumbing/object/commit.go
baloo: Does this version works for you?

baloo:
@pjbgf any chance to get another review here? This bug has been a problem in my everyday workflow and I'd love to get it fixed.

Thanks a lot!

baloo:
Deployed this morning. Works as expected, thank you!

PR #1633, opened 2025-08-21
plumbing: support commits extra headers, support jujutsu signed commit [5.x]

### LINKED ISSUE #1626: EncodeWithoutSignature breaks signatures generated by Jujutsu

baloo:
This adds support for extra headers. While git has a set of  ["standard headers"] (`tree`, `parent`, `author`, `committer`, and `encoding`), it will also support [extra headers] when serializing.

The extra headers must come after the standard ones, but they are otherwise freetyped.

[Jujutsu] takes advantage of that to store its own identifier (`change-id`) as an extra header.

Because signatures will cover the hash of the whole commit (standard headers, extra headers and the message. Everything but the signature itself), if we deserialize a commit and then `EncodeWithoutSignature` to get back the "canonical" representation of a commit, if we don't serialize back the extra headers, the hash will no longer match and the signature will fail to verify.

This adds support for parsing and reencoding the extra headers from the original commit and it's expected to fix support for Jujutsu signed commits.

Fixes #1626

["standard headers"]: https://github.com/git/git/blob/724518f3884d8707c5f51428ba98c115818229b8/commit.c#L1450
[extra headers]: https://github.com/git/git/blob/724518f3884d8707c5f51428ba98c115818229b8/commit.c#L1690
[Jujutsu]: https://github.com/jj-vcs/jj

This is a backport of https://github.com/go-git/go-git/pull/1627 targeting `release-5.x`

### DIFF
--- a/plumbing/object/commit.go
+++ b/plumbing/object/commit.go
@@ -62,8 +62,53 @@ type Commit struct {
 	ParentHashes []plumbing.Hash
 	// Encoding is the encoding of the commit.
 	Encoding MessageEncoding
+	// List of extra headers of the commit
+	ExtraHeaders []ExtraHeader
 
 	s storer.EncodedObjectStorer
+}
+
+// ExtraHeader holds any non-standard header
+type ExtraHeader struct {
+	// Header name
+	Key string
+	// Value of the header
+	Value string
+}
+
+// Implement fmt.Formatter for ExtraHeader
+func (h ExtraHeader) Format(f fmt.State, verb rune) {
+	switch verb {
+	case 'v':
+		fmt.Fprintf(f, "ExtraHeader{Key: %v, Value: %v}", h.Key, h.Value)
+	default:
+		fmt.Fprintf(f, "%s", h.Key)
+		if len(h.Value) > 0 {
+			fmt.Fprint(f, " ")
+			// Content may be spread on multiple lines, if so we need to
+			// prepend each of them with a space for "continuation".
+			value := strings.TrimSuffix(h.Value, "\n")
+			lines := strings.Split(value, "\n")
+			fmt.Fprint(f, strings.Join(lines, "\n "))
+		}
+	}
+}
+
+// Parse an extra header and indicate whether it may be continue on the next line
+func parseExtraHeader(line []byte) (ExtraHeader, bool) {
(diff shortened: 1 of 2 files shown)


baloo:
rebased on https://github.com/go-git/go-git/pull/1646 to fix CI

pjbgf (approved):
@baloo thanks for backporting this. 🙇

PR #1635, opened 2025-08-25
fix: race condition accessing `packConnection.stderrBuf`

jcmfernandes:
Go's built-in data race detector keeps reporting a race between the read of `packConnection.stderrBuf` by `checkError()` and the write to it by `ioutil.CopyBufferPool`.

I must confess that I don't fully understand the rationale here. Can't the goroutine that includes `ioutil.CopyBufferPool(&c.stderrBuf, stderr)` run after `Hashshake()` returns? There's no synchronization between the parent and child goroutines; hence, we may miss the stderr.

### DIFF
--- a/plumbing/transport/pack.go
+++ b/plumbing/transport/pack.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"io"
 	"strings"
+	"sync/atomic"
 
 	"github.com/go-git/go-git/v6/plumbing"
 	"github.com/go-git/go-git/v6/plumbing/protocol"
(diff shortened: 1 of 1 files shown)


aymanbagabas (approved):
LGTM! Thank you @jcmfernandes 🙂

pjbgf (approved):
@jcmfernandes thanks for working on this. 🙇

jcmfernandes:
Thank you both @aymanbagabas and @pjbgf (obrigado :smile:)!

PR #1646, opened 2025-09-08
build: raise timeouts for windows CI tests

baloo:


.github/workflows/cifuzz.yml
```diff
@@ -3,6 +3,10 @@ on: [pull_request]
 permissions: {}
 jobs:
  Fuzzing:
+   # Fuzzing has been disabled on maintenance branches:
```
pjbgf: Instead we can just remove this file altogether. I don't expect we supporting fuzzing on a branch that is in maintenance only.

PR #1670, opened 2025-09-25
Update the clone example with no auth required

Blitz-Cloud:
At the moment writing this  the PlainClone prototype asks for 3 parameters. I am guessing that when the version bump took place the code example wasnt updated.

I also set the isBare parameter to false, because i consider that most of the users will not clone a bare repo.
This might be a bad assumption of mine.

### DIFF
--- a/_examples/clone/main.go
+++ b/_examples/clone/main.go
@@ -16,8 +16,9 @@ func main() {
 
 	// Clone the given repository to the given directory
 	Info("git clone %s %s --recursive", url, directory)
-
-	r, err := git.PlainClone(directory, &git.CloneOptions{
+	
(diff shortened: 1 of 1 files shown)


Blitz-Cloud:
Also i didnt checked the other examples to see if they have the same problem or not. I will be back with an answer .

onee-only:
Hello @Blitz-Cloud! 
Currently `main(v6)` takes `isBare` parameter in `CloneOptions`. This change was introduced in bca6653b99d42178bd2d1e504f3b828019c3db90.
Since it hasn’t been included in go-git’s latest release(`v5.16.2`), you might be confused by the difference between the released version and the source.

Blitz-Cloud:
Ok, i understand

PR #1674, opened 2025-09-26
Force git daemon to listen on IPv4 in tests

### LINKED ISSUE #1669: go test ./... failing
When I clone go-git and run `go test ./...`, then the unit test suite fails. An unclean test suite makes it much harder to contribute to the project.

ianling:
Fixes #1669 for me on MacOS 26.

I guess `git daemon <...>` prefers IPv6, even if IPv6 is not configured properly (or is blocked by corporate IT rules). Changing the command here to specifically listen on the IPv4 address `127.0.0.1` gets all tests passing for me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ profile.out
 .tmp/
 .git-dist/
 .vscode
+.idea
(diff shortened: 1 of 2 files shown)


ianling:
@pjbgf -- done, thanks for reviewing

PR #1675, opened 2025-09-28
fix: open repo in worktree

noahfraiture:
This PR aims to fix the `git.PlainOpenWithOptions` when opening a worktree. Previously the err "no reference found" were returned.

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -358,13 +358,13 @@ func PlainOpenWithOptions(path string, o *PlainOpenOptions) (*Repository, error)
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


pjbgf:
~@noahfraiture please add a test to confirm the behaviour your are trying to fix so that we can avoid future regressions.~

pjbgf:
I'm not sure we need this change. Users can reach the same outcome by using `PlainOpenWithOptions("/repo", &PlainOpenOptions{EnableDotGitCommonDir: true})`.

PR #1689, opened 2025-10-09
git: worktree, add cherry-picking

d-fal:
This is a very basic cherry-picking feature that accepts `theirs` version of files. 
It lacks other strategies for now. 
It's equivalent to 
```
git cherry-pick <commit-hash-1> <commit-hash-2>  --strategy-option theirs
```

It solved my needs and I hope it can be useful. wdyt?

----
Update: 

I added both `ours` and `theirs` strategy options

### DIFF
--- a/worktree_commit.go
+++ b/worktree_commit.go
@@ -149,7 +149,6 @@ func (w *Worktree) CherryPick(commitOpts *CommitOptions, ortStrategyOption OrtMe
 				if _, err := w.Remove(change.From.Name); err != nil {
 					return err
 				}
-
 			case merkletrie.Insert, merkletrie.Modify:
 				_, to, err := change.Files()
 				if err != nil {
@@ -172,8 +171,13 @@ func (w *Worktree) CherryPick(commitOpts *CommitOptions, ortStrategyOption OrtMe
 				}
 			}
 		}
-
-		_, err = w.Commit(commit.Message, commitOpts)
+		_, err = w.Commit(commit.Message, &CommitOptions{
+			Author:            &commit.Author,
+			Committer:         commitOpts.Committer,
+			SignKey:           commitOpts.SignKey,
+			Signer:            commitOpts.Signer,
+			AllowEmptyCommits: commitOpts.AllowEmptyCommits,
+		})
 		if err != nil {
 			return err
 		}
--- a/worktree_commit_test.go
+++ b/worktree_commit_test.go
@@ -695,7 +695,6 @@ func (s *WorktreeSuite) TestCherryPick() {
 	// cherry-pick with no commit options
 	err = w.CherryPick(nil, OursMergeStrategy, rmCommit)
 	s.ErrorIs(err, ErrCannotCherryPickWithoutCommitOptions)
-
 }
 func (s *WorktreeSuite) TestCommitTreeSort() {
 	fs := s.TemporalFilesystem()


worktree_commit.go
```diff
@@ -97,6 +97,56 @@ func (w *Worktree) Commit(msg string, opts *CommitOptions) (plumbing.Hash, error
 	return commit, w.updateHEAD(commit)
 }
 
+// Cherry-pick commits into the worktree
+// It always accepts "theirs" version of the files. the order of commits is important. Each commit sits on the top of worktree's current head
+// it resembles `git cherry-pick <commit-hash-1> <commit-hash-2> ... --strategy-option theirs`
+func (w *Worktree) CherryPick(committer *object.Signature, signKey *openpgp.Entity, commits ...*object.Commit) error {
+	for _, commit := range commits {
```
pjbgf: From an API perspective, I think we are probably better off having a `CommitOptions` instead instead of building it ourselves. That would mean that people not signing the commit won't need to always send a `nil` as the second arg.

```suggestion
func (w *Worktree) CherryPick(commitOpts *CommitOptions, commits ...*object.Commit) error {
	for _, commit := range commits {
```

worktree_commit.go
```diff
@@ -97,6 +97,56 @@ func (w *Worktree) Commit(msg string, opts *CommitOptions) (plumbing.Hash, error
...
+		if err != nil {
+			return err
+		}
+
+		_, err = w.Commit(commit.Message, &CommitOptions{
+			Author:            &commit.Author,
+			Committer:         committer,
+			SignKey:           signKey,
+			AllowEmptyCommits: true,
+		})
```
pjbgf, line 180:
```suggestion
		_, err = w.Commit(commit.Message, commitOpts)
```

worktree_commit.go
```diff
@@ -97,6 +97,56 @@ func (w *Worktree) Commit(msg string, opts *CommitOptions) (plumbing.Hash, error
 	return commit, w.updateHEAD(commit)
 }
 
+// Cherry-pick commits into the worktree
+// It always accepts "theirs" version of the files. the order of commits is important. Each commit sits on the top of worktree's current head
+// it resembles `git cherry-pick <commit-hash-1> <commit-hash-2> ... --strategy-option theirs`
+func (w *Worktree) CherryPick(committer *object.Signature, signKey *openpgp.Entity, commits ...*object.Commit) error {
+	for _, commit := range commits {
+		if commit == nil {
```
pjbgf: Not sure I understand the logic here. Would this not be only when `len(commits) == 0`? If a specific commit is empty, and `AllowEmptyCommits` is true, this should not return an error. Or am I missing something?

d-fal: you are right @pjbgf , that's not necessary

worktree_commit.go
d-fal: like it. just one concern, shouldn't we preserve the original committer ?

d-fal:
@pjbgf I noticed I've missed one thing. If the cherry-picked commit contains a file deletion, we need to delete the same file in the worktree. I added the changes.
I also applied your comments.

worktree_commit.go
d-fal: done

worktree_commit.go
pjbgf: You are absolutely right, we would need to use some of the contents of the `CommitOptions`, such as committer, signKey and AllowEmptyCommits.

options.go
```diff
@@ -112,6 +112,19 @@ const (
 	FastForwardMerge MergeStrategy = iota
 )
 
+// ort is the default Merge strategy.  This strategy can only resolve two heads using a 3-way merge algorithm.
+// since git v2.50.0, ort is a synonym of recursive.
+// ort has two options: theirs and ours
+// in the both cases, conflicts would be auto-resolved by accepting existing or incoming changes
```
pjbgf:
```suggestion
// OrtMergeStrategyOption defines the merge strategy options for the ORT merge strategy, which can only resolve two heads using a 3-way merge algorithm.
// Since Git v2.50.0, ORT is synonym of recursive.
```

options.go
```diff
@@ -112,6 +112,19 @@ const (
...
+type MergeStrategyORTOption int8
+
+const (
+	// "theirs" option auto-resolves the changes accepting the incoming version of the changes
+	Theirs MergeStrategyORTOption = iota
+	// "ours" option favors our version of the changes and it's opposite of "theirs"
+	Ours
```
pjbgf: To align with idiomatic Go, I think we could go for something like:
```suggestion
	// TheirsMergeStrategy is a merge strategy option that auto-resolves the changes by accepting the incoming version of the changes.
	TheirsMergeStrategy MergeStrategyORTOption = iota
	
	// OursMergeStrategy is a merge strategy option that auto-resolves the changes by accepting our version of the changes and rejecting the incoming changes.
	OursMergeStrategy
```

worktree_commit.go
```diff
@@ -97,6 +98,83 @@ func (w *Worktree) Commit(msg string, opts *CommitOptions) (plumbing.Hash, error
 	return commit, w.updateHEAD(commit)
 }
 
+// Cherry-pick commits into the worktree
+// It always accepts "theirs" version of the files. the order of commits is important. Each commit sits on the top of worktree's current head
+// it resembles `git cherry-pick <commit-hash-1> <commit-hash-2> ... --strategy-option [theirs,ours]`
```
pjbgf:
```suggestion
// CherryPick cherry picks commits and merge them into the worktree based on the selected 
// merge strategy. Each commit sits on the top of worktree's current head.
// It resembles `git cherry-pick <commit-hash-1> <commit-hash-2> ... --strategy-option [theirs,ours]`
```

worktree_commit.go
```diff
@@ -97,6 +98,83 @@ func (w *Worktree) Commit(msg string, opts *CommitOptions) (plumbing.Hash, error
...
+func (w *Worktree) CherryPick(commitOpts *CommitOptions, ortStrategyOption MergeStrategyORTOption, commits ...*object.Commit) error {
+	for _, commit := range commits {
```
pjbgf, line 111: We should add some validation here for the commitOpts not being empty.

```suggestion
	if commitOpts == nil {
	  return ErrCannotCherryPickWithoutCommitOptions
	}

	for _, commit := range commits {
```

We would need to define the `ErrCannotCherryPickWithoutCommitOptions`

options.go
```diff
@@ -112,6 +112,19 @@ const (
 	FastForwardMerge MergeStrategy = iota
 )
 
+// ort is the default Merge strategy.  This strategy can only resolve two heads using a 3-way merge algorithm.
+// since git v2.50.0, ort is a synonym of recursive.
+// ort has two options: theirs and ours
+// in the both cases, conflicts would be auto-resolved by accepting existing or incoming changes
+type MergeStrategyORTOption int8
```
pjbgf:
```suggestion
type OrtMergeStrategyOption int8
```

worktree_commit.go
```diff
@@ -97,6 +100,87 @@ func (w *Worktree) Commit(msg string, opts *CommitOptions) (plumbing.Hash, error
...
+				if err != nil {
+					return err
+				}
+				if _, err := w.Add(to.Name); err != nil {
+					return err
+				}
+			}
+		}
+
+		_, err = w.Commit(commit.Message, commitOpts)
```
pjbgf: I think we should be OK only taking Author from the current commit, then everything else from the passed in commit options:

```suggestion
		_, err = w.Commit(commit.Message, &CommitOptions{
			Author:            &commit.Author,
			Committer:         commitOpts.Committer,
			SignKey:           commitOpts.SignKey,
			Signer:           commitOpts.Signer,
			AllowEmptyCommits: commitOpts.AllowEmptyCommits,
		})
```

worktree_commit_test.go
```diff
@@ -584,6 +584,119 @@ func (s *WorktreeSuite) TestCommitSignBadKey() {
...
+	s.NoError(err)
+	fooFileContent, err = util.ReadFile(fs, "foo")
+	s.NoError(err)
+	s.Equal("foo", string(fooFileContent))
+
+	// cherry-pick with no commit options
+	err = w.CherryPick(nil, OursMergeStrategy, rmCommit)
+	s.ErrorIs(err, ErrCannotCherryPickWithoutCommitOptions)
+
+}
```
pjbgf:
```suggestion
}

```

worktree_commit.go
d-fal: good one, 
done!

Issue #1679, opened 2025-09-30
Add() should not add ignored files by default


onee-only:
### Feature Description

From https://git-scm.com/docs/git-add:
> The git add command will not add ignored files by default. If any ignored files were explicitly specified on the command line, git add will fail with a list of ignored files. Ignored files reached by directory recursion or filename globbing performed by Git (quote your globs before the shell) will be silently ignored. The git add command can be used to add ignored files with the -f (force) option.

Currently, [`Add()`](https://github.com/go-git/go-git/blob/145daf2492dd86b1d7e7787555ebd9837b93bff8/worktree_status.go#L299) or [`AddWithOptions()`](https://github.com/go-git/go-git/blob/145daf2492dd86b1d7e7787555ebd9837b93bff8/worktree_status.go#L345) doesn't respect ignored files while upstream git does.
go-git should follow the behavior of upstream. So it should respect ignored files and have `Force` option.

MaestroLockedIn:
I'd like to take a stab at this if it has not been assigned to anyone else

onee-only:
Please don't close this issue!

PR #1690, opened 2025-10-12
feat(worktree): support adding ignored files with --force 

MaestroLockedIn:
Issue: Add() should not add ignored files by default #1679 

This PR adds support for force-adding ignored files using the --force option, mirroring standard Git’s behavior (git add -f).

Changes

Updated AddWithOptions():

Supports Force: true to bypass .gitignore patterns.

Modified doAdd() and doAddDirectory() to correctly include ignored files when forced.

Added comprehensive tests:

TestAddWithForceIgnoredFile

Minor cleanup and improved .gitignore pattern handling.

### DIFF
--- a/worktree_status.go
+++ b/worktree_status.go
@@ -318,14 +318,9 @@ func diffTreeIsEquals(a, b noder.Hasher) bool {
 // no error is returned. When path is a file, the blob.Hash is returned.
 func (w *Worktree) Add(path string) (plumbing.Hash, error) {
 	// TODO(mcuadros): deprecate in favor of AddWithOption in v7.
-	patterns := make([]gitignore.Pattern, 0)
-	dir := filepath.Dir(path)
-	if dir == "." || dir == "" {
-		dirPatterns, _ := gitignore.ReadPatterns(w.Filesystem, nil)
-		patterns = append(patterns, dirPatterns...)
-	} else {
-		dirPatterns, _ := gitignore.ReadPatterns(w.Filesystem, []string{dir})
-		patterns = append(patterns, dirPatterns...)
+	patterns, err := gitignore.ReadPatterns(w.Filesystem, nil)
+	if err != nil {
+		return plumbing.ZeroHash, err
 	}
 	return w.doAdd(path, patterns, false)
 }
@@ -376,23 +371,26 @@ func (w *Worktree) AddWithOptions(opts *AddOptions) error {
 		return err
 	}
 
-	patterns := make([]gitignore.Pattern, 0)
-
-	if opts.All {
-		patterns, _ = gitignore.ReadPatterns(w.Filesystem, nil)
-		_, err := w.doAdd(".", patterns, false)
-		return err
-	}
-
 	if opts.Glob != "" {
 		return w.AddGlob(opts.Glob)
 	}
 
+	path := opts.Path
+	skipStatus := opts.SkipStatus
+	if opts.All {
+		path = "."
+		skipStatus = false
+	}
+
+	patterns := make([]gitignore.Pattern, 0)
 	if !opts.Force {
-		dirPatterns, _ := gitignore.ReadPatterns(w.Filesystem, nil)
+		dirPatterns, err := gitignore.ReadPatterns(w.Filesystem, nil)
+		if err != nil {
+			return err
+		}
 		patterns = append(patterns, dirPatterns...)
 	}
-	_, err := w.doAdd(opts.Path, patterns, opts.SkipStatus)
+	_, err := w.doAdd(path, patterns, skipStatus)
 	return err
 }
 
@@ -465,6 +463,11 @@ func (w *Worktree) AddGlob(pattern string) error {
 	}
 
 	// TODO(mcuadros): deprecate in favor of AddWithOption in v6.
+	patterns, err := gitignore.ReadPatterns(w.Filesystem, nil)
+	if err != nil {
+		return err
+	}
+
 	files, err := util.Glob(w.Filesystem, pattern)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


worktree_status.go
```diff
@@ -297,8 +297,17 @@ func diffTreeIsEquals(a, b noder.Hasher) bool {
 // the worktree to the index. If any of the files is already staged in the index
 // no error is returned. When path is a file, the blob.Hash is returned.
 func (w *Worktree) Add(path string) (plumbing.Hash, error) {
-	// TODO(mcuadros): deprecate in favor of AddWithOption in v6.
-	return w.doAdd(path, make([]gitignore.Pattern, 0), false)
+	// TODO(mcuadros): deprecate in favor of AddWithOption in v7.
```
onee-only, line 320: I think we don't have to delay this to v7. Currently v6 isn't released and API changes can be made. See https://github.com/go-git/go-git/issues/910.

worktree_status.go
```diff
@@ -297,8 +297,17 @@ func diffTreeIsEquals(a, b noder.Hasher) bool {
...
+	patterns := make([]gitignore.Pattern, 0)
+	dir := filepath.Dir(path)
+	if dir == "." || dir == "" {
+		dirPatterns, _ := gitignore.ReadPatterns(w.Filesystem, nil)
+		patterns = append(patterns, dirPatterns...)
+	} else {
+		dirPatterns, _ := gitignore.ReadPatterns(w.Filesystem, []string{dir})
```
onee-only: As far as I know, `gitignore.ReadPatterns` reads the directories recursively starting from the specified directory. So I think the case below won't work as expected:
1. root directory contains `.gitignore` with content: `foo`.
2. user adds `foo/bar.txt`

In this case, since it isn't aware of the root's `.gitignore`, it will add `foo/bar.txt`.
Could you please test the code against this case?

worktree_status.go
pjbgf: > I think we don't have to delay this to v7.

I'd concur with this, ideally we would tackle it as part of v6. I'm planning to raise an RFC on the subject of API consolidation which would affect this (and the general "WithOption" funcs), just haven't got around to it yet. In the mean time, we could leave the comment unchanged:

```suggestion
	// TODO(mcuadros): deprecate in favor of AddWithOption in v6.
```

worktree_status.go
MaestroLockedIn: I see what you are talking about that, this is true 
Should it always read from the parent directory

i think it is outside the scope of what was reported

onee-only: We should align the behavior with the upstream git. Maybe reading all the gitignore patterns is the simplest solution for this?

MaestroLockedIn:
Hello @onee-only & @pjbgf 
I see the PR is still open
Just a reminder and also asking if this satisfy the issue raised or if there are still concerns

worktree_status.go
onee-only: @pjbgf Now that the alpha version of v6 has been released, would this be still valid?

worktree_status.go
```diff
@@ -350,16 +359,23 @@ func (w *Worktree) AddWithOptions(opts *AddOptions) error {
...
 	}
 
+	patterns := make([]gitignore.Pattern, 0)
+
 	if opts.All {
-		_, err := w.doAdd(".", w.Excludes, false)
+		patterns, _ = gitignore.ReadPatterns(w.Filesystem, nil)
+		_, err := w.doAdd(".", patterns, false)
 		return err
 	}
```
onee-only: What if opts.All and opts.Force are both true? Maybe:

```golang
path := opts.Path
skipStatus := opts.SkipStatus
if opts.All {
	path = "."
	skipStatus = false
}

// ...

_, err := w.doAdd(path, patterns, skipStatus)
```

worktree_status.go
```diff
@@ -300,8 +300,17 @@ func diffTreeIsEquals(a, b noder.Hasher) bool {
...
+	// TODO(mcuadros): deprecate in favor of AddWithOption in v7.
+	patterns := make([]gitignore.Pattern, 0)
+	dir := filepath.Dir(path)
+	if dir == "." || dir == "" {
+		dirPatterns, _ := gitignore.ReadPatterns(w.Filesystem, nil)
+		patterns = append(patterns, dirPatterns...)
+	} else {
+		dirPatterns, _ := gitignore.ReadPatterns(w.Filesystem, []string{dir})
+		patterns = append(patterns, dirPatterns...)
+	}
```
onee-only:
```suggestion
	patterns, err := gitignore.ReadPatterns(w.Filesystem, nil)
	if err != nil {
		return err
	}
```

Like `AddWithOptions()`, `Add()` also has to read from the root.

worktree_status.go
onee-only, line 489:
```suggestion
	patterns, err := gitignore.ReadPatterns(w.Filesystem, nil)
	if err != nil {
		return patterns
	}
```

`AddGlob` also needs to respect gitignores.

worktree_status.go
```diff
@@ -350,16 +359,23 @@ func (w *Worktree) AddWithOptions(opts *AddOptions) error {
...
 	}
 
 	if opts.Glob != "" {
 		return w.AddGlob(opts.Glob)
 	}
 
-	_, err := w.doAdd(opts.Path, make([]gitignore.Pattern, 0), opts.SkipStatus)
+	if !opts.Force {
+		dirPatterns, _ := gitignore.ReadPatterns(w.Filesystem, nil)
+		patterns = append(patterns, dirPatterns...)
```
onee-only:
```suggestion
		var err error
		patterns, err = gitignore.ReadPatterns(w.Filesystem, nil)
		if err != nil {
			return err
		}
```

Let's add an error check in case ReadPatterns fails.

MaestroLockedIn:
@onee-only I've updated the Pull Request based on the comments

PR #1694, opened 2025-10-19
WIP: allow use of modification times to speed up equals function in merkletrie  diff 

hansbogert:
For my application i rely heavily on the speed of git status, I noticed that go-git does not use modification times where upstream git *would* do so for a git status. I noticed a large performance discrepancy between the mentioned application using -go-git, and a roughly 30 line bash script using upstream `git` which should be doing the functional equivalent.

Depending on the size of the files in a worktree, the use of modtimes instead of rehashing everything results in substantial time savings. 

The existing benchmark for git status, with a rather trivial size for a repo, already shows a ~6x performance increase. I expect this performance increase to be larger with larger repo's.

```
go test -bench=.                      
2025/10/19 22:21:25 comparing mod time: 2025-10-19 22:21:25.886774837 +0200 CEST == 2025-10-19 22:21:25.886774837 +0200 CEST on linux (amd64)
goos: linux
goarch: amd64
cpu: Intel(R) Core(TM) i5-10400F CPU @ 2.90GHz
...
BenchmarkWorktreeStatus-12                                                      	     612	   1919965 ns/op
PASS
ok  	github.com/go-git/go-git/v6	79.787s

```
With this change applied:

```
go test -bench=.
....
BenchmarkWorktreeStatus-12                                                      	    3792	    302731 ns/op
PASS
ok  	github.com/go-git/go-git/v6	47.303s

```

Basically this change first checks if modification time is the same between the index and the worktree, if not proceed as usual by rehashing, if the same, mark as equal.

This could well be very naive and some expert review is  needed. Note that this 'fasthpath' would become the default when doing a `worktree.Status()`. Another option would be to introduce an `Option` for `Status(...)`.

### DIFF
--- a/plumbing/object/treenoder.go
+++ b/plumbing/object/treenoder.go
@@ -2,6 +2,7 @@ package object
 
 import (
 	"io"
+	"time"
 
 	"github.com/go-git/go-git/v6/plumbing"
 	"github.com/go-git/go-git/v6/plumbing/filemode"
@@ -55,6 +56,10 @@ func (t *treeNoder) Hash() []byte {
 		return append(t.hash.Bytes(), filemode.Regular.Bytes()...)
 	}
 	return append(t.hash.Bytes(), t.mode.Bytes()...)
+}
+
+func (t *treeNoder) ModTime() time.Time {
+	return time.Time{}
 }
 
 func (t *treeNoder) Name() string {
--- a/utils/merkletrie/filesystem/node.go
+++ b/utils/merkletrie/filesystem/node.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"time"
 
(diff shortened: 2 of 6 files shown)


hansbogert:
@pjbgf Can this workflow be started? :angel:

pjbgf:
@hansbogert thanks for looking into this. Status indeed has some room for improvement in terms of performance.

At present, I'm not sure how aligned this is with upstream, I'd need to take a proper look at that. If this diverges from upstream, it would need to be an opt-in feature as opposed to changing the default behaviour.

hansbogert:
Yeah please let me know in time. If this is a wanted optimization i can look into the windows flakyness. My first bet is lower time resolution, like milliseconds, maybe you know more?

hansbogert:
No longer needed thanks to https://github.com/go-git/go-git/pull/1747

PR #1697, opened 2025-10-21
status: preloadWithStatus should return the full path, not just file name

pjbgf:
Supersedes #1576.

Fixes https://github.com/go-git/go-git/issues/1384.


pjbgf:
@davidrios thanks for working on this. :bow:

PR #1706, opened 2025-10-30
plumbing: transport, refactor transport Endpoint to use url.URL

aymanbagabas:
There's no need to duplicate the logic already present in url.URL and the standard library to parse and handle URLs. This refactor simplifies the codebase by removing redundant fields and methods, and leverages the existing functionality of url.URL.

This also remove unused code in transport/http ModifyEndpointIfRedirect and its related tests.

### DIFF
--- a/plumbing/transport/transport.go
+++ b/plumbing/transport/transport.go
@@ -13,13 +13,12 @@
 package transport
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
+	"net"
 	"net/url"
 	"regexp"
 	"runtime"
-	"strconv"
 	"strings"
 
 	giturl "github.com/go-git/go-git/v6/internal/url"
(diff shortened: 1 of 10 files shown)


plumbing/transport/transport.go
```diff
@@ -108,56 +96,8 @@ func (o *ProxyOptions) FullURL() (*url.URL, error) {
...
-
-		if u.Host != "" {
-			buf.WriteString(u.Host)
-
-			if u.Port != 0 {
-				port, ok := defaultPorts[strings.ToLower(u.Protocol)]
-				if !ok || ok && port != u.Port {
-					fmt.Fprintf(&buf, ":%d", u.Port)
-				}
-			}
```
pjbgf, line 149: @aymanbagabas would this not cause a regression when `git://` is being used without a port being set?

aymanbagabas: It shouldn't, that's handled by the git transport [here](https://github.com/go-git/go-git/pull/1706/files#diff-120b7b981c492124fc93e76812964c0b59f99446513e1b9eb3c237daebfbb41bR72) and in the `c.connect` function

pjbgf: Nice, thanks for pointing that out. 👍

pjbgf (approved):

PR #1707, opened 2025-10-31
plumbing: Fix NewKnownHostsCallback panic

wrouesnel:
NewKnownHostsCallback attempts to return db.HostKeyCallback() without checking if db is nil, leading to a panic when there is an error initializing the database.


pjbgf (approved):
@wrouesnel thanks for fixing this. 🙇

PR #1708, opened 2025-11-01
git: worktree, Status() takes config from local, global and system

kayoch1n:
On Windows, if the `core.autocrlf` was not set, Worktree.Status will display unchanged files as modified. The Worktree.Status function currently takes config from local file only, and even if `core.autocrlf` is set globally, the `core.autocrlf` won't take effect.

```go
package main

import "github.com/go-git/go-git/v6"

func main() {
	repo, err := git.PlainOpen(".")
	if err != nil {
		panic(err)
	}
	wt, err := repo.Worktree()
	if err != nil {
		panic(err)
	}
	status, err := wt.Status()
	if err != nil {
		panic(err)
	}
	println(status.String())
}

```
屏幕截图 2025-11-01 094916

In order to make Worktree.Status take global config and work like `git status`, we should use ConfigScoped(config.SystemScope)

屏幕截图 2025-11-01 094812

### DIFF
--- a/plumbing/transport/ssh/common.go
+++ b/plumbing/transport/ssh/common.go
@@ -227,7 +227,13 @@ func (c *command) getHostWithPort() string {
 		return addr
 	}
 
-	return c.endpoint.Host
+	host := c.endpoint.Hostname()
+	port := c.endpoint.Port()
+	if port == "" {
+		port = strconv.Itoa(DefaultPort)
+	}
+
+	return net.JoinHostPort(host, port)
 }
 
 func (c *command) doGetHostWithPortFromSSHConfig() (addr string, found bool) {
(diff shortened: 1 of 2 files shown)


worktree_status.go
```diff
@@ -143,7 +144,7 @@ func (w *Worktree) diffStagingWithWorktree(reverse, excludeIgnoredChanges bool)
 		return nil, err
 	}
 
-	cfg, err := w.r.Config()
+	cfg, err := w.r.ConfigScoped(config.SystemScope)
```
pjbgf: @kayoch1n similar to the other PR, this could cause regression - if this PR is rebased the tests for wasip1 are likely to fail for example.

What are your thoughts on leaning on changing `StatusWithOptions` instead, potentially adding a new option to `StatusOptions`?

PR #1709, opened 2025-11-02
_examples: Add missing update server info example

BytesHaven:
At some point with the update server info cli example was removed and not reflected within the compatibility document (404 error). Since the update server info still available within the transport package and the fact I was looking for the feature myself. I've included a update server info example to _examples.


BytesHaven:
@pjbgf CI seems to of fail on a test that isn't related to any of my changes. Would you be able to retry the CI action?

PR #1712, opened 2025-11-03
git: worktree, implement core.filemode. Related to #771

kayoch1n:
This PR tries to implement the `core.filemode` related to the Windows [issue 771](https://github.com/go-git/go-git/issues/771). In this PR, the `core.filemode` option in config will be honored and only take effect during hash comparison. When an index entry is an 
executable and `core.filemode` is `false`, a mode value of 0644 will be used for hashing purposes—while the actual mode value stored in the index remains unchanged. 

It is `true` by default and hence the actual mode value will be used in hash. 

AFAIK, C git will [automatically set this option before `init`/`clone`](https://github.com/git/git/blob/a99f379adf116d53eb11957af5bab5214915f91d/setup.c#L2370-L2379C2), and `git init` results in a value of `false` by default on Windows. Unlike C git, there is no such auto detection in go-git by now. For windows users, if they want to use go-git to init or clone a repo on disk, they should explicitly set this config option and flush the config to disk.

```go
    repository, err := git.PlainInit(repositoryPath, false)
	if err != nil {
		// error handling
	}
	cfg, err := repository.Config()
	if err != nil {
		// error handling
	}
	cfg.Core.FileMode = false
	err = repository.SetConfig(cfg)
	if err != nil {
		// error handling
	}
```

### DIFF
--- a/blame.go
+++ b/blame.go
@@ -469,20 +469,6 @@ func (b BlameResult) maxAuthorLength() int {
 	return m
 }
 
-func min(a, b int) int {
-	if a < b {
-		return a
-	}
-	return b
-}
-
-func max(a, b int) int {
-	if a > b {
-		return a
-	}
-	return b
-}
-
 type childToNeedsMap struct {
 	Child            *queueItem
 	NeedsMap         []lineMap
--- a/plumbing/object/rename.go
+++ b/plumbing/object/rename.go
@@ -389,29 +389,12 @@ type similarityPair struct {
 	score int
 }
 
-func max(a, b int) int {
-	if a > b {
-		return a
-	}
-	return b
-}
-
-func min(a, b int) int {
-	if a < b {
-		return a
-	}
-	return b
-}
-
 const maxMatrixSize = 10000
 
 func buildSimilarityMatrix(srcs, dsts []*Change, renameScore int) (similarityMatrix, error) {
 	// Allocate for the worst-case scenario where every pair has a score
 	// that we need to consider. We might not need that many.
-	matrixSize := len(srcs) * len(dsts)
-	if matrixSize > maxMatrixSize {
-		matrixSize = maxMatrixSize
-	}
+	matrixSize := min(len(srcs)*len(dsts), maxMatrixSize)
 	matrix := make(similarityMatrix, 0, matrixSize)
 	srcSizes := make([]int64, len(srcs))
 	dstSizes := make([]int64, len(dsts))
@@ -464,13 +447,12 @@ outerLoop:
 				dstSizes[dstIdx] = dstSize
 			}
 
-			min, max := srcSize, dstSize
+			minSize, maxSize := srcSize, dstSize
 			if dstSize < srcSize {
-				min = dstSize
-				max = srcSize
+				minSize, maxSize = dstSize, srcSize
 			}
 
-			if int(min*100/max) < renameScore {
+			if int(minSize*100/maxSize) < renameScore {
 				// File sizes are too different to be a match
 				continue
 			}


kayoch1n:
Hi @pjbgf , would you mind taking a look at this PR when you have a moment?

config/config.go
```diff
@@ -73,6 +73,11 @@ type Config struct {
 		// converted to LF when added to the repository, and vice versa on checkout.
 		// If set to "input", only worktree-to-repository conversion is performed.
 		AutoCRLF string
+		// Tells Git if the executable bit of files in the working tree is to be honored.
+		// If "false", when an index node is an Executable and is comparing hash 
+		// against local file, 0644 will be used as the value of its mode. The original
+		// value of mode is left unchanged in the index.
```
pjbgf:
```suggestion
		// FileMode defines whether the executable bit of working tree files is to be honored.
		// If "false", when an index node is an Executable and is comparing hash 
		// against local file, 0644 will be used as the value of its mode. The original
		// value of mode is left unchanged in the index.
```

config/config.go
```diff
@@ -73,6 +73,11 @@ type Config struct {
...
+		FileMode string
```
pjbgf: We don't need to keep this as string. Let's do the same as it is done for `IsBare`, but ensuring that this defaults to `true` as per upstream.

```suggestion
		FileMode bool
```

utils/merkletrie/index/node.go
```diff
@@ -20,12 +21,16 @@ type node struct {
 	children []noder.Noder
 	isDir    bool
 	skip     bool
+
+	trustExecutableBit bool
```
pjbgf: Not sure the term `trust` is a great match here. What about `uphold` or `honor`? Or even `ignore`, in that case inverting its meaning (e.g. `ignoreExecutableBit = !fileMode`).

```suggestion
	upholdExecutableBit bool
```

repository.go
```diff
@@ -328,6 +328,17 @@ func PlainInit(path string, isBare bool, options ...InitOption) (*Repository, er
...
+		cfg.Core.FileMode = "true"
+	} else {
+		cfg.Core.FileMode = "false"
+	}
+
+	err = r.Storer.SetConfig(cfg)
+	if err != nil {
+		return nil, err
+	}
+
```
pjbgf: Not sure why we need this. Is there a specific problem you are trying to solve?

```suggestion
```

repository.go
```diff
@@ -1905,3 +1916,61 @@ func expandPartialHash(st storer.EncodedObjectStorer, prefix []byte) (hashes []p
...
+	var executable bool
+
+	configPath := filepath.Join(gitRoot, "config")
+	if fileMode, executable = checkFileModeTrustable(configPath); fileMode {
+		if !isReinit(gitRoot) && executable {
+			fileMode = false
+		}
+	}
+	return
+}
```
pjbgf: Instead of using this heuritstics, what about we simply default to `true`, unless the consolidated config is `false`?

```suggestion
```

utils/merkletrie/index/node.go
```diff
@@ -20,12 +21,16 @@ type node struct {
...
 }
 
 // NewRootNode returns the root node of a computed tree from a index.Index,
-func NewRootNode(idx *index.Index) noder.Noder {
+func NewRootNode(idx *index.Index, fileMode string) noder.Noder {
```
pjbgf: This is part of the public API, so for improved extensibility it would be better to be passed as a `RootNodeOption`.

worktree_status.go
```diff
@@ -137,13 +138,13 @@ func (w *Worktree) diffStagingWithWorktree(reverse, excludeIgnoredChanges bool)
 		return nil, err
 	}
 
-	from := mindex.NewRootNode(idx)
-	submodules, err := w.getSubmodulesStatus()
+	cfg, err := w.r.ConfigScoped(config.SystemScope)
```
pjbgf: This may lead to regression. Instead, I'd rather we add a  `IgnoreFileMode` (or `upholdFileMode`) to the `StatusOptions`.

This way, users can call `ConfigScoped(config.SystemScope)` and based on its result set that value - this will also fix the test failures.

repository.go
kayoch1n: > Instead of using this heuritstics, what about we simply default to `true`, unless the consolidated config is `false`?

Hi. This references the implementation of [C git](https://github.com/git/git/blob/a99f379adf116d53eb11957af5bab5214915f91d/setup.c#L2370-L2379C2). During `git init` or `git clone`, C git infers the value of core.filemode by flipping the executable bit — my goal was to align this behavior as closely as possible with C git for consistency.

```c
	if (TEST_FILEMODE && !lstat(path.buf, &st1)) {
		struct stat st2;
		filemode = (!chmod(path.buf, st1.st_mode ^ S_IXUSR) &&
				!lstat(path.buf, &st2) &&
				st1.st_mode != st2.st_mode &&
				!chmod(path.buf, st1.st_mode));
		if (filemode && !reinit && (st1.st_mode & S_IXUSR))
			filemode = 0;
	}
```

If go-git doesn’t require this feature, I’m happy to remove it! I will instead init core.filemode based on the platform: set it to false for Windows, and true for all other OSes. Please let me know which approach you prefer. I’m happy to adjust accordingly!

worktree_status.go
kayoch1n: I have reverted this change, because I think it relates to the Status API and shall be done in a new PR

repository.go
pjbgf: Thanks for clarifying @kayoch1n. If I understood this correctly, that logic seems to be enabled based off `TEST_FILEMODE`, which in turn relies on [NO_TRUSTABLE_FILEMODE](https://github.com/git/git/blob/fd372d9b1a69a01a676398882bbe3840bf51fe72/Makefile#L153-L154). Combined it looks like it is a way to automatically disable `filemode` if the underlying filesystem do not support it properly.

Given that we abstract our FS access with go-billy, we are probably better off trying to use something like [billy.Capability](https://github.com/go-git/go-billy/blob/main/fs.go#L22) for this.

Based on that, I'd rather we first introduce the `core.filemode` logic as part of this PR, and then handle those cases. wdyt?

kayoch1n: > Thanks for clarifying @kayoch1n. If I understood this correctly, that logic seems to be enabled based off `TEST_FILEMODE`, which in turn relies on [NO_TRUSTABLE_FILEMODE](https://github.com/git/git/blob/fd372d9b1a69a01a676398882bbe3840bf51fe72/Makefile#L153-L154). Combined it looks like it is a way to automatically disable `filemode` if the underlying filesystem do not support it properly.
> 
> Given that we abstract our FS access with go-billy, we are probably better off trying to use something like [billy.Capability](https://github.com/go-git/go-billy/blob/main/fs.go#L22) for this.
> 
> Based on that, I'd rather we first introduce the `core.filemode` logic as part of this PR, and then handle those cases. wdyt?

Done!

PR #1714, opened 2025-11-05
Add `GOOS=wasip1` support

whitequark:


whitequark:
Fixed commit message.

pjbgf:
@whitequark thanks for proposing this PR. Please add a test (or perhaps a simple wasip1 step/workflow) to ensure that this works and future changes won't break it?

whitequark:
> Please add a test (or perhaps a simple wasip1 step/workflow) to ensure that this works and future changes won't break it?

I've considered it but I could not find a place where e.g. `GOOS=js` would be tested, or other existing OSes. Did I miss it?

pjbgf:
@whitequark Let's create a new GH workflow just for this, that installs wasmtime and run a simple test that confirm some basic workflow for wasm. For example:

```
$ curl https://wasmtime.dev/install.sh -sSf | bash
$ GOOS=wasip1 GOARCH=wasm go test -c -o go-git.wasm github.com/go-git/go-git/v6
$ wasmtime run go-git.wasm -test.v -test.run '^TestWasmInit$' github.com/go-git/go-git/v6
=== RUN   TestWasmInit
--- PASS: TestWasmInit (0.00s)
PASS
```

The test could be something as simple as:
```golang
func TestWasmInit(t *testing.T) {
	st := memory.NewStorage()
	wt := memfs.New()

	r, err := Init(st, WithWorkTree(wt))
	require.NoError(t, err)
	assert.NotNil(t, r)

	h := createCommit(t, r)
	assert.False(t, h.IsZero())

	ref, err := r.Head()
	require.NoError(t, err)
	require.NotNil(t, ref)
	assert.False(t, ref.Hash().IsZero())
	assert.Equal(t, h, ref.Hash())
}
```

But feel free to expand on it.

Also, please note that this PR also needs an additional change:
```golang
func preReceiveHook(_ string) []byte {
	return []byte{}
}
```

worktree_wasi.go
```diff
@@ -0,0 +1,27 @@
...
+			e.CreatedAt = time.Unix(int64(os.Ctime), 0)
+			e.Dev = uint32(os.Dev)
+			e.Inode = uint32(os.Ino)
+			e.GID = os.Gid
+			e.UID = os.Uid
+		}
+	}
+}
+
+func isSymlinkWindowsNonAdmin(err error) bool {
```
pjbgf:
```suggestion
func isSymlinkWindowsNonAdmin(_ error) bool {
```

whitequark:
All done.

whitequark:
Oh sorry, I had an untracked file!

repository_wasi_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/go-git/go-billy/v6/memfs"
+	"github.com/go-git/go-git/v6/storage/memory"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func preReceiveHook(_ string) []byte {
+	return []byte{}
+}
```
pjbgf: This needs to be moved into `worktree_wasi.go` as this is required for the wasi build tag to work.

.github/workflows/test.yml
```diff
@@ -47,3 +47,34 @@ jobs:
...
+      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+    - name: Install Go
+      uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
+      with:
+        go-version: ${{ matrix.go-version }}
+
+    - name: Install Wasmtime
+      run: |
+        curl https://wasmtime.dev/install.sh -sSf | bash
```
pjbgf: Perhaps using their official GHA will ensure the binary lands in the `PATH`:

```suggestion
    - name: Install Wasmtime
      uses: bytecodealliance/actions/wasmtime/setup@v1
```

whitequark:
Done

whitequark:
You're welcome! Sadly I could not get my actual application to work because all of the net/ code is stubbed out... not even experimental wasip1 sockets work.

Issue #1638, opened 2025-08-26
plumbing/transport/common.go: HTTP response error handling disregards response body


Abraxyn:
I'm currently looking into some issues I'm running into trying to clone from a private repository, and have run into "authorization failed" which I've traced back to https://github.com/go-git/go-git/tree/main/plumbing/transport/transport.go#L36.

Frankly, this is not a very useful error message, which could have come from a number of things I'm using to make the clone calls.

After looking into it a bit more, this seems like a repeat of #1097, where the response body gets added to the error message, which was added but it seems it hasn't made its way over to V6 and reverts to the previous functionality of spitting out pre-described error messages for common response codes. This would greatly help me with debugging, thanks.

pjbgf:
@Abraxyn thanks for opening this issue around the V6. 🙇 

We need to re-introduce [this logic](https://github.com/go-git/go-git/blob/4070322588104631421ebb03d80dc293d7d5a54f/plumbing/transport/http/common.go#L419-L440) to the [current implementation](https://github.com/go-git/go-git/blob/d7f8ecf1cfc8363eb9260e5e2dedfca0036d465d/plumbing/transport/http/common.go#L706-L728).

BytesHaven:
@pjbgf Can we close this ticket now?

PR #1722, opened 2025-11-12
plumbing: transport/http, Add Response Body From HTTP To Transport Error

BytesHaven:
Replaced some transport errors with there own error type that wraps errors for more details. #1638
I've also included a test for CheckError in `transport/http/common.go`.

### DIFF
--- a/plumbing/error.go
+++ b/plumbing/error.go
@@ -1,8 +1,6 @@
 package plumbing
 
-import (
-	"fmt"
-)
+import "fmt"
 
 type PermanentError struct {
 	Err error


aymanbagabas (changes requested):
Why do we need these changes? The related issue can be easily fixed with simple `fmt.Errof(...)` wraps. The new error types don't have any extra information other than wrapping other errors, which feels redundant and unnecessary.

EDIT: https://pkg.go.dev/fmt#Errorf

plumbing/transport/http/common.go
```diff
@@ -673,20 +673,22 @@ func checkError(r *http.Response) error {
...
 	switch r.StatusCode {
 	case http.StatusUnauthorized:
-		return transport.ErrAuthenticationRequired
+		return fmt.Errorf("%s %w", transport.ErrAuthenticationRequired.Error(), err)
 	case http.StatusForbidden:
-		return transport.ErrAuthorizationFailed
+		return fmt.Errorf("%s %w", transport.ErrAuthorizationFailed.Error(), err)
 	case http.StatusNotFound:
-		return transport.ErrRepositoryNotFound
+		return fmt.Errorf("%s %w", transport.ErrRepositoryNotFound.Error(), err)
```
aymanbagabas: We should wrap both errors here `fmt.Errorf("%w: %w", ...)`, otherwise, using `errors.Is(...)` would fail on transport errors.

plumbing/error.go
```diff
@@ -33,3 +33,8 @@ func NewUnexpectedError(err error) *UnexpectedError {
 func (e *UnexpectedError) Error() string {
 	return fmt.Sprintf("unexpected client error: %s", e.Err.Error())
 }
+
+func (e *UnexpectedError) Unwrap() error {
```
aymanbagabas: Could you please add a comment here?

plumbing/transport/http/common_test.go
```diff
@@ -168,6 +170,47 @@ func (s *ClientSuite) TestSetAuth() {
...
+			unwrappedErr := errors.Unwrap(err)
+			s.Error(unwrappedErr)
+			s.IsType(unwrappedErr, &Err{})
+
+			var httpErr *Err
+			s.Equal(errors.As(unwrappedErr, &httpErr), true)
+
+			s.Equal(code, httpErr.Status)
+			s.Equal(req.URL, httpErr.URL)
+			s.Equal(reason, httpErr.Reason)
```
aymanbagabas, line 221: Could you also add `errors.Is(transport.ErrAuthenticationRequired, err)` etc tests as well?

aymanbagabas (changes requested):
Looking good! A few comments 🙂

plumbing/error.go
BytesHaven: I don't really have an ideas what to put as a comment since unwrap part of the error interface. What sort of comment do you have in mind?

plumbing/transport/http/common.go
BytesHaven: That would break the unwrap methods. Go one seems to prefer wrapping one error or it will return an nil value. I believe join also causes this issue.

plumbing/error.go
aymanbagabas: You could just put something like `Unwrap implements the [errors.Unwrap] interface.`

plumbing/transport/http/common.go
aymanbagabas: It shouldn't break unwrap. The errors package will look for errors implementing either `Unwrap() error` or `Unwrap() []error`, as described in the [docs](https://pkg.go.dev/fmt#Errorf), `fmt.Errorf` will return an error that implements either one depending on how many wrapped errors it has

BytesHaven: plumbing.NewUnexpectedError has no `Unwrap() []error`, so I'll have to implement it which will just be a rehash of how go errors package is doing it. I'm a bit confused how in this PR custom error structs were rejected but I'm currently having to use `plumbing.NewUnexpectedError`. Should I just remove it as it currently has a small foot print and replace it with `ErrUnexpected = errors.New("...")` to match transport package errors?

aymanbagabas: Yeah, I would remove it altogether and replace it with whatever error was there. For the HTTP case, that would simply be `struct Err`

BytesHaven: Okay, I'll do that then. Yea, I'll keep http.Err within `transport/http` as is, since it being used at the moment.

BytesHaven:
@aymanbagabas Should I wait on @pjbgf before pushing anymore changes?

aymanbagabas:
> @aymanbagabas Should I wait on @pjbgf before pushing anymore changes?

You can go ahead with the changes

PR #1726, opened 2025-11-17
internal: use clear built-in

ferhatelmas:
similar to #1725, clear was added in go1.21.
https://pkg.go.dev/builtin#clear


pjbgf (approved):

PR #1166, opened 2024-08-09
[v6-exp] Consolidate in-flight changes

pjbgf:
A few changes aimed at the `v6-exp` branch were floating around different branches. This PR consolidates those changes so that those other branches can be deleted.

Changes of note:
- 4a05b0beaeb9e96da3dd0405ac93e39d5983fb41 Enable read-concurrency for commits and tags
- 299d3694fe8bf1b13818022899d0dd78acc08b35 v2 protocol enablement
- dc3adbb94e1ca7f9cb904bf1a2743e2d74a08ffb Make `osfs.BoundOS` the default `osfs` (fixes #1155).

PR #1729, opened 2025-11-20
prevent concurrent map read and write panic

swills:
Got a panic on concurrent read/write of this map calling Repository.PushContext(). It's difficult to reproduce but the panic is pretty clear. This should fix it. Was using v5.16.2 at the time. Patch is for main but can back port if preferred.

### DIFF
--- a/plumbing/format/idxfile/idxfile.go
+++ b/plumbing/format/idxfile/idxfile.go
@@ -57,9 +57,9 @@ type MemoryIndex struct {
 	PackfileChecksum plumbing.Hash
 	IdxChecksum      plumbing.Hash
 
-	offsetHash       map[int64]plumbing.Hash
-	offsetHashIsFull bool
-	mu               sync.RWMutex
+	offsetHash      map[int64]plumbing.Hash
+	offsetBuildOnce sync.Once
+	mu              sync.RWMutex
 
 	objectIDSize int
 }
@@ -130,15 +130,13 @@ func (idx *MemoryIndex) FindOffset(h plumbing.Hash) (int64, error) {
 
 	offset := idx.getOffset(k, i)
 
-	if !idx.offsetHashIsFull {
-		// Save the offset for reverse lookup
-		idx.mu.Lock()
-		if idx.offsetHash == nil {
-			idx.offsetHash = make(map[int64]plumbing.Hash)
-		}
-		idx.offsetHash[int64(offset)] = h
-		idx.mu.Unlock()
+	// Save the offset for reverse lookup
+	idx.mu.Lock()
+	if idx.offsetHash == nil {
+		idx.offsetHash = make(map[int64]plumbing.Hash)
 	}
+	idx.offsetHash[int64(offset)] = h
+	idx.mu.Unlock()
 
 	return int64(offset), nil
 }
@@ -188,14 +186,17 @@ func (idx *MemoryIndex) FindHash(o int64) (plumbing.Hash, error) {
 	}
 	idx.mu.RUnlock()
 
-	// Lazily generate the reverse offset/hash map if required.
-	if !idx.offsetHashIsFull || idx.offsetHash == nil {
-		if err := idx.genOffsetHash(); err != nil {
-			return plumbing.ZeroHash, err
-		}
-
-		hash, ok = idx.offsetHash[o]
+	var genErr error
+	idx.offsetBuildOnce.Do(func() {
+		genErr = idx.genOffsetHash()
+	})
+	if genErr != nil {
+		return plumbing.ZeroHash, genErr
 	}
+
+	idx.mu.RLock()
+	hash, ok = idx.offsetHash[o]
+	idx.mu.RUnlock()
 
 	if !ok {
 		return plumbing.ZeroHash, plumbing.ErrObjectNotFound
@@ -206,16 +207,12 @@ func (idx *MemoryIndex) FindHash(o int64) (plumbing.Hash, error) {
 
 // genOffsetHash generates the offset/hash mapping for reverse search.
 func (idx *MemoryIndex) genOffsetHash() error {
-	defer idx.mu.Unlock()
-	idx.mu.Lock()
-
 	count, err := idx.Count()
 	if err != nil {
 		return err
 	}
 
-	idx.offsetHash = make(map[int64]plumbing.Hash, count)
-	idx.offsetHashIsFull = true
+	offsetHash := make(map[int64]plumbing.Hash, count)
 
 	var hash plumbing.Hash
 	i := uint32(0)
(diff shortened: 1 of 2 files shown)


plumbing/format/idxfile/idxfile.go
```diff
@@ -57,9 +57,9 @@ type MemoryIndex struct {
 	PackfileChecksum plumbing.Hash
 	IdxChecksum      plumbing.Hash
 
-	offsetHash       map[int64]plumbing.Hash
-	offsetHashIsFull bool
```
ferhatelmas, line 61: I would convert this to an atomic and keep it because current approach sacrifices performance in  `FindOffset` by keep writing even after full generation. Combining `buildOnce` with atomic would provide better performance

plumbing/format/idxfile/idxfile.go
```diff
@@ -57,9 +57,9 @@ type MemoryIndex struct {
...
-	mu               sync.RWMutex
+	offsetHash map[int64]plumbing.Hash
+	buildOnce  sync.Once
```
ferhatelmas: I would add `offset` prefix

plumbing/format/idxfile/idxfile.go
```diff
@@ -215,7 +216,6 @@ func (idx *MemoryIndex) genOffsetHash() error {
 	}
 
 	idx.offsetHash = make(map[int64]plumbing.Hash, count)
-	idx.offsetHashIsFull = true
```
ferhatelmas, line 218: It might be a good idea to assign to a local variable then at the end assign it to `idx` so that lock contention is minimized

ferhatelmas:
There is clearly a concurrency issue, thanks for fixing it. It would be desirable to see a test that breaks with race detector.  It seems CI isn't using `-race` though, which is something else to improve.

plumbing/format/idxfile/idxfile.go
swills: Ok, I will do that

plumbing/format/idxfile/idxfile.go
swills: When I tested it using BenchmarkFindOffset I saw:

BenchmarkFindOffset-16    	 2548224	       534.7 ns/op

before this change and this after:

BenchmarkFindOffset-16    	 2696178	       459.4 ns/op

This is using the fasted of 3 runs for each of them. But the performance after this change varied a lot less, the performance before varied a lot more, up to >800ns/op at times. But maybe you are thinking it should be faster?

plumbing/format/idxfile/idxfile.go
swills: Ok, will do

swills: Do we need to lock idx in case Fanout (Count()) or other parts of the struct change?

plumbing/format/idxfile/idxfile.go
ferhatelmas: Since we're on it and improving concurrent access, I suggested not to leave any obvious performance on the table but it might really not matter that much with these numbers.

It also seems existing benchmark doesn't exercise concurrent usage.  It would be better to use `RunParallel` to see real effect of the changes.

plumbing/format/idxfile/idxfile.go
ferhatelmas: It's not supposed to change after decoding so I believe no.

plumbing/format/idxfile/idxfile.go
swills: True, but I'm actually not sure how often parallel access happens. If it happened a lot then I suspect likely the concurrency issue would have been more of an issue and fixed before now. On the other hand, panics, even rare ones, are kind of a show stopper.

plumbing/format/idxfile/idxfile.go
swills: Ok, did that but noticed that performance got worse in BenchmarkFindOffset after doing this. Perhaps I've done something wrong though.

plumbing/format/idxfile/idxfile.go
ferhatelmas: I totally agree. 

Right now, there is a mutex that would like to support concurrent access but it's half baked from [experimental changes](https://github.com/go-git/go-git/pull/1166/) some time ago.  I would delegate this nuance and how performant it should be to maintainers.

plumbing/format/idxfile/idxfile.go
swills: Oh, maybe that's due to other issues, FindOffset doesn't call FindHash. Fixed.

ferhatelmas: This looks good to me now.

plumbing/format/idxfile/idxfile.go
swills: Done

pjbgf:
x-ref: related to [concurrency issues](https://github.com/go-git/go-git/issues/773).

swills:
I have a test which demonstrates the issue:

```
func TestOffsetHashConcurrentPopulation(t *testing.T) {
        idx, err := fixtureIndex()
        if err != nil {
                t.Fatalf("failed to build fixture index: %v", err)
        }

        var wg sync.WaitGroup

        for _, h := range fixtureHashes {
                wg.Add(1)
                go func() {
                        defer wg.Done()
                        for i := 0; i < 5000; i++ {
                                _, _ = idx.FindOffset(h)
                        }
                }()
        }

        for _, off := range fixtureOffsets {
                wg.Add(1)
                go func() {
                        defer wg.Done()
                        for i := 0; i < 3000; i++ {
                                _, _ = idx.FindHash(off)
                        }
                }()
        }

        wg.Wait()
}
```

Should I add that as well?

Also, benchmarks in progress, can throw those in as well, but will need a larger dataset to really show the differences. fixtureHashes is only 9 hashes....

pjbgf (changes requested):
@swills thanks for looking into this and @ferhatelmas for the great initial review. 👏 
The changes LGTM.

Please squash the commits and [reword](https://github.com/go-git/go-git/blob/main/CONTRIBUTING.md#format-of-the-commit-message) them so that they align with the contributing guidelines.

Note that `genOffsetHash` is likely going to be impacted by the [introduction of reverse indexes](https://github.com/go-git/go-git/pull/1731) (i.e. `.rev` files) as it will remove the need to preload the offsets into memory.

pjbgf:
@swills no need to add benchmark stats for this specific change. As for the concurrency tests, I think that would be a nice addition to avoid future regressions.

swills:
Squashed, please let me know if I didn't get the formatting right.

PR #1730, opened 2025-11-20
git: worktree, Optimize Checkout for same tree

cedric-appdirect:
Add fast-path optimization to Checkout() that skips expensive worktree scanning when target commit has the same tree as current HEAD.

This matches git CLI behavior which only updates files when trees differ. When creating a new branch from current HEAD (same tree), this eliminates unnecessary file enumeration and hashing that can take seconds on large repositories.

Implementation:
- Add canSkipWorktreeUpdate() helper that compares tree hashes
- Fast path applies when: !Force && !Keep && no sparse directories
- Falls back to existing Reset() logic for different trees
- Graceful error handling with fallback to slow path

Testing:
- Added unit tests for fast path scenarios
- Added benchmarks demonstrating improvement
- All existing checkout tests pass
- No breaking changes to API


cedric-appdirect:
This showed has a real issue for one of the repository I use which has over 30k files in them. Benchmark will still show the speedup with less file. It does mimics how git cli work as far as I could figure out. I have left documentation inside the function just to make sure it doesn't get accidentally removed.

cedric-appdirect:
Sorry, for the first run with some tests issues. I had not noticed that some of them should have passed locally. For some reason, some of the main branch tests are failing on my computer. If it is not expected, I can look at fixing those too in a separate PR.

worktree.go
```diff
@@ -286,6 +307,51 @@ func (w *Worktree) setHEADToBranch(branch plumbing.ReferenceName, commit plumbin
...
+	if err != nil {
+		return false, err
+	}
+
+	return currentCommit.TreeHash == targetCommitObj.TreeHash, nil
+}
+
+// indexMatchesTargetTree checks if the current index already matches the target tree.
+// This is the proper check to determine if we can skip worktree updates during checkout.
+// Git uses the index (staging area) as the source of truth - if index == target tree,
```
pjbgf, line 393: > Git uses the index (staging area)

Please add a link to upstream code/doc around this subject.

worktree_bench_test.go
```diff
@@ -0,0 +1,150 @@
...
+		Branch: branches[0],
+		Force:  true,
+	})
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	// Benchmark switching between branches
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
pjbgf:
```suggestion
	for b.Loop() {
```

worktree_bench_test.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	w, err := r.Worktree()
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	// Benchmark force checkout (should use slow path)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
pjbgf:
```suggestion
	for b.Loop() {
```

worktree_bench_test.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	w, err := r.Worktree()
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	// Benchmark creating new branches from current HEAD (all same tree)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
pjbgf:
```suggestion
	for b.Loop() {
```

worktree_bench_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/go-git/go-git/v6/plumbing/cache"
+	"github.com/go-git/go-git/v6/storage/filesystem"
+
+	"github.com/go-git/go-billy/v6/memfs"
+)
+
+// BenchmarkCheckoutSameTree benchmarks the common case of creating a new
+// branch from current HEAD, which has the same tree.
+// This is the primary use case that benefits from fast path optimization.
+func BenchmarkCheckoutSameTree(b *testing.B) {
```
pjbgf: There is quite a bit of code duplication around the benchmarks, please sub-benchmarks to decrease the code verbosity. [Here](https://github.com/go-git/go-git/blob/d11db7d85708cb0fb40abba8973ef4937fa4342a/plumbing/objectid_test.go#L200)'s an example.

worktree_test.go
```diff
@@ -3745,3 +3745,63 @@ func (s *WorktreeSuite) TestRestoreBoth() {
 		{Worktree: Untracked, Staging: Untracked},
 	})
 }
+
+// TestCheckoutFastPathVerification verifies fast path by checking tree hashes
+// and ensuring no worktree scanning occurs.
+func (s *WorktreeSuite) TestCheckoutFastPathVerification() {
```
pjbgf: The worktree tests are extremely verbose, so it would be nice for new tests to be table driven. But given that we already have over 20 Checkout-related tests, I'm happy for us to merge this as is and then handle that as a separate PR.

pjbgf (changes requested):
@cedric-appdirect thanks for looking into this. Please make the changes on the benchmarks. For the other tests we could do on a follow-up PR.

Please add the results of your benchmarks into the PR body for future reference.

worktree_bench_test.go
cedric-appdirect: Oh, didn't know about that way of writing benchmark. Thanks, will do.

worktree.go
```diff
@@ -286,6 +307,51 @@ func (w *Worktree) setHEADToBranch(branch plumbing.ReferenceName, commit plumbin
 	return w.r.Storer.SetReference(head)
 }
 
+// canSkipWorktreeUpdate determines if we can skip the worktree update
+// during checkout by comparing tree hashes. Returns true if current HEAD
+// and target commit point to the same tree, meaning no files need to change.
+func (w *Worktree) canSkipWorktreeUpdate(targetCommit plumbing.Hash) (bool, error) {
```
ferhatelmas:
```suggestion
func (w *Worktree) canSkipWorktreeUpdate(targetCommitHash plumbing.Hash) (bool, error) {
```

to be consistent with `indexMatchesTargetTree`

worktree.go
```diff
@@ -183,6 +183,27 @@ func (w *Worktree) Checkout(opts *CheckoutOptions) error {
...
+	//
+	// However, we must also verify the index matches the target tree.
+	// Git uses the index (staging area) as the source of truth for what should
+	// be in the worktree. If index != target tree, we need Reset() to update both
+	// the index and worktree, even if HEAD's tree matches the target tree.
+	if !opts.Force && !opts.Keep && len(opts.SparseCheckoutDirectories) == 0 {
+		if canUseFastPath, err := w.canSkipWorktreeUpdate(c); err == nil && canUseFastPath {
+			if indexMatches, err := w.indexMatchesTargetTree(c); err == nil && indexMatches {
+				if !opts.Hash.IsZero() && !opts.Create {
+					return w.setHEADToCommit(c)
```
ferhatelmas, line 263: Resolved hash is different between fast and slow paths. Is this intended?

worktree.go
```diff
@@ -183,6 +183,27 @@ func (w *Worktree) Checkout(opts *CheckoutOptions) error {
...
+	// the expensive Reset() operation that scans all files. This matches
+	// git CLI behavior which only updates files when trees actually differ.
+	//
+	// However, we must also verify the index matches the target tree.
+	// Git uses the index (staging area) as the source of truth for what should
+	// be in the worktree. If index != target tree, we need Reset() to update both
+	// the index and worktree, even if HEAD's tree matches the target tree.
+	if !opts.Force && !opts.Keep && len(opts.SparseCheckoutDirectories) == 0 {
+		if canUseFastPath, err := w.canSkipWorktreeUpdate(c); err == nil && canUseFastPath {
+			if indexMatches, err := w.indexMatchesTargetTree(c); err == nil && indexMatches {
```
ferhatelmas: Seems like both helpers are retrieving same commit. Might be better to get it once and pass to other method or combine them

worktree_test.go
```diff
@@ -3745,3 +3745,63 @@ func (s *WorktreeSuite) TestRestoreBoth() {
...
+	w := &Worktree{
+		r:          s.Repository,
+		Filesystem: memfs.New(),
```
ferhatelmas: Index is empty on this so this is actually falling back to slow path but we're actually trying to short circuit here right?

worktree.go
cedric-appdirect: I find the intent more readable when they are split, will fetch the commit object and pass it instead.

worktree.go
cedric-appdirect: Yes, I had to switch to use the "resolved" hash otherwise worktree_test.go fail at line 960 with error "object not found".

worktree_test.go
cedric-appdirect: Indeed, I missed that. Might be something that disappeared when I was cleaning the commit before submitting the PR. Will correct.

worktree.go
cedric-appdirect: Their is no documentation about this, just code. I have added a link to the code. There is also a bunch of related commit which span a few years. Not to sure what is the best choice here.

pjbgf: Just the link is OK, that's mostly to make it easier for future contributors to refer back to.

worktree_bench_test.go
```diff
@@ -0,0 +1,132 @@
...
+	})
+}
+
+// benchmarkCheckoutSameTree benchmarks the common case of creating a new
+// branch from current HEAD, which has the same tree.
+// This is the primary use case that benefits from fast path optimization.
+func benchmarkCheckoutSameTree(b *testing.B, w *Worktree) {
+	i := 0
+	for b.Loop() {
+		branchName := plumbing.NewBranchReferenceName(fmt.Sprintf("bench-branch-%d", i))
```
pjbgf:
```suggestion
		branchName := plumbing.NewBranchReferenceName(fmt.Sprintf("bench-branch-%d", i++))
```

worktree_bench_test.go
```diff
@@ -0,0 +1,132 @@
...
+			branches = append(branches, ref.Name())
+		}
+		return nil
+	})
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	if len(branches) < 2 {
+		b.Skip("Need at least 2 branches for this benchmark")
```
pjbgf: We are better off failing this benchmark with a `b.Fatal` here instead.

worktree_bench_test.go
```diff
@@ -0,0 +1,132 @@
...
+		benchmarkCheckoutSameTree(b, w)
+	})
+
+	b.Run("SameTreeForce", func(b *testing.B) {
+		benchmarkCheckoutSameTreeForce(b, w)
+	})
+
+	b.Run("DifferentBranch", func(b *testing.B) {
+		benchmarkCheckoutDifferentBranch(b, w, r)
+	})
```
pjbgf: For simplicity:

```suggestion
	b.Run("SameTree", benchmarkCheckoutSameTree(w))
	b.Run("SameTreeForce", benchmarkCheckoutSameTreeForce(w))
	b.Run("DifferentBranch", benchmarkCheckoutDifferentBranch(w, r))
```

worktree_bench_test.go
```diff
@@ -0,0 +1,132 @@
+//go:build integration
+// +build integration
+
```
pjbgf: We are currently not conditionally building the benchmarks, so we  can remove these.

```suggestion
```

worktree_bench_test.go
```diff
@@ -0,0 +1,132 @@
...
+		err := w.Checkout(&CheckoutOptions{
+			Branch: branchName,
+			Create: true,
+		})
+		if err != nil {
+			b.Fatal(err)
+		}
+		i++
+	}
+}
```
pjbgf: Slight change to the signature, which would need to also be applied to the other two benchmarks:

```suggestion
func benchmarkCheckoutSameTree(w *Worktree) func(b *testing.B) {
	return func(b *testing.B) {
		i := 0
		for b.Loop() {
			branchName := plumbing.NewBranchReferenceName(fmt.Sprintf("bench-branch-%d", i))

			err := w.Checkout(&CheckoutOptions{
				Branch: branchName,
				Create: true,
			})
			if err != nil {
				b.Fatal(err)
			}
			i++
		}
	}
}
```

worktree_test.go
```diff
@@ -3745,3 +3745,66 @@ func (s *WorktreeSuite) TestRestoreBoth() {
...
+	})
+	s.NoError(err)
+
+	head, err := w.r.Head()
+	s.NoError(err)
+	s.Equal(plumbing.NewBranchReferenceName("force-test"), head.Name())
+
+	_, err = fs.Stat("dirty-file.txt")
+	s.Error(err, "Force checkout should have cleaned up dirty files")
+}
```
pjbgf, line 4456: Looking at this again, `TestCheckoutFastPathVerification` is redundant, as it largely does the same as `TestCheckoutCreate`. `TestCheckoutSlowPathVerification` does not seem to do what it says, and feels orthogonal to the changes. My recommendation would be to remove them both:

```suggestion
```

These could be replaced by tests for `canSkipWorktreeUpdate` and `indexMatchesTargetTree`.

pjbgf (changes requested):

worktree_bench_test.go
cedric-appdirect: Nice idea. Will do the change, maybe during the weekend.

cedric-appdirect:
Sorry took me long to come back to this. Let me know if there is anything else that I should improve.

Issue #1601, opened 2025-07-03
system freezes when cloning linux


cooperspencer:
I am trying to clone https://github.com/torvalds/linux and my systems RAM fills up to the brim and ends up freezing completely. I am already using v6 of go-git.

```go
        _, err := git.PlainClone("linux", &git.CloneOptions{
                URL:      "https://github.com/torvalds/linux",
                Progress: os.Stdout,
        })
```

pjbgf:
Do you need to do a full clone of the entire repository? Given how large that repository is, it will churn quite a large amount of memory. The [performance focused](https://github.com/go-git/go-git/blob/main/_examples/performance/clone/main.go) example may provide some tricks on handling large repositories with reduced memory/cpu consumption.

pjbgf:
I took a look at this, considering the [incoming](https://github.com/go-git/go-git/pull/1523) changes into `main`, a bare shallow clone of `torvalds/linux` seems to take `~20s` which does not look to bad considering how big that repository is.

Now the generation/reset of the worktree is quite slow, taking over 11 times as long:
```
15:40:04.515751 common.go:31: performance: 21.334990292 s: update_obj_storage
15:43:54.787195 worktree.go:290: performance: 230.271285377 s: reset_worktree
15:43:54.787337 repository.go:268: performance: 252.490496084 s: git command: git clone https://github.com/torvalds/linux
```

Once the PR that is optimising delta processing is merged I will take a look into this.

cooperspencer:
I can add a shallow clone to gickup but many users want to backup the complete repo, therefore we will need a full clone.

pjbgf:
@cooperspencer that's fair, if you need the entire repo leave as-is.

We still need to do some optimising on very large worktrees, so if you don't need them in this specific case, you should be able to clone the `linux` repo in a tenth of the time with a bare clone:

```
        _, err := git.PlainClone("linux", &git.CloneOptions{
                URL:      "https://github.com/torvalds/linux",
                Progress: os.Stdout,
                Bare: true,
        })
```

That aside, the improvements in https://github.com/go-git/go-git/pull/1523 should also drop the memory churn considerably. They should be merged at some point this week.

cooperspencer:
I think you just gave me an idea with the bare repo on how to solve it. Thanks a lot!

pjbgf:
Some of the changes in https://github.com/go-git/go-git/pull/1698 should further improve the memory consumption in this use case.

[...]

Issue #1501, opened 2025-04-09
Suggestion to add golangci-lint to this project


lazysegtree:
Code linters can be strict, but they do help in maintaining code quality.

I just ran golangci-lint with some basic linters enabled. There were some linter reporting that were a bit too strict, but some of them were actual issues that should be fixed, and good optimizations - like unparam, ineffassign, gosimple, goconst, usestdlibvars some suggestions from gocritic, revive, staticcheck

We could incrementally enable linting. Start with only one linter, and incrementally enable them and fix code issues.

Sample linter config I used - https://github.com/lazysegtree/go-git/blob/golangci_lint/.golangci.yaml

Linter reported issues - https://github.com/lazysegtree/go-git/blob/golangci_lint/sample_linter_errors.txt

lazysegtree:
@pjbgf @aymanbagabas What do you think about this ?

Its okay if this is of lower priority, but do you think this project should eventually have linter in its CICD pipeline ?

aymanbagabas:
@lazysegtree Thanks for bringing this up. I think it's important that we have a sane default linting configuration. We could start with the default list of linters `golangci-lint` provide and add more as we go. There would be a ton of code to update, from what I've seen. I would target the v6 branch though, especially since we're planning on making that the `main` branch soon.
Feel free to work on this if you'd like 🙁

pjbgf:
I'd agree with @aymanbagabas. I added golangci-lint on [some](https://github.com/go-git/go-billy/commit/b321518f51aa015c2607430760eb99d628052b97) [of](https://github.com/go-git/go-git-fixtures/commit/43baec7717f3b0c31d34537410f8f204b296d274) the dependencies within the `go-git` org.

We could start with a small number of linters enabled, and expand as we go, as there is just a tone of code to be updated and it would be overwhelming to try fixing it all at once.

If you decide to work on this, please target the `v6-transport` branch.

pjbgf:
I have added `golangci-lint` to `go-billy`, to avoid too many changes at once the linters and formatters are being added gradually. 

> We could start with the default list of linters `golangci-lint` provide and add more as we go.

For `go-git` it will be a stepper climb to start with so many linters. I'm adding the initial `make validate` on #1731 and then we we probably will have to do a linter by linter addition, to avoid a deluge of changes - some of which will be hard to review due to the volume and they will get stale fast.

ferhatelmas:
What is the list of the linters we want to enable? Happy to give a try to enable them

[...]

PR #1731, opened 2025-11-20
storage/filesystem: Add support for reverse index generation

pjbgf:
The generation of the reverse index will enable further performance optimisations around the fetching of objects by offset.

This lays the foundation to tackle the performance issues for large repositories (e.g. https://github.com/go-git/go-git/issues/1601).


onee-only:
For future reference, 6d806e840a2b7d6f2fe659d205af38f59c5b6c28 is related to https://github.com/go-git/go-git/issues/1501.

ferhatelmas:
> For future reference, https://github.com/go-git/go-git/commit/6d806e840a2b7d6f2fe659d205af38f59c5b6c28 is related to https://github.com/go-git/go-git/issues/1501.

It seems these changes are irrelevant to this PR and that's why a comment is needed. Why not add it first separately?

Edit: With gitignore, renovabot and contributing changes.

plumbing/format/revfile/decoder.go
```diff
@@ -0,0 +1,226 @@
...
+	hasher  crypto.Hash
+	hash    hash.Hash
+	nextFn  stateFn
+	version uint32
+
+	objCount     int64
+	packChecksum plumbing.ObjectID
+	out          chan<- uint32
+
+	m sync.Mutex
```
ferhatelmas, line 51: I understand the intention of using mutex to protect against misuse but it complicates and it's inherently single use.

How about having standalone `Decode` function and unexporting `Decoder` such that each `Decode` call would initialize a new decoder where there is no need for mutex, and free of mutation bugs?

Happy to make the change if you agree

plumbing/format/revfile/decoder.go
```diff
@@ -0,0 +1,226 @@
...
+	if d.objCount == 0 {
+		return nil, ErrEmptyReverseIndex
+	}
+
+	var i int64
+	for i = 0; i < d.objCount; i++ {
+		idx, err := binary.ReadUint32(d.reader)
+		if errors.Is(err, io.EOF) {
+			return nil, err
+		}
```
ferhatelmas, line 159: why do we need to handle this case separately?

plumbing/format/revfile/decoder.go
```diff
@@ -0,0 +1,226 @@
...
+	sha1Hash         uint32 = 1
+	sha256Hash       uint32 = 2
+)
+
+// Decoder reads and decodes idx files from an input stream.
+type Decoder struct {
+	reader  *bufio.Reader
+	hasher  crypto.Hash
+	hash    hash.Hash
+	nextFn  stateFn
```
ferhatelmas, line 44: Seems like unused

plumbing/format/revfile/decoder.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	rh := d.hash.Sum(nil)
+	if rev.Compare(rh) != 0 {
+		return nil, fmt.Errorf("%w: rev file checksum mismatch wanted %q got %q",
+			ErrMalformedRevFile, hex.EncodeToString(rh), rev.String())
+	}
+
+	_, err = d.reader.Peek(1)
+	if err == nil {
```
ferhatelmas, line 221: Don't we need to distinguish EOF and any other I/O errors?

pjbgf: Yep, that would be a nice improvement.

plumbing/format/revfile/decoder.go
pjbgf: Nice spot. This is missing wrapping `ErrMalformedRevFile`, as that would be the only reason for an EOF to happen here.

plumbing/format/revfile/decoder.go
pjbgf: Completely agree that the overall API in the `plumbing/format` encoders/decoders can be improved as currently they are a bit awkward.

If you can, please propose a PR with the changes you pointed out for `revfile` and ping me for a review. If the API changes make sense, potentially we could expand that across the other encoders/decoders.

ferhatelmas: I submitted https://github.com/go-git/go-git/pull/1750
Let me know what you think.

plumbing/format/revfile/decoder.go
ferhatelmas: I handled this in my PR and also see the style change there for `errors.Is` vs `==`.

plumbing/format/revfile/decoder.go
ferhatelmas: Dropped

plumbing/format/revfile/decoder.go
ferhatelmas: I handled this case and validated via a test.

PR #1733, opened 2025-11-21
git: worktree, add lazy cache for tree entries in checkoutChange

cedric-appdirect:
Implement a lazy caching mechanism for tree entries to optimize checkout
operations when processing many files. Instead of traversing the entire
tree upfront, the cache is populated on-demand as entries are accessed
during checkout.
    
Previously, FindEntry() was called for every file during checkout,
causing O(n*d) tree traversals where n is the number of files and d is
the directory depth.
    
This optimization provides ~7% performance improvement for repositories
with many files distributed across multiple directories.

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -491,6 +491,11 @@ func (w *Worktree) resetWorktree(t *object.Tree, files []string) error {
 	}
 	b := newIndexBuilder(idx)
 
+	// Create a cache for tree entries that will be populated lazily
+	// as we process changes. This avoids upfront tree traversal costs
+	// and only caches entries we actually need.
+	entryCache := make(map[string]*object.TreeEntry)
+
 	for _, ch := range changes {
 		if err := w.validChange(ch); err != nil {
 			return err
@@ -514,7 +519,7 @@ func (w *Worktree) resetWorktree(t *object.Tree, files []string) error {
 			}
 		}
 
-		if err := w.checkoutChange(ch, t, b); err != nil {
+		if err := w.checkoutChange(ch, t, b, entryCache); err != nil {
 			return err
 		}
 	}
@@ -625,7 +630,7 @@ func (w *Worktree) validChange(ch merkletrie.Change) error {
 	return nil
 }
 
-func (w *Worktree) checkoutChange(ch merkletrie.Change, t *object.Tree, idx *indexBuilder) error {
+func (w *Worktree) checkoutChange(ch merkletrie.Change, t *object.Tree, idx *indexBuilder, entryCache map[string]*object.TreeEntry) error {
 	a, err := ch.Action()
 	if err != nil {
 		return err
@@ -638,9 +643,15 @@ func (w *Worktree) checkoutChange(ch merkletrie.Change, t *object.Tree, idx *ind
 	switch a {
 	case merkletrie.Modify, merkletrie.Insert:
 		name = ch.To.String()
-		e, err = t.FindEntry(name)
-		if err != nil {
-			return err
+
+		var ok bool
+		e, ok = entryCache[name]
+		if !ok {
+			e, err = t.FindEntry(name)
+			if err != nil {
+				return err
+			}
+			entryCache[name] = e
 		}
 
 		isSubmodule = e.Mode == filemode.Submodule


cedric-appdirect:
This is also showing up when you have a repository with a lot of files.

worktree.go
```diff
@@ -491,6 +491,11 @@ func (w *Worktree) resetWorktree(t *object.Tree, files []string) error {
 	}
 	b := newIndexBuilder(idx)
 
+	// Create a cache for tree entries that will be populated lazily
+	// as we process changes. This avoids upfront tree traversal costs
+	// and only caches entries we actually need.
+	entryCache := make(map[string]*object.TreeEntry)
```
pjbgf, line 497: Have you tested the impact of this cache (in terms of memory churn) against a very large repository - e.g. the Linux Kernel?

cedric-appdirect: I haven't tried with the Linux kernel, will do on Monday and report the impact.

cedric-appdirect: Are you able to load the linux kernel with the current code at all? It has been over an hour and I had to gave up. I ran the benchmark before and after and there is actually no measurable difference. I am guessing the reason is that the only added cost is the size of the map as the TreeEntry is just a pointer (no copy) and it is already hold in memory in any case.

For reference:
```
                  │  before.txt  │             after.txt              │
                  │     B/op     │     B/op      vs base              │
CloneLargeRepo-12   1.970Gi ± 0%   1.936Gi ± 0%  -1.69% (p=0.001 n=7)
CloneDeepRepo-12    172.1Mi ± 0%   172.5Mi ± 0%       ~ (p=0.620 n=7)
geomean             589.1Mi        584.8Mi       -0.74%

                  │ before.txt  │             after.txt             │
                  │  allocs/op  │  allocs/op   vs base              │
CloneLargeRepo-12   20.06M ± 0%   20.05M ± 0%  -0.07% (p=0.001 n=7)
CloneDeepRepo-12    1.160M ± 0%   1.160M ± 0%       ~ (p=0.222 n=7)
geomean             4.825M        4.823M       -0.03%
```

cedric-appdirect:
> Hello @cedric-appdirect. It seems like there's already a cache(`Tree.m`) for the entries, but somehow it doesn’t seem to be used inside `FindEntry`. Maybe we can use it instead to encapsulate the cache logic inside of `Tree`?
> 
> This would make the code cleaner. and the cache could also be reused beyond the scope of `resetWorktree`.
> 
> https://github.com/go-git/go-git/blob/5d40cdde8ca0df28c69852d2082df0d420cd2436/plumbing/object/tree.go#L36-L43

Oh, I missed that. Will investigate.

cedric-appdirect:
> Hello @cedric-appdirect. It seems like there's already a cache(`Tree.m`) for the entries, but somehow it doesn’t seem to be used inside `FindEntry`. Maybe we can use it instead to encapsulate the cache logic inside of `Tree`?
> 
> This would make the code cleaner. and the cache could also be reused beyond the scope of `resetWorktree`.
> 
> https://github.com/go-git/go-git/blob/5d40cdde8ca0df28c69852d2082df0d420cd2436/plumbing/object/tree.go#L36-L43

I looked into it and I am not sure actually if that is the right path. The map in the Tree struct only contain the entry for the current tree, not the full path of the object ( https://github.com/go-git/go-git/blob/main/plumbing/object/tree.go#L333 ). So you still need to do the full resolution and go all the way down to tree.entry ( https://github.com/go-git/go-git/blob/main/plumbing/object/tree.go#L164 ) to be able to use it.

Now, the cache is only in the checkout step, so it isn't long living and has no significant impact on memory usage as all the object are in memory anyway during that step already (due to this two caches in the Tree struct). Thinking about it, I could move that short lived cache to the Tree struct and cache things there. That cache wouldn't be short lived anymore, but I am under the impression that we do not clear the Tree struct cache anyway, so making that cache longer living would not change much the memory profile. This would help all FindEntry caller and not just the Checkout step. So should I move the cache to the Tree struct and maybe remove the entry cache if the full cache is actually being hit more.

cedric-appdirect:
This is not necessary anymore with the bug fixed on FindEntry.

PR #1736, opened 2025-11-23
plumbing: optimize Tree.Decode with ReadSlice and io.ReadFull

Bluebugs:
Replace allocating ReadString calls with zero-allocation ReadSlice in
Tree.Decode, add filemode.NewFromBytes to parse modes directly from
byte slices, and replace binary.Read with io.ReadFull in ObjectID.ReadFrom
to avoid reflection overhead.

Benchmark results (n=6, AMD Ryzen 7 6800U):

  TreeDecode/Small/1entry      -19.76%  (424.7ns → 340.8ns)
  TreeDecode/Medium/8entries   -33.36%  (2.268µs → 1.512µs)
  TreeDecode/Large/100entries  -31.15%  (24.71µs → 17.02µs)
  TreeDecode/VeryLarge/1000    -33.41%  (228.1µs → 151.9µs)

  Allocations reduced 20-33%, B/op increased ~8% due to ReadFull.

PR #1738, opened 2025-11-23
perf: optimize ObjectID.ReadFrom with direct io.ReadFull

Bluebugs:
Replace encoding/binary.Read with io.ReadFull in ObjectID.ReadFrom for better performance, eliminating reflection overhead.

Background:
-----------
The ObjectID.ReadFrom method used encoding/binary.Read, which employs reflection to copy bytes into the hash array. This adds unnecessary overhead in hot paths, particularly during Tree.Decode operations that are called thousands of times during clone operations.

Changes:
--------
1. Updated ObjectID.ReadFrom implementation to use io.ReadFull directly instead of encoding/binary.Read

2. Added error handling to zero out the hash on read errors for cleaner state

3. Updated test coverage to verify the optimized implementation:
   - TestReadFrom: validates existing functionality still works
   - TestReadFromOptimization: validates direct io.ReadFull behavior

Technical Notes:
---------------
This is a backward-compatible change. The method signature remains identical and it still implements io.ReaderFrom. For byte arrays (like hashes), binary.Read's BigEndian parameter is a no-op since no byte-order conversion occurs. io.ReadFull produces identical results while being significantly faster by avoiding reflection.

Performance Impact:
------------------
Benchmark results from 10 iterations (AMD Ryzen 7 6800U):

Speed improvements (statistically significant, p<0.05):
  TreeDecode:           -2.91% (1429ns → 1387ns)
  TreeDecodeSmall:      -8.56% (308ns → 282ns)
  TreeDecodeLarge:      -5.47% (16.95µs → 16.02µs)
  HashReadFrom:         -2.47% (65.18ns → 63.57ns)
  Overall geometric:    -4.02%

Memory:
  Allocations: unchanged (same count)
  Bytes/op: +6.81% to +14.45% increase (acceptable tradeoff)

The speed improvements are consistent across all tree sizes, with the most significant gains on small trees (-8.56%). CPU profiling showed hash reading overhead reduced from 13% to 7% of total Tree.Decode time.

This optimization benefits all callers automatically with no code changes required, as the method signature and behavior remain identical.

### DIFF
--- a/plumbing/objectid.go
+++ b/plumbing/objectid.go
@@ -117,14 +117,21 @@ func (s *ObjectID) Write(in []byte) (int, error) {
 	return n, nil
 }
 
-// ReadFrom reads the Big Endian representation of the ObjectID from
-// reader [r].
+// ReadFrom reads the ObjectID from reader [r]. It implements the io.ReaderFrom interface.
+//
+// This method uses io.ReadFull for optimal performance, avoiding the
+// reflection overhead of encoding/binary.Read. For byte arrays (like hashes),
+// no byte-order conversion occurs regardless of the method used.
 func (s *ObjectID) ReadFrom(r io.Reader) (int64, error) {
-	err := binary.Read(r, binary.BigEndian, s.hash[:s.Size()])
+	n, err := io.ReadFull(r, s.hash[:s.Size()])
 	if err != nil {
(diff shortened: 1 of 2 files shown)


Bluebugs:
I am deciding to fold this optimization with #1736 for simplicity of the final code and improvement.

PR #1747, opened 2025-11-24
git: Improve Status() speed with new `index.ModTime` check

cedric-appdirect:
This should help #181. This mimick what git cli does by relying on file mtime and size to decide if we need to compute its hash. This does not address the untracked file case.

### DIFF
--- a/utils/merkletrie/filesystem/node.go
+++ b/utils/merkletrie/filesystem/node.go
@@ -27,6 +27,11 @@ var ignore = map[string]bool{
 type Options struct {
 	// AutoCRLF converts CRLF line endings in text files into LF line endings.
 	AutoCRLF bool
+
+	// Index is used to enable the metadata-first comparison optimization while
+	// correctly handling the "racy git" condition. If no index is provided,
+	// the function works without the optimization.
+	Index *index.Index
 }
 
 // The node represents a file or a directory in a billy.Filesystem. It
@@ -57,54 +62,36 @@ type node struct {
 // should be provided where the key is the path of the submodule and the commit
 // of the submodule HEAD
 //
-// Deprecated: Use NewRootNodeWithIndex instead for better performance.
+// Deprecated: Use NewRootNodeWithOptions instead for better performance.
 // This function is kept for backward compatibility.
 func NewRootNode(
 	fs billy.Filesystem,
 	submodules map[string]plumbing.Hash,
 ) noder.Noder {
-	return NewRootNodeWithIndex(fs, submodules, nil, Options{})
+	return NewRootNodeWithOptions(fs, submodules, Options{Index: nil})
 }
 
 // NewRootNodeWithOptions returns the root node based on a given billy.Filesystem
-// with options for CRLF handling.
-//
-// Deprecated: Use NewRootNodeWithIndex instead for better performance.
-// This function is kept for backward compatibility and now internally calls
-// NewRootNodeWithIndex with a nil index, which disables the metadata
-// optimization but maintains the same functionality.
-func NewRootNodeWithOptions(
-	fs billy.Filesystem,
-	submodules map[string]plumbing.Hash,
-	options Options,
-) noder.Noder {
-	return NewRootNodeWithIndex(fs, submodules, nil, options)
-}
-
-// NewRootNodeWithIndex returns the root node based on a given billy.Filesystem
-// and an index. This enables the metadata-first comparison optimization while
-// correctly handling the "racy git" condition.
+// with options for CRLF handling and an index. Providing an index enables the
+// metadata-first comparison optimization while correctly handling the "racy git"
+// condition. If no index is provided, the function works without the optimization.
 //
 // The index's ModTime field is used to detect the racy git condition. When a file's
 // mtime equals or is newer than the index ModTime, we must hash the file content
 // even if other metadata matches, because the file may have been modified in the
 // same second that the index was written.
 //
-// If idx is nil, the function works without the optimization (equivalent to
-// NewRootNodeWithOptions), which is useful for cases where no index is available.
-//
 // Reference: https://git-scm.com/docs/racy-git
-func NewRootNodeWithIndex(
+func NewRootNodeWithOptions(
 	fs billy.Filesystem,
 	submodules map[string]plumbing.Hash,
-	idx *index.Index,
 	options Options,
 ) noder.Noder {
 	var idxMap map[string]*index.Entry
 
-	if idx != nil {
-		idxMap = make(map[string]*index.Entry, len(idx.Entries))
-		for _, entry := range idx.Entries {
+	if options.Index != nil {
+		idxMap = make(map[string]*index.Entry, len(options.Index.Entries))
+		for _, entry := range options.Index.Entries {
 			idxMap[entry.Name] = entry
 		}
 	}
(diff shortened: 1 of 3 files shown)


hansbogert:
For context, I functionally tried the same in: https://github.com/go-git/go-git/pull/1694

This PR surely is more production ready. However, in the time between making my PR and and now, I've looked at upstream git some more. There is are some edge cases that they catch. I think they are needed in the go-git implementation as well. A big one is documented here https://git-scm.com/docs/racy-git

cedric-appdirect:
> For context, I functionally tried the same in: #1694

Oh, sorry, I missed it. I have just been hammering my local benchmark and going at everything that stick out.
 
> This PR surely is more production ready. However, in the time between making my PR and and now, I've looked at upstream git some more. There is are some edge cases that they catch. I think they are needed in the go-git implementation as well. A big one is documented here https://git-scm.com/docs/racy-git

I was aware of this issue while reading the git code, but that documentation is a lot more understandable with a lot more context. I had assumed that Golang code in https://github.com/golang/go/blob/master/src/time/time.go#L316 was going to be enough. Seems I need to double check. I wonder how I can write a tests case for this would help ensure we don't have this problem. Thanks a lot for the link.

cedric-appdirect:
The solution is actually simple, we just need to carry around the mtime of the index file and fallback to compute hash in case mtime of the file == mtime of the index. In practice, that shouldn't really happen very often, so the performance impact shouldn't be a big deal for anyone and we keep the race condition out in a simple manner.

utils/merkletrie/filesystem/node.go
```diff
@@ -44,29 +48,72 @@ type node struct {
...
 }
 
-// NewRootNodeWithOptions returns the root node based on a given billy.Filesystem with options.
+// NewRootNodeWithOptions returns the root node based on a given billy.Filesystem
+// with options for CRLF handling.
+//
+// Deprecated: Use NewRootNodeWithIndex instead for better performance.
+// This function is kept for backward compatibility and now internally calls
+// NewRootNodeWithIndex with a nil index, which disables the metadata
+// optimization but maintains the same functionality.
```
cedric-appdirect: Should I just keep this function or remove them as this is work for v6 and API break are still acceptable at this point in time?

cedric-appdirect:
Performance improvement on the provided benchmark:
```
                     │ ../../go-git/before-status.txt │             after.txt              │
                     │             sec/op             │   sec/op     vs base               │
Status/Clean-12                           96.77m ± 2%   18.70m ± 5%  -80.68% (p=0.001 n=7)
Status/Modified-12                        98.99m ± 3%   19.08m ± 1%  -80.72% (p=0.001 n=7)
StatusLarge/Clean-12                     251.89m ± 2%   43.95m ± 6%  -82.55% (p=0.001 n=7)
geomean                                   134.1m        25.03m       -81.34%

                     │ ../../go-git/before-status.txt │              after.txt              │
                     │              B/op              │     B/op      vs base               │
Status/Clean-12                         11.497Mi ± 0%   6.586Mi ± 0%  -42.71% (p=0.001 n=7)
Status/Modified-12                      11.505Mi ± 0%   6.695Mi ± 0%  -41.80% (p=0.001 n=7)
StatusLarge/Clean-12                     28.04Mi ± 1%   15.46Mi ± 0%  -44.88% (p=0.001 n=7)
geomean                                  15.48Mi        8.801Mi       -43.15%

                     │ ../../go-git/before-status.txt │             after.txt              │
                     │           allocs/op            │  allocs/op   vs base               │
Status/Clean-12                           217.9k ± 0%   147.9k ± 0%  -32.13% (p=0.001 n=7)
Status/Modified-12                        218.0k ± 0%   148.7k ± 0%  -31.79% (p=0.001 n=7)
StatusLarge/Clean-12                      542.2k ± 0%   367.1k ± 0%  -32.28% (p=0.001 n=7)
geomean                                   295.3k        200.6k       -32.07%
```

cedric-appdirect:
I hadn't covered the in memory storage case previously and recent main test improvement added a tests case that caught that. This solve the problem in both a defensive way (making sure that if we don't have a proper index or modification time set on the index, it will fallback to full status check) and then in an optimized way by setting the index time in memory to match the time we load the index.

utils/merkletrie/filesystem/node.go
pjbgf: Let's keep `NewRootNodeWithOptions` and add the index in the `Options` instead.

utils/merkletrie/filesystem/node.go
```diff
@@ -202,6 +262,40 @@ func (n *node) calculateHash() {
 	n.hash = append(hash.Bytes(), mode.Bytes()...)
 }
 
+func (n *node) metadataMatches(entry *index.Entry) bool {
+	if uint32(n.size) != entry.Size {
```
pjbgf, line 257:
```suggestion
	if entry == nil {
		return false
	}

	if uint32(n.size) != entry.Size {
```

pjbgf (changes requested):
@cedric-appdirect good stuff, thanks for looking into this it largely LGTM. Please rebase and take a look at the changes below.

Thank you @hansbogert for sharing the additional information around the racy index and helping with the initial feedback.  🙇

pjbgf:
For future reference, a `benchstat` comparing `v5`, `main` and this PR:

```
goos: linux
goarch: amd64
cpu: AMD Ryzen AI 9 HX PRO 370 w/ Radeon 890M
                     │   /tmp/v5    │
                     │    sec/op    │
Status/Clean-24        28.83m ± 21%
Status/Modified-24     29.50m ± 18%
StatusLarge/Clean-24   77.68m ± 18%
geomean                40.42m

                     │   /tmp/v5    │
                     │     B/op     │
Status/Clean-24        70.21Mi ± 0%
Status/Modified-24     70.20Mi ± 0%
StatusLarge/Clean-24   174.7Mi ± 0%
geomean                95.14Mi

                     │   /tmp/v5   │
                     │  allocs/op  │
Status/Clean-24        194.2k ± 0%
Status/Modified-24     194.3k ± 0%
StatusLarge/Clean-24   484.1k ± 0%
geomean                263.4k

                     │   /tmp/main   │               /tmp/pr                │
                     │    sec/op     │    sec/op     vs base                │
Status/Clean-24        21.043m ± 12%   9.636m ± 19%  -54.21% (p=0.000 n=10)
Status/Modified-24     20.659m ± 13%   9.681m ± 15%  -53.14% (p=0.000 n=10)
StatusLarge/Clean-24    54.55m ± 18%   25.41m ± 20%  -53.41% (p=0.000 n=10)
geomean                 28.73m         13.33m        -53.59%

                     │  /tmp/main   │               /tmp/pr                │
                     │     B/op     │     B/op      vs base                │
Status/Clean-24        9.861Mi ± 0%   6.514Mi ± 0%  -33.94% (p=0.000 n=10)
Status/Modified-24     9.866Mi ± 0%   6.626Mi ± 0%  -32.84% (p=0.000 n=10)
StatusLarge/Clean-24   24.06Mi ± 0%   15.27Mi ± 0%  -36.53% (p=0.000 n=10)
geomean                13.28Mi        8.702Mi       -34.46%

                     │  /tmp/main  │               /tmp/pr               │
                     │  allocs/op  │  allocs/op   vs base                │
Status/Clean-24        217.8k ± 0%   147.8k ± 0%  -32.14% (p=0.000 n=10)
Status/Modified-24     217.9k ± 0%   148.6k ± 0%  -31.80% (p=0.000 n=10)
StatusLarge/Clean-24   542.0k ± 0%   367.0k ± 0%  -32.29% (p=0.000 n=10)
geomean                295.2k        200.5k       -32.08%
```

PR #1748, opened 2025-11-24
git: worktree, Eliminate duplicate Tree.FindEntry calls in checkout

cedric-appdirect:
During checkout operations, Tree.FindEntry was being called twice per file:
1. In checkoutChange() to obtain the TreeEntry
2. In checkoutChangeRegularFile() via Tree.File(), which internally calls FindEntry again with the same path

This optimization passes the TreeEntry directly from checkoutChange to checkoutChangeRegularFile, eliminating the redundant tree traversal. The blob is now retrieved directly using object.GetBlob() with the hash from the already-found TreeEntry.

Performance impact measured with BenchmarkAlternatesPerformance (40,000 files):
- Tree.FindEntry time reduced from 120.45s to 62.45s (~52% reduction)
- Savings: ~1.55ms per file checkout operation
- Tree.File() usage dropped from 62s+ to negligible (10ms, used elsewhere)

The change maintains identical behavior while significantly improving performance for large checkouts, particularly beneficial for repositories using alternates where object lookups are more expensive.

### DIFF
--- a/plumbing/object/tree.go
+++ b/plumbing/object/tree.go
@@ -103,6 +103,21 @@ func (t *Tree) File(path string) (*File, error) {
 	return NewFile(path, e.Mode, blob), nil
 }
 
+// FileFromEntry creates a File from a TreeEntry without performing a FindEntry lookup.
+// This is useful when you already have a TreeEntry and want to avoid redundant lookups.
+// The path parameter is used for the File's Name field.
+func (t *Tree) FileFromEntry(path string, e *TreeEntry) (*File, error) {
+	blob, err := GetBlob(t.s, e.Hash)
+	if err != nil {
+		if errors.Is(err, plumbing.ErrObjectNotFound) {
+			return nil, ErrFileNotFound
+		}
+		return nil, err
+	}
+
+	return NewFile(path, e.Mode, blob), nil
+}
+
 // Size returns the plaintext size of an object, without reading it
 // into memory.
 func (t *Tree) Size(path string) (int64, error) {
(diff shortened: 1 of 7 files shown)


plumbing/object/tree.go
```diff
@@ -89,6 +89,21 @@ func (t *Tree) File(path string) (*File, error) {
...
+	blob, err := GetBlob(t.s, e.Hash)
+	if err != nil {
+		if errors.Is(err, plumbing.ErrObjectNotFound) {
+			return nil, ErrFileNotFound
+		}
+		return nil, err
+	}
+
+	return NewFile(path, e.Mode, blob), nil
+}
```
onee-only, line 125: `Tree.TreeEntryFile()` does the same thing as this. However, the naming `FileFromEntry` and the error handling seem like good improvements. What are your thoughts on removing `Tree.TreeEntryFile()` and using this instead?

Also, as in `Tree.TreeEntryFile()`, taking only `e *TreeEntry` should be sufficient. The `path` parameter is redundant since it's equivalent to `e.Name`.

https://github.com/go-git/go-git/blob/5d40cdde8ca0df28c69852d2082df0d420cd2436/plumbing/object/tree.go#L119-L127

cedric-appdirect: Absolutely. I had missed TreeEntryFile as the name wasn't in my expectation. I will update FileFromEntry to use e.Name and then replace all user of TreeEntryFile accordingly.

cedric-appdirect: I could not remove the use of path as the *File entry expected by the caller need to have the full path, not just the basename as stored in e.Name. Removing TreeEntryFile make its user actually have a *File that match the one from the checkout step with the full path. I think that make sense.

onee-only: You're right. I thought `TreeEntry.Name` had the full path. 😅

cedric-appdirect:
I reran the benchmark for this PR. The improvement is not much anymore, about 1% after fixing the bug in FindEntry, but I do think it still does improve the code base.

PR #1749, opened 2025-11-25
Parallel checkout

cedric-appdirect:
This PR introduce the use of parallel goroutine to checkout each files. This scale as you would expect by the number of core. It has minimal overhead for very small repository as most of the workload is driven by decompression and syscall which dwarf the use of the channel to send "checkout command".

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -10,21 +10,26 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"runtime"
 	"strings"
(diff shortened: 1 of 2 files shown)


cedric-appdirect:
I have left this PR in draft as I am not sure of the code logic here. One of the area I am not a fan, but couldn't figure out a better approach is the creation of workerTree for each of the goroutine. Will need review and advise from someone that understand and know this code base.

PR #1750, opened 2025-11-25
refactor(plumbing): improve revfile encode/decode

ferhatelmas:
related to #1731

* Make encode and decode a function and unexport encoder/decoder to hide state so that
there is less opportunity to misuse concurrently,
and drop mutex and recover as a result.
* Handle non EOF errors in decode trailing check.
* Add context in object decode for EOF.
* Drop unused stateFn in encoder/decoder.
* Accept io.Reader since it's more idiomatic. For decode, internally, we create buffered reader
if already not buffered but not doing it for
writing since flushing is required and it's
generally controlled by caller (file, buffer, etc).

Keeping explicit check for io.EOF because of
convention https://github.com/golang/go/issues/39155

### DIFF
--- a/plumbing/format/revfile/decoder.go
+++ b/plumbing/format/revfile/decoder.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"hash"
 	"io"
-	"sync"
 
 	"github.com/go-git/go-git/v6/plumbing"
 	"github.com/go-git/go-git/v6/utils/binary"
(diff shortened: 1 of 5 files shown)


plumbing/format/revfile/decoder_test.go
```diff
@@ -15,6 +14,31 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// errorAfterReader wraps a reader and returns an error after n bytes have been read.
+type errorAfterReader struct {
```
pjbgf, line 18: This would certainly be handy in other parts of the code - including other {en,de}coders. But we can extract it on a follow-up PR when it gets used around other tests.

plumbing/format/revfile/decoder.go
```diff
@@ -36,66 +35,60 @@ const (
...
 	}
 
-	defer func() {
-		if r := recover(); r != nil {
-			err = fmt.Errorf("%v", r)
-		}
-	}()
+	br, ok := r.(*bufio.Reader)
+	if !ok {
+		br = bufio.NewReader(r)
```
pjbgf, line 69: I'm not a massive fan of the additional allocation that comes with this trade-off. But considering that on using a `io.Reader` makes it more flexible, it sounds like a reasonable thing to do.

pjbgf:
@ferhatelmas this commit also needs to be reworded so that I can go ahead and approve/merge.

ferhatelmas:
Updated this as well. 

Aside, shall we update commit message validation to include type of work such as docs, fix, refactor, etc. ?

pjbgf:
> shall we update commit message validation to include type of work such as docs, fix, refactor, etc. ?

Feel free to propose an update to the contributing guidelines then we can loop in other folks on it.

PR #1752, opened 2025-11-25
git: fix flaky remote timeout test

ferhatelmas:
Instead of depending an external
service, use local test server.

### DIFF
--- a/remote_test.go
+++ b/remote_test.go
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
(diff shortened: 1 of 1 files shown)


ferhatelmas:
I realized it was failing frequently https://github.com/ferhatelmas/go-git/actions/runs/19685982916/job/56391517276

ferhatelmas:
Updated the commit message and PR title.

pjbgf:
The failure on Git Compatibility seems to also be flaky - just went ahead and re-tried it.

Issue #1753, opened 2025-11-25
golangci-lint: Enable linters


pjbgf:
### Feature Description

We can start with the linters below and expand further as we go:
- [ ] `errcheck`
- [ ] `govet`
- [ ] `gosec`
- [ ] `ineffassign`
- [ ] `staticcheck`
- [ ] `unused`
- [ ] `modernize`
- [ ] `dupl`
- [ ] `misspell`
- [ ] `prealloc`
- [ ] `revive`
- [ ] `unused`
- [ ] `whitespace`

Such changes tend to stale other active PRs as it will touch a large amount of files, so:
- Let's only start this after #1751 is done.
- Do a single linter per commit - for commit message `*:` will make the PR validatio check happy.
- A PR may enable multiple linters if the total amount of files changed are up to 15. If more files are affected, please keep a linter per PR.

ferhatelmas:
@pjbgf you can assign it to me.

ferhatelmas:
I would suggest `gocritic` as well.  https://github.com/go-critic/go-critic

onee-only:
How about [paralleltest](https://github.com/kunwardeep/paralleltest)? It could help speed up the tests.

pjbgf:
Re-opening as we still need to rollout the `gosec`.

PR #1757, opened 2025-11-26
Enable unused and modernize linters

ferhatelmas:
related to #1753

### DIFF
--- a/plumbing/format/config/option.go
+++ b/plumbing/format/config/option.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 )
 
@@ -98,7 +99,7 @@ func (opts Options) withSettedOption(key string, values ...string) Options {
 			continue
 		}
 
-		if contains(values, o.Value) {
+		if slices.Contains(values, o.Value) {
 			added = append(added, o.Value)
 			result = append(result, o)
 			continue
@@ -106,7 +107,7 @@ func (opts Options) withSettedOption(key string, values ...string) Options {
 	}
 
 	for _, value := range values {
-		if contains(added, value) {
+		if slices.Contains(added, value) {
 			continue
 		}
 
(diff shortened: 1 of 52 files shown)


ferhatelmas:
@pjbgf next batch is ready to review.

plumbing/format/config/option.go
```diff
@@ -117,11 +118,5 @@ func (opts Options) withSettedOption(key string, values ...string) Options {
...
 func contains(haystack []string, needle string) bool {
-	for _, s := range haystack {
-		if s == needle {
-			return true
-		}
-	}
-
-	return false
+	return slices.Contains(haystack, needle)
 }
```
pjbgf: Let's drop this func altogether and get its callees to call `slices.Contains` instead.

```suggestion
```

repository_plan9_test.go
```diff
@@ -11,7 +11,7 @@ import (
 // preReceiveHook returns the bytes of a pre-receive hook script
 // that prints m before exiting successfully
 func preReceiveHook(m string) []byte {
-	return []byte(fmt.Sprintf("#!/bin/rc\necho -n %s\n", quote(m)))
+	return fmt.Appendf(nil, "#!/bin/rc\necho -n %s\n", quote(m))
```
pjbgf, line 14: This seems to be orthogonal to the linters being enabled. At first sight the previous version looks like a better alternative. Any specific reason why this change is being proposed?

pjbgf (changes requested):

plumbing/format/config/option.go
ferhatelmas: dropped

repository_plan9_test.go
ferhatelmas: modernize recommends to optimize byte slice and string conversion

pjbgf: When I reverted the changes golangci-lint did not complain about it. Let me take a proper look again and come back to you on this.

ferhatelmas: golangci-lint respects build tags.  this file is only for plan9:

```
//go:build plan9 && !unix && !windows
// +build plan9,!unix,!windows
```

pjbgf (approved):

PR #1760, opened 2025-11-27
plumbing: object, Fix timezone comparison in encode/decode tests.

### LINKED ISSUE #1755: Failing tests when running locally due to issue comparing commit who include time.Time
### Bug Description

When running `go test ./...` locally, I have two failing tests:
- TestCommitEncodeDecodeIdempotent
- TestTagEncodeDecodeIdempotent

The root cause is that the commit comparison are failing due to my timezone not being -07:00 . This fail when doing a deep s.Equal on the commit as time.Time comparison fail for testify when manipulating this type. See https://github.com/stretchr/testify/issues/1010 and https://github.com/stretchr/testify/issues/984 .

### go-git Version

v6 master

### Steps to Reproduce

Change your timezone and use latest golang compiler.

### Additional Information

I can easily make a PR that either change the s.Equal to a series of explicit Equal on selected field, or I can force the timezone of all commits to be UTC, or I can initialize the tests with the current timezone the tests are currently being run with. I have no idea which approach would be preferred.

Bluebugs:
Fixes #1755

time.Parse attaches a rich *time.Location from the system timezone database (with DST rules, named zones, transition tables), while Signature.Decode uses time.FixedZone which creates a minimal fixed-offset zone.

Use time.Unix with time.FixedZone to construct test timestamps that match exactly what the decode function produces, ensuring encode/decode idempotency tests compare equivalent timezone representations.

### DIFF
--- a/plumbing/object/tag_test.go
+++ b/plumbing/object/tag_test.go
@@ -181,7 +181,7 @@ func (s *TagSuite) TestTagDecodeWrongType() {
 }
 
 func (s *TagSuite) TestTagEncodeDecodeIdempotent() {
-	ts, err := time.Parse(time.RFC3339, "2006-01-02T15:04:05-07:00")
+	ts, err := time.ParseInLocation(time.RFC3339, "2006-01-02T15:04:05-07:00", time.UTC)
 	s.NoError(err)
 	tags := []*Tag{
 		{
@@ -200,7 +200,7 @@ func (s *TagSuite) TestTagEncodeDecodeIdempotent() {
 	}
 	for _, tag := range tags {
 		obj := &plumbing.MemoryObject{}
-		err = tag.Encode(obj)
+		err := tag.Encode(obj)
 		s.NoError(err)
 		newTag := &Tag{}
 		err = newTag.Decode(obj)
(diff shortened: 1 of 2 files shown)


cedric-appdirect:
This fix #1755

plumbing/object/tag_test.go
```diff
@@ -180,8 +179,12 @@ func (s *TagSuite) TestTagDecodeWrongType() {
...
 
 func (s *TagSuite) TestTagEncodeDecodeIdempotent() {
-	ts, err := time.Parse(time.RFC3339, "2006-01-02T15:04:05-07:00")
-	s.NoError(err)
+	// Use time.Unix with a fixed zone to avoid timezone database differences.
+	// time.Parse would attach a rich Location from the system timezone database,
+	// but decode uses time.FixedZone which creates a minimal fixed-offset zone.
+	// Using Unix timestamp with FixedZone ensures encode/decode idempotency.
+	tz := time.FixedZone("", -7*60*60) // -07:00
+	ts := time.Unix(1136239445, 0).In(tz) // 2006-01-02T15:04:05-07:00
```
pjbgf: Instead, we can just change `time.Parse` for `time.ParseInLocation`:

```suggestion
	ts, err := time.ParseInLocation(time.RFC3339, "2006-01-02T15:04:05-07:00", time.UTC)
	s.NoError(err)
```

PR #1761, opened 2025-11-27
Enable whitespace and prealloc linters

ferhatelmas:
related to #1753

### DIFF
--- a/plumbing/protocol/packp/capability/list.go
+++ b/plumbing/protocol/packp/capability/list.go
@@ -166,7 +166,11 @@ func (l *List) Delete(capability Capability) {
 
 // All returns a slice with all defined capabilities.
 func (l *List) All() []Capability {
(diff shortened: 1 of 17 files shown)


pjbgf (dismissed):

plumbing/protocol/packp/capability/list.go
```diff
@@ -166,7 +166,7 @@ func (l *List) Delete(capability Capability) {
 
 // All returns a slice with all defined capabilities.
 func (l *List) All() []Capability {
-	var cs []Capability
+	cs := make([]Capability, 0, len(l.sort))
```
pjbgf, line 173: Perhaps we short-circuit this when len equals 0.

ferhatelmas: Yes, test assertion caught it. Updated

PR #1762, opened 2025-11-27
Alternate push bug fix

cedric-appdirect:
This started with a massive memory spike when pushing a single change on a repository using alternate. Issue was due to continuous creation of NewObjectStorage. Additionally I realized that when doing a Clone it was doing a very large amount of data transfer as it wasn't able to find the hash from the alternate as it wasn't doing the lookup at all.

I think the code could be better if I was to use range-over-func, but I am not sure that is ok pattern for this repository. Let me know if that would be preferred or if you have any better idea on how to implement this.


cedric-appdirect:
```
                                            │ ../../go-git/before-alternate.txt │             after.txt              │
                                            │              sec/op               │   sec/op     vs base               │
AlternatesObjectLookup/EncodedObject-12                            1026.4µ ± 2%   189.4µ ± 2%  -81.55% (p=0.001 n=7)
AlternatesObjectLookup/HasEncodedObject-12                                        24.77µ ± 5%
AlternatesObjectLookup/EncodedObjectSize-12                                       26.17µ ± 6%
geomean                                                             1.026m        49.70µ       -81.55%

                                            │ ../../go-git/before-alternate.txt │              after.txt              │
                                            │               B/op                │     B/op      vs base               │
AlternatesObjectLookup/EncodedObject-12                           145.28Ki ± 0%   13.48Ki ± 0%  -90.72% (p=0.001 n=7)
AlternatesObjectLookup/HasEncodedObject-12                                        5.859Ki ± 0%
AlternatesObjectLookup/EncodedObjectSize-12                                       5.859Ki ± 0%
geomean                                                            145.3Ki        7.735Ki       -90.72%

                                            │ ../../go-git/before-alternate.txt │             after.txt             │
                                            │             allocs/op             │ allocs/op   vs base               │
AlternatesObjectLookup/EncodedObject-12                             2515.0 ± 0%   215.0 ± 0%  -91.45% (p=0.001 n=7)
AlternatesObjectLookup/HasEncodedObject-12                                        80.00 ± 0%
AlternatesObjectLookup/EncodedObjectSize-12                                       80.00 ± 0%
geomean                                                             2.515k        111.2       -91.45%
```

storage/filesystem/object.go
```diff
@@ -54,6 +59,24 @@ func NewObjectStorageWithOptions(dir *dotgit.DotGit, objectCache cache.Object, o
 	}
 }
 
+// initAlternates initializes the cached alternate ObjectStorage instances.
+// This is called lazily via sync.Once to ensure thread-safe initialization
+// and to avoid recreating ObjectStorage instances on every object lookup.
+func (s *ObjectStorage) initAlternates() {
+	s.alternatesOnce.Do(func() {
```
pjbgf: This should be reset if `AddAlternate(remote string)` is called.

storage/filesystem/object.go
```diff
@@ -54,6 +59,24 @@ func NewObjectStorageWithOptions(dir *dotgit.DotGit, objectCache cache.Object, o
...
+		dotgits, err := s.dir.Alternates()
+		if err != nil {
```
pjbgf: Let's check if this is `os.ErrNotExist`, if so we probably want to bubble this up.

storage/filesystem/object.go
```diff
@@ -217,10 +240,18 @@ func (s *ObjectStorage) HasEncodedObject(h plumbing.Hash) (err error) {
...
 	}
-	return nil
+
+	// Check cached alternate repositories.
+	s.initAlternates()
+	for _, alt := range s.alternates {
+		if err := alt.HasEncodedObject(h); err == nil {
+			return nil
+		}
+	}
```
pjbgf: We should be able to do this concurrently - first to find wins. It should improve when multiple alternates exist.

storage/filesystem/object.go
```diff
@@ -358,7 +389,22 @@ func (s *ObjectStorage) EncodedObjectSize(h plumbing.Hash) (size int64, err erro
...
+
+	// Check cached alternate repositories.
+	if errors.Is(err, plumbing.ErrObjectNotFound) {
+		s.initAlternates()
+		for _, alt := range s.alternates {
+			size, err = alt.EncodedObjectSize(h)
+			if err == nil {
+				return size, nil
+			}
+		}
```
pjbgf: This is largely the same as the above, so I'd extract into its own helper func. Something like this is the first `alt` I found which has the obj we are looking for.

storage/filesystem/object.go
```diff
@@ -379,20 +425,13 @@ func (s *ObjectStorage) EncodedObject(t plumbing.ObjectType, h plumbing.Hash) (p
...
-			for _, dg := range dotgits {
-				o := NewObjectStorage(dg, s.objectCache)
-				enobj, enerr := o.EncodedObject(t, h)
-				if enerr != nil {
-					continue
-				}
-				return enobj, nil
+		s.initAlternates()
+		for _, alt := range s.alternates {
+			obj, err = alt.EncodedObject(t, h)
```
pjbgf: ditto

storage/filesystem/object.go
```diff
@@ -670,6 +709,13 @@ func (s *ObjectStorage) buildPackfileIters(
 func (s *ObjectStorage) Close() error {
```
pjbgf, line 797: Not sure how often people actually close the ObjectStorage. This could be an issue for long running applications. Potentially we would need to explore other ways around it.

storage/filesystem/object.go
cedric-appdirect: I went ahead and kept a fast path when there is only one alternate quite common and if more use errgroup caped by the number of available core.

storage/filesystem/object.go
cedric-appdirect: I went with a generic, make for clean caller use. It is internal so not impact on user.

storage/filesystem/object.go
cedric-appdirect: In most case, I don't think it has much impact due to the number of pack and the in memory cost. If it become a problem, we can figure something then.

storage/filesystem/object.go
pjbgf: Nice, thanks for that.

PR #1767, opened 2025-12-01
*: prevent panic

### LINKED ISSUE #1051: (*Worktree).Commit will panic if `CommitOptions` parameter is nil.

swills:
Prevent panics if *Options are nil by adding checks in validators. Set default values where appropriate.

Fixes #1051

Hopefully I interpreted the discussion in #1051 properly and this is what is expected here. Please let me know if not.

### DIFF
--- a/options.go
+++ b/options.go
@@ -33,6 +33,8 @@ const (
 
 var ErrMissingURL = errors.New("URL field is required")
 
+var ErrNilOptions = errors.New("nil options")
+
 // CloneOptions describes how a clone should be performed.
 type CloneOptions struct {
 	// The (possibly remote) repository URL to clone from.
@@ -127,6 +129,10 @@ const (
 
 // Validate validates the fields and sets the default values.
 func (o *CloneOptions) Validate() error {
+	if o == nil {
+		return ErrNilOptions
+	}
+
 	if o.URL == "" {
 		return ErrMissingURL
 	}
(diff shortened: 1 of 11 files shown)


onee-only:
In https://github.com/go-git/go-git/pull/1690#discussion_r2463629672, @pjbgf wrote:
> I'm planning to raise an RFC on the subject of API consolidation which would affect this (and the general "WithOption" funcs), just haven't got around to it yet. 

This may conflict with the upcoming API changes. If they introduce the new "WithOption" functions before the next release, these changes could be overwritten.

PR #1782, opened 2025-12-04
transport: fix opening relative paths

### LINKED ISSUE #1723: v6: Clone of parent fails (worked in v5)
### Bug Description

Here is a test which shows the issue (works in v5 fails in v6):
```golang
func TestCloneParentDir(t *testing.T) {
	// clone parent directory to in-memory storage
	_, err := git.Clone(memory.NewStorage(), nil, &git.CloneOptions{
		URL: "../../",
	})
	require.NoError(t, err)
}
```

### go-git Version

v6.0.0-20251112161705-8cc3e21f07a9

### Steps to Reproduce

_No response_

### Additional Information

_No response_

jfontan:
go-billy checks that in a chroot filesystem the paths do not start by `../`. This causes opening or cloning relative paths to fail. This check is good to find repository links that point to files outside the repository but for opening path is not that useful as the default filesystem is based at `/`.

The change to `parseFile` does the same as v5. The test and the embedded function to generate the relative path is quite naive but checks that the change fixes the issue.

Fixes #1723

### DIFF
--- a/plumbing/transport/loader.go
+++ b/plumbing/transport/loader.go
@@ -13,7 +13,7 @@ import (
 )
 
 // DefaultLoader is a filesystem loader ignoring host and resolving paths to /.
-var DefaultLoader = NewFilesystemLoader(osfs.New(""), false)
+var DefaultLoader = NewFilesystemLoader(osfs.New("", osfs.WithBoundOS()), false)
 
 // Loader loads repository's storer.Storer based on an optional host and a path.
 type Loader interface {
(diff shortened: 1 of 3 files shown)


jfontan:
I'll go back to the drawing board as I'm not happy with the test or the solution.

jfontan:
I'm still checking how I can test it and make it work with Windows but I believe BoundOS could be better fitted for the default FS for loader.

jfontan:
The error in Windows tests is given by `securejoin` checking for symlinks. I'm still not sure if the problem is `securejoin` implementation striping the drive letter or some other problem.

```
CreateFile \\Users: The specified path is invalid.
```

Issue #1772, opened 2025-12-03
Configure Renovate to bump go-billy for `releases/v5.x`


pjbgf:
### Feature Description

All non-security related bumps [have been disabled](https://github.com/go-git/go-git/commit/357e586e36e49b01021aa20aa0026dde584c1c74) for `releases/v5.x`, as that release line is in maintenance mode.

We should re-enable the bump of go-billy against that branch for versions below `v6`.

PR #1783, opened 2025-12-05
Re-enable the bump of go-billy

nodivbyzero:
This PR re-enables the bump of go-billy in the `renovaterc.json`

Related to: https://github.com/go-git/go-git/issues/1772

### DIFF
--- a/.renovaterc.json5
+++ b/.renovaterc.json5
@@ -12,6 +12,14 @@
       "matchBaseBranches": [
         "releases/v5.x"
       ]
+    },
+    {
+      "description": "Enable minor and patch updates for go-billy",
+      "matchPackageNames": [
+        "github.com/go-git/go-billy",
+      ],
+      "matchUpdateTypes": ["minor", "patch"],
+      "enabled": true
     }
   ],
   "vulnerabilityAlerts": {


.renovaterc.json5
```diff
@@ -7,8 +7,8 @@
   ],
   "packageRules": [
     {
-      "description": "Disable non-security bumps for the v5 release line",
-      "enabled": false,
+      "description": Enable non-security bumps for the v5 release line",
```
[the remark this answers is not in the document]
nodivbyzero: Fixed

.renovaterc.json5
```diff
@@ -7,8 +7,8 @@
   ],
   "packageRules": [
     {
-      "description": "Disable non-security bumps for the v5 release line",
-      "enabled": false,
```
pjbgf, line 11: This enables bumps for all packages. We want to limit this instead.

We will need to revert these changes, and add a new package rule around the lines of:
```json
    {
      "description": "Enable minor and patch updates for go-billy",
      "matchPackageNames": [
        "github.com/go-git/go-billy",
      ],
      "matchUpdateTypes": ["minor", "patch"],
      "enabled": true
    },
```

pjbgf (changes requested):

.renovaterc.json5
nodivbyzero: @pjbgf Thank you for pointing me toward the right solution. 
I've updated this PR with your suggestion - please take a look when you have a moment.

pjbgf (approved):
@nodivbyzero thanks for working on this.

PR #1786, opened 2025-12-05
*: enable errcheck linter

ferhatelmas:
Related to https://github.com/go-git/go-git/issues/1753

### DIFF
--- a/plumbing/transport/http/common.go
+++ b/plumbing/transport/http/common.go
@@ -33,6 +33,52 @@ func init() {
 	transport.Register("https", DefaultTransport)
 }
 
+// safeHeaders is a list of HTTP headers that are safe to log for debugging
+// the Git protocol. Headers not in this list may contain sensitive information
+// (e.g., Authorization, Cookie, X-Auth-Token) and should not be logged.
+var safeHeaders = map[string]struct{}{
+	"User-Agent":        {},
+	"Host":              {},
+	"Accept":            {},
+	"Content-Type":      {},
+	"Content-Length":    {},
+	"Cache-Control":     {},
+	"Git-Protocol":      {},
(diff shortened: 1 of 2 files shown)


ferhatelmas:
@pjbgf next one is ready. 

CodeQL is failing due to printing user input when tracing is enabled. Are we going to do something here or out of scope of this change?

plumbing/format/packfile/scanner.go
```diff
@@ -164,8 +164,8 @@ func (r *Scanner) Scan() bool {
 // Reset resets the current scanner, enabling it to be used to scan the
 // same Packfile again.
 func (r *Scanner) Reset() {
-	r.Flush()
-	r.Seek(0, io.SeekStart)
+	_ = r.Flush()
+	_, _ = r.Seek(0, io.SeekStart)
```
pjbgf: In these two cases, we want to return an error if it fails - and handle it at the caller accordingly.

pjbgf (changes requested):
@ferhatelmas let's handle that here too please as a separate commit.

The CodeQL alert is reasonable, I think we should change [this](https://github.com/go-git/go-git/blob/593756deab3de8202602f2657f1f8363d0345f14/plumbing/transport/http/common.go#L71-L71) to only trace a limited list of headers that could be useful when debugging the Git protocol, perhaps:
```
User-Agent
Host
Accept
Content-Type
Cache-Control
Git-Protocol
```

plumbing/format/packfile/scanner.go
ferhatelmas: Updated to check errors

ferhatelmas:
> let's handle that here too please as a separate commit.

added filtering

PR #1792, opened 2025-12-09
git: include shallow references in object walker

erka:
When walking objects in a repository, the objectWalker should check shallow references in addition to regular refs. This ensures that shallow clones are properly handled during operations like pruning or repacking.

Similar to #1785

### DIFF
--- a/object_walker.go
+++ b/object_walker.go
@@ -1,7 +1,9 @@
 package git
 
 import (
+	"errors"
 	"fmt"
+	"slices"
 
 	"github.com/go-git/go-git/v6/plumbing"
 	"github.com/go-git/go-git/v6/plumbing/filemode"
@@ -72,6 +74,13 @@ func (p *objectWalker) walkObjectTree(hash plumbing.Hash) error {
 		for _, h := range obj.ParentHashes {
 			err = p.walkObjectTree(h)
 			if err != nil {
+				shallow, serr := p.Storer.Shallow()
+				if serr != nil {
+					return errors.Join(err, serr)
+				}
+				if slices.Contains(shallow, obj.ID()) {
+					return nil
+				}
 				return err
 			}
 		}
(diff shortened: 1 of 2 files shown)


object_walker.go
```diff
@@ -72,6 +74,13 @@ func (p *objectWalker) walkObjectTree(hash plumbing.Hash) error {
 		for _, h := range obj.ParentHashes {
 			err = p.walkObjectTree(h)
 			if err != nil {
+				shallow, serr := p.Storer.Shallow()
```
pjbgf: This should be gated with `errors.Is` for `plumbing.ErrObjectNotFound`.

object_walker.go
pjbgf:
```suggestion
		return fmt.Errorf("getting object %s failed: %w", hash, err)
```

object_walker_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+type objectWalkerSuite struct {
+	BaseSuite
+}
+
+func TestObjectWalkerSuite(t *testing.T) {
+	suite.Run(t, new(objectWalkerSuite))
+}
+
+func (s *objectWalkerSuite) TestNormalClonedRepo() {
```
pjbgf, line 42: Add another test to ensure that other errors are not silently ignored (e.g. I/O, disk permissions, etc).

pjbgf (changes requested):
@erka thanks for proposing this fix. I added some comments above, let me know if anything isn't clear.

erka:
Thanks for your review @pjbgf. I've addressed your points.

pjbgf (dismissed):
@erka thanks for working on this. 🙇 

ℹ️ I added an additional commit to fix two things:
- The regression that replaced `err != nil` with the check for object not found. We needed both workflows, so that was resulting in some errors being ignored.
- Started front-loading the shallow list and verification. That is needed as currently there is no caching in `.Shallow()`.

Issue #1766, opened 2025-11-30
pktline.ReadLine() isn't safe for concurrent usage


onee-only:
### Bug Description

When `pktline.ReadLine()`is called, the function returns its buffer to the pool while keeping the reference of the buffer. This can lead to memory corruption, as the buffer may be reused and overwritten by other goroutines at any time.
https://github.com/go-git/go-git/blob/48f817f578055cf6f4728062cba3091701c09a92/plumbing/format/pktline/pktline.go#L164-L174

### go-git Version

48f817f578055cf6f4728062cba3091701c09a92

### Steps to Reproduce

```go
func TestConcurrentReadLine(t *testing.T) {
	var wg sync.WaitGroup
	for n := range 30 {
		wg.Add(1)
		go func() {
			defer wg.Done()

			data := strconv.Itoa(n)

			var buf bytes.Buffer
			_, err := pktline.Write(&buf, []byte(data))
			require.NoError(t, err)

			_, p, err := pktline.ReadLine(&buf)
			require.NoError(t, err)
			assert.Equal(t, data, string(p))

		}()
	}
	wg.Wait()
}
```

PR #1798, opened 2025-12-11
*: enable paralleltest linter

ferhatelmas:
Related to #1753

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -10,6 +10,7 @@ linters:
     - misspell
     - modernize
     - nolintlint
+    - paralleltest
     - prealloc
     - revive
     - staticcheck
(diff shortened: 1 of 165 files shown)


ferhatelmas:
Seems like 25% speedup

image
image

onee-only:
I'm surprised that the test workflow didn't result in data race as described in https://github.com/go-git/go-git/issues/1766. In my local test, a lot of tests failed due to the race. Is this only happening in my machine?

pjbgf:
@ferhatelmas looking across a range of successful executions, it feels like the overall execution time hasn't changed much:
image

The same seems to be the case when looking solely at the time taken for the Test to run:
image

I'm not sure this is because `GOMAXPROCS` is not set properly within the GH runner, or for other reason. But given that the focus of this PR is to enable the linter and ensure the code complies, I'm happy to merge as-is and any additional improvement on test execution time can be done on a follow-up PR.

pjbgf (approved):

Issue #1413, opened 2025-01-30
Review `go-git` dependencies


pjbgf:
Ahead of the `v6` release, review all existing dependencies evaluating whether they are still needed and well-maintained, or better alternatives are available.

Additional dependency related work:
- [ ] Review [changes](https://github.com/go-git/go-git/pull/1172/commits/c83efa5acd8e64ca78d455a773a2a3a54ba67503) which embedded the upstream dependencies, and look for opportunities to upstream some/all the changes. Affected dependencies: `github.com/Microsoft/go-winio`, `github.com/xanzy/ssh-agent` and `github.com/skeema/knownhosts`. Reasoning behind the initial transition:
  - > The changes enable a deeper level of instrumentation when tracing SSH connections, and it does so by injecting calls to `trace.SSH.Printf` in areas that are deep in the upstream code.

pjbgf:
Hey @evanelias, moving the [thread](https://github.com/go-git/go-git/pull/1346#issuecomment-2627762141) here, to avoid talking on closed issues/PRs. This issue relates to the review of `go-git` dependencies, which includes `github.com/skeema/knownhosts`.

> my takeaway from these interactions is that it appears that the substantial amount of time spent on recent improvements in [skeema/knownhosts](https://github.com/skeema/knownhosts) has been a completely wasted effort: go-git is the largest external direct importer of skeema/knownhosts, but you are no longer using any of the recent knownhosts improvements in go-git v5, and meanwhile in go-git v6 you've made an internal fork of skeema/knownhosts in order to add a single debug printf (but without ever reporting a corresponding problem to skeema/knownhosts).

I [created](https://github.com/go-git/go-git/issues/1417) an issue to track re-adding the feature back to `v5`, I agree it would make sense to add it back.
For `v6`, nothing has been decided as such. As I said, the changes were meant to increase traceability during SSH hand-shake, a follow-up task was to assess upstreaming those changes / on what the future would look like on those changes, and this issue is here to track exactly that. I will review this and file issues upstream to discuss that possibility.

PR #1799, opened 2025-12-12
*: drop goproxy dependency

ferhatelmas:
Assuming go-git is a library and so it's better to use minimal dependencies. github.com/elazarl/goproxy is only used by tests, adding a tiny implementation enables us to drop it.


pjbgf:
@ferhatelmas overall LGTM. Please rebase.

Relates to #1413.

ferhatelmas:
@pjbgf rebased

PR #1801, opened 2025-12-13
storage: filesystem/dotgit, fixing issue where absolute paths to commonDotGitFs ended up in dotGitFs

Soph:
When using the v6 implementation with worktree repos we got this error: 

```
Add failed: "/tmp/gogit-worktree-test/main-repo/.git/objects/pack/tmp_obj_847737165": path outside base dir "/tmp/gogit-worktree-test/main-repo/.git/worktrees/worktree": file does not exist
```

The current implementation assumes path are relativ but for tmp files (in combination with worktree!?) they are absolute. So the current logic that decides which filesystem should be returned isn't working at all and always returns `dotGitFs` but it should return `commonDotGitFs` instead.

This came out of a debugging session with Claude Code first on our code base and then on go-git main branch. I reviewed the code and I'm pretty confident it's a good approach but still wanted to flag it for transparency. Especially the amount of comments is a bit on the verbose side, so happy to remove/reduce if that's too much. 

Here is a script to replicate this: 
<details>
<summary>open for source</summary>

```
package main

  import (
      "fmt"
      "os"
      "os/exec"
      "path/filepath"
      "time"

      "github.com/go-git/go-git/v6"
  )

  func main() {
      // Create test structure
      tmpDir := "/tmp/gogit-worktree-test"
      os.RemoveAll(tmpDir)
      os.MkdirAll(tmpDir, 0755)

      mainRepoDir := filepath.Join(tmpDir, "main-repo")
      worktreeDir := filepath.Join(tmpDir, "worktree")

      // Init main repo
      exec.Command("git", "init", mainRepoDir).Run()

      // Create initial commit
      os.WriteFile(filepath.Join(mainRepoDir, "file.txt"), []byte("test"), 0644)
      exec.Command("git", "-C", mainRepoDir, "add", ".").Run()
      exec.Command("git", "-C", mainRepoDir, "commit", "-m", "initial").Run()

      // Create worktree
      exec.Command("git", "-C", mainRepoDir, "worktree", "add", worktreeDir, "-b", "test-branch").Run()

      // Change to worktree directory
      os.Chdir(worktreeDir)

      // Open with go-git v6
      repo, err := git.PlainOpenWithOptions(".", &git.PlainOpenOptions{
          EnableDotGitCommonDir: true,
      })
      if err != nil {
          fmt.Printf("Open failed: %v\n", err)
          return
      }

      wt, _ := repo.Worktree()

      // Create a new file
      os.WriteFile(filepath.Join(worktreeDir, "newfile.txt"), []byte("content"), 0644)

      // Try to add it - THIS FAILS
      _, err = wt.Add("newfile.txt")
      if err != nil {
          fmt.Printf("Add failed: %v\n", err)
          return
      }

      // Try to commit
      _, err = wt.Commit("test commit", &git.CommitOptions{
          Author: &object.Signature{Name: "Test", Email: "test@test.com", When: time.Now()},
      })
      if err != nil {
          fmt.Printf("Commit failed: %v\n", err)
          return
      }

      fmt.Println("Success!")
  }
```
</details>

### DIFF
--- a/storage/filesystem/dotgit/repository_filesystem.go
+++ b/storage/filesystem/dotgit/repository_filesystem.go
@@ -32,6 +32,24 @@ func (fs *RepositoryFilesystem) mapToRepositoryFsByPath(path string) billy.Files
 
 	cleanPath := filepath.Clean(path)
 
+	// Handle absolute paths by checking if they're under commonDotGitFs or dotGitFs.
+	// This is needed because temp files return absolute paths from Name(), and operations
+	// like Rename need to route to the correct filesystem.
+	if filepath.IsAbs(cleanPath) {
+		commonRoot := fs.commonDotGitFs.Root()
+		dotGitRoot := fs.dotGitFs.Root()
+
+		if strings.HasPrefix(cleanPath, commonRoot+string(filepath.Separator)) || cleanPath == commonRoot {
+			return fs.commonDotGitFs
+		}
+		if strings.HasPrefix(cleanPath, dotGitRoot+string(filepath.Separator)) || cleanPath == dotGitRoot {
+			return fs.dotGitFs
+		}
+		// Absolute path doesn't match either root - default to dotGitFs.
+		// This shouldn't occur in normal usage.
+		return fs.dotGitFs
+	}
+
 	// Check exceptions for commondir (https://git-scm.com/docs/gitrepository-layout#Documentation/gitrepository-layout.txt)
 	switch cleanPath {
 	case fs.dotGitFs.Join(logsPath, "HEAD"):
--- a/storage/filesystem/dotgit/repository_filesystem_test.go
+++ b/storage/filesystem/dotgit/repository_filesystem_test.go
@@ -110,3 +110,127 @@ func (s *SuiteDotGit) TestRepositoryFilesystem() {
 	_, err = dotGitFs.Stat("a/b/c")
 	s.Require().NoError(err)
 }
+
+// TestRepositoryFilesystemTempFileRename tests the TempFile + Rename flow
+// which is used by ObjectWriter.save(). This is critical for worktree support
+// where temp files are created in commonDotGitFs.
+func (s *SuiteDotGit) TestRepositoryFilesystemTempFileRename() {
+	fs := s.EmptyFS()
+
+	err := fs.MkdirAll("dotGit", 0o777)
+	s.Require().NoError(err)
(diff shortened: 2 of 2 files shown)


storage/filesystem/dotgit/repository_filesystem.go
```diff
@@ -32,6 +32,24 @@ func (fs *RepositoryFilesystem) mapToRepositoryFsByPath(path string) billy.Files
...
 	cleanPath := filepath.Clean(path)
 
+	// Handle absolute paths by checking if they're under commonDotGitFs or dotGitFs.
+	// This is needed because temp files return absolute paths from Name(), and operations
+	// like Rename need to route to the correct filesystem.
+	if filepath.IsAbs(cleanPath) {
+		commonRoot := fs.commonDotGitFs.Root()
+		dotGitRoot := fs.dotGitFs.Root()
+
+		if strings.HasPrefix(cleanPath, commonRoot+string(filepath.Separator)) || cleanPath == commonRoot {
```
pjbgf, line 42: This might break depending on the filesystem implementation. Using [osfs.New(WithChrootOS())](https://github.com/go-git/go-billy/blob/8afc3eb64e4db787c3ceeef70bdfda2cd4c87c41/osfs/os.go#L50) the filesystem represents the repo dir as `/`.
Using [WithBoundOS](https://github.com/go-git/go-billy/blob/8afc3eb64e4db787c3ceeef70bdfda2cd4c87c41/osfs/os.go#L43C6-L43C17) they retain their underlying path (e.g. `/path/to/repo`).

In V6 we use the latter for both [commonDir creation](https://github.com/go-git/go-git/blob/e83cbb9651e85c65beb9e807780efc2145452306/repository.go#L495) and the repository [dotGitFs](https://github.com/go-git/go-git/blob/e83cbb9651e85c65beb9e807780efc2145452306/repository.go#L404). This would become a problem once `EnableDotGitCommonDir` is enabled by default or we add support for setting it in "non-Plain" funcs (e.g. `git.Open`).

Let's add a type check in `NewRepositoryFilesystem` that sets an internal boolean field which enables this logic when both filesystems provided have the same type.

pjbgf (changes requested):

storage/filesystem/dotgit/repository_filesystem.go
pjbgf: I've tested the cross-filesystem implementation with this and it seems to work fine, so marking this as resolved.

PR #1804, opened 2025-12-16
*: enable gosec linter

ferhatelmas:
related to #1753

### DIFF
--- a/plumbing/format/commitgraph/encoder.go
+++ b/plumbing/format/commitgraph/encoder.go
@@ -33,6 +33,7 @@ func (e *Encoder) Encode(idx Index) error {
 	hashToIndex, fanout, extraEdgesCount, generationV2OverflowCount := e.prepare(idx, hashes)
 
 	chunkSignatures := [][]byte{OIDFanoutChunk.Signature(), OIDLookupChunk.Signature(), CommitDataChunk.Signature()}
+	//nolint:gosec // G115: len() and hash.Size() are always small positive values
 	chunkSizes := []uint64{szUint32 * lenFanout, uint64(len(hashes) * e.hash.Size()), uint64(len(hashes) * (e.hash.Size() + szCommitData))}
 	if extraEdgesCount > 0 {
 		chunkSignatures = append(chunkSignatures, ExtraEdgeListChunk.Signature())
@@ -86,7 +87,7 @@ func (e *Encoder) prepare(idx Index, hashes []plumbing.Hash) (hashToIndex map[pl
 	hashToIndex = make(map[plumbing.Hash]uint32)
 	fanout = make([]uint32, lenFanout)
 	for i, hash := range hashes {
-		hashToIndex[hash] = uint32(i)
+		hashToIndex[hash] = uint32(i) //nolint:gosec // G115: i is loop index bounded by hashes count
 		fanout[hash.Bytes()[0]]++
 	}
 
@@ -99,9 +100,9 @@ func (e *Encoder) prepare(idx Index, hashes []plumbing.Hash) (hashToIndex map[pl
 
 	// Find out if we will need extra edge table
 	for i := range len(hashes) {
-		v, _ := idx.GetCommitDataByIndex(uint32(i))
+		v, _ := idx.GetCommitDataByIndex(uint32(i)) //nolint:gosec // G115: i is loop index
 		if len(v.ParentHashes) > 2 {
-			extraEdgesCount += uint32(len(v.ParentHashes) - 1)
+			extraEdgesCount += uint32(len(v.ParentHashes) - 1) //nolint:gosec // G115: parent count is small
 		}
 		if hasGenerationV2 && v.GenerationV2Data() > math.MaxUint32 {
 			generationV2OverflowCount++
@@ -114,7 +115,7 @@ func (e *Encoder) prepare(idx Index, hashes []plumbing.Hash) (hashToIndex map[pl
 func (e *Encoder) encodeFileHeader(chunkCount int) (err error) {
 	if _, err = e.Write(commitFileSignature); err == nil {
 		version := byte(1)
-		if crypto.Hash(e.hash.Size()) == crypto.Hash(crypto.SHA256.Size()) {
+		if crypto.Hash(e.hash.Size()) == crypto.Hash(crypto.SHA256.Size()) { //nolint:gosec // G115: hash.Size() is always small positive
 			version = byte(2)
 		}
 		_, err = e.Write([]byte{1, version, byte(chunkCount), 0})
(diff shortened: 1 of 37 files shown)


ferhatelmas:
@pjbgf last one in the series should be good to go

plumbing/format/commitgraph/encoder.go
```diff
@@ -99,9 +100,9 @@ func (e *Encoder) prepare(idx Index, hashes []plumbing.Hash) (hashToIndex map[pl
 
 	// Find out if we will need extra edge table
 	for i := range len(hashes) {
-		v, _ := idx.GetCommitDataByIndex(uint32(i))
+		v, _ := idx.GetCommitDataByIndex(uint32(i)) //nolint:gosec // G115: i is loop index
```
pjbgf, line 103: Slightly orthogonal, but let's take this opportunity and handle the `err` as that would mean `v` is then `nil`.
```suggestion
		v, err := idx.GetCommitDataByIndex(uint32(i)) //nolint:gosec // G115: i is loop index
```

plumbing/format/idxfile/idxfile.go
```diff
@@ -95,14 +95,16 @@ func (idx *MemoryIndex) findHashIndex(h plumbing.Hash) (int, bool) {
 	low := uint64(0)
 	for {
 		mid := (low + high) >> 1
+		//nolint:gosec // G115: idSize() returns small hash size (20 or 32)
 		offset := mid * uint64(idx.idSize())
 
+		//nolint:gosec // G115: idSize() returns small hash size (20 or 32)
```
pjbgf, line 101: Although the hash size is small, we still should verify whether the end position is within `data` bounds.

```suggestion
end := offset+uint64(idx.idSize())
if end > len(data) {
  return 0, false
}
```

worktree.go
```diff
@@ -897,7 +897,7 @@ func (w *Worktree) addIndexFromFile(name string, h plumbing.Hash, idx *indexBuil
 		Name:       name,
 		Mode:       mode,
 		ModifiedAt: fi.ModTime(),
-		Size:       uint32(fi.Size()),
+		Size:       uint32(fi.Size()), //nolint:gosec // G115: file size fits in uint32 for git index
```
pjbgf, line 900: > file size fits in uint32 for git index

I think this is not correct when considering very large files. IIRC upstream handles this by using bitmask and setting an arbitrary value for a size it knows to be rather large.

Instead of ignoring the lint error, we are better off checking `fi.Size()` and erroring if greater than `uint32`. Supporting larger files can be a follow-up issue/PR.

worktree_bsd.go
```diff
@@ -13,8 +13,8 @@ func init() {
 	fillSystemInfo = func(e *index.Entry, sys any) {
 		if os, ok := sys.(*syscall.Stat_t); ok {
 			e.CreatedAt = time.Unix(os.Atimespec.Unix())
-			e.Dev = uint32(os.Dev)
-			e.Inode = uint32(os.Ino)
+			e.Dev = uint32(os.Dev)   //nolint:gosec // G115: dev number fits in uint32
+			e.Inode = uint32(os.Ino) //nolint:gosec // G115: inode fits in uint32
```
pjbgf, line 17: Both `os.Dev` and `os.Ino` are OS-dependant values. Some implementations are [uint32](https://github.com/golang/go/blob/b28808d838682b2911698fcb934813b02f39fa69/src/syscall/syscall_linux_mips64x.go#L107-L109), but [others](https://github.com/golang/go/blob/b28808d838682b2911698fcb934813b02f39fa69/src/syscall/fs_wasip1.go#L640-L641) are `uint64`, which makes me think in some cases this may overflow.

That being said, upstream relies solely on the [uint32 counterpart](https://github.com/git/git/blob/68cb7f9e92a5d8e9824f5b52ac3d0a9d8f653dbe/statinfo.h#L19C2-L20) for it. Due to that, I'm leaning towards:
- Keep `index.Entry` types as they are.
- Add check for value and error if > `math.MaxUint32`.

plumbing/transport/http/common.go
```diff
@@ -271,7 +271,7 @@ func transportWithCABundle(transport *http.Transport, caBundle []byte) error {
 	}
 	rootCAs.AppendCertsFromPEM(caBundle)
 	if transport.TLSClientConfig == nil {
-		transport.TLSClientConfig = &tls.Config{}
+		transport.TLSClientConfig = &tls.Config{} //nolint:gosec // G402: MinVersion will use Go's secure default
```
pjbgf, line 274:
```suggestion
		transport.TLSClientConfig = &tls.Config{} //nolint:gosec // G402: MinVersion will use Go's current default value - for Go 1.18+ that is TLS 1.2.
```

plumbing/object/rename.go
```diff
@@ -775,7 +775,7 @@ func (i *similarityIndex) grow() error {
 
 	// TODO(erizocosmico): find a way to check if it will OOM and return
 	// errIndexFull instead.
-	i.hashes = make([]keyCountPair, 1<<uint(i.hashBits))
+	i.hashes = make([]keyCountPair, 1<<uint(i.hashBits)) //nolint:gosec // G115: hashBits is bounded
```
pjbgf, line 778: I'm not sure about this, I will take a deeper look at it on a follow-up review.

plumbing/format/packfile/patch_delta.go
```diff
@@ -386,10 +386,10 @@ func patchDeltaWriter(dst io.Writer, base io.ReaderAt, delta io.Reader,
 				return 0, plumbing.ZeroHash, err
 			}
 
-			if _, err := sr.Seek(int64(offset), io.SeekStart); err != nil {
+			if _, err := sr.Seek(int64(offset), io.SeekStart); err != nil { //nolint:gosec // G115: offset fits in int64
 				return 0, plumbing.ZeroHash, err
 			}
-			baselr.N = int64(sz)
+			baselr.N = int64(sz) //nolint:gosec // G115: sz fits in int64
```
pjbgf, line 392: Not sure this is correct. `decodeSizeByteReader` returns `uint` which in theory could overflow `int64`.

plumbing/format/packfile/patch_delta.go
```diff
@@ -334,28 +334,28 @@ func patchDeltaWriter(dst io.Writer, base io.ReaderAt, delta io.Reader,
...
+		err = writeHeader(typ, int64(targetSz)) //nolint:gosec // G115: targetSz is delta target size
 		if err != nil {
 			return 0, plumbing.ZeroHash, fmt.Errorf("could not lazy write header: %w", err)
 		}
 	}
 
 	remainingTargetSz := targetSz
 
-	hasher := plumbing.NewHasher(format.SHA1, typ, int64(targetSz))
+	hasher := plumbing.NewHasher(format.SHA1, typ, int64(targetSz)) //nolint:gosec // G115: targetSz fits in int64
```
pjbgf, line 352: Similarly, `int64(targetSz)` in theory could overflow. I guess we could change `DecodeLEB128FromReader`'s return type to avoid the different casts around this package.

pjbgf (changes requested):
@ferhatelmas thanks for looking into this.

PR #1809, opened 2025-12-23
storage: filesystem/readerat, Add PackScanner to handle repository more efficiently in a portable way

Bluebugs:
This is heavily inspired from https://github.com/go-git/go-git/pull/1776 , but using ReaderAt . This is not as effective as mmap for large repository, but it is portable and does not impact goroutine latency.

Performance wise, it performs competitively with the standard packfile for small repository:
```
BenchmarkPackHandlers/packfile-cache-osfs:_basic-fixture-16         	    5436	    186619 ns/op	   92750 B/op	     613 allocs/op
BenchmarkPackHandlers/packfile-nocache-memfs:basic-fixture-16       	    8757	    121795 ns/op	   92632 B/op	     612 allocs/op
BenchmarkPackHandlers/mmap-pack-scanner:basic-fixture-16            	   17748	     69070 ns/op	   91636 B/op	     157 allocs/op
BenchmarkPackHandlers/readerat-pack-scanner:basic-fixture-16        	    5557	    201515 ns/op	   97431 B/op	     238 allocs/op
```

In very large repository, it is slower and will be noticable at scale than the mmap variant, but it is a lot faster than the other alternative.
```
BenchmarkPackHandlers/packfile-cache-osfs:_linux-kernel-16          	       1	2876558304 ns/op	1333904400 B/op	   34439 allocs/op
BenchmarkPackHandlers/packfile-nocache-memfs:linux-kernel-16        	       1	2458490614 ns/op	1333902024 B/op	   34435 allocs/op
BenchmarkPackHandlers/mmap-pack-scanner:linux-kernel-16             	    1318	    942603 ns/op	  182449 B/op	     307 allocs/op
BenchmarkPackHandlers/readerat-pack-scanner:linux-kernel-16         	    1017	   1087961 ns/op	  191833 B/op	     449 allocs/op
```

This PR is a draft as it is obviously duplicating a lot of code from the mmap PR and also because it raise the question of how to refactor it as performance is really not far from the current solution. Guidance on what path should be taken to get this PR in a good state to be merged will be appreciated.

### DIFF
--- a/tests/pack/bench_test.go
+++ b/tests/pack/bench_test.go
@@ -101,6 +101,10 @@ func BenchmarkPackHandlers(b *testing.B) {
 				return newPackScanner(pack, idx, rev)
 			}, data.offsetHashMap)
 		}
+
+		runBenchmark(b, "readerat-pack-scanner:"+data.name, func() packHandler[uint64] {
+			return newPackScannerAt(pack, idx, rev)
+		}, data.offsetHashMap)
 	}
 }
 
--- a/tests/pack/common_test.go
+++ b/tests/pack/common_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/go-git/go-git/v6/plumbing"
 	"github.com/go-git/go-git/v6/plumbing/format/idxfile"
 	"github.com/go-git/go-git/v6/plumbing/format/packfile"
+	"github.com/go-git/go-git/v6/storage/filesystem/readerat"
 )
 
 // The existing implementation uses int64, instead of uint64, which is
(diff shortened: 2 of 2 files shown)


pjbgf:
@Bluebugs thanks for proposing this.

The `mmap` implementation is a bit of an exception as it is platform-specific and would not fit properly into the primitives provided by `go-git`. IMO, `ReadAt` is better applied on an overhaul of the existing implementation within `plumbing/format`. Either as an option (pre-loaded vs on-demand memory index) or as the new norm if that is consistently performant.

My recommendation would be to look into applying the use of `ReadAt` within `plumbing/format/idxfile`. In an on-demand implementation (as per mmap's), we would no longer need `genOffsetHash` which preloads the resolved object into memory. The logic needed for that already exists between those two packages. Feel free to either open a PR to that effect, or reach out via [Discord](https://github.com/go-git/go-git/blob/main/CONTRIBUTING.md#support-channels) to bounce ideas about it.

Bluebugs:
Thanks for the feedback, working on a version that is better integrated.

PR #1811, opened 2025-12-27
feat/readerat for idxfile and revfile

Bluebugs:
This is a follow up after the work on mmap and my initial quick implementation in a separate package. This attempt make it the default behavior as much as possible. MemoryIndex is kept as is for now.

Performance is now better than it was in standalone, not to sure why. Better than memidx for small repository benchmark:
```
BenchmarkPackHandlers/memidx-cache-osfs:____basic-fixture-16         	    7086	    175030 ns/op	   92733 B/op	     613 allocs/op
BenchmarkPackHandlers/memidx-nocache-memfs:_basic-fixture-16         	   10000	    118263 ns/op	   92793 B/op	     612 allocs/op
BenchmarkPackHandlers/readerat-cache-osfs:__basic-fixture-16         	    5845	    190883 ns/op	   78265 B/op	     256 allocs/op
BenchmarkPackHandlers/readerat-nocache-memfs:basic-fixture-16        	    8400	    145296 ns/op	   78290 B/op	     255 allocs/op
BenchmarkPackHandlers/mmap-pack-scanner:____basic-fixture-16         	   19554	     62925 ns/op	   91838 B/op	     157 allocs/op
```

And it is competing with mmap approach for larger repository. Still higher memory usage, but interesting tradeoff that come with portability:
```
BenchmarkPackHandlers/memidx-cache-osfs:____linux-kernel-16          	       1	2746504459 ns/op	1333904496 B/op	   34440 allocs/op
BenchmarkPackHandlers/memidx-nocache-memfs:_linux-kernel-16          	       1	2691410328 ns/op	1333902024 B/op	   34435 allocs/op
BenchmarkPackHandlers/readerat-cache-osfs:__linux-kernel-16          	    1369	    844128 ns/op	  123838 B/op	     530 allocs/op
BenchmarkPackHandlers/readerat-nocache-memfs:linux-kernel-16         	    1501	    848440 ns/op	  124057 B/op	     529 allocs/op
BenchmarkPackHandlers/mmap-pack-scanner:____linux-kernel-16          	    1302	    894901 ns/op	  179149 B/op	     307 allocs/op
```

### DIFF
--- a/plumbing/format/packfile/packfile.go
+++ b/plumbing/format/packfile/packfile.go
@@ -258,6 +258,9 @@ func (p *Packfile) headerFromOffset(offset int64) (*ObjectHeader, error) {
 }
 
 // Close the packfile and its resources.
+// Note: This does not close the embedded Index, as it may be shared
+// with other components. The caller is responsible for closing the Index
+// when it's no longer needed.
 func (p *Packfile) Close() error {
 	p.m.Lock()
 	defer p.m.Unlock()
--- a/plumbing/format/packfile/packfile_test.go
+++ b/plumbing/format/packfile/packfile_test.go
@@ -371,3 +371,36 @@ var expectedEntries = map[plumbing.Hash]int64{
 	plumbing.NewHash("eba74343e2f15d62adedfd8c883ee0262b5c8021"): 84708,
 	plumbing.NewHash("fb72698cab7617ac416264415f13224dfd7a165e"): 84671,
 }
+
+// TestGetWithReaderAtIndex tests that ReaderAtIndex works with Packfile.
+func TestGetWithReaderAtIndex(t *testing.T) {
+	t.Parallel()
+
+	fixture := fixtures.NewOSFixture(fixtures.Basic().One(), t.TempDir())
+	idxFile := fixture.Idx()
+	defer idxFile.Close()
+
+	index, err := idxfile.NewReaderAtIndex(idxFile, crypto.SHA1.Size())
+	require.NoError(t, err)
+	defer index.Close()
+
+	n, err := index.Count()
+	require.NoError(t, err)
+	require.Equal(t, int64(31), n)
+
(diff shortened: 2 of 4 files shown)


Bluebugs:
There is a lot more going on in this PR than in the standalone package approach, so I took a few design decision. I had to introduce a link between RevFile and IdxFile as it really impact performance. I went with an additional function call, but if you do not set that RevFile, the ReaderAt is not going to be fast. That make the API a bit of an easy one to miss. Not to sure if there is a better approach here.

plumbing/format/idxfile/offset_hash.go
```diff
@@ -0,0 +1,91 @@
...
+	if c.cache == nil {
+		c.cache = make(map[int64]plumbing.Hash)
+	}
+	c.cache[offset] = hash
+}
+
+// BuildOnce builds the complete offset->hash map lazily using the provided builder function.
+// The builder is only called once, even if called concurrently from multiple goroutines.
+// Returns any error from the builder function.
+func (c *offsetHashCache) BuildOnce(builder func() (map[int64]plumbing.Hash, error)) error {
```
pjbgf, line 46: Let's remove this concept.

The core logic should always be based off a `.rev` file. If one doesn't exist we could [create one](https://git-scm.com/docs/git-config#Documentation/git-config.txt-packwriteReverseIndex) or if we are [not meant to read the .rev files](https://git-scm.com/docs/git-config#Documentation/git-config.txt-packreadReverseIndex), we could just build one and keep it in memory.

That heuristics is likely part of storage, while `plumbing/format` gets simplified.

Bluebugs: Sounds good. Will adjust the pr accordingly.

Bluebugs:
Not to sure yet, what the best approach might be, will have to think about it. Maybe first introducing readerat for revindex first. We'll have a look tomorrow on what i can do.

Issue #1685, opened 2025-10-08
Fetch error closing writer: EOF


piedshag:
### Bug Description

About 20% of the time I'm getting an error `closing writer: EOF`  when trying to fetch a repo. I have only been able to reproduce this issue with a specific repo `git@github.com:testing-staging/demo-us-west-2.git`.

If you run the script below a few times you should eventually get `panic: closing writer: EOF` from the fetch function.

### go-git Version

v6

### Steps to Reproduce

```go
package main

import (
	"fmt"

	"github.com/go-git/go-git/v6"
	"github.com/go-git/go-git/v6/config"
	"github.com/go-git/go-git/v6/plumbing"
	"github.com/go-git/go-git/v6/storage/memory"
)

func main() {
	url := "git@github.com:testing-staging/demo-us-west-2.git"
	refName := plumbing.NewBranchReferenceName("main")

	repo, err := git.Init(memory.NewStorage())
	if err != nil {
		panic(err)
	}

	if _, err := repo.CreateRemote(&config.RemoteConfig{Name: "origin", URLs: []string{url}}); err != nil {
		panic(err)
	}

	fetchOpts := &git.FetchOptions{
		RefSpecs: []config.RefSpec{config.RefSpec("+" + refName.String() + ":" + refName.String())},
		Depth:    1,
		Tags:     git.NoTags,
	}

	if err := repo.Fetch(fetchOpts); err != nil {
		panic(err)
	}

	r, err := repo.Reference(refName, true)
	if err != nil {
		panic(err)
	}

	fmt.Println(r.Hash().String())
}
```

pjbgf:
cc: @aymanbagabas

aymanbagabas:
Thank you @piedshag!
I tried reproducing but with no luck. However, when I ran the example with `go run -race ...` the race detector immediately reported a race. Applying https://github.com/go-git/go-git/pull/1635 fixed that for me. Could you try and reproduce this with #1635?

piedshag:
Thanks @aymanbagabas for having a look at this. I tried updating to the latest version but it looks like something has changed with how URLs are parsed? I'm now getting this error with the example above

```
panic: dial tcp: address github.com: missing port in address
```

aymanbagabas:
@piedshag Thank you for trying to test this one. What protocol are you using? Is it SSH?
I think I know where the issue at. I will send a fix tomorrow 🙂

piedshag:
[...]

PR #1815, opened 2026-01-09
plumbing: transport, Ignore io.EOF to avoid intermittent errors

suzuki-safie:
## Summary
Fix `closing writer: EOF` error that occurs during fetch operations over SSH when the server closes the connection before the client closes its writer.

This fixes https://github.com/go-git/go-git/issues/1685

In [NegotiatePack](https://github.com/go-git/go-git/blob/29ae690a9f193c642f1e546c6342bd2b07226a52/plumbing/transport/negotiate.go#L219), function may receive `io.EOF` during closing procedure, this may lead to occasional error.

This PR fixes this by ignoring `io.EOF`.

### DIFF
--- a/plumbing/transport/negotiate.go
+++ b/plumbing/transport/negotiate.go
@@ -216,7 +216,7 @@ func NegotiatePack(
 	}
 
 	if !conn.StatelessRPC() {
-		if err := writer.Close(); err != nil && err != io.EOF {
+		if err := writer.Close(); err != nil && !errors.Is(err, io.EOF) {
 			return nil, fmt.Errorf("closing writer: %w", err)
 		}
 	}


suzuki-safie:
commit comment fixed and rebased

nobe4:
FWIW, I am facing the same issue as described, and this change fixes it. 🎉

plumbing/transport/negotiate.go
```diff
@@ -216,7 +216,7 @@ func NegotiatePack(
 	}
 
 	if !conn.StatelessRPC() {
-		if err := writer.Close(); err != nil {
+		if err := writer.Close(); err != nil && err != io.EOF {
```
nobe4: This might be preferable to check any potentially wrapped error:

```suggestion
		if err := writer.Close(); err != nil && ! errors.Is(err, io.EOF) {
```

protomorphine:
facing same problem when trying pull repo

plumbing/transport/negotiate.go
aymanbagabas: Yes, let's use errors.Is instead

suzuki-safie: changed to `errors.Is`

nobe4 (approved):
🚢

Issue #827, opened 2023-08-17
`go-git` sometimes returns `object not found` when querying for the commit object for the `HEAD`


akshit-deepsource:
The following snippet of code returns `object not found` in some cases that I can not reproduce locally. It happens in some of our deployments.

```go
	// Get the repository head commit.
	head, err := repository.Head()
	if err != nil {
		return "", false, err
	}

	headCommit, err := repository.CommitObject(head.Hash())
	if err != nil { // <- returns object not found sometimes
		return "", false, err
	}
```

From my understanding of `git`, this shouldn't be the case. Also, this is not reproducible, and happens randomly. Any pointers to what the issue may be?

pjbgf:
@akshit-deepsource Would you be able to share all the `go-git` related code you are using (from the clone op)? Is there any parallelism on your workflow or it is just cloning one (or more) repository(ies) sequentially?

akshit-deepsource:
@pjbgf sure. We are not cloning using go-git.

Code to open the repository:
```go
	// Get the current working directory.
	cwd, err := os.Getwd()
	if err != nil {
		return nil, false, errors.Wrap(err, "unable to get the cwd")
	}

	// Open the repository.
	repo, err := git.PlainOpen(cwd)
	if err != nil {
		return nil, errors.Is(err, gogit.ErrRepositoryNotExists), errors.Wrap(err, "unable to open the repo")
	}

	return repo, false, nil
```

Fetching the HEAD commit:
```go
	// Get the repository head commit.
	head, err := repository.Head()
	if err != nil {
		return "", false, errors.Wrap(err, "unable to fetch the merge-base")
	}

	headCommit, err := repository.CommitObject(head.Hash())
	if err != nil {
		return "", false, errors.Wrap(err, "unable to fetch the merge-base")
	}
```

pjbgf:
@akshit-deepsource Thanks for the above. Have you identified any pattern that may correlate to the issue? Can you further describe the clone operation you are doing? Is this a shallow clone? Are the repositories large? Anything you could share may help trying to pin-point this.

semmet95:
Hi @akshit-deepsource 
we have run into the same issue. We can get the git commit object just fine when running locally but keep getting this error consistently on our pipelines.
Did you find any fix for this?

pjbgf:
[...]

PR #1825, opened 2026-01-26
Add support for cloning SHA256 repositories

pjbgf:
This PR progresses the [SHA256 support](#706) journey by:
- Enabling the cloning of SHA1 and SHA256 repositories.
- Setting the `object-format` capability from the server side when it is supported for both SHA1 and SHA256.
- Adding new `x/storage.ObjectFormatGetter` and `x/storage.ObjectFormatSetter` interfaces and implementing them on in-tree storage options. External storage will continue to work, however won't be able to support SHA256 until they implement the new interfaces. 
- `go-git` will refuse to Clone or Open non SHA1 repositories when the underlying storage does not implement the interfaces above. This is a fail-safe mechanism to avoid data corruption when using off-tree `storage.Storer` implementations. They will continue to work as expected for SHA1 repositories.

Running the `clone` example against a sha256 repository:
```
$ go run _examples/clone/main.go https://gitlab.com/pjbgf/sha256.git /tmp/test4
git clone https://gitlab.com/pjbgf/sha256.git /tmp/test4 --recursive
commit b2755b63e92d79410f28a33bef90107789c63a139121a457920c668d5805e974
Author: Paulo Gomes <pjbgf@linux.com>
Date:   Thu Sep 25 22:44:22 2025 +0100

    Add dirs

    Signed-off-by: Paulo Gomes <pjbgf@linux.com>
```

During the implementation I stumbled onto other issues which I fixed as part of this PR as it was either blocking the way, or the change was small enough to make sense to do so:
- `format/idxfile` lacked defensive programming to remove (or error) when objects were not fully materialised, which could lead to invalid index files. This resulted into `object not found` errors, which may be related to existing reports of such symptom still happening on v6 (e.g. #207 #827)
- `format/idxfile` was growing the backing arrays for `names`, `offset` and `crc32` as it iterated through objects. This is rather inefficient and has performance impact when processing very large pack files.
- Expanded on test coverage for `filesystem` and `memory` storage, to avoid future regression when handling `ObjectFormat`.

Relates to #706.

:information_source: Some of this work was done as part of [openSUSE's Hack Week 25](https://hackweek.opensuse.org/projects/go-git-unlocking-sha256-based-repository-cloning-ahead-of-git-v3). Thanks SUSE for sponsoring the `go-git` project. :raised_hands:

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -32,6 +32,7 @@ import (
 	"github.com/go-git/go-git/v6/storage/filesystem/dotgit"
 	"github.com/go-git/go-git/v6/utils/ioutil"
 	"github.com/go-git/go-git/v6/utils/trace"
+	xstorage "github.com/go-git/go-git/v6/x/storage"
 )
 
 // GitDirName this is a special folder where all the git stuff is.
@@ -261,6 +262,17 @@ func Open(s storage.Storer, worktree billy.Filesystem) (*Repository, error) {
 
 	if err != nil {
 		return nil, err
+	}
+
+	// Fail-safe to avoid data corruption when off-tree storage implementations
+	// are used to open SHA256 repositories which they may not support.
+	if cfg, err := s.Config(); err == nil {
+		if cfg.Extensions.ObjectFormat != formatcfg.UnsetObjectFormat &&
+			cfg.Extensions.ObjectFormat != formatcfg.SHA1 {
+			if _, ok := s.(xstorage.ObjectFormatGetter); !ok {
+				return nil, errors.New("storage do not support SHA256 object format")
+			}
+		}
 	}
 
 	return newRepository(s, worktree), nil
--- a/repository_test.go
+++ b/repository_test.go
@@ -41,6 +41,7 @@ import (
 	"github.com/go-git/go-git/v6/storage"
 	"github.com/go-git/go-git/v6/storage/filesystem"
 	"github.com/go-git/go-git/v6/storage/memory"
+	xstorage "github.com/go-git/go-git/v6/x/storage"
 )
 
 func TestInit(t *testing.T) {
@@ -447,6 +448,64 @@ func TestFetchMustNotUpdateObjectFormat(t *testing.T) {
 			}
 		})
 	}
+}
+
+// sha1OnlyStorage wraps a storage.Storer to hide any ObjectFormatGetter
+// or ObjectFormatSetter implementations, simulating a storage backend
+// that only supports SHA1.
+type sha1OnlyStorage struct {
+	storage.Storer
+}
+
+func TestFailSafeUnsupportedStorage(t *testing.T) {
+	t.Parallel()
+
+	t.Run("clone", func(t *testing.T) {
+		t.Parallel()
+
+		f := fixtures.ByTag(".git-sha256").One()
+		require.NotNil(t, f, "fixture not found for tag .git-sha256")
+
+		for _, srv := range server.All(server.Loader(t, f)) {
+			endpoint, err := srv.Start()
+			require.NoError(t, err)
+
+			t.Cleanup(func() {
+				require.NoError(t, srv.Close())
+			})
+
+			st := &sha1OnlyStorage{memory.NewStorage()}
+			_, okGetter := storage.Storer(st).(xstorage.ObjectFormatGetter)
+			assert.False(t, okGetter, "sha1OnlyStorage must not implement ObjectFormatGetter")
+
+			_, okSetter := storage.Storer(st).(xstorage.ObjectFormatSetter)
+			assert.False(t, okSetter, "sha1OnlyStorage must not implement ObjectFormatSetter")
+
+			_, err = Clone(st, nil, &CloneOptions{URL: endpoint})
+			require.Error(t, err)
+			assert.Contains(t, err.Error(), "mismatched algorithms")
+		}
+	})
+
+	t.Run("open", func(t *testing.T) {
+		t.Parallel()
+
+		f := fixtures.ByTag(".git-sha256").One()
+		require.NotNil(t, f, "fixture not found for tag .git-sha256")
+
+		st := filesystem.NewStorage(f.DotGit(fixtures.WithMemFS()), cache.NewObjectLRUDefault())
+
+		wrapped := &sha1OnlyStorage{st}
+		_, okGetter := storage.Storer(wrapped).(xstorage.ObjectFormatGetter)
+		assert.False(t, okGetter, "sha1OnlyStorage must not implement ObjectFormatGetter")
+
+		_, okSetter := storage.Storer(wrapped).(xstorage.ObjectFormatSetter)
+		assert.False(t, okSetter, "sha1OnlyStorage must not implement ObjectFormatSetter")
+
+		r, err := Open(wrapped, nil)
+		assert.Error(t, err)
+		assert.Nil(t, r)
+	})
 }
 
 func (s *RepositorySuite) TestCloneContext() {


storage/storer.go
```diff
@@ -26,6 +27,22 @@ type Storer interface {
...
+
+// ObjectFormatSetter is implemented by storage backends that support
+// configuring the object format (hash algorithm) used for the repository.
+// This is typically called during repository initialization or when the
+// first packfile is fetched.
+type ObjectFormatSetter interface {
+	// SetObjectFormat configures the object format (hash algorithm) for this storage.
+	SetObjectFormat(formatcfg.ObjectFormat) error
+}
+
```
aymanbagabas: Why do we want a custom interface for this? This makes it look like a repository object format is a special object in Git. Instead, I would reuse the `config.ConfigStorer` interface to get/set the object format and come up with a `storage.ReadObjectFormat(config.ConfigStorer)` and `storage.SetObjectFormat(config.ConfigStorer, formatcfg.ObjectFormat)` or something similar.

plumbing/format/packfile/common.go
```diff
@@ -24,17 +26,24 @@ const (
...
 	defer func() {
 		trace.Performance.Printf("performance: %.9f s: update_obj_storage", time.Since(start).Seconds())
 	}()
 
+	if setter, ok := s.(storage.ObjectFormatSetter); ok {
+		err := setter.SetObjectFormat(of)
+		if err != nil {
+			return err
+		}
+	}
```
aymanbagabas: What happens if the repository has a different object format from the incoming packfile? I think in that case this should fail with an error. I think we should always use the storage object format and pass that down to the parser

plumbing/format/packfile/parser_options.go
```diff
@@ -27,6 +28,12 @@ func WithScannerObservers(ob ...Observer) ParserOption {
 	}
 }
 
+func WithObjectFormat(of config.ObjectFormat) ParserOption {
```
aymanbagabas, line 31: Shouldn't we infer the object format from the repository storage?

plumbing/transport/fetch.go
```diff
@@ -39,7 +40,12 @@ func FetchPack(
 		reader = demuxer
 	}
 
-	if err := packfile.UpdateObjectStorage(st, reader); err != nil {
+	of := config.SHA1
+	if req.Wants[0].Size() == config.SHA256.Size() {
```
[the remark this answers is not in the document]
pjbgf: Adding a comment about it - but at this point in time we may need to continue to infer it from Wants.

worktree_commit_test.go
```diff
@@ -703,34 +704,35 @@ func (s *WorktreeSuite) TestCommitTreeSort() {
...
-	s.NoError(err)
+	s.Require().NoError(err)
 
-	r, _ := Clone(memory.NewStorage(), memfs.New(), &CloneOptions{
+	r, err := Clone(memory.NewStorage(), memfs.New(), &CloneOptions{
 		URL: fs.Root(),
 	})
+	s.ErrorIs(err, transport.ErrEmptyRemoteRepository)
 
 	w, err := r.Worktree()
```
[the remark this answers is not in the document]
pjbgf: This is expected - the test needs (and confirms) the repository is empty - however it isn't `nil`.

repository.go
```diff
@@ -69,8 +69,6 @@ var (
 	ErrUnableToResolveCommit = errors.New("unable to resolve commit")
 	// ErrPackedObjectsNotSupported is returned when packed objects are not supported.
 	ErrPackedObjectsNotSupported = errors.New("packed objects not supported")
-	// ErrSHA256NotSupported is returned when go-git was not compiled with SHA256 support.
-	ErrSHA256NotSupported = errors.New("go-git was not compiled with SHA256 support")
 	// ErrAlternatePathNotSupported is returned when the alternate path is not a file scheme.
```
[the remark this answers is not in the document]
pjbgf: This will be released on a new major version.

storage/storer.go
pjbgf: This was a compromise, the reasoning being:

1. When we [init](https://github.com/go-git/go-git/blob/f99a98e81ce9a413f8552bb9271a05409f6ebfdf/repository.go#L285) a new repository we don't know what `ObjectFormat` it will have. That information is acquired talking with the server and then the new format is set using `UpdateObjectStorage`. Alternatively, we could hold the repository initialisation to after the fetching process - but that would be a larger change.
2. A key motivation here was backwards compatibility with off-tree storage implementations. This approach means they continue to compile and work when interacting with SHA1 repositories. But safely fail when interacting with a SHA256 repo.

plumbing/format/packfile/common.go
pjbgf: This PR makes both [memory](https://github.com/pjbgf/go-git/blob/75a0b7b5ae9b0d8e32d87ecd31eaeea8eb89d0ca/storage/memory/storage.go#L80-L86) and [filesystem](https://github.com/pjbgf/go-git/blob/75a0b7b5ae9b0d8e32d87ecd31eaeea8eb89d0ca/storage/filesystem/storage.go#L125-L128) implementations to fail if they are not empty when a change of object format is requested. 

This is a fail-safe mechanism to enable single object format repositories to work, while erroring when a repo is trying to handle multiple formats. This should be reviewed as part of the [translation table](https://git-scm.com/docs/hash-function-transition#_translation_table) implementation. I'm copying this thread to the issue so that this discussion can be discoverable.

plumbing/format/packfile/parser_options.go
pjbgf: Overall, I don't think we can:
1. A new repository likely won't have the ObjectFormat set. Even if it does, it may not align with the target pack being fetched. For example, new repo (defaults to SHA1) then fetches a SHA256 packfile.
2. An existing repository set with a different `ObjectFormat` than the one for the packfile being downloaded.

These changes would set the repository objectformat to the target packfile format, provided this is a new/empty repository. This feels like a safe approach until we have multi `ObjectFormat` support. What do you think?

plumbing/format/revfile/encoder.go
```diff
@@ -32,11 +33,19 @@ func Encode(w io.Writer, h hash.Hash, idx *idxfile.MemoryIndex) error {
 	if w == nil {
 		return fmt.Errorf("nil writer")
 	}
+	v := reflect.ValueOf(w)
+	switch v.Kind() {
+	case reflect.Ptr, reflect.Interface:
+		if v.IsNil() {
+			return fmt.Errorf("nil writer")
+		}
+	}
```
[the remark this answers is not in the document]
pjbgf: This is not true, removing the validation causes the existing tests to panic.

storage/storer.go
aymanbagabas: You need to specify the object format for new repositories, otherwise, it defaults to SHA1. That's how git works by default. Migrating a repository from one format to another requires creating a new repository with the new format, and import/export the repository from the old to the new format via `git fast-import` and `git fast-export`.

I think there should be an option to specify the object format for new repositories. UpdateObjectStorage and others should always use the repository format. The transport layer should validate that new requests match the storage format.

AdamMajer: Empty repos are kind of this funny-place where you'd think the hash function is undefined, but it is actually defined  and stored (implicitly or explicitly) as part of the config. The main reason is what happens when you have an empty repository, and then you clone it?

This is a valid operation that must work predictably. An empty repository is initialized by git forges like Gitea, then you clone it, add commits and push them. Waiting until you create the first objects to define the repository ObjectFormat does not work.

The data sent when you transfer a repo includes this type, even if there are no object, to make sure that the ObjectFormat is unambiguous.

pjbgf: Thank you both for the feedback in such a timely manner. This approach was taken to lean on existing primitives while Cloning new repositories as at that point in time you don't know what `ObjectFormat` your server has. As per upstream:

> We have a chicken-and-egg situation between initializing the refdb
	 * and spawning transport helpers:
	 *
	 *   - Initializing the refdb requires us to know about the object
	 *     format. We thus have to spawn the transport helper to learn
	 *     about it.
	 *
	 *   - The transport helper may want to access the Git repository. But
	 *     because the refdb has not been initialized, we don't have "HEAD"
	 *     or "refs/". Thus, the helper cannot find the Git repository.

Upstream [worked around this](https://github.com/git/git/blob/ab380cb80b0727f7f2d7f6b17592ae6783e9820c/builtin/clone.c#L1192-L1223) by introducing a lazy refdb init, which instead of initialising a valid refdb it actually creates an invalid refdb - only finishing off the init process once the target objectformat is agreed with the remote. Let me take that approach into consideration and update this PR accordingly.

plumbing/format/packfile/parser_options.go
pjbgf: For reference, in the filesystem storage, the Parser will be [called here](https://github.com/pjbgf/go-git/blob/42687fdf86ceeec0773685fde57e22aa7f6aee95/storage/filesystem/dotgit/writers.go#L74) with the ObjectFormat coming from either the repository config, for existing underlying repo, or from the storage options otherwise.

storage/storer.go
pjbgf: Following-up from this discussion, I made a few changes to better align with upstream:
- Added the partial init and moved the setting of object format to the pack negotiation logic. That only takes place on the first negotiation during a clone operation, subsequent negotiations won't trigger a object format change in the storage.
- Tests were added to ensure that object format can't be changed on initialised repositories - regardless of it being set explicitly or not.
- The go-git Storage abstraction for filesystem ignores any `ObjectFormat` passed-in to `NewStorageWithOptions` when the underlying fs contains an initialised repository. In such cases, the repository config is taken a single source of truth.

Are there any additional scenarios we may need to cater for on the test cases?

plumbing/storer/storer.go
```diff
@@ -21,3 +25,13 @@ type Initializer interface {
...
+
+// ObjectFormatGetter expands a Storer so that it can support different Object Formats.
+// Note that storage.Storer do not require this as they expose ConfigStorers, which is
+// the source of truth for this information.
+//
+// Storers that don't implement this interface will default to the default Object Format.
+type ObjectFormatGetter interface {
+	// ObjectFormat returns the object format (hash algorithm) used by the Storer.
+	ObjectFormat() formatcfg.ObjectFormat
+}
```
aymanbagabas: Do we need this interface still? If we're storing the object format in config and config is the single source of truth, this seems redundant. On this note, I always found the `storer.Storer` and `storage.Storer` interfaces confusing, and we should deprecate one of them.

plumbing/transport/negotiate.go
```diff
@@ -60,6 +61,54 @@ func NegotiatePack(
...
+			// to avoid code duplication and to better handle off-tree storage implementations.
+			if err == nil && ref.Target().String() == "refs/heads/.invalid" {
+				if setter, ok := st.(storage.ObjectFormatSetter); ok {
+					err := setter.SetObjectFormat(serverFormat)
+					if err != nil {
+						return nil, fmt.Errorf("unable to set object format: %w", err)
+					}
+
+					clientFormat = serverFormat
+				}
```
aymanbagabas:
```suggestion
				clientFormat = serverFormat
				cfg.Extensions.ObjectFormat = serverFormat
				st.SetConfig(cfg)
```

storage/storer.go
```diff
@@ -26,6 +27,13 @@ type Storer interface {
 	ModuleStorer
 }
 
+// ObjectFormatSetter is implemented by storage backends that support
+// configuring the object format (hash algorithm) used for the repository.
+type ObjectFormatSetter interface {
+	// SetObjectFormat configures the object format (hash algorithm) for this storage.
+	SetObjectFormat(formatcfg.ObjectFormat) error
+}
+
```
aymanbagabas:
```suggestion
// ObjectFormat finds the object format for the given [config.ConfigStorer].
// If the config storer doesn't define an object format, it will return
// [config.UnsetObjectFormat].
func ObjectFormat(st config.ConfigStorer) config.ObjectFormat {
		cfg, err := st.Config()
		if err == nil {
			return cfg.Extensions.ObjectFormat
		}
		return config.UnsetObjectFormat
}

// SetObjectFormat sets the object format for the given [config.ConfigStorer].
func SetObjectFormat(st config.ConfigStorer, f config.ObjectFormat) error {
		cfg, err := st.Config()
		if err != nil {
			return err
		}
		cfg.Extensions.ObjectFormat = f
		return st.SetConfig(cfg)
}
```

aymanbagabas (commented):
Looking great! I wonder if we should keep the `ObjectFormatGetter` 🤔

plumbing/storer/storer.go
pjbgf: Long-term we may not need this interface - I will add some comments to that effect. I also agree that `storer.Storer` and `storage.Storer` could be reviewed, so I created an [issue](https://github.com/go-git/go-git/issues/1832) to handle both subjects at once - tracking to the v7 milestone.

For now, I rather we keep `ObjectFormatGetter` in so that we don't break the interface with existing external implementations, but also have a safe to detect lack of support from SHA256 on off-tree storage implementations. As lack of this awareness could result in data corruption - I will add a test to highlight this point.

plumbing/transport/negotiate.go
pjbgf: We can't just change the config, as the storage is already initialised at this point, so we need to reinitilise the (object) hashers with the target format.

If we were to add this additional logic onto `SetConfig`, that would mean that we'd need to verify on each call whether it is OK to change the object format and whether that should trigger a reinitilisation of the (object) hashers. As this would be `storage.Storer` specific, this could lead to a corrupted state on off-tree implementations - as the core logic would be unaware of such support.

aymanbagabas: In this, can why don't we use the [`Initializer`](https://github.com/go-git/go-git/blob/b5117ad1603d4a44b05d7b4d661b8e1e9d035518/plumbing/storer/storer.go#L13) interface to do that? We change the config first, then call the Init function to reinitialize the object hashers. I don't see that interface is being used by existing storage backends and I think this case would be a good use for that.

PR #1828, opened 2026-02-01
GPG/SSH signature improvement

Bluebugs:
This is a draft PR to gather feedback from the community on improving support for signature in go-git. I am interested mostly on ed25519 ssh key used to sign and verify git commit. That is what I focused on. The documentation that I found the most useful on the topic was this blog: https://calebhearth.com/sign-git-with-ssh .

This PR is not ready for inclusion has since I started working on it, I have learned more about the project expectation and I need to go back, split it and clean it to match better the rest of upstream code base. The intent at this point is to see if it would be useful to more people than just me. I hope it could address #400 .

### DIFF
--- a/_examples/verify-tag-signature/main.go
+++ b/_examples/verify-tag-signature/main.go
@@ -1,0 +1,136 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"sort"
+
+	"github.com/go-git/go-git/v6"
+	"github.com/go-git/go-git/v6/config"
+	. "github.com/go-git/go-git/v6/_examples"
+	"github.com/go-git/go-git/v6/plumbing/object"
+)
+
+// Example of how to verify SSH signatures on tags.
+// This demonstrates loading an allowed_signers file and verifying
+// signatures on annotated tags in a repository.
+//
+// Usage:
+//
+//	verify-tag-signature <repository-path> [allowed-signers-file]
+//
+// If allowed-signers-file is not provided, it will be loaded from
+// the git config (gpg.ssh.allowedSignersFile).
+func main() {
+	CheckArgs("<repository-path>")
+	repoPath := os.Args[1]
+
+	// Open the repository
+	Info("Opening repository at %s", repoPath)
+	r, err := git.PlainOpen(repoPath)
+	CheckIfError(err)
+
+	// Load the SSH verifier
+	var verifier *git.SSHVerifier
+
+	if len(os.Args) > 2 {
+		// Use provided allowed_signers file
+		allowedSignersPath := os.Args[2]
+		Info("Loading allowed signers from %s", allowedSignersPath)
+		verifier, err = git.NewSSHVerifierFromFile(allowedSignersPath)
+		CheckIfError(err)
+	} else {
+		// Load from git config (local + global)
+		Info("Loading allowed signers from git config")
+		cfg, err := r.ConfigScoped(config.GlobalScope)
+		CheckIfError(err)
+
+		verifier, err = git.NewSSHVerifierFromConfig(cfg)
+		CheckIfError(err)
+
+		if verifier == nil {
+			fmt.Println("No allowed signers file configured in git config.")
+			fmt.Println("Set gpg.ssh.allowedSignersFile or provide path as argument.")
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -106,6 +106,17 @@ type Config struct {
 		Email string
 	}
 
+	GPG struct {
+		// Format specifies the signature format to use when signing commits and tags.
+		// Valid values are "openpgp" (default) and "ssh".
```
pjbgf, line 111:
```suggestion
		// Valid values are "openpgp" (default), "x509" and "ssh".
```

bb-Ricardo:
I had a look at This PR and https://github.com/go-git/go-git/pull/1860.

@hiddeco already implemented signing and verification for SSH signatures https://github.com/hiddeco/sshsig

Maybe you could add the allowed signer parsing as PR to the project https://github.com/hiddeco/sshsig/pull/7/changes

Then we could use this in plugin as dependency.

plumbing/object/verification.go
```diff
@@ -0,0 +1,113 @@
...
+
+// ErrNoSignature is returned when attempting to verify an unsigned object.
+var ErrNoSignature = errors.New("object has no signature")
+
+// ErrNilVerifier is returned when a nil verifier is passed to a verification method.
+var ErrNilVerifier = errors.New("verifier is nil")
+
+// TrustLevel represents the trust level of a signing key.
+// The levels follow Git's trust model, from lowest to highest.
+type TrustLevel int8
```
pjbgf, line 17: I'm under the impression this aligns with GPG verification, but not SSH. I'd assume the same would apply for potentially other signature methods - e.g. `X.509`, but that I'm not sure.

Ideally the basic verifier API would be agnostic and still be able to provide additional information if needed. I haven't thought about this part of the API yet, but it could be done via typed errors.

bb-Ricardo: Yes, this also crossed my mind that you need different abstractions. Two which were already made are "verified" and "trusted".

Verified: public matches signature
Trusted: additional key properties (owner, trust level of PGP üublic key or virified_signers for SSH) match the committe/author.

Additionally S/MIME again can be done similar. Here we could/need to check if the cert issue timeframe matches the commit timestamp and the CN or eMails in the SAN. Or is signed by a certain CA.

Not sure how we should abstract these layers. A simple verified for SSH and GPG is not that difficult but trust (level) has more attributes to consider.

pjbgf: +1. On the trust levels, I'm not sure we need to go that far. It could be an option (e.g. `WithAcceptGPGPartialTrust` for the specific verifier), but from an API perspective that would just mean we return an error or not - ideally the default failing on the safe side.

cedric-appdirect: I have been wondering about this part too. On github you have also 3 states see https://docs.github.com/en/authentication/managing-commit-signature-verification/displaying-verification-statuses-for-all-of-your-commits , but I am not so much of a fan. I kind of like the idea of:
- Verified: there is a valid signed commit
- Trusted: the signature was valid according to some rules/configuration/... at the time of the commit
- Unverified for commit that either don't have a signature or we could not verify it.

Still, not great as it bake a specific semantic of the verifier into the API. The idea of just relying on error is maybe a better approach. No error would be fully trusted anything else can be drilled into using errors.Is and would allow semantic extensibility for the verifier plugin. Will try to put that in my proposal PR later this week.

pjbgf: Thanks for sharing the GH docs. Overall, that makes absolute sense. I still believe that Trusted is a valid state, and the handling of it would be verifier specific and opt-in/out. For example:

```
verifier := github.NewVerifier(github.WithTrustedAsVerified())
ok, err := verifier.Verify(commit) // commit has Trusted
// ok = true
// err = nil
```

Without the `WithTrustedAsVerified` option:
```
// ok = false
// err = trusted signature: committer mismatch
```

Similar would apply to a `gpg` verifier.

cedric-appdirect: Oh, didn't think about making it an option on the verifier itself. I see. It make sense. Will have another pass at this.

Issue #1822, opened 2026-01-23
RemoveReference should clean up empty directories


ryanapilado:
### Bug Description

When SetReference is called to create a reference with `/` in the name, it creates subdirectories for each part of the name separated by a slash. For example, when creating the ref `bugfix/issue-1` SetReference would create it at the path `<repo>/refs/bugfix/issue-1`, creating the file `issue-1` and the parent directory `bugfix/`.

When RemoveReference is called on the same reference, it deletes the reference but not the parent directories, even though they may be empty. In the example above, it would delete `issue-1`, but the `refs/heads/bugfix/` directory would still exist.

Since the parent directories are not cleaned up, it becomes impossible to create any new references that share the same name. In the example above, it is now not possible to use SetReference to create a ref with the name `bugfix`, since it conflicts with the leftover directory `refs/heads/bugfix/`.

The Git implementation does cleanup parent directories when a branch is deleted.

### go-git Version

v5.16.4

### Steps to Reproduce

```
func TestConflictWithDeletedRef(t *testing.T) {
	repoPath := t.TempDir()
	fs := filesystem.NewStorage(osfs.New(repoPath), cache.NewObjectLRUDefault())
	if _, err := git.Init(fs, nil); err != nil {
		t.Fatalf("Failed to init git: %v", err)
	}

	// Create a new branch named bugfix/1.
	nestedRef := plumbing.NewReferenceFromStrings("refs/heads/bugfix/1", "target1")
	if err := fs.SetReference(nestedRef); err != nil {
		t.Fatalf("Failed to set reference: %v", err)
	}

	if err := fs.RemoveReference(nestedRef.Name()); err != nil {
		t.Fatalf("Failed to remove reference: %v", err)
	}

	// Now we try to create a ref named 'bugfix' but it fails, since the directory refs/heads/bugfix/ exists.
	// Error:
	// Failed to set reference: open /tmp/TestConflictWithDeletedRef4236784409/001/refs/heads/bugfix: is a directory
	conflictingRef := plumbing.NewReferenceFromStrings("refs/heads/bugfix", "target2")
	if err := fs.SetReference(conflictingRef); err != nil {
		t.Fatalf("Failed to set reference: %v", err)
	}
}
```

pjbgf:
@ryanapilado thanks for reporting this issue.

If anyone from the community wants to pick this up, contributions are very welcome - tag me for a review.

BootstrapperSBL:
[...]

PR #1829, opened 2026-02-01
Fixes: #1822

OJuisPlease:
I fix the bug #1822 through adding a new function that removes empty directories recursively. And it won't remove  all subdirectories of `refs` like `refs/heads`. But if there is an existed empty directory, there will be errors. But I am not clever enough to figure out a better solution. 
I am a new coder and poor in English so if there are comments, please email me and point it out directly.
I rewrite the example code in the issue as a unit test case and it passes the test.
@pjbgf Thanks for reviewing.


storage/filesystem/dotgit/dotgit_test.go
```diff
@@ -1085,3 +1085,27 @@ func (s *SuiteDotGit) TestSetPackedRef() {
...
+		"refs/heads/bugfix/1",
+		"e8d3ffab552895c19b9fcf7aa264d277cde33881",
+	)
+	err := dir.SetRef(oldRef, nil)
+	s.Require().NoError(err)
+
+	err = dir.RemoveRef(oldRef.Name())
+	s.Require().NoError(err)
+
+	// if the function works well, the newRef will be created successfully instead of conflicts
```
pjbgf:
```suggestion
```

storage/filesystem/dotgit/dotgit_test.go
```diff
@@ -1085,3 +1085,27 @@ func (s *SuiteDotGit) TestSetPackedRef() {
...
+	newRef := plumbing.NewReferenceFromStrings(
+		"refs/heads/bugfix",
+		"c8d3ffab552895c19b9fcf7aa264d277cde33881",
+	)
+	err = dir.SetRef(newRef,nil)
+	s.Require().NoError(err)
+}
```
pjbgf: Missing line break after `}`:
```suggestion
}

```

storage/filesystem/dotgit/dotgit_test.go
```diff
@@ -1085,3 +1085,27 @@ func (s *SuiteDotGit) TestSetPackedRef() {
 	s.Require().NoError(err)
 	s.Equal(1, looseCount)
 }
+
+func (s *SuiteDotGit) TestPruneEmptyDirectoriesInRefs() {
```
pjbgf, line 1089: The aim of this test is to verify that empty directories are pruned, however it does not assert this. Please add a check that `refs/heads/bugfix` is removed from `fs`.

storage/filesystem/dotgit/dotgit.go
```diff
@@ -1053,6 +1056,40 @@ func (d *DotGit) readReferenceFile(path, name string) (ref *plumbing.Reference,
...
+	if err != nil {
+		return err
+	}
+
+	if len(files) == 0 {
+		err := d.fs.Remove(path)
+		if err != nil {
+			return err
+		}
+		d.pruneEmptyDirectoriesInRefs(filepath.Dir(path))
```
pjbgf, line 1091: We are better off using [util.Walk](https://github.com/go-git/go-billy/blob/main/util/walk.go#L60) to walk through the descending dirs.

storage/filesystem/dotgit/dotgit.go
```diff
@@ -1053,6 +1056,40 @@ func (d *DotGit) readReferenceFile(path, name string) (ref *plumbing.Reference,
...
+// then ascends the directory tree to remove newly emptied parent directories,
+// stopping immediately before any directory whose parent is exactly "refs".
+// This preserves top-level refs subdirectories (e.g., refs/heads, refs/tags)
+// even if they become empty, 
+func (d *DotGit) pruneEmptyDirectoriesInRefs(path string) error {
+	if filepath.Dir(path) == "refs" {
+		return nil
+	}
+	
+	info, err := d.fs.Stat(path)
```
pjbgf: We should `Lstat` instead - as per Walk.

storage/filesystem/dotgit/dotgit.go
```diff
@@ -812,13 +812,16 @@ func (d *DotGit) RemoveRef(name plumbing.ReferenceName) error {
...
 		err = d.fs.Remove(path)
+		if err != nil && !os.IsNotExist(err) {
+			return err
+		}
+		err = d.pruneEmptyDirectoriesInRefs(filepath.Dir(path))
+		if err != nil {
+			return err
+		}
 		// Drop down to remove it from the packed refs file, too.
 	}
```
pjbgf, line 828: Not sure we need to remove the `if err != nil` we had here. Are we not going to conceal this error? Please share your thoughts on this change.

pjbgf (changes requested):
hey @OJuisPlease, thanks for proposing this fix. Here's some feedback:

storage/filesystem/dotgit/dotgit.go
OJuisPlease: Actually, I fail to take  the situation where concurrent changes happen into consideration. So I think a hidden File Not Exist error would be proper.

storage/filesystem/dotgit/dotgit.go
OJuisPlease: Sorry but I can't think out a way to use Walk to walk the dirs from leaves to the root (That's how my solution works).

OJuisPlease: Please take a look. @pjbgf

PR #1847, opened 2026-02-13
git: Clean up API around signing commits and tags

pjbgf:
The support for signatures beyond PGP were introduced as non-breaking enhancements within `v5`. This PR introduces breaking changes to better reflect the existing support.

- Remove `SignKey` from CommitOptions. Users should use the field `Signer` instead, which is supported since `v5`.
- Field `PGPSignature` in both `Tag` and `Commit` structs are now called `Signature`, to reflect the support of signatures beyond PGP.
- [Cherry-pick](https://github.com/go-git/go-git/pull/1171) the removal of support for `BEGIN CERTIFICATE` for x509 signatures. This change aligns with upstream [git](https://github.com/git/git/blob/453e7b744aef0d9dab62dac6ab030bb643a2e117/gpg-interface.c#L66) and [gnupg](https://github.com/gpg/gnupg/blob/c0f9ca47f064a3a401d965f3821e203eef2228c9/sm/sign.c#L651). We will continue to support `BEGIN SIGNED MESSAGE`, as per official [git docs](https://git-scm.com/docs/gitformat-signature#Documentation/gitformat-signature.txt-x509X509).

Follow-up from #690, #996 and #1029.

Relates to #910 #400.

### DIFF
--- a/plumbing/object/commit.go
+++ b/plumbing/object/commit.go
@@ -52,7 +52,7 @@ type Commit struct {
 	// MergeTag is the embedded tag object when a merge commit is created by
 	// merging a signed tag.
 	MergeTag string
-	// Signature is the cryptographic signature of the commit (e.g. SSH, x.509).
+	// Signature is the cryptographic signature of the commit (e.g. SSH, X.509).
 	Signature string
 	// Message is the commit message, contains arbitrary text.
 	Message string
--- a/plumbing/object/signature_test.go
+++ b/plumbing/object/signature_test.go
@@ -34,7 +34,7 @@ MKEQruIQWJb+8HVXwssA4=
 			want: signatureTypeSSH,
 		},
 		{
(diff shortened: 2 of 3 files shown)


hiddeco (approved):
💯

terlar:
After this change, should the `gpgSigner` be public so everyone don't have to copy this code?
https://github.com/go-git/go-git/blob/main/worktree_commit.go#L258

Or is the plan that this kind of logic should be extracted into the openpgp package itself?

Now this code is only used and accessible within the tests for `worktree_commit`.

pjbgf:
@terlar thanks for the ping, I'm working on this and just created an [issue](https://github.com/go-git/go-git/issues/1849) to highlight that.

This PR was just a stepping stone towards a more consistent and user-friendly API around this feature.

PR #1850, opened 2026-02-16
git: Add strict checks for supported extensions

pjbgf:
The upstream Git enforces fail-safe heuristics to ensure that older git versions
will avoid handling repositories using extensions they are unaware of.

The logic is largely based on the value of `core.repositoryformatversion`. As per
official Git docs:
> This version specifies the rules for operating on the on-disk repository data.
> An implementation of git which does not understand a particular version
> advertised by an on-disk repository MUST NOT operate on that repository;
> doing so risks not only producing wrong results, but actually losing data.

Now go-git will ensure that:
- The `git.Open` logic will verify and enforces the extension support rules.
- `go-git` will keep track of built-in extensions that it supports.
- Any extension that is Storer-specific (e.g. `objectformat`, `reftables`),
  will be confirmed by the specific Storer via new `x/storage.ExtensionChecker`
  interface. Failing to do so, it will be assumed that the extension is
  unsupported and the `git.Open` operation will return an error.

This is a breaking change and it will force go-git to not be able to open
repositories that it in fact doesn't really support. Conversaly, the error
messages will be more useful (e.g. unknown extension vs object not found).

Upstream refs:
- https://git-scm.com/docs/git-config#Documentation/git-config.txt-extensions
- https://git-scm.com/docs/gitrepository-layout#_git_repository_format_versions

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -32,7 +32,6 @@ import (
 	"github.com/go-git/go-git/v6/storage/filesystem/dotgit"
 	"github.com/go-git/go-git/v6/utils/ioutil"
 	"github.com/go-git/go-git/v6/utils/trace"
-	xstorage "github.com/go-git/go-git/v6/x/storage"
 )
 
 // GitDirName this is a special folder where all the git stuff is.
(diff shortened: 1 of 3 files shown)


repository.go
```diff
@@ -260,19 +259,14 @@ func Open(s storage.Storer, worktree billy.Filesystem) (*Repository, error) {
...
-	if cfg, err := s.Config(); err == nil {
-		if cfg.Extensions.ObjectFormat != formatcfg.UnsetObjectFormat &&
-			cfg.Extensions.ObjectFormat != formatcfg.SHA1 {
-			if _, ok := s.(xstorage.ObjectFormatGetter); !ok {
-				return nil, errors.New("storage do not support SHA256 object format")
-			}
-		}
+	err = verifyExtensions(s, cfg)
+	if err != nil {
+		return nil, err
```
pjbgf, line 269: @aymanbagabas following-up from the previous discussion on `ObjectFormatGetter`, this PR removes the need for that interface from a data corruption perspective. 

As a separate change, I'll look into dropping the Getter/Setter.

repository_extensions.go
```diff
@@ -0,0 +1,110 @@
...
+			if _, ok := builtinExtensions[ext.name]; ok {
+				continue
+			}
+
+			if sok && storageChecker.SupportsExtension(ext.name, ext.value) {
+				continue
+			}
+
+			missing = append(missing, ext.name)
+		}
```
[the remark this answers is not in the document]
pjbgf: `storage/transactional` is an experimental, and currently does not support this feature - nor sha256 that relies on it.

repository.go
aymanbagabas: Amazing! This looks much better!

aymanbagabas (approved):

Issue #1845, opened 2026-02-12
resetWorktree: O(N²) filepath.Clean calls during Clone cause severe CPU regression in v5.16.5


EladGabay:
### Bug Description

Commit https://github.com/go-git/go-git/commit/600fb139079e3c6886fcfeb20021c707e99e29b4 ("Don't delete local untracked files when resetting worktree") introduced an O(N²) performance regression that affects every Clone operation.

During Clone, go-git internally performs a MergeReset (worktree.go:138). In v5.16.4, resetWorktree was called with a nil file list:
```
  // v5.16.4
  w.resetWorktree(t, opts.Files)  // opts.Files is nil during Clone
```

  In v5.16.5, resetIndex now collects all touched files and passes them to resetWorktree:
```
  // v5.16.5
  removedFiles, _ = w.resetIndex(t, dirs, opts.Files)
  w.resetWorktree(t, removedFiles)
```

  Inside resetWorktree, the inFiles function is called for every worktree change when the file list is non-nil:
```
  func inFiles(files []string, v string) bool {
      v = filepath.Clean(v)
      for _, s := range files {
          if filepath.Clean(s) == v {
              return true
          }
      }
      return false
  }
```

During a fresh Clone, both removedFiles and changes (from diffStagingWithWorktree) contain every file in the repository. This results in N×N calls to filepath.Clean, where every check returns true, the filtering has no practical effect during Clone, but the cost is paid in full.

For a repository with ~50,000 files, this produces ~2.5 billion filepath.Clean calls.
This affects every user of git.Clone / git.PlainClone - the MergeReset is hardcoded in the checkout path (worktree.go:138), not opt-in.

### go-git Version

v5.16.5

EladGabay:
@pjbgf @k-anshul @UnseenBook

k-anshul:
Hey @EladGabay @pjbgf 

I pushed an optimisation in https://github.com/go-git/go-git/pull/1852 that converts this to a o(n) implementation. In my testing doing `file.Clean` for 100K files completes in under 200ms.

EladGabay:
Thanks @k-anshul !
One minor suggestion: map[string]struct{} is more idiomatic than map[string]any for set semantics, avoids allocating an interface header per entry

pjbgf:
@EladGabay thanks for filing the issue and @k-anshul for the proposed PR, just reviewed it and happy with the approach apart from some nits.

pjbgf:
[...]

PR #1852, opened 2026-02-17
git: worktree, optimize infiles function for very large repos

k-anshul:
Relates to https://github.com/go-git/go-git/issues/1845.

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -428,13 +428,7 @@ func (w *Worktree) resetIndex(t *object.Tree, dirs, files []string) ([]string, e
 	}
 
 	removedFiles := make([]string, 0, len(changes))
-	var filesMap map[string]any
-	if len(files) > 0 {
-		filesMap = make(map[string]any, len(files))
-		for _, f := range files {
-			filesMap[filepath.Clean(f)] = nil
-		}
-	}
+	filesMap := buildFilePathMap(files)
 	for _, ch := range changes {
 		a, err := ch.Action()
 		if err != nil {
@@ -485,7 +479,7 @@ func (w *Worktree) resetIndex(t *object.Tree, dirs, files []string) ([]string, e
 }
 
 // inFiles checks if the given file is in the list of files. The incoming filepaths in files should be cleaned before calling this function.
-func inFiles(files map[string]any, v string) bool {
+func inFiles(files map[string]struct{}, v string) bool {
 	v = filepath.Clean(v)
 	_, exists := files[v]
 	return exists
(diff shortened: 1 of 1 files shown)


worktree.go
```diff
@@ -500,6 +500,10 @@ func (w *Worktree) resetWorktree(t *object.Tree, files []string) error {
 	}
 	b := newIndexBuilder(idx)
 
+	filesMap := make(map[string]bool, len(files))
+	for _, f := range files {
+		filesMap[filepath.Clean(f)] = true
+	}
```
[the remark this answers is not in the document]
k-anshul: This is not required. These are simple conversions and meaningful util function can not be extracted.

EladGabay (approved):
minor suggestion: map[string]struct{} is more idiomatic than map[string]any for set semantics, avoids allocating an interface header per entry

worktree.go
```diff
@@ -428,6 +428,13 @@ func (w *Worktree) resetIndex(t *object.Tree, dirs, files []string) ([]string, e
 	}
 
 	removedFiles := make([]string, 0, len(changes))
+	var filesMap map[string]any
```
pjbgf:
```suggestion
	var filesMap map[string]struct{}
```

worktree.go
```diff
@@ -500,6 +503,13 @@ func (w *Worktree) resetWorktree(t *object.Tree, files []string) error {
 	}
 	b := newIndexBuilder(idx)
 
+	var filesMap map[string]any
+	if len(files) > 0 {
+		filesMap = make(map[string]any, len(files))
+		for _, f := range files {
+			filesMap[filepath.Clean(f)] = nil
+		}
+	}
```
pjbgf: Let's extract this into a helper func instead, to avoid defining this twice.

worktree.go
```diff
@@ -428,6 +428,13 @@ func (w *Worktree) resetIndex(t *object.Tree, dirs, files []string) ([]string, e
 	}
 
 	removedFiles := make([]string, 0, len(changes))
+	var filesMap map[string]any
+	if len(files) > 0 {
+		filesMap = make(map[string]any, len(files))
```
pjbgf:
```suggestion
		filesMap = make(map[string]struct{}, len(files))
```

worktree.go
```diff
@@ -428,6 +428,13 @@ func (w *Worktree) resetIndex(t *object.Tree, dirs, files []string) ([]string, e
...
+		for _, f := range files {
+			filesMap[filepath.Clean(f)] = nil
```
pjbgf:
```suggestion
			filesMap[filepath.Clean(f)] = struct{}{}
```

k-anshul:
Thank you for the review @pjbgf and @EladGabay 

I have addressed the review comments. Could you please take a look again ?

pjbgf (approved):
@k-anshul thanks for working on this. 🙇

PR #1853, opened 2026-02-17
git: worktree, optimize infiles function for very large repos

k-anshul:
Backport for https://github.com/go-git/go-git/pull/1852.
Closes https://github.com/go-git/go-git/issues/1845.

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -386,13 +386,7 @@ func (w *Worktree) resetIndex(t *object.Tree, dirs []string, files []string) ([]
 	}
 
 	removedFiles := make([]string, 0, len(changes))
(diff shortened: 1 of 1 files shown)


cedric-appdirect:
Is it possible to have a benchmark or is there an existing benchmark that get better when using this PR?

pjbgf:
> Is it possible to have a benchmark or is there an existing benchmark that get better when using this PR?

@cedric-appdirect I agree benchmarking performance changes is a good practice. On this case specifically I'm slighly less worried due to the change itself. 

That being said, I'm planning to have some level of performance regression tests into CI in the near future.

PR #1862, opened 2026-02-23
backport, git: Improve Status() speed with new index.ModTime check

cedric-appdirect:
This is a backport of https://github.com/go-git/go-git/pull/1747 . There is more change than I would have wanted, but v5 and v6 have diverged a bit. I will try to highlight in the code what I had to change compared to v6 PR.

### DIFF
--- a/storage/test/storage_suite.go
+++ b/storage/test/storage_suite.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"time"
 
 	"github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
@@ -431,11 +430,8 @@ func (s *BaseStorageSuite) TestSetIndexAndIndex(c *C) {
 	c.Assert(err, IsNil)
 
 	// ModTime is set during SetIndex (memory storage) or read-back (filesystem storage).
-	// Verify it was set, then clear both sides for structural comparison.
+	// Verify it was set.
 	c.Assert(idx.ModTime.IsZero(), Equals, false)
-	idx.ModTime = time.Time{}
-	expected.ModTime = time.Time{}
-	c.Assert(idx, DeepEquals, expected)
 }
 
 func (s *BaseStorageSuite) TestSetConfigInvalid(c *C) {


storage/test/storage_suite.go
```diff
@@ -428,6 +429,12 @@ func (s *BaseStorageSuite) TestSetIndexAndIndex(c *C) {
 
 	idx, err := s.Storer.Index()
 	c.Assert(err, IsNil)
+
+	// ModTime is set during SetIndex (memory storage) or read-back (filesystem storage).
+	// Verify it was set, then clear both sides for structural comparison.
+	c.Assert(idx.ModTime.IsZero(), Equals, false)
+	idx.ModTime = time.Time{}
+	expected.ModTime = time.Time{}
```
cedric-appdirect: This tests somehow seems to run with both filesystem and memory, which require this overriding time to make the DeepEquals work.

utils/merkletrie/filesystem/node.go
```diff
@@ -42,7 +47,43 @@ func NewRootNode(
...
+// The index's ModTime field is used to detect the racy git condition. When a file's
+// mtime equals or is newer than the index ModTime, we must hash the file content
+// even if other metadata matches, because the file may have been modified in the
+// same second that the index was written.
+//
+// If idx is nil, the function works without the optimization (equivalent to
+// NewRootNodeWithOptions), which is useful for cases where no index is available.
+//
+// Reference: https://git-scm.com/docs/racy-git
+func NewRootNodeWithIndex(
```
cedric-appdirect: There is no WithOption in v5, so I went with creating WithIndex.

utils/merkletrie/filesystem/node_test.go
```diff
@@ -246,3 +251,138 @@ func IsEquals(a, b noder.Hasher) bool {
 
 	return bytes.Equal(a.Hash(), b.Hash())
 }
+
+func (s *NoderSuite) TestRacyGit(c *C) {
+	td, err := os.MkdirTemp(c.MkDir(), "racy-git")
```
cedric-appdirect, line 256: All the tests had to be modified with new parameters and c.Assert instead of s.Require().

utils/merkletrie/filesystem/node.go
pjbgf: The issue with this is that we are then making an additional breaking change on the API from v5 to v6. I'd rather we keep the API consistent with v6 instead.

storage/filesystem/index.go
```diff
@@ -48,6 +48,11 @@ func (s *IndexStorage) Index() (i *index.Index, err error) {
 
 	defer ioutil.CheckClose(f, &err)
 
+	fi, statErr := s.dir.Fs().Stat(f.Name())
```
pjbgf, line 51: Do you mind clarifying the change in approach as opposed of `f.Stat()` - as per the other PR?

storage/test/storage_suite.go
pjbgf: Yep the Suite goes through the different storage options. Did this break for the `transactional` storage? If that's the case, we can implement this change for it and then forward port that specific commit.

utils/merkletrie/filesystem/node.go
cedric-appdirect: Ok, will make ia WithOption and Option will have just index in it.

storage/filesystem/index.go
cedric-appdirect: Oh, forgot this one. In v6, we have Stat as part of the interface on f, but in v5, we don't. So I searched for a way to access the filesystem and do a stat. It does seems logical, but I definitively prefer the v6 Stat() form. I think that come from change in billy v6 to billy v5. I felt that was too big for a backport.

storage/test/storage_suite.go
cedric-appdirect: With the DeepEqual the time for ModTime for expected and idx must match, but at the time of instantiation of expected we don't know what time.Now will be when SetIndex does its job. The check that ModTime IsZero equal false, make sure that it was set, but we can't make the DeepEqual work without setting expected.ModTime. I could set expected.ModTime to idx.ModTime if that make the test a bit more logical/readable.

storage/test/storage_suite.go
```diff
@@ -428,6 +429,12 @@ func (s *BaseStorageSuite) TestSetIndexAndIndex(c *C) {
...
 	c.Assert(idx, DeepEquals, expected)
```
pjbgf: The below doesn't really test anything, so I think we are better off removing it:

```suggestion
```

storage/test/storage_suite.go
pjbgf: Thanks for clarifying. I'm closing this and adding a separate change instead.

storage/test/storage_suite.go
```diff
@@ -428,6 +429,12 @@ func (s *BaseStorageSuite) TestSetIndexAndIndex(c *C) {
 
 	idx, err := s.Storer.Index()
 	c.Assert(err, IsNil)
+
+	// ModTime is set during SetIndex (memory storage) or read-back (filesystem storage).
+	// Verify it was set, then clear both sides for structural comparison.
```
pjbgf:
```suggestion
	// Verify it was set.
```

storage/test/storage_suite.go
pjbgf:
```suggestion
	// Verify it was set.
	c.Assert(idx.ModTime.IsZero(), Equals, false)
```

pjbgf (dismissed):
@cedric-appdirect thanks for working on this. 🙇

pjbgf (approved):

PR #1873, opened 2026-02-27
storage: filesystem, decode ActualSize from delta object header

pjbgf:
Delta object sizes in go-git are normally resolved by walking the delta chain and materialising all base objects until the target object can be reconstructed and its size determined.

DeltaObject instead decodes the target size directly from the LEB128- encoded delta header, setting ActualSize without resolving the full chain. No verification of the delta or its size is performed.

### DIFF
--- a/storage/filesystem/object_test.go
+++ b/storage/filesystem/object_test.go
@@ -11,6 +11,8 @@ import (
 
 	"github.com/go-git/go-billy/v6"
 	fixtures "github.com/go-git/go-git-fixtures/v5"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 
 	"github.com/go-git/go-git/v6/plumbing"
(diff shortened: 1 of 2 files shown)


runxiyu:
The copilot review comments seem valid here.

storage/filesystem/object_test.go
```diff
@@ -641,3 +643,20 @@ func (s *FsSuite) TestGetFromUnpackedDoesNotCacheLargeObjects() {
...
+
+	obj, err := st.DeltaObject(plumbing.AnyObject, plumbing.NewHash("aa9b383c260e1d05fbbf6b30a02914555e20c725"))
+	require.NoError(t, err)
+
+	assert.Equal(t, plumbing.OFSDeltaObject, obj.Type())
+	assert.Equal(t, int64(4), obj.Size())
+
+	dobj, ok := obj.(plumbing.DeltaObject)
+	assert.True(t, ok)
+	assert.Equal(t, int64(73), dobj.ActualSize())
```
runxiyu, line 661: Unfamiliar with how other go-git tests work but I find it awkward to have this magic number here; could we derive it via other means and compare?

storage/filesystem/object_test.go
runxiyu, line 1: * Edge cases e.g. decoded target size 0 (which should not occur in practice, because it would not make sense to deltify the blob anyway, but it's technically possible)
* Maybe compare through full object materialization? (Or by using git as an oracle, although that would be quite different from go-git's style overall wrt here)

storage/filesystem/object_test.go
pjbgf: Overall those magic numbers come from a fixture state. I'm working on improving the go-git-fixtures to expose those values for specific fixtures, so that we can remove this approach across the board.

storage/filesystem/object_test.go
pjbgf: Given that this handles data from the raw delta, without providing any sort of validation, I'm not sure whether we should provide additional checks on the test itself.

That being said, I can certainly see the value of mapping delta objects from the fixture(s), looping through them and comparing the values against fully materialised objects. That was we can at least confirm the logic is consistent and working properly.

PR #1875, opened 2026-02-28
submodule: add URLRewriter for custom submodule URL resolution

paralin:
Add URLRewriter callback to SubmoduleUpdateOptions and SubmoduleURLRewriter to CloneOptions, allowing callers to rewrite submodule remote URLs before fetching.

This is useful when cloning from a local repository where submodules should be fetched from the source repo's .git/modules/ directory rather than from the remote URLs recorded in .gitmodules.

Add LocalSubmoduleRewriter helper that maps submodule names to .git/modules/<name>/ paths under a given local repository path, falling back to the original URL when the module directory does not exist.

No existing behavior is changed; the rewriter is nil by default.

This is part of a set of optimizations for speeding up the clone operation against a Git repository on disk (file path) which has many and/or large Git submodules.

### DIFF
--- a/_examples/common_test.go
+++ b/_examples/common_test.go
@@ -45,9 +45,10 @@ var args = map[string][]string{
 
 // tests not working / set-up
 var ignored = map[string]bool{
-	"ls":              true,
-	"sha256":          true,
-	"submodule":       true,
+	"local-clone":    true,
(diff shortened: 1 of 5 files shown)


paralin:
comments addressed

PR #1879, opened 2026-03-05
x: worktree, fix worktreeNameRE, tests

0xDEC0DE:
Update the FuzzAdd test in x/plumbing/worktree/worktree_test.go to capture whether the Add command should throw an error or not, rather than leaving it up to a regex expansion.  This can help surface bugs in the regex itself.

Also update the regex to expand what it considers a valid name.

### DIFF
--- a/x/plumbing/worktree/worktree.go
+++ b/x/plumbing/worktree/worktree.go
@@ -37,7 +37,7 @@ const (
 )
 
 var (
-	worktreeNameRE = regexp.MustCompile(`^[a-zA-Z0-9\-]+$`)
+	worktreeNameRE = regexp.MustCompile(`^[a-zA-Z0-9._\-]+$`)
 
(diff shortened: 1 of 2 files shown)


x/plumbing/worktree/worktree.go
```diff
@@ -37,7 +37,7 @@ const (
 )
 
 var (
-	worktreeNameRE = regexp.MustCompile(`^[a-zA-Z0-9\-]+$`)
+	worktreeNameRE = regexp.MustCompile(`^[a-zA-Z0-9._\-]+$`)
```
0xDEC0DE, line 40: I'm honestly at a loss as to why this guard is here.  `git worktree add` is profoundly forgiving when it comes to the names of worktrees, I'm pretty sure it accepts anything that is allowed in a pathname, i.e., nearly the full Unicode character set is allowed.

But at minimum, accepting all the common characters you're likely to find in a Unix pathname is a better place to be in.

PR #1880, opened 2026-03-05
storage: filesystem: add stat-based index cache

pjbgf:
Introduce an `IndexCache` interface and a default stat-based implementation that avoids redundant disk reads and decoding of the git index file. On each `Index()` call the file is stat'd and the cached copy is returned when modification time and size match, falling back to a full read+decode on a miss. `SetIndex()` populates the cache via write-through so the immediately following Index() is served from memory.

Both read and write paths use `copyIndex()` which copies the Index struct value and clones the Entries slice, ensuring callers cannot corrupt the cached data through post-call mutations.

The cache can be injected via `Options.IndexCache` for testing or custom invalidation strategies; a stat-based default is created automatically when none is provided.

A new benchmark to assess commits was introduced, there is a large variance on sec/op and B/op between the first call and the subsequent ones - due to other performance issues that will be looked as a follow-up. However, the impact of the new cache has a clear impact on allocations:

```
goos: linux
goarch: amd64
cpu: AMD Ryzen AI 9 HX PRO 370 w/ Radeon 890M
                 │ /tmp/before  │           /tmp/after           │
                 │    sec/op    │    sec/op     vs base          │
Commit/Commit-24   6.299m ± 42%   5.719m ± 27%  ~ (p=0.579 n=10)

                 │  /tmp/before  │           /tmp/after            │
                 │     B/op      │     B/op       vs base          │
Commit/Commit-24   2.963Mi ± 41%   2.498Mi ± 29%  ~ (p=0.280 n=10)

                 │ /tmp/before  │              /tmp/after              │
                 │  allocs/op   │  allocs/op    vs base                │
Commit/Commit-24   73.38k ± 40%   29.82k ± 29%  -59.36% (p=0.001 n=10)
```

The impact is observable by running the performance trace for the [commit example](https://github.com/go-git/go-git/blob/main/_examples/commit/main.go) in a very large repository (i.e. Linux Kernel).

Before these changes, the 3 calls to `Store.Index()` accounted for `0.3s`:
```
11:06:24.362677 index.go:49: performance: 0.104357395 s: storage/filesystem: get index()
11:06:24.448323 index.go:49: performance: 0.085250701 s: storage/filesystem: get index()
11:06:26.925873 index.go:49: performance: 0.060172693 s: storage/filesystem: get index()
11:06:27.542282 index.go:49: performance: 0.060106399 s: storage/filesystem: get index()
```

With the new cache, the cost for subsequent calls are considerably reduced. In this example, the total time dropped to `0.11s`:
```
11:06:00.865018 index.go:61: performance: 0.107215545 s: storage/filesystem: get index()
11:06:00.867021 index.go:61: performance: 0.001620462 s: storage/filesystem: get index()
11:06:03.346510 index.go:61: performance: 0.000107721 s: storage/filesystem: get index()
11:06:03.958044 index.go:61: performance: 0.000076523 s: storage/filesystem: get index()
```

Follow-up from #1747.

### DIFF
--- a/storage/filesystem/dotgit/dotgit.go
+++ b/storage/filesystem/dotgit/dotgit.go
@@ -206,6 +206,11 @@ func (d *DotGit) Index() (billy.File, error) {
 	return d.fs.Open(indexPath)
 }
 
+// StatIndex returns the os.FileInfo for the index file without opening it.
+func (d *DotGit) StatIndex() (os.FileInfo, error) {
+	return d.fs.Stat(indexPath)
+}
+
 // ShallowWriter returns a file pointer for write to the shallow file
 func (d *DotGit) ShallowWriter() (billy.File, error) {
 	return d.fs.Create(shallowPath)
--- a/storage/filesystem/index.go
+++ b/storage/filesystem/index.go
@@ -2,6 +2,7 @@ package filesystem
 
 import (
 	"bufio"
+	"errors"
 	"hash"
 	"os"
 	"time"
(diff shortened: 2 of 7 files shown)


pjbgf:
@Bluebugs this expands on the recent changes you made. PTAL

cedric-appdirect:
Looks interesting and make me wonder how we can make this help with commit across worktree. I might look into this at some point.

Issue #1869, opened 2026-02-27
Improve UX for verifying commits and Tags


pjbgf:
### Feature Description

The `go-git` signing primitives are being reviewed as part of #1849. The same exercise is needed for the verification counterpart. 

This issue tracks some of the improvements:
- [ ] Provide trustworthy out of tree implementations for essential verifiers (e.g. `ssh` and `X.509`).
- [ ] Enable `go-git` implementations to be replaceable.

bb-Ricardo:
Hi,

is this actually happening? I was waiting for this since pretty much two years but now there seems to be some traction, which makes me quite happy.

I worked the the last two days on implementing SSH signature validation for FluxCD: https://github.com/fluxcd/pkg/pull/1141

Two (known to me) signing implementations I was looking into getting implemented are:
* https://github.com/github/smimesign (also not sure how alive this project still is)
* https://github.com/sigstore/gitsign/tree/main

Is it planned to add verifiers for both of these implementations? I guess they differ in how the certificate is validated but the S/MIME envelope should be the same.

Looking through the PRs and issues, there seems to be quite a bit of contributions around signing and valuation. Is there a plan for a design around this topic? Sort of an interface to have a common abstraction for types signatures. Also the `Commit.Signuature`/`Tag.Signature` should be a structure with handlers instead of just a string.

What do you think?

Bluebugs:
I have made a draft PR very early on when I started working on this project #1828 . That PR is absolutely not ready to merge as I have learned since a lot about this project, but it provide a proof of concept with some API that can be made to align with the project. My understanding at this point is to add an interface for a "verifier" and trust level like in that draft, use the plugin interface introduced by #1860 and the in go-git/x add the verifier implementation for GPG and SSH. I am personally interested in only the ssh side, so that's the one I have tested/played with so far in that draft PR, but the GPG code should work as well.

There is definitively things that are not correct/ideal in that draft PR like it parse the allowed ssh key files, but ignore the after/before information associated with a key. So if anything, at this point it is mostly a discussion about the API that would be useful. I was hoping to have time to propose a draft PR with the Verifier API + plugin as a follow up to #1860, earlier, but I think I won't have time until next weekend.

Still, if you can share your opinion mostly on the API in plumbing/object/verification.go from #1828 that would be helpful. Glad to know other are interested in this.

pjbgf:
[...]

PR #1883, opened 2026-03-06
Verify plugin

Bluebugs:
Added on top of #1860 a verify plugin that could be used like so:
```go
import (
    "errors"
    "fmt"
    "os"

    "github.com/go-git/go-git/v6/x/plugin"

    // Out-of-tree verifier implementations (e.g. from github.com/go-git/x)
    gossh "github.com/go-git/x/plugin/objectverifier/ssh"
    // "github.com/go-git/x/plugin/objectverifier/gpg"
)

func main() {
    // Register a verifier plugin globally — once at startup.
    // All commits/tags fetched via Repository methods will auto-verify.
    plugin.Register(plugin.ObjectVerifier(),
        func() plugin.Verifier {
            v, _ := gossh.FromAllowedSignersFile("~/.ssh/allowed_signers")
            return v
            // return gpg.FromArmoredKeyRing(keyring)
        })

    r, _ := git.PlainOpen("/path/to/repo")

    // Commits retrieved via the repo auto-carry the verifier
    commit, _ := r.CommitObject(someHash)

    result, err := commit.Verify()
    switch {
    case err == nil:
        fmt.Printf("Verified: %s (signed by %s)\n", result.KeyID, result.Signer)
    case errors.Is(err, object.ErrKeyNotTrusted):
        fmt.Printf("Valid signature from unknown key: %s\n", result.KeyID)
    case errors.Is(err, object.ErrSignatureInvalid):
        fmt.Println("Signature is invalid — content may have been tampered with")
    case errors.Is(err, object.ErrNoSignature):
        fmt.Println("Commit is unsigned")
    default:
        fmt.Printf("Verification failed: %v\n", err)
    }
}
```

This is trying to address the discussion from #1869 . I couldn't figure out a way to do this PR against PR #1860, so it does have all the commit from it too :-(

### DIFF
--- a/plumbing/object/signature.go
+++ b/plumbing/object/signature.go
@@ -2,12 +2,33 @@ package object
 
 import "bytes"
 
+// SignatureType represents the type of a cryptographic signature.
+type SignatureType int8
+
 const (
-	signatureTypeUnknown signatureType = iota
-	signatureTypeOpenPGP
-	signatureTypeX509
-	signatureTypeSSH
+	// SignatureTypeUnknown indicates an unrecognized signature format.
+	SignatureTypeUnknown SignatureType = iota
+	// SignatureTypeOpenPGP indicates an OpenPGP (GPG) signature.
+	SignatureTypeOpenPGP
+	// SignatureTypeX509 indicates an X.509 (S/MIME) signature.
+	SignatureTypeX509
+	// SignatureTypeSSH indicates an SSH signature.
+	SignatureTypeSSH
 )
+
+// String returns the string representation of the signature type.
+func (t SignatureType) String() string {
+	switch t {
+	case SignatureTypeOpenPGP:
+		return "OpenPGP"
+	case SignatureTypeX509:
+		return "X.509"
+	case SignatureTypeSSH:
+		return "SSH"
+	default:
+		return "unknown"
+	}
+}
 
 var (
 	// openPGPSignatureFormat is the format of an OpenPGP signature.
@@ -19,6 +40,7 @@ var (
 	// a PKCS#7 (S/MIME) signature.
 	x509SignatureFormat = signatureFormat{
 		[]byte("-----BEGIN SIGNED MESSAGE-----"),
+		[]byte("-----BEGIN CERTIFICATE-----"),
 	}
 
 	// sshSignatureFormat is the format of an SSH signature.
@@ -28,21 +50,23 @@ var (
 )
 
 // knownSignatureFormats is a map of known signature formats, indexed by
-// their signatureType.
-var knownSignatureFormats = map[signatureType]signatureFormat{
-	signatureTypeOpenPGP: openPGPSignatureFormat,
-	signatureTypeX509:    x509SignatureFormat,
-	signatureTypeSSH:     sshSignatureFormat,
+// their SignatureType.
+var knownSignatureFormats = map[SignatureType]signatureFormat{
+	SignatureTypeOpenPGP: openPGPSignatureFormat,
+	SignatureTypeX509:    x509SignatureFormat,
+	SignatureTypeSSH:     sshSignatureFormat,
 }
-
-// signatureType represents the type of the signature.
-type signatureType int8
 
 // signatureFormat represents the beginning of a signature.
 type signatureFormat [][]byte
 
+// DetectSignatureType returns the SignatureType for the given signature bytes.
+func DetectSignatureType(sig []byte) SignatureType {
+	return typeForSignature(sig)
+}
+
 // typeForSignature returns the type of the signature based on its format.
-func typeForSignature(b []byte) signatureType {
+func typeForSignature(b []byte) SignatureType {
 	for t, i := range knownSignatureFormats {
 		for _, begin := range i {
 			if bytes.HasPrefix(b, begin) {
@@ -50,7 +74,7 @@ func typeForSignature(b []byte) signatureType {
 			}
 		}
 	}
-	return signatureTypeUnknown
+	return SignatureTypeUnknown
 }
 
 // parseSignedBytes returns the position of the last signature block found in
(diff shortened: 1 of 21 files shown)


plumbing/object/signature.go
```diff
@@ -19,6 +40,7 @@ var (
 	// a PKCS#7 (S/MIME) signature.
 	x509SignatureFormat = signatureFormat{
 		[]byte("-----BEGIN SIGNED MESSAGE-----"),
+		[]byte("-----BEGIN CERTIFICATE-----"),
```
bb-Ricardo, line 43: Is this supported by git, or implemented by any other git implementation? https://github.com/git/git/blob/795c338de725e13bd361214c6b768019fc45a2c1/gpg-interface.c#L65

bb-Ricardo:
I had a quick look at it. From the signer part it makes sense to use a configured signer. But from the verifier it seems to make sense to be able to iterated over a list from preconfigured verifiers. As git commits in a tree can have different types of signatures I would like to be able to apply a preconfigured verifier based on the signature type.

Or the verifiers would implement a certain type of signature and then be based on the type.

Bluebugs:
> I had a quick look at it. From the signer part it makes sense to use a configured signer. But from the verifier it seems to make sense to be able to iterated over a list from preconfigured verifiers. As git commits in a tree can have different types of signatures I would like to be able to apply a preconfigured verifier based on the signature type.
> 
> Or the verifiers would implement a certain type of signature and then be based on the type.

I was thinking of implementing chaining of verifiers out of tree following this pattern, but that would be in `go-git/x/`. I can see pro and cons to having it in tree. Pros is that the mechanism is always there included directly in the verifiers API. The cons is that now you aren't quite necessarily sure of the chain of plugin registered for doing verification. That was my reasoning for moving it out of tree, it make it explicit when you have a chain.

bb-Ricardo:
> I was thinking of implementing chaining of verifiers out of tree following this pattern, but that would be in `go-git/x/`. I can see pro and cons to having it in tree. Pros is that the mechanism is always there included directly in the verifiers API. The cons is that now you aren't quite necessarily sure of the chain of plugin registered for doing verification. That was my reasoning for moving it out of tree, it make it explicit when you have a chain.

You don't really need chaining as the type of signatures requires a certain verifier. And as we currently only have three types of signatures this would result in a GPG verifier, SSH verifier and a x509 verifier. And you would leave it to the go-git user to setup the verifier with the desired set of verification config.

```go
// simple example
plugin.Register(plugin.ObjectSSHVerifier(),
    func() plugin.Verifier {
        v, _ := gossh.FromAllowedSignersFile("~/.ssh/allowed_signers")
        return v
    }
)

plugin.Register(plugin.ObjectGPGVerifier(),
    func() plugin.Verifier {
        v, _ := plugin.GPGVerifierWithKeyring("my-trusted-keyring.asc")
        return v
    }
)
```

I might have missed it but in order to verify for example a trust level, the verifier needs access to the commit/tag attributes like author, committer, timestamp, id without parsing it from the signed message again. Or is this intentional?

cedric-appdirect:
> You don't really need chaining as the type of signatures requires a certain verifier. And as we currently only have three types of signatures this would result in a GPG verifier, SSH verifier and a x509 verifier. And you would leave it to the go-git user to setup the verifier with the desired set of verification config.

Indeed, that simplify things in a way. For my use case, that would work. Ideally, you would want the interface to not be interchangeable to avoid the user to make mistake that don't get caught at compilation time. Do you see some difference between this verifier that I could codify in the API?

Otherwise I can add just a `SupportsSignatureType(object.SignatureType) bool` to the plugin verifier API and this avoid the problem as go-git can ensure to call the right plugin for the right type.

> I might have missed it but in order to verify for example a trust level, the verifier needs access to the commit/tag attributes like author, committer, timestamp, id without parsing it from the signed message again. Or is this intentional?

That's a good point. I could pass the commit itself, not just the []byte. My interest being mostly on ssh, it is good to have additional review indeed. I will add that to the Verifier API.

plumbing/object/verification_test.go
```diff
@@ -0,0 +1,143 @@
+package object
+
+import (
+	"testing"
+)
+
+func TestTrustLevel_String(t *testing.T) {
+	t.Parallel()
```
pjbgf, line 8: Not sure whether trust levels are required for this level of abstraction. For core-tree, I'd rather we keep it agnostic to the verification method and trust level is a concept from GPG - IIRC.

plumbing/object/verification_test.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	for _, tt := range tests {
+		if got := tt.st.String(); got != tt.want {
+			t.Errorf("SignatureType(%d).String() = %q, want %q", tt.st, got, tt.want)
+		}
+	}
+}
+
+func TestDetectSignatureType(t *testing.T) {
```
pjbgf, line 118: For the abstraction level of the core, I'd proactively not use actual verification methods the important thing is: 
- Is the new API flexible enough to account for any verification method?
- Do we provide ways to handle failure scenarios in a sensible manner?

plumbing/object/verification.go
```diff
@@ -0,0 +1,115 @@
...
+	Type SignatureType
+
+	// KeyID is the identifier of the signing key.
+	// For OpenPGP: the key ID (last 16 hex chars of fingerprint)
+	// For SSH: the key fingerprint (SHA256:...)
+	KeyID string
+
+	// PrimaryKeyFingerprint is the full fingerprint of the primary key.
+	// For OpenPGP subkeys, this differs from the signing key fingerprint.
+	PrimaryKeyFingerprint string
```
pjbgf, line 78: This feels like implementation details. Would a `X.509` verifier require either of this?

The question is, how to enable that use case without hard-coding it on our exposed types?

plumbing/object/signature.go
pjbgf: We had that in the past, we no longer have - I just don't recall if my PR removed it or if that is already in `main`.

cedric-appdirect: It may have come back because I started working on this PR month ago. Will make sure to remove it.

pjbgf:
Thank you both for looking at this. I should get the other PR ready soonish. Here's my initial feedback on the discussions I saw here (as well as the changes):

> I was thinking of implementing chaining of verifiers out of tree following this pattern, but that would be in `go-git/x/`

@Bluebugs  +1 on this approach. In theory any verifier can be a chain in itself. Meaning - we could have ssh/gpg static verifiers, each as different implementation. And a third one that is a chain that goes through whatever verifiers you pass at it. 


> we currently only have three types of signatures 

@bb-Ricardo ideally the in-tree code would be oblivious to verifier implementation. If someone believes that they have a better way of signing/verifying commits, I'd like to enable that use case without having to change `go-git's` code base - pretty much along the lines of [OCP](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle). I think we may have achieved that for Signers, it would be great to do the same for verifiers.

PR #1884, opened 2026-03-06
plumbing: transport, fix shallow boundaries not persisted; remote, skip bare-hash dst refs

AriehSchneier:
## Summary

Two bugs caused fetching a specific commit by hash (using `+<hash>:<hash>`) into an existing shallow-cloned repository to silently fail or corrupt state.

### Regression test

`TestFetchByHashThenResolveRevision` in `repository_test.go` covers both bugs end-to-end:
1. Clone `git-fixtures/basic.git` at `depth=1`
2. Fetch a deep ancestor commit by SHA using `+<hash>:<hash>`
3. Assert the commit object is present, `ResolveRevision` returns the correct SHA, and `Worktree.Checkout` succeeds

### Bug 1 — shallow boundaries not persisted after clone (`plumbing/transport/negotiate.go`)

`readShallows` checked `if shallowInfo == nil` where `shallowInfo` is a `**packp.ShallowUpdate`. This pointer is always non-nil (it is the address of a local variable in the caller), so the body that assigns `*shallowInfo = &shupd` never executed.

**Fix:** check `*shallowInfo == nil`, which dereferences the outer pointer and tests whether a shallow update has already been captured.

Without this fix, `st.Shallow()` returns an empty slice on every subsequent fetch. `NegotiatePack` therefore sends no `shallow` lines to the server. The server sees `have 6ecf0ef2` (the shallow tip), infers the client has its full ancestry, concludes the wanted commit (a deep ancestor of the tip) is already present, and returns a 32-byte empty packfile.

### Bug 2 — spurious branch ref created for bare-hash dst (`remote.go`, `updateLocalReferenceStorage`)

When the dst portion of a refspec is a raw SHA (e.g. `+<hash>:<hash>`), the code fell into the "treat as branch" path and called `plumbing.NewBranchReferenceName(<sha>)`, creating `refs/heads/<sha>` pointing to the zero hash. `ResolveRevision` then resolved via that spurious ref and returned the zero hash instead of the real commit hash.

**Fix:** detect `plumbing.IsHash(localName.String())` and skip ref creation — a bare-hash dst means "fetch the object only".

### Complementary fix (`remote.go`, `fetch`)

When building the HAVE list for a shallow fetch, shallow-boundary commits are now excluded. This matches the behaviour of the reference git client and avoids the server treating their ancestors as implicitly owned by the client.

### DIFF
--- a/repository_test.go
+++ b/repository_test.go
@@ -450,6 +450,109 @@ func TestFetchMustNotUpdateObjectFormat(t *testing.T) {
 	}
 }
 
+// TestFetchByHashThenResolveRevision is a regression test for two bugs that
+// affected fetching a specific commit by hash into an existing repository
+// using the force-refspec "+<hash>:<hash>".
+//
+// Regression 1 (negotiate.go): shallow boundaries were not persisted after a
+// clone, so subsequent fetches sent no "shallow" lines to the server. The
+// server then inferred the client already owned the wanted commit (a deep
+// ancestor of the shallow tip) and returned an empty packfile, leaving the
+// object absent from the store.
+//
+// Regression 2 (remote.go): when the dst of a refspec is a bare SHA, the
+// code created a branch named after the hash (refs/heads/<sha>) pointing to
+// the zero hash. ResolveRevision resolved via that spurious ref and returned
+// the zero hash instead of the real commit hash, breaking all downstream
+// callers such as Worktree.Checkout.
+//
+// Note: this test uses the live GitHub URL rather than a local fixture server
+// because the fixture HTTP server does not implement depth filtering in
+// upload-pack. Without depth support the server sends all objects on a
+// depth=1 clone, making commitSHA present from the start and preventing
+// regression 1 from being reproduced.
+func TestFetchByHashThenResolveRevision(t *testing.T) {
+	t.Parallel()
+
+	// git-fixtures/basic.git commit graph (abbreviated):
+	//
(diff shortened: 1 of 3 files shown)


repository_test.go
```diff
@@ -450,6 +450,94 @@ func TestFetchMustNotUpdateObjectFormat(t *testing.T) {
...
+		commitSHA = "35e85108805c84807bc66a02d91535e1e24b38b9"
+	)
+
+	tmp := t.TempDir()
+
+	// Step 1: clone the default branch at depth=1.
+	// commitSHA is a deep ancestor excluded by the shallow depth limit.
+	r, err := PlainClone(tmp, &CloneOptions{
+		URL:   repoURL,
+		Depth: 1,
```
pjbgf, line 501: Instead of cloning the repository out from the internet, we should lean on the go-git fixtures. The `TestCloneAll` is a good example from a non `file://` fixture clone.

aymanbagabas (approved):

repository_test.go
AriehSchneier: I have rebased the change onto main and updated the comment to explain why we can't use the fixtures and need to pull directly from the internet. (No change to the code other than updating comments in the test)

pjbgf: We could change this to lean on the `git daemon` as the source of truth, but will be merging as is for the time being. I will come up with some new option to E2E test our server implementation.

Thanks for that @AriehSchneier.

PR #1887, opened 2026-03-07
git: remote: handle shallow clone in isFastForward

### LINKED ISSUE #207: Pull returns object not found

Br1an67:
Fixes #207

## Summary

When doing a shallow clone and then pulling with depth 0 (unlimited), the `isFastForward` function would return `plumbing.ErrObjectNotFound` because the commit iterator cannot find parent commits that don't exist in the shallow clone.

This change makes `isFastForward` return `false` (not a fast-forward) instead of an error when encountering missing objects during iteration. This matches the behavior already implemented in `getHavesFromRef` which ignores errors from the commit iterator for the same reason.

## Changes

- Modified `isFastForward` in `remote.go` to handle `plumbing.ErrObjectNotFound` gracefully
- When the error occurs during commit iteration, return `false, nil` instead of propagating the error
- Moved the explanatory comment to where the error is actually handled

## Testing

- Verified existing test `TestFetchAfterShallowClone` passes
- Verified all `RemoteSuite` tests pass
- The fix follows the same pattern already used in `getHavesFromRef` (line 849)

```
 remote.go | 10 +++++++---
 1 file changed, 7 insertions(+), 3 deletions(-)
```

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -1072,9 +1072,6 @@ func isFastForward(s storer.EncodedObjectStorer, old, newHash plumbing.Hash, ear
 
 	found := false
 	// stop iterating at the earliest shallow commit, ignoring its parents
-	// note: when pull depth is smaller than the number of new changes on the remote, this fails due to missing parents.
-	//       as far as i can tell, without the commits in-between the shallow pull and the earliest shallow, there's no
-	//       real way of telling whether it will be a fast-forward merge.
 	iter := object.NewCommitPreorderIter(c, nil, parentsToIgnore)
 	err = iter.ForEach(func(c *object.Commit) error {
 		if c.Hash != old {
@@ -1084,6 +1081,13 @@ func isFastForward(s storer.EncodedObjectStorer, old, newHash plumbing.Hash, ear
 		found = true
 		return storer.ErrStop
 	})
+	if errors.Is(err, plumbing.ErrObjectNotFound) {
+		// When pull depth is smaller than the number of new changes on the remote,
+		// this fails due to missing parents. Without the commits in-between the
+		// shallow pull and the earliest shallow, there's no real way of telling
+		// whether it will be a fast-forward merge, so return false.
+		return false, nil
+	}
 	return found, err
 }
 


AriehSchneier:
AI checking your work:

The fix is directionally right but there's a meaningful behavioral problem. Here's the analysis:

**What the PR does**: when the ancestry walk in `isFastForward` hits `ErrObjectNotFound` (missing intermediate commits in a shallow clone), return `false, nil` instead of propagating the error.

**The core problem with returning `false`**: it produces the wrong observable failure. Consider:
- Shallow clone at depth 1 → local tip is `Z`
- Remote has 3 new commits: `Z→A→B→C`
- `git pull --depth=1` fetches `C` only, not `A` or `B`
- Walk from `C` → `B` (missing) → `ErrObjectNotFound` before reaching `Z`
- PR returns `false` → caller in `checkFastForwardUpdate` returns `"non-fast-forward update: refs/heads/main"` to the user

That's a misleading error — the update *is* a fast-forward, we just can't prove it because we didn't fetch deep enough. **Real git** reaches the same conclusion (can't confirm ancestry → conservative refusal), so the behavior is broadly consistent with git, but the *root cause* is the missing `earliestShallow` argument.

**The real fix**: both callers pass `nil` for `earliestShallow`:

```go
// line 1067
ff, err := isFastForward(s, cmd.Old, cmd.New, nil)

// line 1172  
ff, err := isFastForward(r.s, old.Hash(), newRef.Hash(), nil)
```

The `earliestShallow` mechanism already exists to stop the walk at the known shallow boundary — but it's never used. The correct fix would wire the actual shallow boundary in from `r.s` (the storer implements `storer.ShallowStorer`) so the walk stops at the right point, allowing the answer to be `true` for legitimate FF updates in shallow repos.

**The `getHavesFromRef` comparison is misleading**: that function fully discards errors (`_ = walker.ForEach(...)`) because it's just building a best-effort "haves" advertisement — correctness doesn't matter there. `isFastForward` is making a hard allow/deny decision, so blindly returning `false` on error is not the same pattern.

**Bottom line**: the PR fixes the crash/error surface (object-not-found → non-fast-forward error instead of an unhandled error), but it's a band-aid. The proper fix is to pass the shallow boundaries from `r.s` into `isFastForward` so shallow pulls that are genuine fast-forwards are correctly identified as such.

AriehSchneier:
@Br1an67 Please tell me what you think of https://github.com/go-git/go-git/pull/1892 as a replacement?

pjbgf:
@Br1an67 thanks for working on this, but unfortunately we ended up merging #1892. I'm closing this PR as it is no longer needed.

PR #1890, opened 2026-03-08
plumbing: transport/serve, fix detached HEAD advertised as broken symref

AriehSchneier:
## Summary

Fix a bug where cloning from a source repository with a detached HEAD fails with `ErrReferenceNotFound`.

## Root Cause

When the source has a detached HEAD (a `HashReference`), `AdvertiseReferences` was unconditionally emitting a `symref=HEAD:` capability with an empty target — because `r.Target()` returns `""` for a `HashReference`. The client's `addSymbolicRefs` then stored `HEAD` as a symbolic ref pointing to the empty string. Any subsequent `storer.ResolveReference(HEAD)` followed that broken symref and returned `ErrReferenceNotFound`.

## Fix

Guard the `symref` capability emission behind `r.Type() == plumbing.SymbolicReference`. A detached HEAD is now advertised only as a hash in `ar.Head` and `ar.References`, matching the behaviour of the reference `git` implementation.

## Changes

- `plumbing/transport/serve.go`: only emit `symref=HEAD:<target>` when HEAD is a symbolic reference.
- `repository_test.go`: adds `TestPlainCloneContext_DetachedHeadSource` as a regression test (self-contained, no network access).

### DIFF
--- a/repository_test.go
+++ b/repository_test.go
@@ -561,6 +561,79 @@ func TestPlainCloneContext_EmptyRemoteDoesNotCleanup(t *testing.T) {
 	assert.Equal(t, "origin", reopenedRemotes[0].Config().Name)
 }
 
+// TestPlainCloneContext_DetachedHeadSource is a regression test for a bug
+// where PlainCloneContext returns plumbing.ErrReferenceNotFound when the
+// source repository has a detached HEAD.
+//
+// A detached HEAD is the normal state of a repository after
+// Worktree.Checkout is called with CheckoutOptions{Hash: someHash}.
+// Cloning FROM such a repo (e.g. when using a local filesystem directory
+// as a "remote") must succeed, just as `git clone` does.
+func TestPlainCloneContext_DetachedHeadSource(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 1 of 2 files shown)


repository_test.go
```diff
@@ -498,6 +498,64 @@ func TestPlainCloneContext_FailedClonePreservesPreexistingEmptyDirectory(t *test
...
+	// store; only HEAD is detached. PlainCloneContext must succeed — just as
+	// `git clone` does — by advertising the available refs rather than
+	// requiring HEAD to be symbolic.
+	dstDir := filepath.Join(t.TempDir(), "dst")
+	_, err = PlainCloneContext(context.Background(), dstDir, &CloneOptions{
+		URL: srcDir,
+	})
+	assert.NoError(t, err,
+		"PlainCloneContext must succeed when the source repo has a detached HEAD")
+}
```
pjbgf, line 635: Let's add additional checks that the cloned repo:
1. Has a detached HEAD.
2. The HEAD points to the same commit as the server.

AriehSchneier: @pjbgf Done

PR #1891, opened 2026-03-08
*: clone, resolve relative local URLs against CWD in CloneOptions.Validate

### LINKED ISSUE #1723: v6: Clone of parent fails (worked in v5)

AriehSchneier:
When a caller passes a URL like "../../some-repo" to Clone, the billy chroot-based FilesystemLoader rejects any path whose cleaned form starts with ".." ("chroot boundary crossed"), regardless of the chroot root.

Fix: in CloneOptions.Validate, detect local URLs whose cleaned form starts with ".." and resolve them to absolute paths via filepath.Abs before the URL reaches the transport layer. This matches the behaviour of `git clone`: relative paths are resolved against the caller's working directory at the command level, before any transport code runs.

Adds TestCloneOptionsValidate_RelativeDotDot as a regression test.

Note: This fix only applies to code paths that call CloneOptions.Validate (e.g., Repository.Clone). Direct calls to transport.NewEndpoint or transport.ParseURL will not have ../-relative paths normalized. A future PR may add normalization at the transport layer to cover all code paths.

Fixes: https://github.com/go-git/go-git/issues/1723

### DIFF
--- a/options.go
+++ b/options.go
@@ -3,6 +3,7 @@ package git
 import (
 	"errors"
 	"fmt"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"time"
@@ -10,6 +11,7 @@ import (
 	"github.com/go-git/go-billy/v6"
 
 	"github.com/go-git/go-git/v6/config"
+	giturl "github.com/go-git/go-git/v6/internal/url"
 	"github.com/go-git/go-git/v6/plumbing"
 	"github.com/go-git/go-git/v6/plumbing/client"
 	"github.com/go-git/go-git/v6/plumbing/object"
@@ -134,6 +136,22 @@ const (
 func (o *CloneOptions) Validate() error {
 	if o.URL == "" {
 		return ErrMissingURL
+	}
+
+	// Resolve ..-relative local paths to absolute paths before they reach
+	// the transport layer. The billy chroot-based FilesystemLoader rejects
+	// any path whose clean form starts with ".." with "chroot boundary
+	// crossed". Real git resolves such paths against the working directory
+	// at the command level, before any transport code runs.
+	if giturl.IsLocalEndpoint(o.URL) && !filepath.IsAbs(o.URL) {
+		cleaned := filepath.Clean(o.URL)
+		if cleaned == ".." || strings.HasPrefix(cleaned, ".."+string(filepath.Separator)) {
+			abs, err := filepath.Abs(o.URL)
+			if err != nil {
+				return fmt.Errorf("resolve local clone URL %q to absolute path: %w", o.URL, err)
+			}
+			o.URL = abs
+		}
 	}
 
 	if o.RemoteName == "" {
--- a/options_test.go
+++ b/options_test.go
@@ -1,6 +1,7 @@
 package git
 
 import (
+	"path/filepath"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


plumbing/transport/transport.go
```diff
@@ -176,6 +177,17 @@ func parseFile(endpoint string) (*Endpoint, bool) {
 	}
 
 	path := endpoint
+	// Paths that navigate upward (e.g. "../../repo", "../sibling") would be
+	// rejected by the chroot-based filesystem loader with "chroot boundary
+	// crossed". Resolve them to absolute paths now, using the process's
+	// working directory, which is exactly what `git clone` does.
+	cleaned := filepath.Clean(path)
+	if !filepath.IsAbs(path) &&
+		(cleaned == ".." || strings.HasPrefix(cleaned, ".."+string(filepath.Separator))) {
```
pjbgf: @AriehSchneier I'm under the impression this needs to be handled higher up the stack e.g. Clone/Open.

AriehSchneier: I have moved it to `CloneOptions.Validate()` which fixes it in the `Clone` case (which is where I was hitting the error), but what if someone calls `transport.NewEndpoint()` directly it will no longer resolve.

AriehSchneier:
@pjbgf any chance you can review this again, save me rebasing it any more...

(PS the windows fail would probably be fixed by #1999 or #2001)

plumbing/transport/transport.go
pjbgf: Overall, transport should enforce abs paths across the board so let's go ahead with the Clone fix for now.

options.go
```diff
@@ -136,6 +138,20 @@ func (o *CloneOptions) Validate() error {
 		return ErrMissingURL
 	}
 
+	// Resolve ..-relative local paths to absolute paths before they reach
+	// the transport layer. The billy chroot-based FilesystemLoader rejects
+	// any path whose clean form starts with ".." with "chroot boundary
+	// crossed". Real git resolves such paths against the working directory
+	// at the command level, before any transport code runs.
+	if giturl.IsLocalEndpoint(o.URL) && !filepath.IsAbs(o.URL) {
```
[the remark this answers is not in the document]
AriehSchneier: Updated PR title/description

options.go
```diff
@@ -136,6 +138,20 @@ func (o *CloneOptions) Validate() error {
...
+		cleaned := filepath.Clean(o.URL)
+		if cleaned == ".." || strings.HasPrefix(cleaned, ".."+string(filepath.Separator)) {
+			if abs, err := filepath.Abs(o.URL); err == nil {
+				o.URL = abs
+			}
```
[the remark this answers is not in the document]
AriehSchneier: Done

pjbgf (approved):

PR #1900, opened 2026-03-14
Add reflog support

Soph:
Add support for reading and writing Git reflogs across the storage layer.

- `plumbing/format/reflog` - Encoder/decoder for the reflog file format. Handles the `<old> <new> <name> <<email>> <timestamp> <tz>\t<message>\n` line format, with proper error reporting for malformed entries and message normalization matching Git's behavior
- `plumbing/storer` - New `ReflogStorer` interface (`Reflog`, `AppendReflog`, `DeleteReflog`), kept separate from `storage.Storer` so reflog support is opt-in.
- `storage/filesystem` - `ReflogStorage` backed by `.git/logs/` files.
- `storage/memory` - In-memory `ReflogStorer` implementation.
- `storage/transactional` - Transactional reflog with proper isolation - tracks appends and deletes independently and flushes on `Commit()`. Conditionally created only when both base and temporal storers support reflogs.

Design decisions

- `ReflogStorer` is not part of `storage.Storer` - callers opt in via type assertion, so existing code is unaffected.
- `Reflog()` returns entries oldest-first (file order), matching the on-disk format.
- Message normalization on encode (collapse whitespace, strip newlines) matches Git's behavior and prevents format corruption.
- Empty messages omit the TAB separator, matching Git.

Test plan

- Unit tests for encode/decode round-trips, edge cases (empty messages, long lines, malformed timestamps)
- Transactional storage test suite covering isolation, commit, and concurrent base/temporal operations
- Verified output matches git reflog on real repositories

### DIFF
--- a/plumbing/format/reflog/reflog.go
+++ b/plumbing/format/reflog/reflog.go
@@ -36,32 +36,52 @@ type Entry struct {
 	Message string
 }
 
+// Decoder reads reflog entries from a reader one at a time.
+type Decoder struct {
+	r *bufio.Reader
+}
+
+// NewDecoder creates a Decoder that reads reflog entries from r.
+func NewDecoder(r io.Reader) *Decoder {
+	return &Decoder{r: bufio.NewReader(r)}
+}
+
+// Next returns the next reflog entry. It returns io.EOF when there are no more entries.
+func (d *Decoder) Next() (*Entry, error) {
+	for {
+		line, err := d.r.ReadBytes('\n')
+		if err != nil && err != io.EOF {
+			return nil, err
+		}
+
+		line = bytes.TrimSuffix(line, []byte{'\n'})
+		if len(line) != 0 {
+			return decodeLine(line)
+		}
+
+		if err == io.EOF {
+			return nil, io.EOF
+		}
+	}
+}
+
 // Decode reads all reflog entries from the reader.
 // Entries are returned in file order (oldest first).
 func Decode(r io.Reader) ([]*Entry, error) {
 	if r == nil {
 		return nil, fmt.Errorf("reader is nil")
 	}
+	d := NewDecoder(r)
 	var entries []*Entry
-	reader := bufio.NewReader(r)
 	for {
-		line, err := reader.ReadBytes('\n')
-		if err != nil && err != io.EOF {
-			return entries, err
-		}
-
-		line = bytes.TrimSuffix(line, []byte{'\n'})
-		if len(line) != 0 {
-			e, decodeErr := decodeLine(line)
-			if decodeErr != nil {
-				return entries, decodeErr
-			}
-			entries = append(entries, e)
-		}
-
+		e, err := d.Next()
 		if err == io.EOF {
 			return entries, nil
 		}
+		if err != nil {
+			return entries, err
+		}
+		entries = append(entries, e)
 	}
 }
 
--- a/plumbing/format/reflog/reflog_test.go
+++ b/plumbing/format/reflog/reflog_test.go
@@ -2,6 +2,7 @@ package reflog
 
 import (
 	"bytes"
+	"io"
 	"strings"
 	"testing"
 	"time"
@@ -33,6 +34,28 @@ func TestDecode(t *testing.T) {
 	assert.Equal(t, "Another Author", entries[1].Committer.Name)
 	assert.Equal(t, "another@example.com", entries[1].Committer.Email)
 	assert.Equal(t, "commit: Second commit", entries[1].Message)
+}
+
+func TestDecoderNext(t *testing.T) {
+	t.Parallel()
+
+	input := "0000000000000000000000000000000000000000 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa Author Name <author@example.com> 1234567890 +0000\tcommit (initial): Initial commit\n" +
+		"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb Another Author <another@example.com> 1234567891 +0100\tcommit: Second commit\n"
+
+	d := NewDecoder(strings.NewReader(input))
+
+	e, err := d.Next()
+	require.NoError(t, err)
+	assert.Equal(t, "Author Name", e.Committer.Name)
+	assert.Equal(t, "commit (initial): Initial commit", e.Message)
+
+	e, err = d.Next()
+	require.NoError(t, err)
+	assert.Equal(t, "Another Author", e.Committer.Name)
+	assert.Equal(t, "commit: Second commit", e.Message)
+
+	_, err = d.Next()
+	assert.ErrorIs(t, err, io.EOF)
 }
 
 func TestDecodeNilReader(t *testing.T) {


plumbing/format/reflog/doc.go
```diff
@@ -0,0 +1,11 @@
+// Package reflog implements encoding and decoding of git reflog entries.
+//
+// Git stores reflog entries in .git/logs/<ref> files, one entry per line.
+// Each line has the format:
+//
+//	<old-hash> <new-hash> <name> <<email>> <unix-timestamp> <timezone>\t<message>\n
```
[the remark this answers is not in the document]
Soph: this is correct, the `<>` are both placeholders for values but `email` has actual `<>` wrapping too in a reflog. See for example `.git/logs/HEAD`

plumbing/format/reflog/reflog.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+
+		if err == io.EOF {
+			return entries, nil
+		}
+	}
+}
+
+// decodeLine parses a single reflog line.
+// Format: <old-hash> <new-hash> <name> <<email>> <unix-timestamp> <timezone>\t<message>
```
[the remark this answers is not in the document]
Soph: this is correct, the `<>` are both placeholders for values but `email` has actual `<>` wrapping too in a reflog. See for example `.git/logs/HEAD`

plumbing/format/reflog/reflog_test.go
```diff
@@ -0,0 +1,223 @@
...
+	t.Parallel()
+
+	message := strings.Repeat("a", 70*1024)
+	line := "0000000000000000000000000000000000000000 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa Author <a@b.com> 1234567890 +0000\t" + message + "\n"
+
+	entries, err := Decode(strings.NewReader(line))
+	require.NoError(t, err)
+	require.Len(t, entries, 1)
+	assert.Equal(t, message, entries[0].Message)
+}
```
pjbgf, line 295: Let's add a Fuzz test for it as well.

Something like the one [for worktree](https://github.com/go-git/go-git/blob/5b5182acfe97427c3245f53e71afc114e79e4fc6/x/plumbing/worktree/worktree_test.go#L1244), whereby we add some valid and invalid data to the fuzzing corpus.

plumbing/format/reflog/doc.go
pjbgf, line 6: For clarity:
```suggestion
//	<old-hash> <new-hash> <committer> <<email>> <unix-timestamp> <timezone>\t<message>\n
```

plumbing/format/reflog/reflog.go
```diff
@@ -0,0 +1,191 @@
...
+	// OldHash is the hash the reference pointed to before the change.
+	OldHash plumbing.Hash
+	// NewHash is the hash the reference points to after the change.
+	NewHash plumbing.Hash
+	// Name is the name of the person who made the change.
+	Name string
+	// Email is the email of the person who made the change.
+	Email string
+	// When is the timestamp of the change.
+	When time.Time
```
pjbgf: Align with the approach in `Commit`, encapsulating the fields `Name`, `Email` and `When` into a single object:

```suggestion
	// Committer holds the signature for the entry, including name, email and when it was created.
	Committer object.Signature
```

plumbing/format/reflog/reflog.go
```diff
@@ -0,0 +1,191 @@
...
+	Message string
+}
+
+// Decode reads all reflog entries from the reader.
+// Entries are returned in file order (oldest first).
+func Decode(r io.Reader) ([]*Entry, error) {
+	var entries []*Entry
+	reader := bufio.NewReader(r)
+	for {
+		line, err := reader.ReadBytes('\n')
```
pjbgf: I suspect that if `r` is `nil`, this will panic. Please add a test to that effect and a `nil` check if that's the case.

plumbing/format/reflog/reflog_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+func TestDecodeInvalidNewHash(t *testing.T) {
+	t.Parallel()
+
+	line := "0000000000000000000000000000000000000000 not-a-valid-hash Author <a@b.com> 1234567890 +0000\tcommit: test\n"
+	_, err := Decode(strings.NewReader(line))
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "invalid new hash")
+}
+
```
pjbgf, line 259: Please add a test whereby both commits are 64 characters to represent `SHA256`.

plumbing/format/reflog/reflog.go
```diff
@@ -0,0 +1,191 @@
...
+func normalizeMessage(msg string) string {
+	msg = strings.ReplaceAll(msg, "\n", " ")
+	msg = strings.ReplaceAll(msg, "\r", " ")
+	fields := strings.Fields(msg)
+	return strings.Join(fields, " ")
+}
+
+// Encode writes a single reflog entry to the writer.
+func Encode(w io.Writer, e *Entry) error {
+	_, offset := e.When.Zone()
```
pjbgf: Please add `nil` checks for `w` and `e`.

plumbing/format/reflog/reflog_test.go
Soph: added.

plumbing/format/reflog/doc.go
Soph: changed

plumbing/format/reflog/reflog.go
Soph: Couldn't use `object.Signature` because that would have created a dependency cycle: `reflog -> object -> storer -> reflog` but I think this is now fine too.

plumbing/format/reflog/reflog.go
Soph: added.

plumbing/format/reflog/reflog_test.go
Soph: added.

plumbing/format/reflog/reflog.go
Soph: added

plumbing/format/reflog/reflog.go
pjbgf: As per off-line chat, let's remove this - the idea was to rely on existing patterns. Let's keep as is instead.

plumbing/format/reflog/reflog.go
```diff
@@ -0,0 +1,208 @@
...
+	NewHash plumbing.Hash
+	// Committer holds the signature for the entry, including name, email and when it was created.
+	Committer Signature
+	// Message describes the action that caused the change (e.g. "commit: Add feature").
+	Message string
+}
+
+// Decode reads all reflog entries from the reader.
+// Entries are returned in file order (oldest first).
+func Decode(r io.Reader) ([]*Entry, error) {
```
onee-only: Maybe we can make it return an iterator? Some people might not want to load the whole entries in their memory.

plumbing/format/reflog/reflog.go
```diff
@@ -0,0 +1,208 @@
...
+
+// normalizeMessage normalizes a reflog message the same way Git does:
+// collapse consecutive whitespace to a single space, strip leading/trailing
+// whitespace, and remove newlines.
+func normalizeMessage(msg string) string {
+	msg = strings.ReplaceAll(msg, "\n", " ")
+	msg = strings.ReplaceAll(msg, "\r", " ")
+	fields := strings.Fields(msg)
+	return strings.Join(fields, " ")
+}
```
onee-only, line 193: Would you kindly share a reference of this implementation?

Soph: Good point. Added a comment with the source.

plumbing/format/reflog/reflog.go
Soph: Added a iterator and kept the method to return all.

plumbing/format/reflog/reflog.go
```diff
@@ -0,0 +1,183 @@
...
+	e.OldHash = plumbing.NewHash(string(line[:spaceIdx]))
+	line = line[spaceIdx+1:]
+
+	// Parse new hash (up to next space)
+	spaceIdx = bytes.IndexByte(line, ' ')
+	if spaceIdx == -1 {
+		return nil, fmt.Errorf("expected space after new hash")
+	}
+	e.NewHash = plumbing.NewHash(string(line[:spaceIdx]))
+	line = line[spaceIdx+1:]
```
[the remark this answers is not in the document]
pjbgf: This would be a valid feedback if not preceded by `plumbing.IsHash(newHashStr)`.

pjbgf (approved):

Issue #1719, opened 2025-11-11
Pull removes untracked files in shallow repository


dvachaiev-hdx:
### Bug Description

Pull command removes all untracked files from the repo that was cloned with `Depth: 1` even if they are present in `.gitignore` file. 
Repos cloned with whole history aren't affected. 

### go-git Version

v5.16.3

### Steps to Reproduce

1. Clone the repo: 
```
git.PlainCloneContext(ctx, path, false, &git.CloneOptions{
		URL: url,
		Depth:    1,
		Auth:     &auth,
	})
```
2. Create any untracked file inside the repo
3. Run `Pull`:
```
wt.PullContext(ctx, &git.PullOptions{
		Auth: &auth,
		Force:    true,
	})
```
Result: file from step 2 doesn't exist anymore.

Issue #633, opened 2022-12-13
ignored files are deleted after a checkout


robertkowalski:
- Example Repo with executable go code: https://github.com/robertkowalski/testgogit-runner
 - Used code: https://github.com/robertkowalski/testgogit-runner/blob/master/main.go
 - Cloned Repo that leads to the error: https://github.com/robertkowalski/testgogit

**Bug Report** When we have a repository where the .gitignore on main/master ignores a file, and this file was added with `git add --force` to a branch, e.g. a release branch, the file will be deleted after checking out the branch.

**Expected Behaviour**: Similar to using git on the command line, the file should be there

run example with: `go run main.go && cd test`

<img width="1309" alt="Screenshot 2022-12-13 at 15 40 49" src="https://user-images.githubusercontent.com/298166/207363178-a1809e76-ea8a-486e-81a1-6a9f5fc1e695.png">

robertkowalski:
this only happens when the files are in the gitignore on the default branch

LuBingtan:
Got same issue. any updates or workaround for this?

LuBingtan:
After read this part of code, I guess it's been fixed in the latest version. Let me take a test, will give a feedback after that.
https://github.com/go-git/go-git/blob/1e2b0d67ae859dc3c81fc4455290287235395d09/worktree.go#L383

LuBingtan:
> After reading this part of code, I guess it's been fixed in the latest version. Let me take a test, will give a feedback after that.
> 
> https://github.com/go-git/go-git/blob/1e2b0d67ae859dc3c81fc4455290287235395d09/worktree.go#L383

It's been fixed in the latest version, this issue could be closed.

alexgeorgousis:
> In cases where ignored files are not added, the files will still be deleted after checkout.
> 
> Here's how to reproduce it:
> 
> 1. clone repro repo: https://github.com/frozenbonito/go-git-repro
> 2. create ignored file
>    ```
>    $ touch ignored.txt
>    ```
> 3. run `main.go`
>    ```
>    $ go run main.go
>    ```
> 4. `ignored.txt` is deleted

Ignored files shouldn't be able to be added (i.e. staged), right? Because they're ignored.

I'm experiencing the same problem reported in this issue with files that aren't in the `.gitignore` file but that I've explicitly added to the `Excludes` array of the repo `Worktree` via `wt.Excludes = append(wt.Excludes, gitignore.ParsePattern(fileIWannaIgnore, nil))`

[...]

Issue #365, opened 2021-08-20
Why does `Reset` delete files that Unstaged? Command line Git is not deleted.


yueqiumao:
Why does `Reset` delete files that Unstaged? Command line Git is not deleted.
w.Reset(&git.ResetOptions{
    Mode: git.HardReset,
})

breathbath:
+1, it should be at least a configurable option

kodawah:
see https://github.com/go-git/go-git/issues/1031

kodawah:
^

AriehSchneier:
Fixed in #1903

Issue #1031, opened 2024-02-20
Add Keep to ResetOptions


kkrime:
When running:

```err = wt.Reset(&git.ResetOptions{ Mode: git.HardReset  })```

Untracked files are over written.

 when running the git equivalent
```git reset HEAD --hard```

untracked files are left alone. 

Also, when running:
 ``` wt.Checkout(&git.CheckoutOptions{Branch: plumbing.ReferenceName("refs/heads/main"), Force: true})```
 Untracked files are over written.

 when running the git equivalent
```git checkout main -f```

untracked files are left alone. 

I think that the behaviour should be the same, as untracked files are not apart of the repo and this is causing me issues with my work.

pjbgf:
#### git checkout

As per [official documentation](https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt---force) the `-f` / `--force`:
> This is used to throw away local changes and any untracked files or directories that are in the way.

Which should be the same behaviour as you get with [Force](https://github.com/go-git/go-git/blob/7959a42552a99b2e2df21a6aacafc97b2b5c7457/options.go#L357-L359). To keep the untracked files, use [Keep](https://github.com/go-git/go-git/blob/7959a42552a99b2e2df21a6aacafc97b2b5c7457/options.go#L360-L364) instead.


#### git reset

We could explore adding a `Keep` option for `Reset` so they are largely aligned.

pjbgf:
Renaming the issue to align with the gap we have in `go-git`.

fawick:
> #### git checkout
> As per [official documentation](https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt---force) the `-f` / `--force`:
> 
> > This is used to throw away local changes and any untracked files or directories that are in the way.
> 
> Which should be the same behaviour as you get with [Force](https://github.com/go-git/go-git/blob/7959a42552a99b2e2df21a6aacafc97b2b5c7457/options.go#L357-L359). 

I would argue that the suffix "that are in the way" is neglected in go-git. as all untracked files are discarded.

Instead, `git checkout -f` would modify untracked files if (and only if) they are also in the tree of the target commit. All other untracked files are left alone.

kkrime:
I agree with @fawick 

That aside I would argue everyone (or very very close to everyone) that uses `go-git` is coming from `git` and therefore expecting the same experience and these inconsistancies get in the way.

[...]

PR #1903, opened 2026-03-17
git: worktree, don't delete untracked files on HardReset / Force checkout

### LINKED ISSUE #796: git.Worktree.Checkout with Force = true: behavior change in v5.7.0 (ignored files deleted)
Our team has noticed a behavior change when upgrading from go-git v5.6.1 to v5.7.0 regarding the usage of `Checkout()` and the `Force: true` option.

In v5.6.1 and previous versions, if the current branch had any ignored files (.gitignore) and we used the `Force: true` option when checking out another branch, the ignored files were left in place.

In v5.7.0, using `Force: true` option, files that are untracked AND ignored are all deleted from the filesystem after branch switch with `Checkout()`.

We have fixed the issue by using the option `Keep: true` instead of using the `Force` option, and now our ignored files are left alone on filesystem after switch of branches.

We did not know if we were incorrectly using `Force` to accomplish this before and a bug had been fixed in v5.7.0, or if this is not intended and a new bug has been introduced? If someone could confirm?

Example:
(Assume the current worktree directory has an ignored file that exists, and we are on a branch different than "test-branch")
```go
resultErr := w.Checkout(&git.CheckoutOptions{
    Branch: plumbing.NewBranchReferenceName("test-branch"),
    Force:   true})
```
in v5.6.1 and previous, the ignored file will be left alone, in v5.7.0 it will be deleted.

AriehSchneier:
## Summary

`Checkout` with `Force: true` (and `Reset` in `HardReset` mode) was deleting untracked files from the worktree. Real `git checkout -f` and `git reset --hard` never remove untracked files -- only `git clean -f` does.

Fixes #796, #1719, #365, and #633 (HardReset / Force checkout path).

### Root cause

`resetWorktree` computed a reverse diff between the worktree filesystem and the index (`diffStagingWithWorktree(reverse=true)`). A `Delete` action in this diff means "file exists on disk but is not in the index" -- covering both:

1. A tracked file deleted from the new target tree -- should be removed.
2. An untracked file -- must not be removed.

Previously all `Delete` actions were processed, so untracked files were deleted.

### Fix -- matching real git

Real `git reset --hard` / `git checkout -f` operates in three phases on tracked-file sets only:

1. **Tree-to-tree diff** (`fromTree -> toTree`): delete files tracked in `fromTree` but removed in `toTree`. Untracked files are invisible to this diff.

2. **Index-to-worktree**: write files in the new index but absent or stale on disk. `Delete` actions (file on disk, absent from the index node tree) are always truly untracked -- `SkipWorktree` entries are marked `skip=true` in the merkletrie and invisible to this diff, never surfaced as Delete actions here.

3. **SkipWorktree cleanup**: iterate all index entries directly; for any entry with `SkipWorktree=true` that still exists on disk, remove it. `diffStagingWithWorktree` never surfaces these as Delete actions (see phase 2). Untracked files in the same directories are untouched.

This PR introduces `diffTrees` and a new `resetWorktreeToTree` implementing this three-phase approach for `HardReset`. The old `resetWorktree` (used by `MergeReset`) is unchanged.

### Also fixed: `Index.SkipUnless`

`Index.SkipUnless` only ever set `SkipWorktree=true` for newly-excluded files but never cleared it for files moving back into the active set. Fixed by also setting `SkipWorktree=false` for included entries.

### Tests added

- `TestCheckoutForceUntracked` -- `Checkout{Force: true}` must not delete an untracked file.
- `TestPullAfterShallowClone_UntrackedFilesPreserved` -- untracked file must survive a subsequent `Pull`.
- `TestCheckoutForceSparseUntrackedPreserved` -- switching `SparseCheckoutDirectories` with `Force: true` must (a) remove tracked files outside the new sparse set from disk and (b) preserve untracked files inside those directories.

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -349,6 +349,18 @@ func (w *Worktree) Reset(opts *ResetOptions) error {
 		}
 	}
 
+	// For HardReset, capture the current HEAD tree before resetting HEAD.
+	// resetWorktreeToTree will diff prevTree→t and apply only those changes
+	// to the worktree. Since the diff is tree-to-tree, untracked files are
+	// invisible and are never deleted — matching real git reset --hard.
+	var prevTree *object.Tree
+	if opts.Mode == HardReset {
+		prevTree, err = w.headTree()
+		if err != nil {
+			return err
+		}
+	}
+
 	if err := w.setHEADCommit(opts.Commit); err != nil {
 		return err
 	}
@@ -367,7 +379,7 @@ func (w *Worktree) Reset(opts *ResetOptions) error {
 	}
 
 	if opts.Mode == HardReset {
-		if err := w.resetWorktree(t, opts.Files); err != nil {
+		if err := w.resetWorktreeToTree(prevTree, t, opts.Files); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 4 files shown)


onee-only:
Possibly related: https://github.com/go-git/go-git/issues/365, https://github.com/go-git/go-git/issues/1031, https://github.com/go-git/go-git/issues/633

AriehSchneier:
> #365, #1031, #633

Added #365 and #633 as fixed, but for #1031 the untracked-files-deleted bug described in the body is fixed, but the actual feature request (Keep mode on ResetOptions) is a separate enhancement not addressed here

worktree.go
```diff
@@ -501,6 +513,109 @@ func inFiles(files map[string]struct{}, v string) bool {
...
+		if errors.Is(err, plumbing.ErrReferenceNotFound) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	c, err := w.r.CommitObject(head.Hash())
+	if err != nil {
+		return nil, err
+	}
+	return c.Tree()
```
onee-only:
```suggestion
	return w.r.getTreeFromCommitHash(head.Hash())
```
Could be simplified to this.

worktree_test.go
```diff
@@ -344,6 +344,70 @@ func (s *WorktreeSuite) TestPullAfterShallowClone() {
...
+
+	// The untracked file must survive the pull (mirrors real git behaviour).
+	_, statErr := os.Stat(untrackedPath)
+	s.NoError(statErr, "untracked file was removed by pull (no Force) on a shallow clone")
+
+	// Push another commit and test with Force: true.
+	_ = CommitNewFile(s.T(), remote, "File5")
+
+	// Re-create the untracked file in case it was removed.
+	s.Require().NoError(os.WriteFile(untrackedPath, []byte("do not delete me"), 0o644))
```
pjbgf: Seems redundant - the untracked should fail on `L390`.

```suggestion
```

worktree_test.go
```diff
@@ -344,6 +344,70 @@ func (s *WorktreeSuite) TestPullAfterShallowClone() {
...
+	err = w.Pull(&PullOptions{
+		RemoteName:    DefaultRemoteName,
+		SingleBranch:  true,
+		ReferenceName: plumbing.NewBranchReferenceName("master"),
+	})
+	s.Require().NoError(err)
+
+	// The untracked file must survive the pull (mirrors real git behaviour).
+	_, statErr := os.Stat(untrackedPath)
+	s.NoError(statErr, "untracked file was removed by pull (no Force) on a shallow clone")
```
pjbgf:
```suggestion
	s.Require().NoError(statErr, "untracked file was removed by pull (no Force) on a shallow clone")
```

worktree.go
```diff
@@ -501,6 +513,109 @@ func inFiles(files map[string]struct{}, v string) bool {
...
+	}
+	b := newIndexBuilder(idx)
+
+	for _, ch := range worktreeChanges {
+		a, err := ch.Action()
+		if err != nil {
+			return err
+		}
+		if a == merkletrie.Delete {
+			// File is on disk but not in new index = untracked. Never delete.
```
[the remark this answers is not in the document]
pjbgf: Let's add a test case that caters for `untracked` file on a dir that is being removed from `SparseCheckoutDirectories`. Upstream carries on with the operation but shows a warning:

```
$ git sparse-checkout init
$ git sparse-checkout set plumbing
$ touch plumbing/abc2
$ git sparse-checkout set _examples
$ git s
On branch main
Your branch is behind 'origin/main' by 4 commits, and can be fast-forwarded.
  (use "git pull" to update your local branch)

You are in a sparse checkout with 18% of tracked files present.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	plumbing/abc2
```

AriehSchneier:
@pjbgf Updated and rebased (the new test found an existing bug which I also fixed)

PR #1904, opened 2026-03-17
git: worktree, add KeepReset mode (git reset --keep)

### LINKED ISSUE #1031: Add Keep to ResetOptions

AriehSchneier:
## Summary

Adds `KeepReset` as a new `ResetMode`, implementing the behaviour of `git reset --keep`.
Fixes #1031.

### What `git reset --keep` does

`git reset --keep <commit>` resets the index and worktree to `<commit>`, but **aborts** if any file that would be changed by the reset has local staged or unstaged modifications. Untracked files are never deleted.

### Implementation

1. **Pre-flight check** (`checkKeepResetConflicts`): diffs `fromTree -> toTree` to find files that will change; returns `ErrLocalChanges` (wrapping the conflicting path) if any of those files have staged or unstaged modifications. `SkipWorktree` (sparse-excluded) files are not present in the worktree and are never counted as conflicting.

2. **Apply**: if the check passes, the reset proceeds identically to `HardReset` -- index updated via `resetIndex`, worktree updated via `resetWorktreeToTree` (tree-to-tree, untracked files untouched, SkipWorktree files cleaned up by phase 3 of `resetWorktreeToTree`).

### Tests added

- `TestResetKeep` -- basic success: no local conflicts, worktree updated.
- `TestResetKeepConflict` -- aborted when a to-be-changed file has local modifications; HEAD does not move.
- `TestResetKeepUntrackedPreserved` -- untracked files survive the reset.
- `TestResetKeepSparselyUntracked` -- switching `SparseDirs` via `KeepReset` removes tracked files from the old sparse dir, populates the new sparse dir, and preserves untracked files in the removed dir.
- `TestResetKeepConflictNotOnSparseExcluded` -- `SkipWorktree` files that change between commits do not falsely trigger the conflict check.

closes #1031

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -41,6 +41,9 @@ var (
 	ErrSubmoduleNotFound = errors.New("submodule not found")
 	// ErrUnstagedChanges is returned when the worktree has unstaged changes.
 	ErrUnstagedChanges = errors.New("worktree contains unstaged changes")
+	// ErrLocalChanges is returned when a KeepReset is attempted but a file
+	// that would be changed by the reset has local modifications.
+	ErrLocalChanges = errors.New("worktree contains local changes that would be overwritten by reset")
 	// ErrGitModulesSymlink is returned when .gitmodules is a symlink.
 	ErrGitModulesSymlink = errors.New(gitmodulesFile + " is a symlink")
 	// ErrNonFastForwardUpdate is returned when a non-fast-forward update is attempted.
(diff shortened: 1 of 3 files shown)


worktree.go
```diff
@@ -525,6 +534,113 @@ func (w *Worktree) headTree() (*object.Tree, error) {
 	return c.Tree()
 }
 
+// checkKeepResetConflicts implements the safety check for KeepReset
+// (git reset --keep): it aborts if any file that would be modified by the
+// reset has local staged or unstaged modifications in the worktree.
+//
+// Three categories of "touched" paths are checked:
```
pjbgf: @AriehSchneier the comment refers to three categories but only have two below. Please update the comment.

PR #1906, opened 2026-03-18
git: add Repository.Close(), use it in tests to fix Windows file handle leak

AriehSchneier:
Adds `Repository.Close()` which releases any resources held by the underlying storage (a no-op for memory-backed repositories).

This fixes intermittent Windows CI failures where tests using submodule repositories backed by filesystem storage left pack index (`.idx`) file handles open. On Windows these prevent `TempDir RemoveAll` cleanup:

```
TempDir RemoveAll cleanup: remove ...pack-....idx: The process cannot access the file because it is being used by another process.
```

The root cause is that `ObjectStorage` holds `LazyIndex` instances that keep the `.idx` file open via `sharedFile`. These are only released via `ObjectStorage.Close()`. Since `Repository` had no `Close()` method, callers had no standard way to release these resources.

This PR:
- Adds `Repository.Close() error` to `repository.go`, delegating to `Storer` if it implements `io.Closer`
- Updates all test call sites where `Submodule.Repository()` returns a filesystem-backed repo to `defer r.Close()`:
  - `TestSubmodulesFetchDepth` (submodule_test.go)
  - `TestUpdate` (submodule_test.go)
  - `TestUpdateWithInitAndUpdate` (submodule_test.go)
  - `TestPlainCloneWithShallowSubmodules` (repository_test.go)
  - `TestCheckoutRelativePathSubmoduleInitialized` (worktree_test.go)


repository.go
```diff
@@ -659,6 +659,16 @@ func checkTargetDirIsEmpty(path string) (empty bool, err error) {
 	return false, nil
 }
 
+// Close releases any open resources held by the repository. It must be called
+// when the repository is no longer needed. It is safe to call Close on a
+// repository backed by memory storage, where it is a no-op.
+func (r *Repository) Close() error {
```
pjbgf, line 664: @AriehSchneier thanks for looking into this.

I understand the reasoning behind this, but I'm slightly concerned on changing the UX on the repository handling - this would require all use of `r.Open` to be followed by a `defer r.Close()`.

If there is no other way, we can go ahead with it, but I'd like to make sure we explored all options before then. Any additional ideas on that front?

AriehSchneier: AI had suggested that each case where we had the issue, where we had a repo (`r`) that we add:
```
	if c, ok := r.Storer.(io.Closer); ok {
		return c.Close()
	}
```
Which I thought exposes the implementation details (although `Storer` is already exposed). I suggested adding the `r.Close()` to make it more idiomatic and clearer that it needs to be called.

The problem is I think you already need to call close everywhere, we just don't, and hence we are getting leaked file handles on Windows.

AriehSchneier:
Another instance:
https://github.com/go-git/go-git/actions/runs/23794964041/job/69339762456

repository.go
pjbgf: That's fair, let's go ahead with it and document it as part of https://github.com/go-git/go-git/issues/910.

pjbgf:
@AriehSchneier please rebase and we will go ahead with the merge.

AriehSchneier:
> @AriehSchneier please rebase and we will go ahead with the merge.

@pjbgf Done

PR #1913, opened 2026-03-20
Update docs and introduce AI Policy

pjbgf:
This documentation update include:
- Update `COMPATIBILITY.md` to include features introduced in `v6`.
- Add `AI_POLICY.md` and `AGENTS.md` to define the project's AI Policy as well as clarify to agents how to work with the repository.
- Update `CONTRIBUTING.md` to add a link to the AI policy.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     branches:
       - main
+    paths-ignore:
+      - '**.md'
+      - 'rfcs/**'
 
 permissions: {}
 
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -4,7 +4,13 @@ on:
     branches:
     - releases/v5.x
     - main
+    paths-ignore:
+    - '**.md'
+    - 'rfcs/**'
   pull_request:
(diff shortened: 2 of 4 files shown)


AI_POLICY.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Disclosure
+
+If AI assisted in producing any part of your contribution, disclose it in the
+PR description. Add an `Assisted-by:` trailer to each affected commit:
+
+```
+Assisted-by: GitHub Copilot
+Assisted-by: Claude Sonnet 4.6
+Assisted-by: ChatGPT o3
```
cedric-appdirect, line 51: This might require clarification. As the example mix model and tools. In practice, I have really started to use AI for PRs since February and I rely on a subagent workflow with Claude code which means I have the main agent using opus while I have a coder, a reviewer and a commit contribution agent that use mostly sonnet but also haiku. Anyway, I think specifying tools is the logical and right things to do here.

pjbgf: @cedric-appdirect good point. The example aims to highlight that you can specify multiple. However, I'd only refer to the agents that actively changed the given commit (or proposed changes to it). I'll add a comment to that effect.

PR #1915, opened 2026-03-21
config: add Core.HooksPath field for core.hooksPath setting

### LINKED ISSUE #1429: Missing keys in config.Config
I would like to set the config option [`core.hooksPath`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corehooksPath) via go-git. Unfortunately the `config.Config` struct does not seem to support so.

I could probably go about modifying `config.Config.Raw`, or by calling `git config` with Exec, or by parsing and modifying `repo.git/config` directly. However, IMO this should be something in go-git.

mvanhorn:
## Summary

Adds `HooksPath` field to `Config.Core` struct, supporting the git `core.hooksPath` configuration option. This allows setting a custom hooks directory via the go-git API instead of requiring raw config manipulation.

## Why this matters

Users who need to set `core.hooksPath` currently have to use `config.Raw.SetOption()` as a workaround (see #1429). Other Core fields like `Worktree`, `CommentChar`, and `AutoCRLF` are already supported as struct fields - `HooksPath` follows the same pattern.

## Changes

- `config/config.go`: Added `HooksPath string` field to `Core` struct, `hooksPathKey` constant, unmarshal via `s.Options.Get()`, and conditional marshal
- `config/config_test.go`: Added `hooksPath` to both unmarshal and marshal test fixtures with assertions

## Testing

`go test ./config/... -run "TestUnmarshal|TestMarshal"` - all pass.

Fixes #1429

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -84,6 +84,8 @@ type Config struct {
 		// against local file, 0644 will be used as the value of its mode. The original
 		// value of mode is left unchanged in the index.
 		FileMode bool
+		// HooksPath is the path to look for hooks instead of $GIT_DIR/hooks.
+		HooksPath string
 	}
 
 	User struct {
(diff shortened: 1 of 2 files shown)


mvanhorn:
Thanks for the review and merge, @pjbgf! Glad to contribute the core.hooksPath support.

PR #1114, opened 2024-06-14
plumbing: gitignore, Fix loading of ignored .gitignore files.

Achilleshiel:
Stop loading .gitignore files from ignored directories, as this can cause files to be included that should not be.

See https://git-scm.com/docs/gitignore#_pattern_format

PR #1925, opened 2026-03-25
plumbing: gitignore, Fix negation in subdirectory with path-based parent ignore. Fixes #877

### LINKED ISSUE #877: bug: gitignore behavior mismatch to the git in specific scenarios
If the root `.gitignore` has `abc`, and the `abc/.ignore` has `!def`

Then, the ignore status for `abc/def` between git and go-git mismatch:
- In go-git: the `abc/def` is not ignored
- In git: the `abc/def` is ignored

Test Folder: [ignore-check.zip](https://github.com/go-git/go-git/files/12890595/ignore-check.zip)

Test Code:

```go
import (
	"testing"
	
	"github.com/go-git/go-billy/v5/osfs"
	"github.com/go-git/go-git/v5/plumbing/format/gitignore"
	"github.com/stretchr/testify/assert"
)

func TestGitIgnore(t *testing.T) {
	fs := osfs.New("/tmp/ignore-check")

	ps, err := gitignore.ReadPatterns(fs, nil)
	assert.NoError(t, err)

	assert.Equal(t, 2, len(ps))

	matcher := gitignore.NewMatcher(ps)

	assert.True(t, matcher.Match([]string{"abc"}, true), "abc should be ignored")
	assert.True(t, matcher.Match([]string{"abc/def"}, true), "abc/def should be ignored") # this test failed
	assert.False(t, matcher.Match([]string{"abc/ghi"}, true), "abc/ghi should not be ignored")
	assert.False(t, matcher.Match([]string{"def"}, true), "def (root) should not be ignored")
}

```

queelius:
## Summary

- Fix `ReadPatterns` to pass parent ignore patterns down through recursive calls, so path-based patterns from ancestor `.gitignore` files (e.g., `a/b` in root `.gitignore`) correctly prevent descending into ignored directories
- Without this fix, `ReadPatterns` would enter the ignored directory, load its `.gitignore` negation patterns (e.g., `!file`), and incorrectly un-ignore files that git treats as ignored
- Add test case `TestDir_ReadPatterns_NestedPathIgnoreWithNegation` that reproduces the bug and verifies the fix

## Problem

When a root `.gitignore` uses a **path pattern** to ignore a nested directory (e.g., `a/b`), and that directory's own `.gitignore` contains a negation (e.g., `!file`), go-git incorrectly treats `a/b/file` as not-ignored. Real git treats it as ignored, per the [gitignore spec](https://git-scm.com/docs/gitignore#_pattern_format): "It is not possible to re-include a file if a parent directory of that file is excluded."

The prior fix in #1114 handled the case where the ignore pattern matched at the **same directory level** (e.g., simple name `abc` in root `.gitignore`). However, path-based patterns that match at a deeper level (e.g., `a/b`) were not visible during the recursive call into `a/`, so the check on directory `b` would miss the ignore and recurse into it.

## Solution

Introduce an internal `readPatterns` helper that threads accumulated parent patterns down through the recursion. The public `ReadPatterns` signature is unchanged. The directory-skip check now uses the combined parent + local patterns, so path-based patterns from any ancestor are visible at every recursion level.

## Test plan

- [x] New test `TestDir_ReadPatterns_NestedPathIgnoreWithNegation` fails without fix, passes with fix
- [x] All existing `gitignore` tests pass: `go test ./plumbing/format/gitignore/... -v`
- [x] `go vet` and `gofmt` clean

Fixes #877

### DIFF
--- a/plumbing/format/gitignore/dir.go
+++ b/plumbing/format/gitignore/dir.go
@@ -52,6 +52,10 @@ func readIgnoreFile(fs billy.Filesystem, path []string, ignoreFile string) (ps [
 // recursively traversing through the directory structure. The result is in
 // the ascending order of priority (last higher).
 func ReadPatterns(fs billy.Filesystem, path []string) (ps []Pattern, err error) {
+	return readPatterns(fs, path, nil)
+}
+
+func readPatterns(fs billy.Filesystem, path []string, parentPatterns []Pattern) (ps []Pattern, err error) {
 	ps, _ = readIgnoreFile(fs, path, infoExcludeFile)
 
 	subps, _ := readIgnoreFile(fs, path, gitignoreFile)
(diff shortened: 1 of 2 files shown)


queelius:
Closing this to keep your queue clean. Happy to reopen if useful, or feel free to pick up any of it. Thanks!

PR #1940, opened 2026-03-31
Implement Git-compatible wildmatch for gitignore patterns

### LINKED ISSUE #877: bug: gitignore behavior mismatch to the git in specific scenarios

AriehSchneier:
This PR implements a Git-compatible wildmatch function to fix gitignore pattern matching compatibility issues.

## Summary

Replaces `filepath.Match` with a custom wildmatch implementation that provides full Git compatibility for gitignore pattern matching.

### Commit 1: Add comprehensive gitignore compatibility test suite
- Complete test suite based on Git's official tests (t3070-wildmatch.sh and t0008-ignores.sh)
- 84 tests covering all edge cases and compatibility scenarios
- Tests for POSIX character classes, bracket expressions, double star patterns, etc.

### Commit 2: Implement Git-compatible wildmatch for gitignore patterns
- Custom wildmatch() function with star backtracking
- Full POSIX character class support ([:digit:], [:upper:], [:alpha:], [:xdigit:], etc.)
- Proper bracket expression negation with [!...] and [^...]
- Escaped characters in patterns and bracket expressions
- Complex bracket ranges and edge cases (e.g., [\\]], [[-\\]], [--A])
- Malformed character class detection
- Patterns with ** embedded in segments (e.g., foo**/bar, **/bar**)

## Test Results

**84/84 tests passing (100%)**
- TestGitCanonicalSuite ✅
- TestWildmatchCanonicalSuite ✅
- TestCompleteWildmatchSuite ✅
- TestPatternSuite ✅
- TestMatcherSuite ✅

## Implementation

Based directly on Git's wildmatch.c implementation from the Git source code, ensuring full compatibility with Git's gitignore pattern matching behavior.

Fixes #877

---


plumbing/format/gitignore/complete_wildmatch_test.go
```diff
@@ -0,0 +1,450 @@
...
+		{"", "foo", false, "empty pattern vs text"},
+		{"**/t[o]", "foo/bar/baz/to", true, "double star with bracket"},
+	}
+
+	for _, tt := range tests {
+		s.testMatch(tt.pattern, tt.text, tt.expected, tt.desc)
+	}
+}
+
+func (s *CompleteWildmatchSuite) TestCharacterClassTests() {
```
pjbgf:
```
  func (s *ConformanceSuite) TestPOSIXClassesAreASCIIOnly() {
      // Single high-bit bytes that classify as letters/digits/etc. in Unicode
      // but not in ASCII.
      upperA_grave := "\xc0"  // 'À' as a single Latin-1 byte
      superscript2 := "\xb2"  // '²' as a single Latin-1 byte (Unicode digit)
      nbsp := "\xa0"          // NBSP — Unicode space, not ASCII space

      cases := []struct {
          pattern, text string
          want          bool
          desc          string
      }{
          {`[[:alpha:]]`, upperA_grave, false, "0xC0 not ASCII alpha"},
          {`[[:upper:]]`, upperA_grave, false, "0xC0 not ASCII upper"},
          {`[[:digit:]]`, superscript2, false, "0xB2 not ASCII digit"},
          {`[[:space:]]`, nbsp, false, "0xA0 not ASCII space"},
          {`[[:print:]]`, upperA_grave, false, "0xC0 not ASCII printable"},
          {`[[:punct:]]`, upperA_grave, false, "0xC0 not ASCII punct"},
          // Sanity: ASCII counterparts still match.
          {`[[:alpha:]]`, "A", true, "ASCII A matches alpha"},
          {`[[:digit:]]`, "5", true, "ASCII 5 matches digit"},
          {`[[:space:]]`, " ", true, "ASCII space matches space"},
      }
      for _, c := range cases {
          s.testMatch(c.pattern, c.text, c.want, c.desc)
      }
  }
```

pjbgf:
@AriehSchneier please note that commits assisted by AI must have a `Assisted-By:` trailer as opposed to `Co-Authored-By:` as per [AI Policy](https://github.com/go-git/go-git/blob/main/AI_POLICY.md#disclosure). Please update that and rebase the changes. 

I have some additional changes to add to this PR which I will at some point today.

pjbgf (approved):
@AriehSchneier thanks for working on this, a much needed addition to go-git. 🙇 🙇 🙇 

I made a few additional changes while I was checking for upstream conformance:
- Consolidated the different upstream-test cases on a single file.
- Trailing `\` fix in wildmatch  a pattern ending in `\` falls through to backtracking and fails to match, matching `wildmatch.c`'s escape.
- POSIX classes are now ASCII-only. `matchCharClass` replaces `unicode.Is*` with byte-range checks, mirroring `sane-ctype.h` so high-bit bytes never satisfy `[[:alpha:]]/[[:digit:]]/etc`.
- Shared `parsePosixClass` helper. The `[:name:]` header parser used to be duplicated in `findBracketEnd` and `matchBracket`, now this is a helper function.
- When `!testing.Short()` several test cases will go on and also verify if the matching expectations aligns with upstream Git.
- Added checks for Git `v2.11.0` (due to compatibility checks) to ensure we accommodate for cases whereby upstream behaviour changed over time.
- Added checks to enforce that MacOS and Windows behaviour aligns with Git with `core.ignorecase=false`, at least until we add support for `core.ignorecase=true`.
- Added `FuzzMatch` to fuzz test this logic going forwards.

PR #1947, opened 2026-04-01
Faster send pack for remote

stiak:
I noticed while using `remote.PushContext` that there was a significant delay between getting the remote refs and the push, even when pushing a relatively small branch (~12 files changed).

It looks like the current `revlist.Objects()` function walks a lot more than it needs to.  This PR introduces a new `objectWalk` struct with functions that aim to improve the algorithm, though native git has other solutions which are likely faster again.

This wires the `Objects` and `ObjectsWithRef` functions to use the new path.  It also removes the `ObjectsWithStorageForIgnores` path, which I didn't understand.  Happy to bring it back if you can explain it to me.

### DIFF
--- a/plumbing/revlist/revlist.go
+++ b/plumbing/revlist/revlist.go
@@ -7,14 +7,27 @@ import (
 	"github.com/go-git/go-git/v6/plumbing/storer"
 )
 
+// objectWalker can be implemented by storers that provide a specialized
+// revlist object walk for a wants/haves query.
+type objectWalker interface {
+	RevListObjects(wants, haves []plumbing.Hash) ([]plumbing.Hash, error)
+}
+
 // Objects computes object hashes reachable from wants while excluding
-// commits reachable from haves. It expands haves first to establish commit
-// boundaries, then walks wants in the same object store.
+// commits reachable from haves.
+//
+// If s implements objectWalker, its RevListObjects method is used.
+// Otherwise, Objects expands haves first to establish commit boundaries,
+// then walks wants in the same object store.
 func Objects(
 	s storer.EncodedObjectStorer,
 	wants,
 	haves []plumbing.Hash,
 ) ([]plumbing.Hash, error) {
+	if walker, ok := s.(objectWalker); ok {
+		return walker.RevListObjects(wants, haves)
+	}
+
 	w, err := newObjectWalk(s)
 	if err != nil {
 		return nil, err
--- a/plumbing/revlist/revlist_test.go
+++ b/plumbing/revlist/revlist_test.go
@@ -25,6 +25,21 @@ type RevListSuite struct {
 	Storer storer.EncodedObjectStorer
 }
 
+type delegatedObjectStorer struct {
+	storer.EncodedObjectStorer
+	called bool
+	wants  []plumbing.Hash
+	haves  []plumbing.Hash
+	result []plumbing.Hash
+}
+
+func (s *delegatedObjectStorer) RevListObjects(wants, haves []plumbing.Hash) ([]plumbing.Hash, error) {
+	s.called = true
+	s.wants = append([]plumbing.Hash(nil), wants...)
+	s.haves = append([]plumbing.Hash(nil), haves...)
+	return append([]plumbing.Hash(nil), s.result...), nil
+}
+
 func TestRevListSuite(t *testing.T) {
 	t.Parallel()
 	suite.Run(t, new(RevListSuite))
@@ -82,6 +97,25 @@ func (s *RevListSuite) TestRevListObjects_Submodules() {
 	for _, h := range revList {
 		s.False(submodules[h.String()])
 	}
+}
+
+func (s *RevListSuite) TestRevListObjects_DelegatesToObjectWalker() {
+	want := plumbing.NewHash(initialCommit)
+	have := plumbing.NewHash(secondCommit)
+	expected := []plumbing.Hash{plumbing.NewHash(someCommit)}
+
+	sto := &delegatedObjectStorer{
+		EncodedObjectStorer: memory.NewStorage(),
+		result:              expected,
+	}
+
+	got, err := Objects(sto, []plumbing.Hash{want}, []plumbing.Hash{have})
+	s.Require().NoError(err)
+
+	s.True(sto.called)
+	s.Equal([]plumbing.Hash{want}, sto.wants)
+	s.Equal([]plumbing.Hash{have}, sto.haves)
+	s.Equal(expected, got)
 }
 
 // ---


remote.go
```diff
@@ -215,7 +215,7 @@ func (r *Remote) sendPack(ctx context.Context, conn transport.Connection, remote
 			hashesToPush, err = revlist.ObjectsWithStorageForIgnores(
 				r.s, localStorer, objects, haves)
 		} else {
-			hashesToPush, err = revlist.Objects(r.s, objects, haves)
+			hashesToPush, err = revlist.ObjectsDiff(r.s, objects, haves)
 		}
```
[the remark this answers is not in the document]
stiak: I can see commit b416a3f02ce80c776680f0e21bf0257caf7a3a95 introduced the possibility of resolving ref specs to arbitrary objects.  I'm not sure that's actually valid git behaviour though.

pjbgf:
This implementation of `ObjectsDiff` seems to fix https://github.com/go-git/go-git/issues/1953.

plumbing/revlist/diff.go
```diff
@@ -0,0 +1,192 @@
...
+)
+
+// ObjectsDiff computes the set of object hashes reachable from localCommits but not
+// reachable from remoteCommits. It walks local and remote commits simultaneously
+// in committer-time order, only advancing the remote side far enough to
+// determine which local commits are new.
+func ObjectsDiff(
+	s storer.EncodedObjectStorer,
+	localCommits []plumbing.Hash,
+	remoteCommits []plumbing.Hash,
```
pjbgf: Let's rename this to align with the caller as that is the primary use case. 

```suggestion
	wants []plumbing.Hash,
	haves []plumbing.Hash,
```

plumbing/revlist/diff.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/go-git/go-git/v6/plumbing/filemode"
+	"github.com/go-git/go-git/v6/plumbing/object"
+	"github.com/go-git/go-git/v6/plumbing/storer"
+)
+
+// ObjectsDiff computes the set of object hashes reachable from localCommits but not
+// reachable from remoteCommits. It walks local and remote commits simultaneously
+// in committer-time order, only advancing the remote side far enough to
+// determine which local commits are new.
+func ObjectsDiff(
```
pjbgf: `ObjectsDiff` seems to provide the same functionality as `Objects`. I think we are better off amending the latter directly so that existing users benefit from it and we don't need to add a new func to the public API.

If we need to add a specific change in behaviour, that could be done via the functional options pattern.

plumbing/revlist/diff.go
```diff
@@ -0,0 +1,192 @@
...
+	s storer.EncodedObjectStorer,
+	localCommits []plumbing.Hash,
+	remoteCommits []plumbing.Hash,
+) ([]plumbing.Hash, error) {
+	var localQueue []*object.Commit
+	var remoteQueue []*object.Commit
+	localSeen := make(map[plumbing.Hash]bool)
+	remoteSeen := make(map[plumbing.Hash]bool)
```
pjbgf: nit: let's use `map[plumbing.Hash]struct{}` instead

plumbing/revlist/diff.go
```diff
@@ -0,0 +1,192 @@
...
+	return result, nil
+}
+
+// collectChangedTreeObjects walks newTree, comparing entry hashes against
+// oldTree. Subtrees with matching hashes are skipped entirely. Only new or
+// modified tree and blob hashes are added to result.
+func collectChangedTreeObjects(
+	s storer.EncodedObjectStorer,
+	newTree, oldTree *object.Tree,
+	seen map[plumbing.Hash]bool,
```
pjbgf: Same across the use of `bool` in maps.

```suggestion
	seen map[plumbing.Hash]struct{},
```

plumbing/revlist/diff_test.go
```diff
@@ -0,0 +1,493 @@
...
+		TreeHash:     treeHash,
+		ParentHashes: parents,
+	}
+	obj := h.s.NewEncodedObject()
+	obj.SetType(plumbing.CommitObject)
+	require.NoError(h.t, c.Encode(obj))
+	hash, err := h.s.SetEncodedObject(obj)
+	require.NoError(h.t, err)
+	return hash
+}
```
pjbgf: Instead, we can lean on fixtures, they have different repos we can test from and they generally would have the amount of objects they contain via `.ObjectsCount()`.

You can create a storage with:
```golang
    dotgit := fixtures.Basic().One().DotGit()
	storage := filesystem.NewStorage(dotgit, cache.NewObjectLRUDefault())
```

For a list of the fixtures, you can take a look here: https://github.com/go-git/go-git-fixtures/blob/0622d7459a67/fixtures.go#L24

Maybe something like [this](https://github.com/go-git/go-git-fixtures/blob/0622d7459a67/fixtures.go#L81-L86)  could be helpful given the amount of objects it has.

plumbing/revlist/diff_test.go
```diff
@@ -0,0 +1,493 @@
...
+		Hash: mkBlob("branch-change"),
+	}
+	branchTree := mkTree(entries)
+	branchTip := mkCommit(branchTree, mainTip)
+
+	return s, mainTip, branchTip
+}
+
+func BenchmarkObjects(b *testing.B) {
+	s, mainTip, branchTip := buildBenchRepo(b, 200, 100)
```
pjbgf: We could use the src-d from fixtures as that has over 2k objects.

plumbing/revlist/diff_test.go
```diff
@@ -0,0 +1,493 @@
...
+	require.False(t, gotSet[r2], "must not include remote main tip")
+	require.False(t, gotSet[r4], "must not include remote release tip")
+	require.False(t, gotSet[b2], "must not include remote main blob")
+	require.False(t, gotSet[b5], "must not include remote release blob")
+
+	// Must not include shared ancestors.
+	require.False(t, gotSet[r1], "must not include shared ancestor r1")
+	require.False(t, gotSet[r3], "must not include shared ancestor r3")
+}
+
```
squishykid: I was playing with this with claude, and this test seems to capture the current issue which is causing objects to be missing from the diff:

```
func TestObjectsDiff_RevertedSubtreeMatchesRoot(t *testing.T) {
	// Regression test: a subtree S in the root commit reappears in a
	// descendant commit C after being modified in an intermediate commit P.
	// Because C is walked first (newest-first), S goes into `seen` with
	// only the entries that changed relative to P. When the root commit is
	// walked (full walk, no parent), the old buggy code skipped S entirely
	// via the `seen[S]` early return, so entries that were unchanged in
	// C-vs-P (but only reachable through the root's copy of S) were lost.
	//
	// Root: {dir/ → S = {a → A, b → B}}
	// P (parent Root): {dir/ → S_P = {a → A, b → B_new}}
	// C (parent P):    {dir/ → S = {a → A, b → B}, x → X}
	//
	// Walk C vs P: S ≠ S_P → walk S: b changed (collect B), a unchanged (skip).
	// Walk P vs Root: S_P ≠ S → walk S_P: b changed (collect B_new), a unchanged (skip).
	// Walk Root (no parent): dir/ → S is in seen → BUG: skip! A never collected.
	t.Parallel()
	h := newMemHelper(t)

	blobA := h.blob("a-content")
	blobB := h.blob("b-content")
	blobBNew := h.blob("b-new-content")
	blobX := h.blob("x-content")

	// Root's subtree S.
	subS := h.tree([]object.TreeEntry{
		{Name: "a.txt", Mode: filemode.Regular, Hash: blobA},
		{Name: "b.txt", Mode: filemode.Regular, Hash: blobB},
	})
	treeRoot := h.tree([]object.TreeEntry{
		{Name: "dir", Mode: filemode.Dir, Hash: subS},
	})
	cRoot := h.commit(treeRoot)

	// P modifies b inside dir/.
	subSP := h.tree([]object.TreeEntry{
		{Name: "a.txt", Mode: filemode.Regular, Hash: blobA},
		{Name: "b.txt", Mode: filemode.Regular, Hash: blobBNew},
	})
	treeP := h.tree([]object.TreeEntry{
		{Name: "dir", Mode: filemode.Dir, Hash: subSP},
	})
	cP := h.commit(treeP, cRoot)

	// C reverts b back to original AND adds a new top-level file.
	// dir/ → S (same hash as Root's dir/).
	treeC := h.tree([]object.TreeEntry{
		{Name: "dir", Mode: filemode.Dir, Hash: subS},
		{Name: "x.txt", Mode: filemode.Regular, Hash: blobX},
	})
	cC := h.commit(treeC, cP)

	assertObjectsDiffSubset(t, h.s, []plumbing.Hash{cC}, nil)

	got, err := ObjectsDiff(h.s, []plumbing.Hash{cC}, nil)
	require.NoError(t, err)

	gotSet := make(map[plumbing.Hash]bool, len(got))
	for _, gh := range got {
		gotSet[gh] = true
	}

	require.True(t, gotSet[blobA], "must include blobA (reachable from root, inside shared subtree S)")
	require.True(t, gotSet[blobB], "must include blobB (reverted in C, changed vs P)")
	require.True(t, gotSet[blobBNew], "must include blobBNew (introduced in P)")
	require.True(t, gotSet[blobX], "must include blobX (new in C)")
}
```

stiak:
@pjbgf I think I have the correctness issues solved now.  

- The missing objects were due to an early return, meaning we didn't evaluate the diff again if we'd seen the tree before.  Early return was removed.
- The extra objects were due to objects being re-introduced by the commits between haves and wants.  Solved by capturing "seen" objects at the tip before starting the commit walk.


I'll look into the benchmarking and broader replacement next.

plumbing/revlist/diff_test.go
stiak: Thanks, I've included this in the tests.

plumbing/revlist/diff_test.go
stiak: I've moved most tests over to fixtures, including the benchmarks.  Many of the tests were duplicated from revlist_test.go.  For the new scenarios (`TestObjectsDiff_ReintroducedBlobInHaves`, `TestObjectsDiff_SeenSubtreeDifferentParent`), how do I find an appropriate fixture to use?

plumbing/revlist/diff.go
```diff
@@ -0,0 +1,192 @@
...
+
+// ObjectsDiff computes the set of object hashes reachable from localCommits but not
+// reachable from remoteCommits. It walks local and remote commits simultaneously
+// in committer-time order, only advancing the remote side far enough to
+// determine which local commits are new.
+func ObjectsDiff(
+	s storer.EncodedObjectStorer,
+	localCommits []plumbing.Hash,
+	remoteCommits []plumbing.Hash,
+) ([]plumbing.Hash, error) {
```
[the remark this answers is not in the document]
stiak: [git rev-list](https://git-scm.com/docs/git-rev-list) only supports commit-ish references.  I'm not sure why the code currently supports arbitrary objects or what the expected behaviour would be in that case.

plumbing/revlist/diff_test.go
pjbgf: I'd generate start from a fixture (e.g. `fixtures.Basic()`), then create the different subtrees from that point.

stiak: Done, it's building trees/commits on top the basic fixture and re-using the existing objects.

stiak:
Latest timing output, also including the changes from https://github.com/squishykid/go-git/tree/faster-send-pack as `ObjectsTopo`.  This PR implementation is `Objects` and current main branch is `ObjectsOld`:

```
     local:  refs/tags/v5.13.0 (94bd4af1deb15a64e90c6287eaf9e9f09b192a1f)
     remote: refs/tags/v5.12.0 (302dddeda962e4bb3477a8e4080bc6f5a253e2bb)
     git rev-list refs/tags/v5.13.0 --objects --count: 15607
     revlist.ObjectsOld ref1
       objects: 15607  time: 557.486125ms  alloc: 10 MB  total-alloc: 169 MB  gc-cycles: 38
     revlist.Objects ref1
       objects: 15607  time: 391.296958ms  alloc: 9 MB  total-alloc: 135 MB  gc-cycles: 15
     revlist.ObjectsTopo ref1
       objects: 15607  time: 566.089792ms  alloc: 9 MB  total-alloc: 197 MB  gc-cycles: 15

     git rev-list refs/tags/v5.13.0..refs/tags/v5.12.0 --objects --count: 738
     revlist.ObjectsOld ref1..ref2
       objects: 738  time: 401.096708ms  alloc: 9 MB  total-alloc: 145 MB  gc-cycles: 14
     revlist.Objects ref1..ref2
       objects: 738  time: 82.081542ms  alloc: 9 MB  total-alloc: 30 MB  gc-cycles: 4
     revlist.ObjectsTopo ref1..ref2
       objects: 738  time: 52.568834ms  alloc: 9 MB  total-alloc: 18 MB  gc-cycles: 3
```

pjbgf:
I've tested this and ended up adding a few commits to better align behaviour with upstream.

Existing problems:
- Support for shallow clones (fixes #1953).
- Objects should error when it encounters missing parents or missing parent trees - unless it encounters a shallow boundary (above).

Fixes to the previous commit:
- Remove assumptions on having previous commits based on a newer commit date.

Additionally, object storers are able to completely override the object walker by adding a func `RevListObjects(wants, haves []plumbing.Hash) ([]plumbing.Hash, error)`. In the future this could be exposed as its own public interface.

remote.go
```diff
@@ -206,17 +201,7 @@ func (r *Remote) sendPack(ctx context.Context, conn transport.Connection, remote
...
-			// faster to use a local storage layer to get the commits
-			// to ignore, when calculating the object revlist.
-			localStorer := filesystem.NewStorage(
-				osfs.New(o.RemoteURL, osfs.WithBoundOS()), cache.NewObjectLRUDefault())
-			hashesToPush, err = revlist.ObjectsWithStorageForIgnores(
-				r.s, localStorer, objects, haves)
-		} else {
-			hashesToPush, err = revlist.Objects(r.s, objects, haves)
-		}
+		hashesToPush, err = revlist.Objects(r.s, objects, haves)
```
pjbgf, line 204: For future reference: the performance gain on a local-specific approach wasn't worth the additional complexity.

plumbing/revlist/object_walk.go
```diff
@@ -0,0 +1,271 @@
...
+			w.seen[tag.Hash] = struct{}{}
+			w.result = append(w.result, tag.Hash)
+			c, err = object.GetCommit(w.s, tag.Target)
+			if err != nil {
+				return fmt.Errorf("getting commit for tag %s: %w", h, err)
+			}
+		}
+		w.wantsSeen[h] = struct{}{}
+		insertSorted(&w.wantsQueue, c)
+	}
```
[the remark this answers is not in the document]
pjbgf: This is no longer the case.

PR #1949, opened 2026-04-01
git: worktree, fix Add silently failing for absolute paths

### LINKED ISSUE #1460: `Worktree.Add()` silently fails for absolute paths

tmchow:
#### Description

`Worktree.Add()` silently fails when given an absolute path. No error is returned, but the file is not added to the index because the index stores relative paths.

The fix converts absolute paths to worktree-relative paths using `filepath.Rel()` before updating the index. If the path is outside the worktree root, an error is returned.

#### Link to tracking issue
Fixes #1460

#### Testing

Added `TestAddAbsolutePath` which:
1. Creates a repo with a real filesystem
2. Writes a file
3. Calls `w.Add()` with the absolute path
4. Verifies the index contains the relative path
5. Verifies status shows the file as staged

`go test ./...` and `go vet ./...` pass.

#### Documentation

N/A - behavior now matches `git add /absolute/path`.

### DIFF
--- a/worktree_status.go
+++ b/worktree_status.go
@@ -3,6 +3,7 @@ package git
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	"path"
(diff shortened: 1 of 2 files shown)


tmchow:
> @tmchow thanks for working on this. 🙇 

You bet!

PR #1951, opened 2026-04-02
git: repository, fix DeleteBranch failing with full ref name

### LINKED ISSUE #1189: Can't delete a branch that for sure exists
Repro
``` go
repoDir = "pathtoexistingrepo"
r, err := git.PlainOpen(repoDir)
if err != nil {
	return err
}

branchToDelete = "test"
branches, err := r.Branches()
err = branches.ForEach(func(b *plumbing.Reference) error {
	if b.Name().String() == branchToDelete {
		fmt.Println("Found the branch")
		err = r.DeleteBranch(b.Name().String())
		if err != nil {
			return err
		}
	}
	return nil
})
if err != nil {
    fmt.Println(err)
}
```

Output:
```
Found the branch // from the ForEach
Branch not found // err check at the end
```

Tried with b.Name().Short() as well, same result

tmchow:
#### Description

`DeleteBranch("refs/heads/test")` returns `ErrBranchNotFound` even when the branch exists. This happens because `Branches()` returns full reference names but `DeleteBranch()` looks up `cfg.Branches` by short name only.

The fix strips the `refs/heads/` prefix before the config lookup, so both `DeleteBranch("test")` and `DeleteBranch("refs/heads/test")` work. This matches `git branch -d` which accepts both forms.

#### Link to tracking issue
Fixes #1189

#### Testing

Added `TestDeleteBranchFullRefName` which creates a branch and deletes it using the full `refs/heads/` name.

`go test ./...` and `go vet ./...` pass.

#### Documentation

N/A

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -860,6 +860,8 @@ func (r *Repository) CreateBranch(c *config.Branch) error {
 
 // DeleteBranch delete a Branch from the repository and delete the config
 func (r *Repository) DeleteBranch(name string) error {
+	name = strings.TrimPrefix(name, "refs/heads/")
+
 	cfg, err := r.Config()
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


tmchow:
Thanks for the quick merge!

PR #1957, opened 2026-04-03
internal: revision, fix resolution for tags containing @ and /

### LINKED ISSUE #1612: Unable to resolve revision of valid tag "@name/subname@v1.0.1"
**This is a copy of issue #204 which is still present.**

<hr>

# Steps to reproduce

main.go

```go
package main

import (
	"github.com/go-git/go-git/v5"
	"github.com/go-git/go-git/v5/plumbing"
)

func main() {
	repo, err := git.PlainOpen(".")
	if err != nil {
		panic(err)
	}

	tagName := "@name/subname@v1.0.1"
	_, err = repo.ResolveRevision(plumbing.Revision(tagName))
	if err != nil {
		panic(err)
	}
}
```

```
$ git tag                                                                                                     
@name/subname@v1.0.1
$ git rev-list -n 1 @name/subname@v1.0.1
c93521b3316ee00c2d8f190b1420ed0adde79f6c          
$ ./main
panic: Revision invalid : reference must be defined once at the beginning                    
```

Workaround

As a workaround, I resolve the tag by reference hash instead of its name.

```go
package main

import (
	"github.com/go-git/go-git/v5"
	"github.com/go-git/go-git/v5/plumbing"
)

func main() {
	repo, err := git.PlainOpen(".")
	if err != nil {
		panic(err)
	}

	tagName := "@name/subname@v1.0.1"
	reference, err := repo.Tag(tagName)
	if err != nil {
		panic(err)
	}

	_, err = repo.ResolveRevision(plumbing.Revision(reference.Hash().String()))
	if err != nil {
		panic(err)
	}
}
```

tmchow:
The revision scanner unconditionally tokenized `@` as an `at` operator token, which broke tag names like `@scope/package@v1.0.1`. The parser then interpreted the leading `@` as a reflog operator and failed to resolve the reference.

Now `@` is only emitted as an operator token when followed by `{` (reflog syntax like `@{1}`, `@{upstream}`). Otherwise it's treated as part of the reference name. The parser maps a standalone `@` ref to HEAD to preserve the git shorthand (`@` = `HEAD`).

### Changes

- `internal/revision/scanner.go`: peek at the next character after `@`. Emit `at` token only for `@{`, otherwise emit `word` so the `@` is included in the ref name.
- `internal/revision/parser.go`: map a parsed ref of exactly `"@"` to `Ref("HEAD")` to preserve the `@` shorthand.
- Tests added for `@scope/package@v1.0.1`, `@scope/package@v1.0.1~1`, and scanner tokenization of `@{1}` vs `@scope`.

Fixes #1612

### DIFF
--- a/internal/revision/parser.go
+++ b/internal/revision/parser.go
@@ -565,6 +565,10 @@ func (p *Parser) parseRef() (Revisioner, error) {
 
 		if endOfRef {
 			p.unscan()
+			if buf == "@" {
+				return Ref("HEAD"), nil
+			}
+
 			return Ref(buf), nil
 		}
 
--- a/internal/revision/parser_test.go
+++ b/internal/revision/parser_test.go
@@ -177,6 +177,13 @@ func (s *ParserSuite) TestParseWithValidExpression() {
 			CaretPath{1},
 			CaretPath{1},
 		},
+		"@scope/package@v1.0.1": []Revisioner{
+			Ref("@scope/package@v1.0.1"),
+		},
+		"@scope/package@v1.0.1~1": []Revisioner{
+			Ref("@scope/package@v1.0.1"),
+			TildePath{1},
+		},
 	}
 
 	for d, expected := range datas {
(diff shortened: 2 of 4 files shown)


pjbgf (approved):
@tmchow Thanks for working on this. 🙇

PR #1976, opened 2026-04-09
Fix Windows file handle leak in file transport

AriehSchneier:
## Summary

Fixes intermittent Windows CI failures in `TestFetchWithFilters` caused by file handle leaks in the file transport.

## Problem

The file transport's `Loader.Load()` creates a `storage.Storer` instance when executing git operations, but the storage was never being closed. This left file handles open on pack index (`.idx`) files until garbage collection occurred.

On Windows, open file handles prevent file/directory deletion, causing test cleanup to fail with:
```
The process cannot access the file because it is being used by another process.
```

## Solution

Modified the `closeAll` function in `file.go` to properly close the loaded storage if it implements `io.Closer`. This ensures file handles are released when the connection is closed, rather than waiting for garbage collection.

## Test plan

- [x] Existing tests pass locally
- [x] Linting passes
- [x] Rebased onto latest main (after #1983 refactored file transport)
- [ ] Verify Windows CI no longer fails intermittently

### DIFF
--- a/plumbing/transport/file/file.go
+++ b/plumbing/transport/file/file.go
@@ -83,9 +83,22 @@ func (t *Transport) connect(ctx context.Context, req *transport.Request) (io.Rea
 	pr, pw := io.Pipe()
 	sr, sw := io.Pipe()
 
-	closeAll := func() error { _ = pw.Close(); return sr.Close() }
+	done := make(chan struct{})
+
+	closeAll := func() error {
+		_ = pw.Close()
+		closeErr := sr.Close()
+		<-done // Wait for server goroutine to finish
+		if closer, ok := st.(io.Closer); ok {
+			if err := closer.Close(); err != nil && closeErr == nil {
+				closeErr = err
+			}
+		}
+		return closeErr
+	}
 
 	go func() {
+		defer close(done)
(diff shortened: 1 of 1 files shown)


pjbgf:
@AriehSchneier I'm not sure this is still needed after the recent changes in `transport`. If so, this will need rebasing.

AriehSchneier:
Rebased onto latest main (commit 65857034). The file transport was refactored in #1983, so the fix has been adapted to work with the new `file.go` implementation. The same bug (not closing loaded storage) existed in the refactored code, so this fix is still needed.

AriehSchneier:
Updated the fix to properly synchronize with the server goroutine. The original fix had a race condition where the storage could be closed while the background goroutine was still using it. The fix now waits for the goroutine to finish before closing the storage.

AriehSchneier:
Examples of the Windows file handle leak error in CI:
- https://github.com/go-git/go-git/actions/runs/24375131581
- https://github.com/go-git/go-git/actions/runs/24350947979
- https://github.com/go-git/go-git/actions/runs/24384591183

Both show intermittent failures with "The process cannot access the file because it is being used by another process" on .idx files during temp directory cleanup.

pjbgf (approved):
@AriehSchneier thank you for fixing this issue and resolving the flaky tests. 🙇

PR #1979, opened 2026-04-10
SHA256: Support submodules in SHA256 repositories

pjbgf:
This PR now enables cloning of SHA-256 repositories with submodules.
Several other tests were updated to align with latest changes from go-git-fixtures, which are now `ObjectFormat`-aware.

Relates to #706.


plumbing/transport/http/upload_pack_test.go
```diff
@@ -23,14 +23,14 @@ type uploadPackSuite struct {
...
 	emptyFS := prepareRepo(s.T(), fixtures.ByTag("empty").One(), base, "empty.git")
 
 	s.Endpoint = httpEndpoint(addr, "basic.git")
 	s.EmptyEndpoint = httpEndpoint(addr, "empty.git")
 	s.NonExistentEndpoint = httpEndpoint(addr, "non-existent.git")
 
-	s.Storer = filesystem.NewStorage(basicFS, nil)
+	s.Storer = memory.NewStorage()
 	s.EmptyStorer = filesystem.NewStorage(emptyFS, nil)
 	s.NonExistentStorer = memory.NewStorage()
```
aymanbagabas: Is this intentional? Do we want different storer types?

plumbing/transport/ssh/pack_test.go
```diff
@@ -74,7 +74,7 @@ func (s *uploadPackSuite) SetupTest() {
 	s.Endpoint = env.Endpoint
 	s.EmptyEndpoint = env.EmptyEndpoint
 	s.NonExistentEndpoint = env.NonExistentEndpoint
-	s.Storer = env.Storer
+	s.Storer = memory.NewStorage()
 	s.EmptyStorer = env.EmptyStorer
 	s.NonExistentStorer = env.NonExistentStorer
```
aymanbagabas, line 79: Same here

pjbgf: As per off-line chat, that was to fix a bug in those tests. Let's handle that separately on its own PR.

PR #1981, opened 2026-04-12
Add `compatObjectFormat` repository support

Soph:
**Summary**
This adds support for Git repositories using `extensions.compatObjectFormat`, with bidirectional object ID mapping between the repository’s native object format and a configured compatibility format.

The implementation is modeled on upstream Git’s hash transition design and covers:
- config parsing and validation for `extensions.compatObjectFormat`
- native<->compat hash mapping
- object content translation for blobs, trees, commits, and tags
- persistence of compat mappings in memory and filesystem storages
- repository open/init/repack integration
- compat-aware fetch import
- compat-aware push export
- upstream parity tests and end-to-end interop coverage

**What’s Included**
- New `plumbing/compat` package for:
  - mapping storage
  - object translation
  - fetch import
  - push export
- Storage integration:
  - memory storage compat lookups and translation hooks
  - filesystem storage compat lookups, persistence, and compaction
- Repository integration:
  - compat-aware init/open options
  - compat mapping compaction during repack
- Transport integration:
  - compat fetch when the remote advertises the repository’s compat format
  - compat push when the remote expects the repository’s compat format
- Validation:
  - upstream Git parity tests for translated hashes and object bytes
  - signed object parity coverage
  - end-to-end repository interop matrix

**Behavior**
When `compatObjectFormat` is configured:
- the repository remains native in its configured `extensions.objectFormat`
- compat mappings are maintained between native and compat object IDs
- objects can be translated between the native and compat namespaces
- fetch can import compat-format objects into native storage
- push can export native objects in compat format when required by the remote

When compat is not configured:
- the feature is inert aside from a few nil/capability checks

**Config Validation**
This change enforces the core config invariants for repository extensions:
- extensions require `core.repositoryformatversion = 1`
- `compatObjectFormat` requires `extensions.objectFormat`
- `compatObjectFormat` must differ from `objectFormat`

**On-Disk Mapping Format**
Filesystem mapping support:
- always reads both:
  - legacy `objects/loose-object-idx`
  - `objects/object-map/map-*.map`
- writes legacy format by default
- supports opt-in `object-map` writes
- compacts mappings during repack
- uses atomic temp-file-and-rename writes for persisted mapping files

**Push / Fetch Scope**
Compat push and fetch are intentionally scoped to the repository’s configured native/compat pair:
- push translation activates when the remote expects `CompatObjectFormat`
- fetch import activates when the remote advertises `CompatObjectFormat`

This is not a general arbitrary-format conversion layer.

**Testing**
Coverage includes:
- unit tests for translation, mapping, import, export, and persistence
- upstream Git parity tests for:
  - object ID translation
  - translated object bytes
  - signed commits and tags
- repository interop matrix covering:
  - init/open/repack/reopen
  - incremental fetch
  - upstream `git gc`
  - alternates
  - push create / fast-forward / no-op / delete
  - push non-fast-forward rejection
  - force
  - force-with-lease
  - cross-format fetch import

**Known Limits**
- compat push/fetch only handle the repository’s configured native/compat pair
- object-map support is practical and tested, but upstream source/docs are still not fully converged on one persisted mapping format
- write+mapping update is not a single atomic transaction across the whole object ingest path; missing compat mappings are reconciled when needed
- broader lifecycle scenarios beyond the current matrix can still be added over time

**Implementation Notes**
To make compat/native-format interop correct, this change also removes a few SHA-1-specific assumptions in lower-level object storage paths:
- pack parser/scanner now propagate `RawObjectWriter.Close()` failures
- SHA-256 packed-object iteration uses the correct object ID width
- filesystem loose-object writes hash using the repository’s configured object format


plumbing/compat/translate.go
```diff
@@ -0,0 +1,334 @@
...
+// back into native format. This is used when importing objects from a compat
+// remote during fetch.
+func (t *Translator) TranslateCompatContent(objType plumbing.ObjectType, compatContent []byte) ([]byte, error) {
+	switch objType {
+	case plumbing.BlobObject:
+		return compatContent, nil
+	case plumbing.TreeObject:
+		return t.translateCompatTree(compatContent)
+	case plumbing.CommitObject:
+		return t.translateCompatTextObject(compatContent, []string{"tree", "parent"})
```
pjbgf: We need to also translate `mergetag` headers.

x/compat/fetch_import.go
```diff
@@ -0,0 +1,360 @@
...
+
+// SetEncodedObject translates a compat-format object into native storage and
+// returns the native hash once the object is fully imported.
+func (s *ImportStorer) SetEncodedObject(obj plumbing.EncodedObject) (plumbing.Hash, error) {
+	reader, err := obj.Reader()
+	if err != nil {
+		return plumbing.ZeroHash, err
+	}
+
+	content, err := io.ReadAll(reader)
```
pjbgf, line 67: Instead of `io.ReadAll` we should use the reader so that we don't load the entire contents of each object into memory.

x/compat/oidmap/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package oidmap stores bidirectional mappings between native and compat
```
pjbgf, line 1: Let's move the entire `plumbing/compat` to `x/compat`.

x/compat/oidmap/file_objectmap.go
```diff
@@ -0,0 +1,409 @@
+package oidmap
+
```
pjbgf, line 2: Add a link to upstream spec:
https://github.com/git/git/blob/94f057755b7941b321fd11fec1b2e3ca5313a4e0/Documentation/gitformat-loose.adoc

plumbing/compat/oidmap/file_objectmap.go
```diff
@@ -0,0 +1,409 @@
...
+
+	"github.com/go-git/go-git/v6/plumbing"
+	formatcfg "github.com/go-git/go-git/v6/plumbing/format/config"
+	gitbinary "github.com/go-git/go-git/v6/utils/binary"
+)
+
+const (
+	objectMapDirName  = "object-map"
+	mapFilePrefix     = "map-"
+	mapSnapshotPrefix = "map-zsnapshot-"
```
pjbgf: We may be better off not sharing the same prefix as `mapFilePrefix`.


```suggestion
	mapSnapshotPrefix = "zsnapshot-map"
```

Issue #1943, opened 2026-03-31
Worktreeconfig support for filesystem storage not included in 5.17.x


charithe:
### Bug Description

Hi. v5.17.0 includes the strict extensions check change (#1861) but, without the changes from #1877, opening a repo with go-git fails with `core.repositoryformatversion does not support extension: worktreeconfig`. We kept go-git pinned to v5.16.5 until the change made it to v5.17.x but v5.17.1 doesn't include it. Is that on purpose or just an oversight? Given that 5.17.1 includes a security fix, it'd be great if #1877 can be included in a 5.17.x release soon so that we can upgrade. 

Thanks.

### go-git Version

v5.17.1

pjbgf:
@charithe thanks for opening this issue. Please note that v5 never supported that feature, and incorrectly would open a repository without being aware of it - this is now blocked to align with Git upstream requirements.

The release line `v5` is maintenance only as we are working hard on getting `v6` released. My recommendation would be to start using `v6` and feedback on any issues you may encounter.

charithe:
@pjbgf thanks for the quick reply. If I am not mistaken, v6 doesn't have any tagged releases yet? That makes it tricky for us with security scans and so on. I understand that v5 is in maintenance mode. However, given that #1861 was added to v5 and it breaks things without the additional changes in #1877, would you consider backporting it until v6 is production-ready?

pjbgf:
@charithe we have recently cut the first tagged release for v6: [v6.0.0-alpha.1](https://github.com/go-git/go-git/releases/tag/v6.0.0-alpha.1). It contains a lot of bug fixes which are not in v5 - the current recommendation is to start using that. The only reason why we haven't cut a final release yet is that some API changes are still in flux.

As for backporting the support for `WorktreeConfig`, a recently proposed [change](https://github.com/go-git/go-git/blob/c0d5afbdceb5a64b44ea7c2becd63b3e5bfa9885/CONTRIBUTING.md#sustainability) in `CONTRIBUTING.md` aims to touch this subject:

> The majority of the work on go-git comes from individual contributors volunteering their own time. This limits the amount of capacity available for activities like backporting features to v5, triaging issues, and expanding test coverage.
>
> If your company relies on go-git, please consider contributing engineering hours to help sustain the project.

Backporting a feature requires support and bug fixes for the duration of the given release line, which diverts the community from implementing new features or improving existing ones.

charithe:
[...]

Issue #1939, opened 2026-03-30
Case-sensitivity bug in extension validation rejects worktreeConfig


migmartri:
### Bug Description

In v5.17.0 (#1861), repository_extensions.go lowercases extension names via strings.ToLower in the extensions() function, but extensionsValidForV0 uses mixed-case keys ("worktreeConfig", "partialClone", "preciousObjects"). Since Go map lookups are case-sensitive, the lowercased names never match the allowlist.

Additionally, worktreeConfig is missing from builtinExtensions, so it also fails for repositoryFormatVersion=1 repos via ErrUnknownExtension.

This breaks PlainOpen on any repository using git worktree (which sets extensions.worktreeConfig = true

```
opening repository: core.repositoryformatversion does not support extension: worktreeconfig
```

see https://github.com/chainloop-dev/chainloop/issues/2966

### go-git Version

v5.17.0

pjbgf:
Thanks for opening this issue. Unfortunately, there is already an existing issue that covers it - although newer - so I'll be closing this in favour of #1943.

The TL;DR is that the error is correct - go-git `v6` never supported that extension. This has recently been added to `v6`.

PR #1982, opened 2026-04-12
[v5] repository: fix extension validation for worktreeConfig repositories

cgrdavies:
## What

Fixes the v5.17.0 regression in which opening a repository that has `extensions.worktreeConfig = true` fails with:

```
core.repositoryformatversion does not support extension: worktreeconfig
```

…or, after the first case bug is fixed:

```
unknown extension: worktreeconfig
```

## Why

`extensions.worktreeConfig = true` is set automatically by modern git on `git sparse-checkout set` and `git sparse-checkout disable` (see `git-sparse-checkout(1)`). Any consumer of go-git v5 that opens a repository touched by those commands is currently broken on v5.17.x — this includes at least [`tea`](https://gitea.com/gitea/tea), [Google Antigravity](https://discuss.ai.google.dev/t/antigravity-agent-breaks-on-git-worktree-repos-with-extensions-worktreeconfig/137246), and CI environments that use `actions/checkout` (which [had to add an explicit `git config --unset`](https://github.com/actions/checkout/pull/1692) to work around similar tooling breakage).

## The two bugs

**1. Case mismatch in `extensionsValidForV0`.** `extensions()` normalises every extension name with `strings.ToLower` before lookup, but the map keys were stored in camelCase (`"worktreeConfig"`, `"partialClone"`, `"preciousObjects"`). Go map lookups are case-sensitive, so the V0 gate rejected every one of these extensions even though they were explicitly intended to be allowed.

**2. `worktreeconfig` missing from `builtinExtensions`.** After passing the V0 gate, `verifyExtensions` also checks every extension against `builtinExtensions`. This patch adds `worktreeconfig` so v5 will open the repository. Per-worktree config files (`.git/config.worktree`) are not read — that matches the behaviour the library had prior to v5.17.0 and avoids pulling in the larger storage changes from #1877.

## Scope

This is the **minimum viable backport** for v5. It is not a port of #1877 — that's a much larger change involving the `ExtensionChecker` interface and filesystem storage support, and the maintainer has noted v5 is in maintenance mode. This patch restores the pre-v5.17.0 behaviour of opening worktreeConfig-enabled repositories without breaking anything else.

## Testing

Three new table-driven cases added to `TestVerifyExtensions`:

- `repositoryformatversion=0: allows worktreeConfig (camelCase on disk)` — the primary regression case. Verified to fail on unpatched `v5.17.2` with the original error (`core.repositoryformatversion does not support extension: worktreeconfig`) and pass with this patch.
- `repositoryformatversion=1: allows worktreeConfig` — covers the v1 path through `builtinExtensions`.
- `repositoryformatversion=0: allows partialClone and preciousObjects` — covers the other two camelCase v0 extensions affected by bug #1. They still hit `unknown extension` at the builtin-extensions gate (pre-existing behaviour, out of scope), but the assertion confirms they no longer trip the v0 gate with the wrong error.

`go build ./...` and `go vet ./...` pass clean. Full `go test .` pack still green.

## Refs

- #1861 — introduced the regression
- #1939 — closed as duplicate, identified the case-sensitivity bug
- #1943 — open, reported by a consumer affected by this
- #1877 — full fix on `main` (v6), not backported

### DIFF
--- a/repository_extensions.go
+++ b/repository_extensions.go
@@ -40,16 +40,30 @@ var (
 		// noop-v1 does not change git’s behavior at all.
 		// It is useful only for testing format-1 compatibility.
 		"noop-v1": {},
+
+		// worktreeconfig signals that per-worktree config files
+		// (.git/config.worktree) may be present. go-git reads the
+		// shared config like upstream git would when the extension
+		// is absent; worktree-specific overrides are ignored but do
+		// not prevent the repository from being opened.
+		//
+		// Modern git enables this extension automatically on
+		// `git sparse-checkout set` and `git sparse-checkout disable`,
+		// so rejecting it would break common workflows.
+		"worktreeconfig": {},
 	}
 
 	// Some Git extensions were supported upstream before the introduction
 	// of repositoryformatversion. These are the only extensions that can be
 	// enabled while core.repositoryformatversion is unset or set to 0.
+	//
+	// Keys must be lowercase to match the output of extensions(), which
+	// normalises extension names with strings.ToLower.
 	extensionsValidForV0 = map[string]struct{}{
 		"noop":            {},
-		"partialClone":    {},
-		"preciousObjects": {},
-		"worktreeConfig":  {},
+		"partialclone":    {},
+		"preciousobjects": {},
+		"worktreeconfig":  {},
 	}
 )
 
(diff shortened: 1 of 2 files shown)


migmartri:
I am running into this issue, is there any way we could go ahead with the review? 

Thanks

pjbgf:
Thanks for proposing this PR. Unfortunately I'll need to close it as it tackles the wrong issue.
More info at: https://github.com/go-git/go-git/issues/1943#issuecomment-4232656963

PR #1985, opened 2026-04-12
plumbing: transport, add git-upload-archive support

aymanbagabas:
This PR adds support for the `git-upload-archive` service to the transport layer, allowing fetching of archives from remote git repositories over SSH, HTTP, and local file transports.

## Background

The `git-upload-archive` protocol command is used by `git archive --remote` to fetch archives from remote repositories without cloning them. This is useful for CI/CD pipelines, release automation, and tools that need specific files from a repo.

### Client-Side Types

**Archiver interface** - Sessions that implement this can perform archive operations:
```go
type Archiver interface {
    Archive(ctx context.Context, req *ArchiveRequest) (io.ReadCloser, error)
}
```

**ArchiveRequest** - Client archive request:
- `Args []string`: arguments sent as "argument <arg>" pkt-lines (format, prefix, tree-ish, paths)
- `Progress sideband.Progress`: optional progress callback for status messages

### Server-Side Types

**UploadArchiveRequest** - Server archive request configuration:
- `AllowUnreachable bool`: disables security restrictions allowing arbitrary refs

**UploadArchive function** - Server-side handler for the git-upload-archive protocol.

### Backend Support

**SSH/TCP/Unix transports** - Backend now dispatches to `UploadArchive`:
```go
case transport.UploadArchiveService:
    return transport.UploadArchive(...)
```

**HTTP** - Left for future protocol-v2 work (HTTP archive requires command-based protocol)

### Wire Protocol

Implements the git-upload-archive protocol:
- Client sends "argument <arg>" pkt-lines for format, prefix, tree-ish, etc.
- Server sends "ACK" or "NACK <reason>"
- Archive data streamed via sideband (channel 1 = data, channel 2 = progress)

### Supported Formats

Supported formats parsed from args:
- `tar`, `tar.gz`, `tgz`, `zip`
- `--prefix=path/` adds prefix to archive entries
- `--list` lists available archives
- Tree-ish (commit, tag, branch) specifies what to archive

## Testing

- `TestArchive_Tar`, `TestArchive_TarGz`, `TestArchive_Zip` (file transport)
- `TestGitTransport_Archive` (git daemon transport)
- `TestSSHTransport_Archive` (SSH transport)

All tests verify archive contents against git's output.

## Security

The `uploadArchive.allowUnreachable` config option controls whether arbitrary SHA-1 expressions are allowed. By default, only ref targets (e.g., `v1.0`, `main`) and `ref:path` sub-tree syntax are allowed.

## Follow-up Work

[...]

PR #1986, opened 2026-04-13
plumbing: transport, add git-upload-archive support

aymanbagabas:
This PR adds support for the `git-upload-archive` service to the transport layer, allowing fetching of archives from remote git repositories over SSH, HTTP, and local file transports.

## Background

The `git-upload-archive` protocol command is used by `git archive --remote` to fetch archives from remote repositories without cloning them. This is useful for CI/CD pipelines, release automation, and tools that need specific files from a repo.

### Client-Side Types

**Archiver interface** - Sessions that implement this can perform archive operations:
```go
type Archiver interface {
    Archive(ctx context.Context, req *ArchiveRequest) (io.ReadCloser, error)
}
```

**ArchiveRequest** - Client archive request:
- `Args []string`: arguments sent as "argument <arg>" pkt-lines (format, prefix, tree-ish, paths)
- `Progress sideband.Progress`: optional progress callback for status messages

### Server-Side Types

**UploadArchiveRequest** - Server archive request configuration:
- `AllowUnreachable bool`: disables security restrictions allowing arbitrary refs

**UploadArchive function** - Server-side handler for the git-upload-archive protocol.

### Backend Support

**SSH/TCP/Unix transports** - Backend now dispatches to `UploadArchive`:
```go
case transport.UploadArchiveService:
    return transport.UploadArchive(...)
```

**HTTP** - Left for future protocol-v2 work (HTTP archive requires command-based protocol)

### Wire Protocol

Implements the git-upload-archive protocol:
- Client sends "argument <arg>" pkt-lines for format, prefix, tree-ish, etc.
- Server sends "ACK" or "NACK <reason>"
- Archive data streamed via sideband (channel 1 = data, channel 2 = progress)

### Supported Formats

Supported formats parsed from args:
- `tar`, `tar.gz`, `tgz`, `zip`
- `--prefix=path/` adds prefix to archive entries
- `--list` lists available archives
- Tree-ish (commit, tag, branch) specifies what to archive

## Testing

- `TestArchive_Tar`, `TestArchive_TarGz`, `TestArchive_Zip` (file transport)
- `TestGitTransport_Archive` (git daemon transport)
- `TestSSHTransport_Archive` (SSH transport)

All tests verify archive contents against git's output.

## Security

The `uploadArchive.allowUnreachable` config option controls whether arbitrary SHA-1 expressions are allowed. By default, only ref targets (e.g., `v1.0`, `main`) and `ref:path` sub-tree syntax are allowed.

## Follow-up Work

A future PR will add porcelain-level support:
- `git.Archive()` convenience method
- `ArchiveOptions` with parametrized fields:
  - `Format`: tar, tar.gz, zip
  - `Prefix`: prefix path for archive entries
  - `Remote`: tree-ish reference to archive
  - `Paths`: optional path filter
  - Additional `git archive --remote` flags

## Usage Example

```go
sess, _ := client.Dial("git@github.com:go-git/go-git.git")
req := &transport.ArchiveRequest{
    Args: []string{"--format=tar.gz", "--prefix=go-git/", "v5.12.0"},
}
rc, _ := sess.Archive(ctx, req)
defer rc.Close()
// Write rc to file or pipe to tar/gzip
```

Supersedes: https://github.com/go-git/go-git/pull/1985

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -381,6 +381,17 @@ func WriteZipArchive(st storage.Storer, w io.Writer, tree *object.Tree, commitHa
 }
 
 // MatchesPathFilter checks if a name matches any of the path filters.
+//
+// Note: This function assumes paths use forward slashes (/), which is the
+// format used by Git internally. The TreeWalker produces paths with forward
+// slashes regardless of the operating system, so this function is
+// platform-independent.
+//
+// Supported patterns:
+//   - Exact match: "README.md"
+//   - Prefix match: "docs/" matches "docs/guide.md"
+//   - Glob patterns: "*.go" matches "main.go"
+//   - Parent-child: "docs/guide.md" matches parent "docs"
 func MatchesPathFilter(name string, filters []string) bool {
 	for _, f := range filters {
 		if name == f || strings.HasPrefix(name, f+"/") || strings.HasPrefix(f, name+"/") {
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"io/fs"
 	"path"
+	"slices"
 	"strings"
 	"time"
 
@@ -56,6 +57,10 @@ var (
 	// ErrSymlinkTargetTooLarge is returned when a symlink blob exceeds the
 	// maximum size accepted by the tar archive writer.
 	ErrSymlinkTargetTooLarge = errors.New("symlink target too large")
+
+	// ErrInvalidPrefix is returned when the requested archive prefix contains
+	// path traversal sequences.
+	ErrInvalidPrefix = errors.New("invalid archive prefix")
 )
 
 const maxTarSymlinkTargetSize = 64 * 1024
@@ -454,6 +459,9 @@ func WriteArchive(st storage.Storer, w io.Writer, treeish, format, prefix string
 	if treeish == "" {
 		return fmt.Errorf("no tree-ish specified")
 	}
+	if hasInvalidArchivePrefix(prefix) {
+		return fmt.Errorf("%w: %s", ErrInvalidPrefix, prefix)
+	}
 
 	tree, commitHash, commitTime, err := ResolveTreeish(st, treeish, allowUnreachable)
 	if err != nil {
@@ -474,4 +482,13 @@ func WriteArchive(st storage.Storer, w io.Writer, treeish, format, prefix string
 	default:
 		return fmt.Errorf("unsupported archive format: %s", format)
 	}
+}
+
+func hasInvalidArchivePrefix(prefix string) bool {
+	if strings.HasPrefix(prefix, "/") || strings.HasPrefix(prefix, "\\") {
+		return true
+	}
+	return slices.Contains(strings.FieldsFunc(prefix, func(r rune) bool {
+		return r == '/' || r == '\\'
+	}), "..")
 }
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -52,7 +52,13 @@ var (
 
 	// ErrPathspecNoMatch is returned when pathspec filters don't match any files.
 	ErrPathspecNoMatch = errors.New("pathspec did not match any files")
+
+	// ErrSymlinkTargetTooLarge is returned when a symlink blob exceeds the
+	// maximum size accepted by the tar archive writer.
+	ErrSymlinkTargetTooLarge = errors.New("symlink target too large")
 )
+
+const maxTarSymlinkTargetSize = 64 * 1024
 
 // SupportedFormats returns the list of supported archive formats.
 func SupportedFormats() []string {
@@ -263,10 +269,13 @@ func WriteTarArchive(st storage.Storer, w io.Writer, tree *object.Tree, commitHa
 			if err != nil {
 				return err
 			}
-			target, err := io.ReadAll(rc)
+			target, err := io.ReadAll(io.LimitReader(rc, maxTarSymlinkTargetSize+1))
 			_ = rc.Close()
 			if err != nil {
 				return err
+			}
+			if len(target) > maxTarSymlinkTargetSize {
+				return fmt.Errorf("%w: %s (%d bytes)", ErrSymlinkTargetTooLarge, fullName, len(target))
 			}
 			hdr.Typeflag = tar.TypeSymlink
 			hdr.Linkname = string(target)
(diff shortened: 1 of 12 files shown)


plumbing/transport/upload_archive.go
```diff
@@ -0,0 +1,523 @@
...
+	// AllowUnreachable when true disables the default security restrictions
+	// and allows clients to use arbitrary SHA-1 expressions. By default,
+	// only direct ref targets (e.g. v1.0, main) and ref:path sub-tree
+	// syntax (e.g. v1.0:Documentation) are allowed.
+	//
+	// This serves as the default value. When the repository-level config
+	// uploadArchive.allowUnreachable is explicitly set, it always overrides
+	// this value (both true → false and false → true).
+	// See https://git-scm.com/docs/git-upload-archive
+	AllowUnreachable bool
```
aymanbagabas: This is usually derived from global/system config then overridden by repository config. Alternatively, we'd use config.LoadConfig.

cc/ @pjbgf

pjbgf: Let's start with the current `cfg, err := st.Config()` which is already in place. As that's the safest approach to start with.

plumbing/transport/upload_archive.go
```diff
@@ -0,0 +1,523 @@
...
+	}
+}
+
+// resolveTreeish resolves a tree-ish expression to a tree object.
+//
+// Security: By default, only direct ref names (v1.0, main) and ref:path
+// sub-tree syntax (v1.0:Documentation) are allowed. Raw SHA-1 hashes and
+// relative expressions (main^, HEAD~2) are rejected unless
+// allowUnreachable is true. See https://git-scm.com/docs/git-upload-archive
+func resolveTreeish(st storage.Storer, treeish string, allowUnreachable bool) (*object.Tree, time.Time, error) {
```
pjbgf: We need a wide range of tests around this to ensure we are enforcing the `allowUnreachable=false` correctly.

plumbing/transport/upload_archive.go
```diff
@@ -0,0 +1,523 @@
...
+		_ = rc.Close()
+		if err != nil {
+			return err
+		}
+	}
+
+	return zw.Close()
+}
+
+func matchesPathFilter(name string, filters []string) bool {
```
pjbgf: We need tests for this. How is this affected cross-platform?

plumbing/transport/upload_archive.go
```diff
@@ -0,0 +1,523 @@
...
+			return nil, time.Time{}, err
+		}
+		commitTime = commit.Committer.When
+		tree, err = commit.Tree()
+		if err != nil {
+			return nil, time.Time{}, err
+		}
+	case *object.Tree:
+		tree = o
+		commitTime = time.Now()
```
pjbgf: This behaviour aligns, but let's add a comment to highlight that and a ref to `git-archive(1)` or upstream:

> git archive behaves differently when given a tree ID versus when given a commit ID or tag ID. In the first case the current time is used as the modification time of each file in the archive. In the latter case the commit time as recorded in the referenced commit object is used instead.

plumbing/transport/upload_archive.go
```diff
@@ -0,0 +1,523 @@
...
+	return (mode | 0o666) &^ defaultUmask
+}
+
+// applyUmaskDir applies umask to directories.
+// Directories always get full permissions minus umask.
+func applyUmaskDir(mode int64) int64 {
+	return (mode | 0o777) &^ defaultUmask
+}
+
+func writeTarArchive(st storage.Storer, w io.Writer, tree *object.Tree, prefix string, pathFilter []string, modTime time.Time) error {
```
pjbgf: I think we are missing this:

> Additionally the commit ID is stored in a global extended pax header if the tar format is used; it can be extracted using git get-tar-commit-id. In ZIP files it is stored as a file comment.

plumbing/transport/upload_archive.go
```diff
@@ -0,0 +1,523 @@
...
+	for {
+		name, entry, err := walker.Next()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return err
+		}
+
+		if len(pathFilter) > 0 && !matchesPathFilter(name, pathFilter) {
```
pjbgf: If not entries match pathFilter [upstream](https://github.com/git/git/blob/9f223ef1c026d91c7ac68cc0211bde255dda6199/archive.c#L479) fails with: `pathspec '%s' did not match any files`

plumbing/transport/upload_archive.go
```diff
@@ -0,0 +1,523 @@
...
+		tree, err = object.GetTree(st, entry.Hash)
+		if err != nil {
+			return nil, time.Time{}, err
+		}
+	}
+
+	return tree, commitTime, nil
+}
+
+func resolveRef(st storage.Storer, name string, allowHash bool) (plumbing.Hash, error) {
```
pjbgf: Need some tests for this.

pjbgf (changes requested):
@aymanbagabas thanks for looking into this, here's some feedback:

plumbing/transport/upload_archive.go
aymanbagabas: Cool, I will drop this option and add a comment to investigate this later and derive it from global/system config before repository config.

plumbing/transport/upload_archive.go
aymanbagabas: Git stores the paths using `/` internally, it shouldn't be a problem. I also added tests to verify that.

internal/archive/archive.go
```diff
@@ -0,0 +1,467 @@
...
+	switch o := obj.(type) {
+	case *object.Commit:
+		commitHash = &o.Hash
+		commitTime = o.Committer.When
+		tree, err = o.Tree()
+		if err != nil {
+			return nil, nil, time.Time{}, err
+		}
+	case *object.Tag:
+		commit, err := object.GetCommit(st, o.Target)
```
pjbgf: This would not work for annotated tags pointing to a tree for example (which is supported upstream).

Given that annotated tags could be pointing to other annotated tag, we will need to resolved it to a non-tag object, then verify whether it is a commit or a tree. Any other object should fail with an error.

Let's support this use case and add a test to confirm it.

pjbgf: Closing as I'll be adding commits to this effect shortly.

pjbgf (approved):
@aymanbagabas thanks for the great effort putting this together. 🙇🙇🙇

PR #1987, opened 2026-04-13
plumbing: packp, refactor AdvRefs and capability packages

aymanbagabas:
This PR refactors the `plumbing/protocol/packp` package to simplify the AdvRefs encoder/decoder, clean up the capability package, and remove unnecessary constructors.

### AdvRefs loses information
The old `AdvRefs` stored references in two maps (`References` and `Peeled`) that lost wire-order information. Most consumers called `MakeReferenceSlice()` immediately—the maps were an intermediate representation nobody wanted. The git pack protocol defines advertised-refs as an ordered sequence where peeled refs follow their base ref. The current representation didn't preserve this.

### Capability type forces conversions
`Capability` was a defined type requiring conversions like `capability.Capability("custom-cap")`. Protocol v2's arbitrary command capabilities multiply this friction. A type alias eliminates conversions while keeping the named constants.

### capability.List validation is counterproductive
`List.Add()` and `Set()` returned errors that callers universally discarded. Every call site discards the error. The validation maps enforce rules that callers already know. Worse, the `known` map becomes a liability for v2: unknown capabilities bypass validation silently.

### Constructors exist only to initialize maps
Every packp constructor (`NewAdvRefs`, `NewUploadRequest`, etc.) existed primarily to call `capability.NewList()`, which existed to `make(map[...])`. If `List` is zero-value safe, all constructors become unnecessary.

### capability.List zero-value safety
- Added lazy init to `List.Add()`, `Set()`, and other methods
- Made `List` zero-value safe—first write creates the internal map
- Removed `NewList()`—use `&List{}` or zero value

### Capability type alias
- Changed `type Capability string` to `type Capability = string`
- Eliminated conversion friction for protocol v2
- Removed validation maps and error types
- Made `Add`/`Set` infallible (no error return)

### Wire format moved out of List
- Replaced `List.Decode` with package-level `DecodeList`
- v0/v1 uses space-separated capabilities; v2 uses line-separated
- Keeping `List` protocol-agnostic: it's now a pure container type

### AdvRefs slice migration
- Replaced `References map[string]Hash` and `Peeled map[string]Hash` with single `References []*Reference` slice
- Added `ReferenceName.IsPeeled()` helper (consistent with `IsTag()`, `IsBranch()`)
- Removed `NewAdvRefs()`, `AddReference()`, `MakeReferenceSlice()`, `IsEmpty()`

### AdvRefs encoder/decoder simplification
- Removed state machine pattern from encoder (5 state functions → single `Encode()` method)
- Removed state machine pattern from decoder (8 state functions → single `Decode()` method with loop)
- Net reduction of 145 lines of code

### Constructor removal and zero-value safety
- Removed `NewUploadRequest`, `NewUpdateRequests`, `NewReportStatus`
- Types are zero-value safe: nil slices work with `append`, `Capabilities` is value type

### DepthRequest struct replaces Depth interface
The old `Depth` interface used a sum-type pattern (`DepthCommits`, `DepthSince`, `DepthReference`) that couldn't represent the full git protocol. Specifically:
- Couldn't combine `deepen-since` with `deepen-not` (git allows this)
- Couldn't specify multiple `deepen-not` refs (git uses an oidset)
- Required a constructor just to set a non-nil default

The new `DepthRequest` struct matches how canonical git C handles it:

```go
type DepthRequest struct {
    Commits int           // "deepen <n>" - 0 means no limit
    Since   time.Time     // "deepen-since <timestamp>"
    NotRefs []string      // Multiple "deepen-not <ref>" lines
}
```

This matches git's `upload_pack_data` struct which stores depth as separate fields, not a sum type. The zero value (all fields empty/zero) means "no depth constraint" which is the correct default for infinite depth.

**Added validation:** Following canonical git, the encoder and decoder now reject invalid combinations:
- `deepen <n>` cannot be combined with `deepen-since` or `deepen-not`
- Error message matches git: `"deepen and deepen-since (or deepen-not) cannot be used together"`

## Benchmarks

**Before (transport-migrate/base: 3 runs):**
```
BenchmarkPlainClone-16    3    341770083 ns/op    3742216 B/op    34587 allocs/op
BenchmarkPlainClone-16    3    345509806 ns/op    3472776 B/op    34530 allocs/op
BenchmarkPlainClone-16    4    329814698 ns/op    3562374 B/op    34564 allocs/op
```

**After (packp-refactoring: 3 runs):**
```
BenchmarkPlainClone-16    3    346557236 ns/op    2984754 B/op    29572 allocs/op
BenchmarkPlainClone-16    3    335943014 ns/op    2855368 B/op    29477 allocs/op
BenchmarkPlainClone-16    3    349759278 ns/op    2876133 B/op    29320 allocs/op
```

**Improvement:**
- Memory: ~3.6MB → ~2.9MB (~19% reduction in allocations)
- Allocs: ~34,500 → ~29,500 (~14% reduction)

## Migration Guide

| Before | After |
|--------|-------|
| `packp.NewAdvRefs()` | `&packp.AdvRefs{}` |
| `ar.References["name"] = hash` | `ar.References = append(ar.References, plumbing.NewHashReference("name", hash))` |
| `capability.NewList()` | `&capability.List{}` |
| `caps.Decode(data)` | `capability.DecodeList(data, &caps)` |
| `depth := DepthCommits(3)` | `depth := DepthRequest{Commits: 3}` |

## Testing
All existing tests pass:
- `go test ./plumbing/protocol/packp/...`
- `go test ./plumbing/protocol/capability/...`
- `make validate` passes (0 lint issues)

Supersedes: https://github.com/go-git/go-git/pull/1984


plumbing/protocol/capability/list.go
```diff
@@ -0,0 +1,156 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+)
+
+// List represents a list of capabilities. The zero value is safe to use;
+// the internal map is lazily initialized on first write.
+type List struct {
```
pjbgf, line 13: We should either make this thread-safe, or highlight that it isn't.

plumbing/protocol/capability/list_test.go
```diff
@@ -0,0 +1,187 @@
...
+	s.Len(caps.Get(Agent), 1)
+}
+
+func (s *SuiteCapabilities) TestSetDuplicate() {
+	caps := &List{}
+	caps.Set(Agent, "baz")
+	caps.Set(Agent, "bar")
+
+	s.Equal("agent=bar", caps.String())
+}
```
pjbgf, line 221: Let's use table driven to avoid unnecessary bloating.

plumbing/protocol/packp/advrefs_decode_test.go
```diff
@@ -70,7 +70,8 @@ func (s *AdvRefsDecodeSuite) TestZeroId() {
 		"",
 	}
 	ar := s.testDecodeOK(payloads)
-	s.Nil(ar.Head)
+	_, err := ar.Head()
+	s.Equal(plumbing.ErrReferenceNotFound, err)
```
pjbgf:
```suggestion
	s.ErrorIs(plumbing.ErrReferenceNotFound, err)
```

plumbing/protocol/packp/ulreq_decode.go
```diff
@@ -9,244 +9,179 @@ import (
...
+		line = bytes.TrimPrefix(line, want)
+		h, err := readHash()
+		if err != nil {
+			return err
+		}
+		req.Wants = append(req.Wants, h)
+	}
+
+	// --- Shallow lines ---
+
```
pjbgf:
```suggestion
```

plumbing/protocol/packp/ulreq_decode.go
```diff
@@ -9,244 +9,179 @@ import (
...
+		if err != nil {
+			return err
+		}
+		if !ok || len(line) == 0 {
+			return nil
+		}
+	}
+
+	// --- Deepen lines ---
+
```
pjbgf:
```suggestion
```

plumbing/protocol/packp/ulreq_encode.go
```diff
@@ -15,140 +14,79 @@ import (
...
-		}
-	case DepthSince:
-		when := time.Time(depth).UTC()
-		if _, err := pktline.Writef(e.w, "deepen-since %d\n", when.Unix()); err != nil {
-			e.err = fmt.Errorf("encoding depth %s: %s", when, err)
-			return nil
+	// Depth
+	depth := req.Depth
+	if depth.Deepen > 0 && (!depth.DeepenSince.IsZero() || len(depth.DeepenNot) > 0) {
+		return fmt.Errorf("deepen and deepen-since (or deepen-not) cannot be used together")
```
pjbgf: Let's create a typed error for this. Perhaps: `ErrDeepenMutuallyExclusive = errors.New("deepen is mutually exclusive with deepen-since and deepen-not")`.

```suggestion
		return ErrDeepenMutuallyExclusive
```

plumbing/protocol/packp/ulreq_decode.go
```diff
@@ -9,244 +9,179 @@ import (
...
+		return h, nil
 	}
-	d.line = bytes.TrimPrefix(d.line, want)
 
-	hash, ok := d.readHash()
-	if !ok {
-		return nil
-	}
-	d.data.Wants = append(d.data.Wants, hash)
+	// --- Want lines ---
```
pjbgf:
```suggestion
```

plumbing/protocol/packp/ulreq_decode.go
```diff
@@ -9,244 +9,179 @@ import (
...
+		// After deepen <n>, only flush-pkt is valid
+		if req.Depth.Deepen > 0 {
+			if bytes.HasPrefix(line, deepenSince) || bytes.HasPrefix(line, deepenReference) {
+				return fmt.Errorf("deepen and deepen-since (or deepen-not) cannot be used together")
+			}
+			return decodeError("unexpected payload while expecting a flush-pkt: %q", line)
+		}
+		// After deepen-since/deepen-not, only deepen-since/deepen-not or flush is valid
+		if deepenRevList && bytes.HasPrefix(line, deepen) && !bytes.HasPrefix(line, deepenSince) && !bytes.HasPrefix(line, deepenReference) {
+			return fmt.Errorf("deepen and deepen-since (or deepen-not) cannot be used together")
```
pjbgf:
```suggestion
			return ErrDeepenMutuallyExclusive
```

plumbing/protocol/packp/ulreq_decode.go
```diff
@@ -9,244 +9,179 @@ import (
...
+			return err
+		}
+		if !ok || len(line) == 0 {
+			return nil
+		}
+
+		// After deepen <n>, only flush-pkt is valid
+		if req.Depth.Deepen > 0 {
+			if bytes.HasPrefix(line, deepenSince) || bytes.HasPrefix(line, deepenReference) {
+				return fmt.Errorf("deepen and deepen-since (or deepen-not) cannot be used together")
```
pjbgf:
```suggestion
				return ErrDeepenMutuallyExclusive
```

plumbing/protocol/packp/ulreq_decode.go
```diff
@@ -9,244 +9,179 @@ import (
...
+			line = bytes.TrimPrefix(line, deepenSince)
+			secs, err := strconv.ParseInt(string(line), 10, 64)
+			if err != nil {
+				return err
+			}
+			req.Depth.DeepenSince = time.Unix(secs, 0).UTC()
+			deepenRevList = true
+		case bytes.HasPrefix(line, deepenReference):
+			if req.Depth.Deepen > 0 {
+				return fmt.Errorf("deepen and deepen-since (or deepen-not) cannot be used together")
```
pjbgf:
```suggestion
				return ErrDeepenMutuallyExclusive
```

plumbing/protocol/packp/ulreq_decode.go
```diff
@@ -9,244 +9,179 @@ import (
...
+			if err != nil {
+				return err
+			}
+			if n < 0 {
+				return fmt.Errorf("negative depth")
+			}
+			req.Depth = DepthRequest{Deepen: n}
+		case bytes.HasPrefix(line, deepenSince):
+			if req.Depth.Deepen > 0 {
+				return fmt.Errorf("deepen and deepen-since (or deepen-not) cannot be used together")
```
pjbgf:
```suggestion
				return ErrDeepenMutuallyExclusive
```

plumbing/protocol/packp/ulreq_decode.go
```diff
@@ -9,244 +9,179 @@ import (
...
+		}
+	}
+
+	// --- Deepen lines ---
+
+	for bytes.HasPrefix(line, deepen) {
+		switch {
+		case bytes.HasPrefix(line, deepenCommits):
+			if deepenRevList {
+				return fmt.Errorf("deepen and deepen-since (or deepen-not) cannot be used together")
```
pjbgf:
```suggestion
				return ErrDeepenMutuallyExclusive
```

plumbing/protocol/packp/updreq_decode.go
```diff
@@ -69,74 +70,45 @@ func errMalformedCommand(err error) error {
...
-
-type updReqDecoder struct {
-	r   io.ReadCloser
-	pr  io.Reader
-	req *UpdateRequests
-
-	payload []byte
-	length  int // length of the pktline payload
-}
+	defer rc.Close() //nolint:errcheck
```
pjbgf: Not sure we were closing the reader before. However, closing an `io.Reader` may catch caller off-guard. If this is needed, we should add a line to Decode's docs to highlight this behaviour.

I wonder whether we are better off receiving an `io.ReadCloser` instead? We may have discussed this in the previous refactor, but I don't seem able to find that thread.

plumbing/protocol/capability/list.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func (l *List) init() {
+	if l.m == nil {
+		l.m = make(map[string]*entry)
+	}
+}
+
+// Add adds a capability, values are optional
+func (l *List) Add(c string, values ...string) {
```
pjbgf, line 85: The latest changes dropped some of the validation. The [specs](https://github.com/git/htmldocs/blob/gh-pages/gitprotocol-capabilities.adoc) are fairly strict around capabilities:

> server MUST NOT advertise capabilities it does not understand.

> [session-id] ... must fit within a packet-line, and must not contain non-printable or whitespace characters

Let's bring back validation and ensure that we check:
- Argument-less caps
- Caps that require arguments
- Any argument specific validation defined by the spec
- Only known caps are accepted

Adding tests to confirm the above.

plumbing/protocol/packp/ulreq_decode.go
```diff
@@ -9,244 +9,179 @@ import (
...
+			}
+			req.Depth.DeepenSince = time.Unix(secs, 0).UTC()
+			deepenRevList = true
+		case bytes.HasPrefix(line, deepenReference):
+			if req.Depth.Deepen > 0 {
+				return fmt.Errorf("deepen and deepen-since (or deepen-not) cannot be used together")
+			}
+			line = bytes.TrimPrefix(line, deepenReference)
+			req.Depth.DeepenNot = append(req.Depth.DeepenNot, string(line))
+			deepenRevList = true
```
pjbgf, line 155: [Upstream](https://github.com/git/git/blob/94f057755b7941b321fd11fec1b2e3ca5313a4e0/upload-pack.c#L1013) does some validation here for ambiguity. Probably not something we'd at decoding time, but may be worth including as part of this PR.

pjbgf (changes requested):
@aymanbagabas thanks for working on this. Here's some feedback on the refactoring.

plumbing/protocol/capability/list.go
aymanbagabas: I will keep it as is right now and highlight that it's not thread-safe. We can revisit this later in a future change.

plumbing/protocol/capability/list.go
aymanbagabas: I added a `capability.Validate` method to validate v0/v1 capabilities, added the missing `session-id` capability with tests, and updated client-side and server-side capability validation during handshake.

plumbing/protocol/packp/ulreq_decode.go
aymanbagabas: IIRC, we still don't fully support `deepen-not`. We can revisit this later when we add support.

plumbing/protocol/packp/updreq_decode.go
aymanbagabas: Good catch! I think this is a relic from the past pre v5 transport refactor. We no longer need to close anything here, the transport handles that. The [main branch](https://github.com/aymanbagabas/go-git/blob/6fa62ef48d1f7f4bc6990b22398c096f115261e4/plumbing/protocol/packp/updreq_decode.go#L73) doesn't do anything with the closer.

aymanbagabas:
I think the existing `idxfile` issue that is shown in the [CI](https://github.com/go-git/go-git/actions/runs/24908334127/job/72949435719?pr=1987) should be fixed by https://github.com/go-git/go-git/pull/2027

plumbing/protocol/packp/advrefs_decode_test.go
```diff
@@ -70,7 +70,8 @@ func (s *AdvRefsDecodeSuite) TestZeroId() {
 		"",
 	}
 	ar := s.testDecodeOK(payloads)
-	s.Nil(ar.Head)
+	_, err := ar.Head()
+	s.ErrorIs(plumbing.ErrReferenceNotFound, err)
```
pjbgf, line 74:
```suggestion
	s.ErrorIs(err, plumbing.ErrReferenceNotFound)
```

pjbgf (approved):

PR #1988, opened 2026-04-13
plumbing: transport, correct the test to use the correct receive-pack…

aymanbagabas:
… service name

The receive-pack test was incorrectly using the upload-pack service name. This commit corrects the test to use the receive-pack service name, ensuring that the test is validating the correct service behavior.


pjbgf (approved):

PR #1989, opened 2026-04-13
plumbing/transport: don't emit a second NAK after encoding empty ACKs

Soph:
## Summary

`UploadPack` wrote **two** NAK pktlines in front of the sideband-wrapped pack when a stateless-RPC client sent `haves` that weren't reachable from any `want`:

```
0008NAK\n        ← ServerResponse{ACKs: nil}.Encode  (len(haves) > 0 branch)
0008NAK\n        ← ServerResponse{}.Encode           (ack.Hash.IsZero() branch)
0009\x01PACK…    ← sideband-wrapped pack
```

`ServerResponse.Decode` only consumes the first NAK. The standard go-git client pipeline (`Decode` + `sideband.Demuxer`) then reads the trailing `"NAK\n"` as a sideband frame, sees channel byte `'N'` (0x4E), and fails with `unknown channel NAK`. This reliably breaks fetches where the target's advertised ref isn't an ancestor of the want — rewinds, divergent overwrites, force-push retargets — even between a go-git client and a go-git server.

## Fix

Only emit the terminal NAK in the `ack.Hash.IsZero()` branch when no haves were sent. When haves were sent with no common found, the `len(haves) > 0` branch already wrote the NAK via `ServerResponse{ACKs: empty}.Encode`.

## Behavior matrix

| haves | common found? | multi-ack? | Output (before → after) |
|---|---|---|---|
| empty | — | — | 1 NAK → 1 NAK |
| present | yes | no | 1 ACK → 1 ACK |
| present | yes | yes | ACKs-with-status + final ACK → same |
| **present** | **no** | **no** | **2 NAKs → 1 NAK** |
| present | no | yes | ACKs + 1 NAK → same |

## Test

Adds `TestUploadPackStatelessRPCUnreachableHavesEmitsSingleNAK` in the new `ReceivePackServeSuite` file, asserting both:

1. Response does not begin with `0008NAK\n0008NAK\n`.
2. `ServerResponse.Decode` + `sideband.NewDemuxer(Sideband64k)` can read the `PACK` signature end-to-end.

## Verification

- New test: FAIL before, PASS after.
- `./plumbing/transport/...`: all green.
- `./...` full module: all green.

### DIFF
--- a/plumbing/transport/upload_pack.go
+++ b/plumbing/transport/upload_pack.go
@@ -228,8 +228,17 @@ func UploadPack(
 					writec <- fmt.Errorf("sending final ack server-response: %w", err)
 					return
 				}
-			case ack.Hash.IsZero():
-				// We don't have multi-ack and there are no haves. Encode a NAK.
+			case ack.Hash.IsZero() && len(haves) == 0:
+				// We don't have multi-ack and no haves were sent. Emit the
+				// single terminal NAK.
+				//
+				// When haves *were* sent, the ServerResponse{ACKs: acks}
+				// write above already emitted a NAK (encodeServerResponse
+				// writes NAK when ACKs is empty). Emitting another one here
+				// would produce two consecutive "0008NAK\n" pktlines;
+				// ServerResponse.Decode consumes only the first, and the
+				// second would then be misread by the sideband demuxer as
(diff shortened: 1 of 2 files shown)


pjbgf (approved):

PR #1997, opened 2026-04-14
plumbing: transport/http, harden redirect handling to match canonical git

aymanbagabas:
## Summary

The HTTP transport's redirect handling had several security gaps compared to canonical git's default `http.followRedirects = initial` policy. This PR closes them.

### Problem

When go-git's HTTP transport follows server redirects during `Handshake` (the `GET /info/refs` discovery request), several things could go wrong:

1. **No redirect policy** — Go's default `http.Client` follows redirects on *every* request (GET, POST, etc.). A malicious server could redirect a pack-data POST to an attacker-controlled host and the client would silently follow it. Canonical git only follows redirects on the initial `/info/refs` request and treats 3xx on subsequent requests as errors.

2. **No protocol/scheme restriction** — A redirect to `file://`, `gopher://`, or any other scheme was accepted. This is an SSRF vector in server-side environments. Canonical git restricts redirect targets to `http`/`https`/`ftp`/`ftps` via `CURLOPT_REDIR_PROTOCOLS`.

3. **Silent fallback on path mismatch** — If the redirect target's path didn't end with `/info/refs`, `applyRedirect` silently returned the original base URL instead of erroring. Canonical git's `update_url_from_redirect()` calls `die()` here because a tail mismatch could indicate a malicious redirect attempting to rewrite the base URL to an unrelated repository.

4. **Credential leakage to redirect target** — Both the URL-embedded `Userinfo` credentials and the `Authorizer` callback were stored in the session and re-applied on every subsequent POST. After a cross-host redirect, the original host's credentials would be sent to the new host. Canonical git re-derives credentials from the new URL via `credential_from_url()`, effectively wiping the old ones.

### Changes

- **`common.go`** — `applyRedirect` now returns `(*url.URL, error)`. Validates the redirect target's scheme is `http` or `https`. Returns a hard error when the redirected path doesn't end with `/info/refs` (matching canonical git's `die()` behavior).

- **`http.go`** — Adds a `CheckRedirect` policy (`checkRedirect`) implementing canonical git's `initial` mode: only the `/info/refs` GET (tagged via context key) is allowed to follow redirects. Non-initial requests, unsupported schemes, and >10 hops are rejected. Go's stdlib already strips `Authorization` on cross-host redirects and preserves it for same-host redirects, so no manual header manipulation is needed.

- **`handshake.go`** — Tags the `/info/refs` request with `withInitialRequest(ctx)`. After `applyRedirect`, clears both `URL.User` and the `Authorizer` callback when the host changed — both are credential sources that would otherwise leak to the redirect target.

### Canonical git reference

- Redirect policy: `http.c:130` (`HTTP_FOLLOW_INITIAL` default), `http.c:1583-1591` (CURLOPT_FOLLOWLOCATION per-request)
- Protocol whitelist: `http.c:991-1013` (`get_curl_allowed_protocols`)
- Tail validation: `http.c:2243-2293` (`update_url_from_redirect`)
- Credential wipe: `http.c:2354-2360` (`credential_from_url` after redirect)

## Test plan

- [x] `TestCheckRedirectPolicy` — unit tests for the `checkRedirect` function: blocks non-initial requests, allows initial requests, blocks unsupported schemes, blocks >10 redirects
- [x] `TestApplyRedirect` — unit tests for `applyRedirect`: no-op detection, host/scheme/path updates, scheme rejection, tail mismatch error
- [x] `TestRedirectPostBlocked` — integration test: server proxies GET but 307-redirects POST; verifies fetch fails with "non-initial request"
- [x] `TestRedirectStripsCredentials` — integration test: cross-host redirect clears `URL.User` from the session's base URL
- [x] `TestRedirectPath`, `TestRedirectSchema`, `TestRedirectPathWithFetch` — existing integration tests still pass (legitimate redirects during handshake continue to work)
- [x] Full `go test ./plumbing/transport/http/...` passes

### DIFF
--- a/plumbing/transport/http/http.go
+++ b/plumbing/transport/http/http.go
@@ -2,12 +2,31 @@
 package http
 
 import (
+	"context"
 	"crypto/tls"
+	"fmt"
 	"net/http"
 	"net/url"
 
 	"github.com/go-git/go-git/v6/plumbing/transport"
 )
+
+// contextKey is an unexported type for context keys in this package.
+type contextKey int
+
+const initialRequestKey contextKey = iota
+
+// withInitialRequest marks a context so that checkRedirect allows
+// the HTTP client to follow redirects. Only the /info/refs discovery
+// request should carry this flag.
+func withInitialRequest(ctx context.Context) context.Context {
+	return context.WithValue(ctx, initialRequestKey, true)
+}
+
+func isInitialRequest(req *http.Request) bool {
+	v, _ := req.Context().Value(initialRequestKey).(bool)
+	return v
+}
 
 // Options configures the HTTP transport.
 type Options struct {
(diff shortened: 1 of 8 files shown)


plumbing/transport/http/http.go
```diff
@@ -2,20 +2,62 @@
...
+
+const initialRequestKey contextKey = iota
+
+// FollowRedirects controls how the HTTP transport follows redirects.
+//
+// The values mirror Git's http.followRedirects config:
+// "true" follows redirects for all requests, "false" treats any redirect
+// as an error, and "initial" follows redirects only for the initial
+// /info/refs discovery request. The zero value defaults to "initial".
+type FollowRedirects string
```
aymanbagabas: @pjbgf I wouldn't make this a type. Instead, this should be derived from system/global/repo gitconfig. I initially skipped this and only implemented "initial" follow redirect for that reason.

pjbgf: @aymanbagabas Fair point, let me change that.

Given the change in behaviour, I think we need to support opt-in/out, as per upstream, so will wire in config based setting.

aymanbagabas (commented):
LGTM

pjbgf (approved):

plumbing/transport/http/http.go
pjbgf: As discussed off-line, we may refine the API in the coming weeks.

Issue #1956, opened 2026-04-02
worktree: Optimize checkout performance for worktree creation and branch switching


cedric-appdirect:
### Feature Description

Creating a new linked worktree (or checking out a branch) currently decompresses every file from the object store, even when most files are already materialized in an existing worktree. This is the same approach used by Git C's git worktree add (which runs git reset --hard internally), but Git C compensates with parallel checkout workers and streaming blob writes — optimizations go-git currently lacks.

For large repositories, this makes worktree creation and clone checkout significantly slower than necessary.

#### Background

When git worktree add -b feature HEAD is executed, the new branch points to the **same commit** as the source. Every file in the target tree is byte-for-byte identical to the existing worktree. Despite this, both Git C and go-git decompress every blob from the object store individually.

Git C mitigates this with:
- **Parallel checkout** (parallel-checkout.c) — multiple worker processes write files concurrently.
- **Streaming blob writes** (odb_stream_blob_to_fd) — blobs stream directly to disk without full memory buffering.
- **Cache-tree fast path** (traverse_by_cache_tree) — skips entire subtrees when the index cache-tree matches the target tree OID.
- **CE_UPTODATE / FSMonitor** — avoids redundant lstat calls for files known to be clean.

go-git's current checkout path (resetWorktreeToTree → checkoutChange → checkoutFile → copyObjectToWorktree) has none of these, plus a few additional per-file overheads:
- r.Config() is called inside copyObjectToWorktree for **every file** to check AutoCRLF.
- When AutoCRLF is enabled, each blob is decompressed **twice** (once for binary detection, once for writing).

#### Proposed Improvements

**Tier 1 — Low-hanging fruit:**
- [ ] Cache r.Config() result outside the checkout file loop, pass it through as a parameter.
- [ ] Fix AutoCRLF double-decompression: buffer the initial peek for binary detection and reuse it for writing (or use a TeeReader).

**Tier 2 — Parallel file extraction:**
- [ ] Introduce a bounded goroutine pool for concurrent file writes during checkout — see #1749.

**Tier 3 — Copy from existing worktree (novel, beyond Git C):**
- [ ] When creating a worktree from the same commit, copy files directly from the source worktree instead of decompressing from the object store. The source worktree's index provides the expected OID — if stat data matches, the file is known clean and safe to copy.
[...]

PR #1998, opened 2026-04-14
plumbing: format/packfile, use ReadAt in FSObject.Reader() for concurrency safety.

cedric-appdirect:
Replace Seek+Read with io.SectionReader backed by ReadAt in FSObject.Reader(). ReadAt does not modify the file's seek cursor (it maps to pread(2) on POSIX), so multiple goroutines can safely call Reader() on FSObjects that share the same packfile handle. The billy.File interface already requires io.ReaderAt, and both osfs (via os.File) and memfs implementations are documented as safe for concurrent ReadAt calls.

The closed-file recovery path now uses a 1-byte ReadAt probe instead of Seek to detect stale descriptors without mutating shared state.

A zero-length ReadAt is insufficient because os.File returns (0, nil) for empty reads even on closed files.

This PR is necessary to enable parallel checkout from PR #1749 which will help resolving #1956

Assisted-by: OpenCode with Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/plumbing/format/packfile/fsobject.go
+++ b/plumbing/format/packfile/fsobject.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"errors"
 	"io"
+	"math"
 	"os"
 
 	billy "github.com/go-git/go-billy/v6"
@@ -54,6 +55,10 @@ func NewFSObject(
 }
 
 // Reader implements the plumbing.EncodedObject interface.
+//
+// Reader is safe for concurrent use: it uses ReadAt (which does not modify the
+// file's seek cursor) instead of Seek+Read, so multiple goroutines can call
+// Reader on FSObjects that share the same underlying packfile handle.
 func (o *FSObject) Reader() (io.ReadCloser, error) {
 	obj, ok := o.cache.Get(o.hash)
 	if ok && obj != o {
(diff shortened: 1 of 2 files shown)


plumbing/format/packfile/fsobject_test.go
```diff
@@ -0,0 +1,71 @@
...
+	objects := make([]plumbing.EncodedObject, 0, len(entries))
+	for h := range entries {
+		obj, err := p.Get(h)
+		require.NoError(t, err)
+		objects = append(objects, obj)
+	}
+	require.NotEmpty(t, objects)
+
+	// Read all objects concurrently. Before the fix, this triggered a data
+	// race on the shared Seek cursor inside the pack file handle.
```
pjbgf:
```suggestion
```

plumbing/format/packfile/fsobject_test.go
```diff
@@ -0,0 +1,71 @@
...
+	const iterations = 5
+	var wg sync.WaitGroup
```
pjbgf: Let's use `testing/synctest` here.

plumbing/format/packfile/fsobject_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/go-git/go-git/v6/internal/fixtureutil"
+	"github.com/go-git/go-git/v6/plumbing"
+	"github.com/go-git/go-git/v6/plumbing/format/packfile"
+)
+
+// TestFSObjectConcurrentReader verifies that multiple goroutines can call
+// Reader() on FSObjects backed by the same packfile handle without racing.
+// Before the ReadAt fix, FSObject.Reader() used Seek+Read on a shared file
+// descriptor, causing data races under concurrent access.
```
pjbgf, line 20:
```suggestion
```

pjbgf (commented):
@cedric-appdirect thanks for working on this, below are a few nits.

plumbing/format/packfile/fsobject.go
```diff
@@ -65,27 +70,29 @@ func (o *FSObject) Reader() (io.ReadCloser, error) {
...
-	// In some cases that descriptor would already be closed, in such cases,
-	// open the packfile again and close it when the reader is closed.
-	if err != nil && errors.Is(err, os.ErrClosed) {
-		o.pack, err = o.fs.Open(o.packPath)
+
+	// Probe with a 1-byte ReadAt to detect a closed file descriptor without
+	// modifying any shared state. A zero-length read cannot be used because
+	// some implementations (e.g. os.File) return (0, nil) for empty reads
+	// even on closed files.
+	if _, err := pack.ReadAt(make([]byte, 1), o.offset); err != nil && errors.Is(err, os.ErrClosed) {
```
pjbgf:
```suggestion
	_, err := pack.ReadAt(make([]byte, 1), o.offset);
	if err != nil && errors.Is(err, os.ErrClosed) {
```

Then we should handle the case of `err` not being `os.ErrClosed` - as before, which was to return the error.

pjbgf (approved):
@cedric-appdirect thanks for working on this. 🙇

PR #2005, opened 2026-04-17
Add support for negative refspecs

utafrali:
Adds support for negative refspecs, which use the `^` prefix to exclude certain refs from fetch operations. Go-git wasn't handling these, so opening repos with them in the config would fail. Added IsNegative() and MatchNegative() methods to detect and match them properly, plus skipped validation for negative specs.

### DIFF
--- a/config/refspec.go
+++ b/config/refspec.go
@@ -11,6 +11,7 @@ const (
 	refSpecWildcard  = "*"
 	refSpecForce     = "+"
 	refSpecSeparator = ":"
+	refSpecNegative  = "^"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


BootstrapperSBL:
Hey @utafrali — just noticed my #2008 overlaps with this (didn't catch yours when I opened mine; sorry for the duplication). Looks like yours is broader: adds `IsNegative()`/`MatchNegative()` for the full matching flow, while mine only widens `RefSpec.Validate` to stop rejecting `^pattern` specs on repo open. Happy to close mine if this one gets merged first, or to rebase the Validate-only slice on top if maintainers prefer to split the work. No strong preference — whatever's easiest for @pjbgf to land.

Issue #1927, opened 2026-03-27
Fail to open repository with negative fetch spec


pjbgf:
### Bug Description

`go-git` is unable the open a repository with negative fetch spec due to strict validation, resulting in `failed to open repository: read config: malformed refspec, separators are wrong`

To fix this we need to:
- [ ] Loose the existing [validation](https://github.com/go-git/go-git/blob/main/config/refspec.go#L34) to support negative ref specs.
- [ ] Amend the Fetch logic so that it can support fetch wildcard refspec, after removing the negative ones.

Relates to https://github.com/entireio/cli/issues/778.
Upstream information on negative refspec: https://git-scm.com/docs/git-fetch/2.31.0#Documentation/git-fetch.txt-refspec.

### go-git Version

v6@main

### Steps to Reproduce

The repository config:

```
  [core]
        repositoryformatversion = 1
        filemode = true
        bare = false
        logallrefupdates = true
        ignorecase = true
        precomposeunicode = true
  [remote "origin"]
        url = git@github.com:example/example.git
        promisor = true
        partialclonefilter = blob:none
        fetch = +refs/heads/*:refs/remotes/origin/*
        fetch = ^refs/heads/some-excluded-branch # <== The offending negation pattern!
        fetch = ^refs/heads/another-excluded-branch # <== (again)
  [branch "main"]
        remote = origin
        merge = refs/heads/main
```

BootstrapperSBL:
Took a look — for step 1 the existing `Validate()` in `config/refspec.go:34` rejects negative refspecs early because it requires exactly one `:` separator (and negative specs have none). Minimal change is a short-circuit on leading `^`:

```go
func (s RefSpec) Validate() error {
    if s.IsNegative() {
        return nil
    }
    // existing validation...
}
```

plus a new `IsNegative()` helper, and adjusting `Src()` to strip the leading `^` / `Dst()` to return `""` for negative specs.

Happy to send the validation-side PR (step 1) with table tests covering the reporter's config — both `^refs/heads/some-excluded-branch` and wildcard forms. Step 2 (fetch logic filtering) feels cleaner as a follow-up PR once this lands.

PR #2008, opened 2026-04-17
config: accept negative refspecs in RefSpec.Validate

BootstrapperSBL:
## What

Teaches `RefSpec.Validate` about negative refspecs (`^<pattern>`) so that repositories with exclusion rules in their fetch config can be opened without `malformed refspec, separators are wrong`. Refs #1927.

## Why

Git's refspec grammar has three shapes:

| Shape | Example |
|---|---|
| positive | `refs/heads/main:refs/remotes/origin/main` |
| force | `+refs/heads/*:refs/remotes/origin/*` |
| **negative** | `^refs/heads/some-excluded-branch` |

Only the first two have a destination, so `Validate()` insists on exactly one `:` separator. That means a repo with a perfectly normal exclusion config like

```ini
[remote "origin"]
    fetch = +refs/heads/*:refs/remotes/origin/*
    fetch = ^refs/heads/some-excluded-branch
    fetch = ^refs/heads/another-excluded-branch
```

fails on `git.PlainOpen`. This is the bug described in #1927.

## What changed

1. `refSpecNegative = "^"` constant + `RefSpec.IsNegative()` predicate.
2. `Validate()` short-circuits with `nil` for negative refspecs — they're valid by construction, no separator checks apply.
3. `Src()` strips the leading `^` so callers get the pattern in the same shape wildcard/exact refspecs produce.
4. `Dst()` returns an empty `ReferenceName` for negative refspecs (they have no destination).
5. Test coverage for `IsNegative`, `Validate`, `Src`/`Dst`, and negative-spec interaction with the existing `IsForceUpdate` / `IsDelete` / `IsExactSHA1` predicates.

I deliberately stopped here rather than also touching `Match` or the `remote.Fetch` plumbing — the issue describes that as a separate step, and I'd rather keep this PR tight so it can land first. Happy to follow up with step 2 once this is in.

## Before / after

```go
// Before
RefSpec("^refs/heads/excluded").Validate()
// → ErrRefSpecMalformedSeparator

// After
RefSpec("^refs/heads/excluded").Validate() // → nil
RefSpec("^refs/heads/*").IsNegative()      // → true
RefSpec("^refs/heads/*").Src()              // → "refs/heads/*"
```

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l config/refspec.go config/refspec_test.go` — clean
- [x] `go test -race ./config/...` — all pass
- [x] `go test -race -short ./...` — all pass (the SSH-transport suite wants `SSH_KNOWN_HOSTS` set locally; passes with it and CI sets it up per-platform)
- [x] `golangci-lint run ./...` — 0 issues

### DIFF
--- a/config/refspec.go
+++ b/config/refspec.go
@@ -10,6 +10,7 @@ import (
 const (
 	refSpecWildcard  = "*"
 	refSpecForce     = "+"
+	refSpecNegative  = "^"
 	refSpecSeparator = ":"
 )
 
@@ -32,6 +33,13 @@ type RefSpec string
 
 // Validate validates the RefSpec
 func (s RefSpec) Validate() error {
+	// Negative refspecs ("^<pattern>") exclude matching refs from fetch/push
+	// and have no destination, so the usual "<src>:<dst>" shape does not apply.
+	// See git-fetch(1).
+	if s.IsNegative() {
+		return nil
+	}
+
 	spec := string(s)
 	if strings.Count(spec, refSpecSeparator) != 1 {
 		return ErrRefSpecMalformedSeparator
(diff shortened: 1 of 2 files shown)


BootstrapperSBL:
Hey, just checking if you've had a chance to look at this — happy to adjust anything if needed.

PR #2012, opened 2026-04-22
utils: sync, Make zlib compression pluggable via x/plugin

stiak:
Adds a pluggable zlib compression path to go-git by wiring it into the existing `x/plugin` registry. The stdlib `compress/zlib` is registered as the default provider during `x/plugin`'s init, so existing behaviour is unchanged — but applications can now register an alternative (for example `github.com/klauspost/compress/zlib`) without go-git taking a direct dependency on it.

See EXTENDING.md for the full example.

Registration follows the same freeze-on-first-use lifecycle as other plugins: once go-git has resolved the provider (on the first pool miss or NewZlibWriter call) subsequent plugin.Register calls return plugin.ErrFrozen and the active provider is kept.

### DIFF
--- a/EXTENDING.md
+++ b/EXTENDING.md
@@ -104,7 +104,7 @@ func (klauspostProvider) NewReader(r io.Reader) (plugin.ZlibReader, error) {
     }
     zlr, ok := zr.(plugin.ZlibReader)
     if !ok {
-        return nil, fmt.Errorf("klauspost reader %T does not implement Resetter", zr)
+        return nil, fmt.Errorf("klauspost reader %T does not implement plugin.ZlibReader", zr)
     }
     return zlr, nil
 }
@@ -114,9 +114,12 @@ func (klauspostProvider) NewWriter(w io.Writer) plugin.ZlibWriter {
 }
 
 func init() {
-    plugin.Register(plugin.Zlib(), func() plugin.ZlibProvider {
+    err := plugin.Register(plugin.Zlib(), func() plugin.ZlibProvider {
         return klauspostProvider{}
     })
+    if err != nil {
+        panic(err)
+    }
 }
 ```
 
--- a/EXTENDING.md
+++ b/EXTENDING.md
@@ -85,7 +85,10 @@ Install a provider during program init, before any `go-git` operation runs:
 
 ```go
 import (
+    "io"
+
     kpzlib "github.com/klauspost/compress/zlib"
+
     gogitsync "github.com/go-git/go-git/v6/utils/sync"
 )
 
--- a/EXTENDING.md
+++ b/EXTENDING.md
@@ -77,6 +77,41 @@ The default hash functions can be changed by calling `hash.RegisterHash`.
 
 New `SHA1` or `SHA256` hash functions that implement the `hash.RegisterHash` interface can be registered by calling `RegisterHash`.
 
+## Compression
+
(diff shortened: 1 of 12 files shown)


utils/sync/zlib.go
```diff
@@ -3,37 +3,167 @@ package sync
...
+// compressing or decompressing data is undefined: existing sync.Pool
+// entries remain on the old provider until the garbage collector
+// evicts them, and callers that have already acquired a reader or
+// writer continue to use the old one.
+//
+// SetZlibProvider panics if p is nil. A typed-nil provider (a non-nil
+// interface wrapping a nil concrete value) cannot be detected here
+// and will panic on first use.
+func SetZlibProvider(p ZlibProvider) ZlibProvider {
+	if p == nil {
```
pjbgf: On the new implementation, we are probably better off ignoring registration and falling back to a [default/built-in provider](https://github.com/go-git/go-git/blob/137874facf9259abed21263b16fae92bae3b4f21/x/plugin/plugin_config.go#L9-L13).

utils/sync/zlib.go
```diff
@@ -3,37 +3,167 @@ package sync
...
+// factory methods (often via internal sync.Pool instances) whenever it
+// needs to read or write zlib-compressed data. Install a non-default
+// provider with SetZlibProvider — for example to swap in
+// github.com/klauspost/compress/zlib without go-git taking a direct
+// dependency on it.
+//
+// Implementations must be safe for concurrent calls to NewReader and
+// NewWriter. Returned readers and writers need not be concurrency-safe
+// themselves; each caller gets its own instance.
+type ZlibProvider interface {
```
pjbgf: Let's build this based on the new plugin approach, an example can be found [here](https://github.com/go-git/go-git/blob/main/x/plugin/plugin_config.go) for definition, and [here](https://github.com/go-git/go-git/tree/main/x/plugin/config) for implementations.

pjbgf (changes requested):
@stiak thanks for proposing this: I `100%` agree this will be a nice addition to the project.

Let's just align the implementation of the factor/provider registration with the [x/plugin](https://github.com/go-git/go-git/tree/main/x/plugin).

utils/sync/zlib.go
stiak: Thanks, I'll move it across.

stiak: This should be done now.

EXTENDING.md
```diff
@@ -77,6 +77,52 @@ The default hash functions can be changed by calling `hash.RegisterHash`.
...
 New `SHA1` or `SHA256` hash functions that implement the `hash.RegisterHash` interface can be registered by calling `RegisterHash`.
 
+## Compression
+
+`go-git` uses zlib compression for loose objects and packfile entries. By default it uses the Go standard library's `compress/zlib`, registered at init time as the [`plugin.Zlib()`](x/plugin/plugin_zlib.go) provider. Register an alternative `plugin.ZlibProvider` — for example [`github.com/klauspost/compress/zlib`](https://github.com/klauspost/compress) — to swap the implementation without `go-git` taking a direct dependency on it.
+
+Register a provider during program init, before any `go-git` operation runs. Registration uses the [plugin system](#plugin-system-experimental) so it follows the same freeze-on-first-use lifecycle as other plugins:
+
+```go
+import (
```
pjbgf, line 87: As a follow-up, we could add this implementation to https://github.com/go-git/x.

pjbgf (approved):
@stiak thanks for working on this! 🙇 

I've rebased and made a few additional changes to your original PR:
- Added a new validation primitive so that plugins can be rejected at registration time. This further decreases the chances of an invalid zlib plugin causing a panic at runtime.
- In the unlike case of the zlib not being registered, falls back to default.
- A few nits around documentation.

stiak:
Brilliant, thanks @pjbgf !

PR #2013, opened 2026-04-24
git: add repository and remote archive support

aymanbagabas:
This implements the high level porcelain end of `git-upload-archive` and the `git archive` equivalent APIs. It adds a `repo.Archive(opts)` and `git.ArchiveRemote(url, opts)` methods to create archives of local and remote repositories.

This continues the work that was introduced in https://github.com/go-git/go-git/pull/1986

### DIFF
--- a/archive.go
+++ b/archive.go
@@ -1,0 +1,67 @@
+package git
+
+import (
+	"context"
+	"errors"
+	"io"
+
+	"github.com/go-git/go-git/v6/plumbing/transport"
+)
+
+// ArchiveRemote creates an archive from a remote repository.
+// It returns an io.ReadCloser that yields the archive data.
+// The caller must close the returned ReadCloser.
+func ArchiveRemote(url string, o *ArchiveOptions) (io.ReadCloser, error) {
+	return ArchiveRemoteContext(context.Background(), url, o)
+}
+
+// ArchiveRemoteContext creates an archive from a remote repository.
+// The provided Context can be used to cancel the operation.
+func ArchiveRemoteContext(ctx context.Context, url string, o *ArchiveOptions) (io.ReadCloser, error) {
+	if o == nil {
+		o = &ArchiveOptions{}
+	}
+	if err := o.Validate(); err != nil {
+		return nil, err
+	}
+	if url == "" {
+		return nil, errors.New("remote URL is required")
+	}
+
+	cl, req, err := newClient(url, o.ClientOptions)
+	if err != nil {
+		return nil, err
+	}
+
+	req.Command = transport.UploadArchiveService
+	sess, err := cl.Handshake(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	// Check if session implements Archiver
+	arch, ok := sess.(transport.Archiver)
+	if !ok {
+		_ = sess.Close()
+		return nil, transport.ErrArchiveUnsupported
+	}
+
+	// Build arguments
+	var args []string
+	if o.Format != "" {
+		args = append(args, "--format="+o.Format)
+	}
+	if o.Prefix != "" {
+		args = append(args, "--prefix="+o.Prefix)
+	}
+	args = append(args, o.Treeish)
+	if len(o.Paths) > 0 {
+		args = append(args, "--")
+		args = append(args, o.Paths...)
+	}
+
+	return arch.Archive(ctx, &transport.ArchiveRequest{
+		Args:     args,
+		Progress: o.Progress,
+	})
+}
--- a/archive_test.go
+++ b/archive_test.go
@@ -1,0 +1,239 @@
+package git
+
+import (
+	"archive/tar"
+	"bytes"
+	"io"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	fixtures "github.com/go-git/go-git-fixtures/v6"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/go-git/go-git/v6/internal/transport/test"
+)
+
+func TestArchiveOptionsValidate(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name    string
+		opts    *ArchiveOptions
+		wantErr string
+	}{
+		{
+			name:    "valid hash treeish",
+			opts:    &ArchiveOptions{Format: "tar", Treeish: "8ab686eafeb1f44702738c8b0f24f2567c36da6d"},
+			wantErr: "",
+		},
+		{
+			name:    "valid branch treeish",
+			opts:    &ArchiveOptions{Format: "zip", Treeish: "master"},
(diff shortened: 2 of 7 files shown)


archive_test.go
```diff
@@ -0,0 +1,41 @@
+package git
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestArchiveRemote(t *testing.T) {
```
aymanbagabas, line 103: We already test the remote transport logic in our [git transport tests](https://github.com/aymanbagabas/go-git/blob/f9a4d6228aebc7151131369ac032ab6e7acda94e/plumbing/transport/git/git_test.go#L146). This ensures the high-level API does validation.

repository.go
```diff
@@ -1432,6 +1436,62 @@ func (r *Repository) PushContext(ctx context.Context, o *PushOptions) error {
...
+	Paths []string
+	// ClientOptions are options for the transport client (used by ArchiveRemote).
+	ClientOptions []client.Option
+	// Progress receives human-readable status from the remote server.
+	// Only used by ArchiveRemote, ignored by Archive.
+	Progress sideband.Progress
+}
+
+// Validate validates the ArchiveOptions.
+func (o *ArchiveOptions) Validate() error {
```
pjbgf, line 1462: Let's add stricter rules here please, for example:
- `Format` must be empty or one of `tar`, `tar.gz`, `tgz`, `zip`.
- `Prefix` must be empty or a valid path component, not containing `..`.
- `Treeish` must be a sha1/sha256 valid hex, or a valid `heads` name.
- `Paths` must be empty, or a valid relative path not containing `..`.

pjbgf (changes requested):
Overall, LGTM. Let's just tighten the validation on `ArchiveOptions` please as per comment below.

repository.go
aymanbagabas: Hey @pjbgf,

I made some changes to validate the treeish during `git.Archive` and before calling `WriteArchive`. The rules here are intentional for different reasons.
- For `Format`, we do the validation in `git.Archive` because a remote server and `git.ArchiveRemote` can have different supported archive formats than local.
- The `Prefix` here is the prepended prefix component added to the files being archived. [Canonical git allows path](https://git-scm.com/docs/git-archive#Documentation/git-archive.txt---prefixprefix) components containing `../` and leave unsafe non-normalized prefixes to be the responsibility of the extractor. Many `tar` and `unzip` tools reject extracting to `../` by default.
- Since we moved resolving the tree before actually calling `WriteArchive`, we validate the tree is valid and has either valid sha1/sha256 or valid heads.
- Paths are just filters, if a path doesn't exist, we already return `pathspec did not match any files` similar to canonical git. This validation cannot happen during `Validate` without resolving the tree and start writing the archive.

aymanbagabas: @pjbgf Prefix now rejects paths with `..` components and absolute paths.

archive_test.go
```diff
@@ -0,0 +1,138 @@
...
+			url:     "",
+			opts:    &ArchiveOptions{Format: "tar", Treeish: "master"},
+			wantErr: "remote URL is required",
+		},
+		{
+			name:    "remote allows custom format",
+			url:     "file:///tmp/repo.git",
+			opts:    &ArchiveOptions{Format: "tar.xz", Treeish: "master"},
+			wantErr: "",
+		},
```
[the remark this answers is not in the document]
aymanbagabas: File here is the transport and our client resolved the url to our default file transport. Custom transports can be added via `ClientOptions`. If a transport is not supported or valid, we error during the construction of client.

archive_test.go
```diff
@@ -0,0 +1,239 @@
...
+			opts:    &ArchiveOptions{Format: "rar", Treeish: "HEAD"},
+			wantErr: "",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			err := tt.opts.Validate()
+			if tt.wantErr == "" {
```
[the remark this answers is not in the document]
pjbgf: No longer the case due to Go version.

archive_test.go
```diff
@@ -0,0 +1,133 @@
...
+			name:    "unsupported format",
+			opts:    &ArchiveOptions{Format: "rar", Treeish: "HEAD"},
+			wantErr: "",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			err := tt.opts.Validate()
```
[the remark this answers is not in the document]
pjbgf: No longer the case due to Go version.

archive.go
```diff
@@ -0,0 +1,67 @@
...
+	req.Command = transport.UploadArchiveService
+	sess, err := cl.Handshake(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	// Check if session implements Archiver
+	arch, ok := sess.(transport.Archiver)
+	if !ok {
+		_ = sess.Close()
```
pjbgf, line 45: Shall we just `defer` this instead on `L41`?

repository_test.go
```diff
@@ -4076,3 +4080,223 @@ func TestCreateTagSignerSelection(t *testing.T) { //nolint:paralleltest // modif
...
+			wantErr: "cannot resolve",
+		},
+		{
+			name:     "pathspec no match",
+			opts:     ArchiveOptions{Format: "tar", Treeish: "master", Paths: []string{"nonexistent/path/xyzzy"}},
+			wantRead: "pathspec",
+		},
+	}
+
+	for _, tt := range tests {
```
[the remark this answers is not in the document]
pjbgf: No longer the case due to Go version.

repository_test.go
```diff
@@ -4076,3 +4080,223 @@ func TestCreateTagSignerSelection(t *testing.T) { //nolint:paralleltest // modif
...
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			r := setupArchiveRepo(t)
```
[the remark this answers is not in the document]
pjbgf: No longer the case due to Go version.

pjbgf (commented):
Tiny nit on `defer sess.Close()`. Apart from that, please rebase it.

archive.go
aymanbagabas: Oh, we can't close the session as that would abort the archive on the server side. The `transport.StreamSession.Archive` passes a read closer that closes the session after the archive has been read. https://github.com/aymanbagabas/go-git/blob/104343e4ca3f8ec693b074ed9ce9a694eee770c1/plumbing/transport/pack_stream.go#L150

pjbgf (approved):

archive.go
pjbgf: Fair point, thanks for calling it out.

PR #2014, opened 2026-04-24
internal: servers, add git server implementation

aymanbagabas:
This adds a new Git TCP server implementation based on the existing backend package. All the tests that use the `internal/server.All()` for testing now will also use a git daemon `git://` TCP server in addition to the existing HTTP server.

I assume this can eventually grow to offer a full `git-daemon` implementation that can replace [go-git/cli/server/git](https://github.com/go-git/cli/tree/main/server/git)

### DIFF
--- a/internal/server/git/git.go
+++ b/internal/server/git/git.go
@@ -26,6 +26,8 @@ import (
 
 const defaultAddr = "127.0.0.1:0"
 
+var errAlreadyStarted = errors.New("git: server already started")
+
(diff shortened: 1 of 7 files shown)


pjbgf (dismissed):

I added two additional commits:
1. Enforce safe flows (e.g. can't call `Start()` on a started server) for both git and http implementation.
2. Moved some of  the tests into a suite to exercise across all implementations.

pjbgf (approved):

Issue #1796, opened 2025-12-10
SshBackend


manland:
### Feature Description

Today, go-git supports [http](https://github.com/go-git/go-git/blob/main/backend/http/http.go) and [git](https://github.com/go-git/go-git/blob/main/backend/git/git.go) backend. I want to add ssh.

Context: this discussion follows https://github.com/go-git/go-git/issues/1614 between me, @aymanbagabas and @pjbgf 

In [this small project](https://github.com/manland/mingogit) I implement a monorepository server, using ssh as the sole transport protocol. The "decoration" required for it to work, could be integrated directly in go-git. It is the [git](https://github.com/go-git/go-git/blob/main/backend/git/git.go) backend with ssh support.

I agree [with @aymanbagabas](https://github.com/go-git/go-git/issues/1614#issuecomment-3637589251) that we could have a `stateless` and a `statefull` backend. The first one for http and the second one for ssh and git.

## What problems should we discuss?

I see multiple problem in the current `backend` architecture I want to talk about.

- How an implementer can accept/refuse a command from a user?
- How an implementer can know what the user has pushed/pulled in the repo?

To summarize:

What are the limits of the `backend`? What is its responsibility? Should it handle a lot or a minimum amount of logic?

Happy to discuss this with you.

pjbgf:
I think we should be able to support a basic feature set and ensure they are capable to have middleware injected to support things like auth, logging, etc - those not necessarily implemented in-tree.

Key use-cases I can see:
- Enable low-volume stand-alone Git servers.
- Improve testing inside `go-git`, replacing the [existing third-party dependency](https://github.com/go-git/go-git/blob/8cea26cfef496d17267edd9c3f43fdb8e171881f/plumbing/transport/ssh/common_test.go#L7) for ssh for example.

It would be interesting to understand the specific use cases you are trying to accomodate. For things just as "accept/refuse" commands from a user, what did you have in mind? Overall, this could fit into the middleware bill depending on the API we define for it.

aymanbagabas:
Hi @manland, I've proposed this unified backend API to make it protocol agnostic but also implements `http.Handler` as a convenience. https://github.com/aymanbagabas/go-git/pull/16. Please let me know what you think

manland:
@aymanbagabas That sounds good, I'll give it a go to see if it works and if it's easy to do with ssh.

[...]

PR #2016, opened 2026-04-24
plumbing: transport/upload_pack multi round bugs

manland:
if client/server concensus need multi round
- deadlock was caused by not listening to writec
- final ack was not send if it was found in previous round

related to #1796


pjbgf:
@aymanbagabas any thoughts on this one?

aymanbagabas:
@pjbgf Looks legit. @manland Could you write a test for this to avoid future regressions?

pjbgf (approved):
@manland thanks for working on this. 🙇 

I added a test to avoid future regressions.

PR #2017, opened 2026-04-24
backend: receive_pack hook implem

manland:
first implementation of what could look like hooks in backend, only for receive_pack for now, byt easily extandable with opts

related to https://github.com/go-git/go-git/issues/1796

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -13,6 +13,7 @@ import (
 	"log"
 	"net/url"
 
+	"github.com/go-git/go-git/v6/plumbing"
 	"github.com/go-git/go-git/v6/plumbing/protocol/packp"
 	"github.com/go-git/go-git/v6/plumbing/transport"
 )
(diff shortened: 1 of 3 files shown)


pjbgf (commented):
Added @aymanbagabas for thoughts on the approach. I'll give this a try and ponder a bit around the API.

aymanbagabas:
I think we should implement git quarantine storers before working on hooks. Canonical git uses quarantine environments when receiving packs. See https://git-scm.com/docs/git-receive-pack#_quarantine_environment

From an API perspective, it makes sense to keep it as an option to `transport.ReceivePack`. I did some work [here](https://gist.github.com/aymanbagabas/0a25bca2af87b4ad8c720ed7ee365bff) related to the API and I think the best way is to define some sort of `type HookEnv struct` that all hooks receive as a parameter. Something like this I think would be better and modular.

```go
// HookEnv carries context for a server-side hook invocation.
type HookEnv struct {
    // Storer is the repository storage the hook operates on.
    // For pre-receive, this is the transaction storer — objects are
    // readable (falls through to base) but not yet committed to
    // permanent storage.
    // For all other hooks, this is the base storer.
    Storer storage.Storer
    // GitProtocol is the GIT_PROTOCOL value (e.g. "version=2").
    GitProtocol string
    // PushOptions are the push options sent by the client.
    // Nil means push options were not negotiated.
    PushOptions *packp.PushOptions
    // Progress is where the hook can write progress at. Nil means
    // progress is not supported.
    Progress io.Writer
}

// Hooks aggregates server-side hook implementations.
// Nil fields mean no hook is configured for that event.
type Hooks struct {
    PreReceive  PreReceiveHook
    Update      UpdateHook
    PostReceive PostReceiveHook
    PostUpdate  PostUpdateHook
}

// PreReceiveHook is called once before any references are updated.
// Receives all commands and push options. If it returns a non-nil error,
// the entire push is rejected and quarantined objects are discarded.
type PreReceiveHook interface {
    PreReceive(ctx context.Context, env HookEnv, cmds []*packp.Command) error
}

// PreReceiveHookFunc is a function adapter for PreReceiveHook.
type PreReceiveHookFunc func(ctx context.Context, env HookEnv, cmds []*packp.Command) error

func (f PreReceiveHookFunc) PreReceive(ctx context.Context, env HookEnv, cmds []*packp.Command) error {
    return f(ctx, env, cmds)
}
```

We would then have a default `FilesystemHooks() *Hooks` implementation that uses `filesystem.Storage` and the `.hooks` directory and the new local git config based hooks to execute hooks.

manland:
Yeah @aymanbagabas, it was more of a PoC to discuss the API rather than a finished dev. I definitely love what you did in https://gist.github.com/aymanbagabas/0a25bca2af87b4ad8c720ed7ee365bff, so let's continue the discussion there.

I need it for my project, but it's far from being stable for go-git, so I will continue to keep my implementation in my fork, keeping it as close as possible to your proposal so that I can use it as soon as it is merged.

If you need my help, please ping me! I don't have enough experience to implement the transactions, but I can help with the API or test it in real-world usage if you need.

PR #2041, opened 2026-04-29
docs: update comment to reflect support for the last 2 stable Go versions

alexandear:
As per [rfc-0001](https://github.com/go-git/go-git/blob/main/rfcs/0001-supported-go-versions.md), go-git supports only the last two stable Go versions.

Currently, the stable Go versions are 1.25 and 1.26.


pjbgf (approved):
@alexandear nice catch, thanks for that. 🙇

PR #2042, opened 2026-04-29
storage: filesystem, dedupe object iteration

pjbgf:
The shared seen map in lazyPackfilesIter was never written to by packfileIter.Next, so the same object hash could be returned once per packfile it appeared in. Mark each returned hash as seen so cross-pack iteration over IterEncodedObjects no longer double-counts.

Also tighten resource handling on iterator error paths: defer Close in ForEach for both packfileIter and objectsIter, close the inner iter and clear cur in lazyPackfilesIter on Next errors, and close the pack file in newPackfileIter when GetByType fails (when keepPack is false).

Update the upload-pack test suite to fetch into s.EmptyStorer with absolute pre/post counts. The previous before/after diff against s.Storer relied on the duplicate-counting bug, since s.Storer aliased the server-side fixture filesystem.

Add table-driven tests for the iterator behaviour: seen-map skipping, shared-seen dedupe across iterators, ForEach closing the pack on every return path (including callback error), and newPackfileIter closing the pack file when GetByType returns an error.

### DIFF
--- a/internal/transport/test/upload_pack.go
+++ b/internal/transport/test/upload_pack.go
@@ -10,9 +10,11 @@ import (
 	"github.com/stretchr/testify/suite"
 
 	"github.com/go-git/go-git/v6/plumbing"
+	"github.com/go-git/go-git/v6/plumbing/format/config"
 	"github.com/go-git/go-git/v6/plumbing/protocol/packp/capability"
 	"github.com/go-git/go-git/v6/plumbing/transport"
 	"github.com/go-git/go-git/v6/storage"
+	"github.com/go-git/go-git/v6/storage/memory"
 )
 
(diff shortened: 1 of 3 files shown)


aymanbagabas (commented):
The difference between `s.Storer`and `s.EmptyStorer` is the referenced git repository. One uses `basic.git` and the other uses `empty.git`. Both refer to actual repositories. I think for clarity, we should replace the empty storer with a non-fixture `storage.Storer` initialization.

FWIW, I opened https://github.com/go-git/go-git/pull/2043 to try and fix the ssh knownhosts concurrent test issue found here https://github.com/go-git/go-git/actions/runs/25137758637/job/73680183936?pr=2042

PR #2043, opened 2026-04-30
plumbing: ssh, fix concurrent test failures due to cert key generation

aymanbagabas:
Before we added t.Parallel to run test in parallel, the test known_hosts file was generated once and memoized for reuse across test functions. This was done to avoid the overhead of generating new keys for each test function, which can be slow. However, with tests now running in parallel, this shared state caused concurrent test failures when multiple tests tried to generate and memoize cert keys at the same time.

Moreover, the key generation in general takes a negligible amount of time compared to the overall test runtime, so the memoization optimization is not as critical as it once was.

Fixes: https://github.com/go-git/go-git/actions/runs/25137758637/job/73680183936?pr=2042


pjbgf (approved):

PR #2046, opened 2026-04-30
_examples: remove deprecated usage of io/ioutil

alexandear:
[io/ioutil](https://pkg.go.dev/io/ioutil) is deprecated:

> Deprecated: As of Go 1.16, the same functionality is now provided by package [io](https://pkg.go.dev/io) or package [os](https://pkg.go.dev/os), and those implementations should be preferred in new code. See the specific function documentation for details.

### DIFF
--- a/_examples/restore/main.go
+++ b/_examples/restore/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
-	"io/ioutil"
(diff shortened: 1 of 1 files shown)


pjbgf (approved):
@alexandear thanks for working on this. 🙇

Issue #1917, opened 2026-03-23
Race condition in `v6/plumbing/object.(*Tree).FindEntry()`


pjbgf:
### Bug Description

A recent pipeline execution failed due to the data race below:

```
==================
WARNING: DATA RACE
Read at 0x00c00042a060 by goroutine 82:
  github.com/go-git/go-git/v6/plumbing/object.(*Tree).FindEntry()
      /home/runner/work/go-git/go-git/plumbing/object/tree.go:131 +0x64
  github.com/go-git/go-git/v6.treeContainsDirs()
      /home/runner/work/go-git/go-git/worktree.go:398 +0xa4
  github.com/go-git/go-git/v6.TestTreeContainsDirs.func1()
      /home/runner/work/go-git/go-git/worktree_test.go:3516 +0x56
  testing.tRunner()
      /opt/hostedtoolcache/go/1.24.13/x64/src/testing/testing.go:1792 +0x225
  testing.(*T).Run.gowrap1()
      /opt/hostedtoolcache/go/1.24.13/x64/src/testing/testing.go:1851 +0x44

Previous write at 0x00c00042a060 by goroutine 83:
  github.com/go-git/go-git/v6/plumbing/object.(*Tree).FindEntry()
      /home/runner/work/go-git/go-git/plumbing/object/tree.go:132 +0x8d
  github.com/go-git/go-git/v6.treeContainsDirs()
      /home/runner/work/go-git/go-git/worktree.go:398 +0xa4
  github.com/go-git/go-git/v6.TestTreeContainsDirs.func1()
      /home/runner/work/go-git/go-git/worktree_test.go:3516 +0x56
  testing.tRunner()
      /opt/hostedtoolcache/go/1.24.13/x64/src/testing/testing.go:1792 +0x225
  testing.(*T).Run.gowrap1()
      /opt/hostedtoolcache/go/1.24.13/x64/src/testing/testing.go:1851 +0x44

Goroutine 82 (running) created at:
  testing.(*T).Run()
      /opt/hostedtoolcache/go/1.24.13/x64/src/testing/testing.go:1851 +0x8f2
  github.com/go-git/go-git/v6.TestTreeContainsDirs()
      /home/runner/work/go-git/go-git/worktree_test.go:3514 +0x473
  testing.tRunner()
      /opt/hostedtoolcache/go/1.24.13/x64/src/testing/testing.go:1792 +0x225
  testing.(*T).Run.gowrap1()
[...]

PR #2047, opened 2026-04-30
git: worktree, fix treeContainsDirs test to build a new tree for each…

aymanbagabas:
… test case

This makes the TestTreeContainsDirs test more robust by ensuring that each test case operates on a fresh tree instance. This prevents any unintended side effects from previous test cases.

The treeContainsDirs function may write to the tree and cause a race condition when a single tree is shared across multiple test cases. By building a new tree for each test case, we ensure that the tests are isolated and can be run in parallel without interference.

See tree.go:185-189:
```go
  func (t *Tree) entry(baseName string) (*TreeEntry, error) {
      if t.m == nil {       // read t.m
          t.buildMap()      // write t.m
      }
      entry, ok := t.m[baseName]  // read t.m
```

Fixes: https://github.com/go-git/go-git/actions/runs/25163491857/job/73763612343?pr=2014

### DIFF
--- a/worktree_test.go
+++ b/worktree_test.go
@@ -3760,13 +3760,15 @@ func (s *WorktreeSuite) TestLinkedWorktree() {
 
 func TestTreeContainsDirs(t *testing.T) {
 	t.Parallel()
-	tree := &object.Tree{
-		Entries: []object.TreeEntry{
-			{Name: "foo", Mode: filemode.Dir},
-			{Name: "bar", Mode: filemode.Dir},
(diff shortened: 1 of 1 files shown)


onee-only (approved):
Related: https://github.com/go-git/go-git/issues/1917
Although this wouldn't fix the root cause of the issue, test isolation still is a great improvement.

pjbgf (approved):

@onee-only thanks for the initial review. 🥇

PR #2048, opened 2026-04-30
worktree: skip ignored directories during Status walk

Soph:
## Summary

`Worktree.Status()` walks the working tree via the filesystem merkletrie noder and then drops gitignored entries in `excludeIgnoredChanges` (worktree_status.go:175-177). Because the walker has no concept of `.gitignore`, every file under a large ignored directory like `node_modules` is `lstat`'d and hashed before being filtered out. CLI git skips such directories at the directory level via `treat_directory` / `read_directory_recursive` in [dir.c](https://github.com/git/git/blob/master/dir.c), which is why `git status` is effectively constant-time in the size of `node_modules`.

This PR plumbs an `IgnoreMatcher` into `filesystem.Options` so the walker can skip ignored entries during traversal, while preserving the "tracked entries are still walked" semantics that match `git status`.

## Design

- New exported interface `filesystem.IgnoreMatcher` (single method, satisfied by `gitignore.Matcher` — kept as an interface to avoid pulling `gitignore` into a leaf utility package).
- New optional field `filesystem.Options.IgnoreMatcher`.
- At root construction, when both `Index` and `IgnoreMatcher` are set, build a `trackedDirs` set of directory prefixes that contain index entries.
- In `calculateChildren`, skip a child entry when:
  - it matches the matcher, **and**
  - it has no entry in the index (files), or contains no tracked entry (directories).
- In `worktree_status.go`, load `gitignore.ReadPatterns` once per Status call and reuse the slice for both the matcher and the post-walk `excludeIgnoredChanges` filter (it previously re-read patterns from disk).

The post-walk filter is kept as a defense-in-depth backstop; with the new path it has no work to do but the cost is negligible.

## Correctness vs. upstream git

Behavior preserved:

- **Untracked + ignored → not reported.** Same as `git status`.
- **Tracked + matches ignore → still reported.** Mirrors `git status` behavior for files that were committed before an ignore rule was added (e.g. `git add -f` then later ignored). The walker descends into directories that contain tracked entries even if the directory matches an ignore pattern.
- **Untracked siblings of a tracked-but-ignored file → still filtered.** When the walker descends into an ignored directory because a tracked file lives there, untracked siblings are filtered per-entry.

No change to:
- `diffCommitWithStaging` (left side of Status) — only the worktree walk is affected.
- `excludeIgnoredChanges` semantics — still drops `len(ch.From) == 0` entries that match.
- Any code path that doesn't pass an `IgnoreMatcher` (this is opt-in via Options).

## Benchmark

`BenchmarkStatusIgnoredDir` (added in this PR) times `Status()` over a tree with 100 tracked source files plus N untracked files inside a gitignored `vendor_ignored/` directory. Apple M4 Max, `-benchtime=10x`:

| Untracked files | Before | After | Speedup |
|---|---|---|---|
| 0 | 1.5 ms | 1.9 ms | — |
| 1,000 | 7.0 ms | 4.4 ms | 1.6× |
| 5,000 | 18.2 ms | 4.4 ms | 4.2× |
| 20,000 | 58.4 ms | 4.4 ms | **13.3×** |

The "After" column is flat in the size of the ignored tree — the walker no longer descends. The ~0.4 ms overhead at N=0 is the unchanged `gitignore.ReadPatterns` walk plus the small `trackedDirs` map; it scales with the *tracked* tree, not the ignored one.

## Tests

Unit tests in `utils/merkletrie/filesystem/node_ignore_test.go`:
- `TestIgnoredDirIsSkipped` — directory matching the matcher with no tracked content is not walked.
- `TestTrackedFileInIgnoredDirReportsModify` — a tracked file inside an ignored directory still surfaces as `Modify` when its content changes.
- `TestUntrackedSiblingsInIgnoredDirAreSkipped` — when the walker descends into an ignored directory because of a tracked file, untracked siblings are still filtered.

Integration test in `worktree_status_test.go`:
- `TestStatusReportsModifiedTrackedFileInIgnoredDirectory` — same invariant as the second unit test, exercised end-to-end through `Worktree.Status()`.

Existing test suites (`Status*`, `TestNoder*`, gitignore) remain green.

## API surface

Additive only:
- New exported interface `filesystem.IgnoreMatcher`.
- New field `filesystem.Options.IgnoreMatcher`.

`gitignore.Matcher` satisfies the new interface implicitly, so existing callers can opt in by setting the field.

## AI disclosure

Per `AI_POLICY.md`, this contribution was assisted by Claude Opus 4.7 (commit trailers). The author has reviewed every line and is accountable for correctness; the design (interface, `trackedDirs` building, the "tracked → still walked" invariant) was specified by the author and validated against upstream git's `dir.c` behavior.


worktree_status_ignored_bench_test.go
```diff
@@ -0,0 +1,102 @@
...
+	require.NoError(b, err)
+
+	wt, err := repo.Worktree()
+	require.NoError(b, err)
+
+	for i := range tracked {
+		path := filepath.Join("src", fmt.Sprintf("dir%02d", i%10), fmt.Sprintf("file%04d.go", i))
+		require.NoError(b, wt.Filesystem.MkdirAll(filepath.Dir(path), 0o755))
+		require.NoError(b, util.WriteFile(wt.Filesystem, path, []byte("package main\n"), 0o644))
+	}
```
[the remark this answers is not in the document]
Soph: Go 1.22+ supports for i := range int

worktree_status_ignored_bench_test.go
```diff
@@ -0,0 +1,102 @@
...
+	_, err = wt.Commit("initial", &CommitOptions{Author: sig, Committer: sig})
+	require.NoError(b, err)
+
+	// Drop a large *gitignored* untracked tree. None of these files affect
+	// status, but go-git lstat's and hashes each one.
+	for i := range untracked {
+		path := filepath.Join(ignoredDir, fmt.Sprintf("sub%02d", i%20), fmt.Sprintf("dep%05d.txt", i))
+		require.NoError(b, wt.Filesystem.MkdirAll(filepath.Dir(path), 0o755))
+		require.NoError(b, util.WriteFile(wt.Filesystem, path, []byte("ignored\n"), 0o644))
+	}
```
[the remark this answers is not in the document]
Soph: Go 1.22+ supports for i := range int

utils/merkletrie/filesystem/node.go
```diff
@@ -32,6 +40,14 @@ type Options struct {
 	// correctly handling the "racy git" condition. If no index is provided,
 	// the function works without the optimization.
 	Index *index.Index
+
+	// IgnoreMatcher, if non-nil, is consulted while walking the tree.
+	// Untracked entries (files or directories) that match the matcher are
+	// excluded from the walk so callers do not have to descend into large
+	// gitignored directories like node_modules. Tracked entries are always
+	// walked even if they match, so modifications to them are still
+	// reported. When Index is nil this option is ignored.
```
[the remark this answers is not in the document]
Soph: fixed

utils/merkletrie/filesystem/node.go
```diff
@@ -192,15 +231,38 @@ func (n *node) calculateChildren() error {
...
+	}
+	childPath := path.Join(n.path, name)
+	if !n.options.IgnoreMatcher.Match(strings.Split(childPath, "/"), isDir) {
+		return false
+	}
+	if isDir {
+		return !n.trackedDirs[childPath]
+	}
+	if n.idxMap == nil {
+		return true
```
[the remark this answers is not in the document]
Soph: fixed

utils/merkletrie/filesystem/node.go
```diff
@@ -19,6 +20,13 @@ import (
 	"github.com/go-git/go-git/v6/utils/sync"
 )
 
+// IgnoreMatcher matches paths against a set of ignore rules. It is
+// satisfied by gitignore.Matcher; the indirection avoids a direct
+// dependency on the gitignore package from this leaf utility.
+type IgnoreMatcher interface {
+	Match(path []string, isDir bool) bool
+}
```
pjbgf: Let's use type alias instead:

```suggestion
// IgnoreMatcher defines a global multi-pattern matcher for gitignore patterns.
type IgnoreMatcher interface = gitignore.Matcher
```

utils/merkletrie/filesystem/node.go
```diff
@@ -44,6 +61,11 @@ type node struct {
 	submodules map[string]plumbing.Hash
 	idx        *index.Index
 	idxMap     map[string]*index.Entry
+	// trackedDirs holds every directory path that has at least one entry
+	// in the index. It is populated only when IgnoreMatcher is set so the
+	// walker can keep tracked entries even if their parent directory
+	// matches an ignore rule.
+	trackedDirs map[string]bool
```
pjbgf:
```suggestion
	trackedDirs map[string]struct{}
```

utils/merkletrie/filesystem/node.go
```diff
@@ -88,21 +110,35 @@ func NewRootNodeWithOptions(
 	options Options,
 ) noder.Noder {
 	var idxMap map[string]*index.Entry
+	var trackedDirs map[string]bool
```
pjbgf:
```suggestion
	var trackedDirs map[string]struct{}
```

worktree_status.go
```diff
@@ -173,19 +185,20 @@ func (w *Worktree) diffStagingWithWorktree(reverse, excludeIgnoredChanges bool)
 	}
 
 	if excludeIgnoredChanges {
-		return w.excludeIgnoredChanges(c), nil
+		return w.excludeIgnoredChanges(c, ignorePatterns), nil
 	}
```
pjbgf: The PR changes results in ignored patterns being verified eagerly, which makes me think this then becomes a redundant call. Should we not drop this call and the `excludeIgnoredChanges` func altogether?

worktree_status.go
```diff
@@ -173,19 +185,20 @@ func (w *Worktree) diffStagingWithWorktree(reverse, excludeIgnoredChanges bool)
...
 	if err != nil {
-		return changes
+		patterns = nil
 	}
+	return append(patterns, w.Excludes...)
+}
 
-	patterns = append(patterns, w.Excludes...)
-
+func (w *Worktree) excludeIgnoredChanges(changes merkletrie.Changes, patterns []gitignore.Pattern) merkletrie.Changes {
```
pjbgf: As per above comment, this `func` seems to be redundant now.


```suggestion
```

pjbgf (changes requested):
@Soph thanks for looking into this. The performance results look great. 👏 

The PR is largely LGTM apart from the small bits below.

worktree_status_ignored_bench_test.go
```diff
@@ -0,0 +1,102 @@
+package git
```
pjbgf: Let's move the contents of this file into `worktree_status_bench_test.go` so that we have a single `bench_test.go` for worktree status.

onee-only:
Related issue: https://github.com/go-git/go-git/issues/181. The PR https://github.com/go-git/go-git/pull/1379 originally attempted to fix this using a similar approach. Thankfully, the author agreed to let us supersede their PR.

utils/merkletrie/filesystem/node.go
onee-only: Why don't we just use it directly?

Soph: A type alias still requires that import, so it gives up the layering benefit while introducing a second name for the type. I think then rather what @onee-only says and drop the abstraction and use `gitignore.Matcher` directly? 🤔

pjbgf: Yep, let's do that.

Soph: ok, changed.

Soph:
@pjbgf @onee-only I had the thought to check which tests the git cli has around this scenario. It had 4 tests with coverage this PR hadn't yet. So I added them and the submodule case found an actual bug.

utils/merkletrie/filesystem/node.go
```diff
@@ -88,21 +104,35 @@ func NewRootNodeWithOptions(
...
+			for _, entry := range options.Index.Entries {
+				for parent := path.Dir(entry.Name); parent != "." && parent != "/"; parent = path.Dir(parent) {
+					if _, ok := trackedDirs[parent]; ok {
+						break
+					}
+					trackedDirs[parent] = struct{}{}
+				}
+			}
+		}
 	}
```
onee-only, line 126: Maybe we can combine these `idxMap` and `trackdDirs` initializations into one loop? I'm not sure how big this would affect to the performance, but it could take advantage of locality of reference I think.

onee-only (approved):
LGTM. Thanks for the update! I left a small suggestion on `trackedDirs` initialization. I'd appreciate it if you could share your thoughts.

utils/merkletrie/filesystem/node.go
Soph: 🤓 I let claude work a bit for it's money: 

Ran a microbenchmark over `NewRootNodeWithOptions` setup at 1k/10k/100k index entries, 10 samples each, compared via `benchstat`:

| Entries | Path | Two-loop (current) | Fused | Δ |
|---|---|---:|---:|---:|
| 1k | WithMatcher | 147.7 µs | 146.6 µs | **-0.78%** |
| 10k | WithMatcher | 1.027 ms | 1.012 ms | **-1.46%** |
| 100k | WithMatcher | 7.552 ms | 7.454 ms | **-1.30%** |
| 1k | NoMatcher | 11.32 µs | 11.46 µs | **+1.32%** |
| 10k | NoMatcher | 146.1 µs | 149.1 µs | **+2.06%** |
| 100k | NoMatcher | 1.861 ms | 1.911 ms | **+2.71%** |
| **geomean** | | 390.1 µs | 391.7 µs | **+0.41%** |

All deltas are statistically significant (p ≤ 0.01). Allocations are identical.

Fusing wins 1–1.5% on the matcher path but loses 1–3% on the no-matcher path — the per-iteration `trackedDirs != nil` check costs a comparison every iteration even when nothing else runs. Geomean is +0.41%, essentially flat.

Since the no-matcher path is the common one (only `Status` with `excludeIgnoredChanges` populates `trackedDirs`), I'd rather keep the two passes — the rare-path win doesn't pay for the common-path regression, and the absolute setup cost is well under 1% of `Status()` either way on realistic repos.

pjbgf: That sounds good. Thanks for looking into it. :bow:

pjbgf (approved):

PR #1999, opened 2026-04-15
Fix file handle leaks by calling Repository.Close() and adding transactional.Storage.Close()

AriehSchneier:
## Summary
- Fix 400+ repository leaks across the test suite by calling Repository.Close()
- Add transactional.Storage.Close() method and fix 30+ storage leaks across test files
- Update all examples to demonstrate proper resource cleanup
- Add resource management documentation to CONTRIBUTING.md and copilot-instructions.md

## Problem
Unclosed repositories and storage instances leak file handles, particularly critical on Windows where LazyIndex keeps .idx files open. This causes intermittent test failures and can accumulate in long-running applications.

### Commit 1: Call Repository.Close() to fix file handle leaks
Repository.Close() closes the underlying storage if it implements io.Closer. This is critical on Windows where unclosed repositories leak file handles.

Fixed leaks in:
- Suite infrastructure (BaseSuite, RepositorySuite, RemoteSuite, WorktreeSuite, etc.)
- Error test cases requiring nil checks before Close()
- Variable reassignment patterns needing explicit Close()
- Production code error paths in Clone, Init, and Submodule methods
- All 33 _examples files to demonstrate proper Close() usage
- example_test.go examples

### Commit 2: Add transactional.Storage.Close() and fix storage handle leaks
Added Close() method to transactional.Storage to properly close both the base and temporal storages if they implement io.Closer. This prevents file handle leaks on Windows.

Note: filesystem.Storage already has Close() via embedded ObjectStorage.

Fixed leaks in:
- 30+ test files across plumbing/object, plumbing/transport, storage/filesystem
- Test helper functions that create storage instances
- Suite setup/teardown methods using t.Cleanup instead of defer in anonymous functions
- Loader tests using io.Closer type assertions
- Added nil checks before closing cached repositories

## Key Patterns
```go
// Standard repository pattern
r, err := PlainClone(...)
if err != nil {
    return err
}
defer func() { _ = r.Close() }()

// Filesystem storage pattern
sto := filesystem.NewStorage(fs, cache)
defer func() { _ = sto.Close() }()

// Transactional storage pattern
txStore := transactional.NewStorage(base, temporal)
defer func() { _ = txStore.Close() }()

// Error test cases
r, err := Init(...)
if r != nil {
    defer func() { _ = r.Close() }()
}
s.ErrorIs(err, expectedError)
```

## Testing
- All tests pass with no leaks
- Verified Windows file handle cleanup patterns

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2053, opened 2026-05-03
Add build tag-gated leak detection for Repository and Storage

AriehSchneier:
## Summary
Add optional leak detection using `runtime.SetFinalizer` to catch unclosed Repository, filesystem.Storage, and transactional.Storage instances. Detection is enabled via `-tags leakcheck` build tag to avoid overhead in production.

⚠️ **This PR builds on #1999 and should be reviewed after that PR is merged.**

## Features
- Captures stack traces at creation time for debugging
- Panics when GC collects an unclosed resource  
- Filters stack traces to show only relevant go-git/test frames
- Zero overhead when disabled (default)
- Detects double-close bugs via closed tracking

## Implementation
Build tag-gated files:
- `repository_leakcheck.go` / `repository_noleakcheck.go`
- `storage/filesystem/storage_leakcheck.go` / `storage_noleakcheck.go`  
- `storage/transactional/storage_leakcheck.go` / `storage_noleakcheck.go`

Documentation updates:
- Updated Makefile to run tests with `-tags leakcheck`
- Updated CONTRIBUTING.md with leak detection instructions
- Updated .github/copilot-instructions.md with leak detection guidelines

## How It Works

When built with `-tags leakcheck`:
```go
// setupLeakCheck is called from _leakcheck.go
func (r *Repository) setupLeakCheck() {
    stack := getStackTrace()
    runtime.SetFinalizer(r, func(r *Repository) {
        if !r.closed {
            panic("Repository was garbage collected without calling Close()\n" + stack)
        }
    })
}
```

When built without the tag (default), it's a no-op:
```go
// _noleakcheck.go
func (r *Repository) setupLeakCheck() {}
```

## Usage

```bash
# Run tests with leak detection
go test -tags leakcheck ./...

# Or use make target
make test
```

## Testing
- All tests pass with leak detection enabled
- No leaks detected in full test suite
- Verified zero overhead when disabled

## Depends On
- #1999 - Fix file handle leaks by calling Repository.Close() and adding transactional.Storage.Close()

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -45,6 +45,7 @@ When reviewing pull requests in this repository, focus on correctness, maintaina
 - **Repository cleanup**: All `Repository` instances created with `PlainClone`, `PlainInit`, `PlainOpen`, `Clone`, or `Open` must have a corresponding `defer func() { _ = repo.Close() }()` immediately after error checking.
   - Flag any repository creation where the instance is discarded with `_`. These must assign to a variable and add `defer Close()`.
   - Rationale: Prevents file handle leaks that cause intermittent Windows test failures.
+  - Leak detection is available via `-tags leakcheck` which will panic with a clear message if repositories are garbage collected without calling `Close()`.
 - **Storage cleanup**: All `Storage` instances created with `filesystem.NewStorage` must have a corresponding `defer func() { _ = storage.Close() }()` immediately after creation, **except** when the storage is passed to a repository creation function.
   - **Repository takes ownership**: When storage is passed to `Init`, `Open`, `PlainInit`, `PlainOpen`, `Clone`, `PlainClone`, or `newRepository`, do NOT add a separate `defer storage.Close()`. The repository takes ownership and will close the storage when `repo.Close()` is called.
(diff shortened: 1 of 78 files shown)


AriehSchneier:
@aymanbagabas test failures in the new server code?

aymanbagabas:
> @aymanbagabas test failures in the new server code?

Yeah, I think there is a race in there where it's not measuring the time properly.

AriehSchneier:
> > @aymanbagabas test failures in the new server code?
> 
> Yeah, I think there is a race in there where it's not measuring the time properly.

This is Claudes thoughts on the issue:

The server is setting the idle timeout deadline immediately upon accepting a connection (line 250), before any I/O occurred. On loaded CI runners, there could be significant scheduling delay between:

Server accepting connection and setting deadline to T + 200ms
Client connecting and starting its timer at T + X
When the deadline expired at T + 200ms, the client only measured 200ms - X elapsed time. With X ≈ 102ms due to CI load, the client measured ~98ms instead of the expected ≥170ms.

Fix: Only set an initial deadline when InitTimeout or MaxTimeout are configured (absolute deadlines that should start from connection accept). For idle timeout alone, let the first I/O operation trigger the deadline via the Read() method. This makes timeout behavior predictable and aligned with the test's expectations.

PR #2055, opened 2026-05-03
plumbing: transport, close storage in tests

aymanbagabas:
This ensures that the storage is closed after each test, which can help prevent resource leaks especially in Windows where tests may fail to clean up temporary directories if storage resources are still open.

Tries to fix https://github.com/go-git/go-git/actions/runs/25163782952/job/73769629975

### DIFF
--- a/plumbing/transport/serve_test.go
+++ b/plumbing/transport/serve_test.go
@@ -58,6 +58,7 @@ func testAdvertise[T UploadPackRequest | ReceivePackRequest](
 		t.Fatal(err)
 	}
 	st := filesystem.NewStorage(dot, cache.NewObjectLRUDefault())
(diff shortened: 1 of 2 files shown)


AriehSchneier:
These are probably closed in #1999

pjbgf:
@AriehSchneier one of the cases here are not catered for on #1999, so it would make sense to do both.

@aymanbagabas please rebase.

AriehSchneier:
@pjbgf If you are talking about the fix on line 141, then its already in main, the others are in my PR. I think you should try to merge the other so the failures will stop and we wont need to keep looking into them...

pjbgf:
Closing now that we already merged the changes from #1999.

PR #2064, opened 2026-05-06
internal: git, add server timeout test small tolerance

aymanbagabas:
This adds a small tolerance to workaround OS related time precision jitter. Without this, time-based tests might fail with really small deltas like 199.3ms < 200ms.


pjbgf (approved):
@aymanbagabas thanks for fixing this. 🙇

PR #2066, opened 2026-05-06
Align object encoding with upstream

pjbgf:
- Ensures that unmodified `Commit` and `Tag` objects return their raw bytes (minus `gpgsign` and `gpgsign-sha256` headers).
- Aligns `Commit` and `Tag` interpretation with upstream: first instance of standard headers wins.
- Drop `MergeTag` in favour of `ExtraHeaders`.
- Add conformance tests to ensure verification aligns with `git`. Note that duplicate signatures were valid back on `v2.11`, which is why that test is skipped during capability mode.
- Aligns `Tree` interpretation with upstream: first wins and unsorted tree is short-circuited.

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Check Package Prefix
         uses: gsactions/commit-message-checker@16fa2d5de096ae0d35626443bcd24f1e756cafee # v2.0.0
         with:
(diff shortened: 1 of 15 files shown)


pjbgf:
Git Compatibility failure is orthogonal:

```
--- FAIL: TestGitServer_Timeout (0.26s)
    git_test.go:187: 
        	Error Trace:	/home/runner/work/go-git/go-git/internal/server/git/git_test.go:187
        	Error:      	"199.310309ms" is not greater than or equal to "200ms"
        	Test:       	TestGitServer_Timeout
        	Messages:   	server should not close before Timeout elapses
FAIL
```

Will be fixed by #2013.

PR #2067, opened 2026-05-06
plumbing: transport/ssh, Shell-quote path and args

hiddeco:
The SSH transport interpolates `req.URL.Path` and each entry of `req.Args` into the remote-exec command line. The previous formatter wrapped each value in literal single quotes via `'%s'`, which diverges from canonical Git's wire format for any value containing characters that the shell treats specially.

Port `sq_quote_buf`[1] from canonical Git into a `shellQuote` helper and route both the path and each `req.Args` entry through it. The helper escapes the same character set that upstream Git escapes (`'` and `!`), so the bytes go-git puts on the wire match what `git` itself would send for the same input. Benign inputs are byte-identical to the previous output; values containing `'` or `!` now round-trip correctly through any POSIX shell and through `git-shell`'s `sq_dequote_to_argv`.

Add a table-driven `TestBuildCommand` covering the plain case, both characters that `sq_quote_buf` escapes, mixed inputs, and inert metacharacters that pass through unchanged.

[1]: https://github.com/git/git/blob/v2.54.0/quote.c#L28

### DIFF
--- a/plumbing/transport/ssh/ssh.go
+++ b/plumbing/transport/ssh/ssh.go
@@ -5,7 +5,6 @@ import (
 	"bytes"
 	"context"
 	"errors"
-	"fmt"
 	"io"
 	"net"
 	"strconv"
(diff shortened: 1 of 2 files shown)


aymanbagabas (commented):
This looks good!

aymanbagabas:
Why don't we pass a generic `io.Writer` and reuse our string builder from `buildCommand`?
```suggestion
func writeShellQuote(w io.Writer, s string) string {
```

hiddeco:
@aymanbagabas yes, good shout, and this refinement was still on my todo list. Drafting this was primarily to show @pjbgf. I will update once 👧 is 💤.

pjbgf (approved):

PR #2068, opened 2026-05-06
v5: plumbing: transport/ssh, Shell-quote path

hiddeco:
The SSH transport interpolates `ep.Path` into the remote-exec command line via `endpointToCommand`. The previous formatter wrapped the value in literal single quotes via `'%s'`, which diverges from canonical Git's wire format for any path containing characters the shell treats specially.

Port `sq_quote_buf`[1] from canonical Git into a `writeShellQuote` helper and route the path through it. The helper escapes the same character set that upstream Git escapes (`'` and `!`), so the bytes go-git puts on the wire match what `git` itself would send for the same input. Benign paths are byte-identical to the previous output; paths containing `'` or `!` now round-trip correctly through any POSIX shell and through `git-shell`'s `sq_dequote_to_argv`.

Add a table-driven `TestEndpointToCommand` covering the plain case, both characters that `sq_quote_buf` escapes, mixed inputs, and inert metacharacters that pass through unchanged.

Back-port from #2067.

[1]: https://github.com/git/git/blob/v2.54.0/quote.c#L28

### DIFF
--- a/plumbing/transport/ssh/common.go
+++ b/plumbing/transport/ssh/common.go
@@ -252,7 +252,39 @@ func (c *command) setAuthFromEndpoint() error {
 }
 
 func endpointToCommand(cmd string, ep *transport.Endpoint) string {
-	return fmt.Sprintf("%s '%s'", cmd, ep.Path)
+	var b strings.Builder
+	b.WriteString(cmd)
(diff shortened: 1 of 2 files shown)


pjbgf (approved):

PR #2070, opened 2026-05-07
v5: git: submodule, Fix relative URL resolution

hiddeco:
Reported by @N0zoM1z0, thank you! 🙇 

---

A relative submodule URL such as `url = ../X.git` in `.gitmodules` should be resolved against the parent repository's remote URL, per git-submodule(1)[1] and canonical Git. Currently it is resolved against the process's current working directory and persisted as a `file://` remote: an incorrect, working-directory-dependent target.

The cause is in `Submodule.Repository()`. The configured URL is fed through `transport.NewEndpoint`, which hands scheme-less paths to `parseFile`, which calls `filepath.Abs` unconditionally. The follow-up gate `!path.IsAbs(moduleEndpoint.Path)` is therefore always false for any input that wasn't already absolute, and the relative-URL branch never fires. The eager absolutization was introduced when the relative-URL handling was rewritten to use `transport.NewEndpoint` in 9706315a; the original `net/url.Parse` form added in 81d429bc did not exhibit the issue.

Detect the relative form from the raw `s.c.URL` string before `transport.NewEndpoint` normalizes it, using `IsLocalEndpoint` plus both `path.IsAbs` and `filepath.IsAbs` to handle POSIX and Windows absolute paths. When the form is relative, look up the parent's remote and join the raw URL against `rootEndpoint.Path` with `path.Join`. While here, replace the implicit panic on `remotes[0]` with an explicit error when the parent has no remote configured.

Add unit tests covering relative URLs against HTTPS and SSH parents, deep traversal (`../../`), absolute local URLs (whose behaviour is preserved), and the no-parent-remote error.

[1]: https://git-scm.com/docs/git-submodule

### DIFF
--- a/submodule.go
+++ b/submodule.go
@@ -6,9 +6,11 @@ import (
 	"errors"
 	"fmt"
 	"path"
+	"path/filepath"
 
 	"github.com/go-git/go-billy/v5"
 	"github.com/go-git/go-git/v5/config"
+	giturl "github.com/go-git/go-git/v5/internal/url"
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/format/index"
 	"github.com/go-git/go-git/v5/plumbing/transport"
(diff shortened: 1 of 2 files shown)


pjbgf (approved):

PR #2071, opened 2026-05-07
git: submodule, canonical remote for relative URLs

hiddeco:
A relative submodule URL like `../sibling` has to be joined onto a base URL — the superproject's remote — to produce the actual URL to clone. Submodule.Repository() picked `Repository.Remotes()[0]` as the base. That slice is built by iterating a Go map, so its order is randomized: in any superproject with more than one remote (the common `origin` + `upstream` fork setup), the chosenbase flipped between runs and the same `git submodule update` could clone the sibling from either namespace.

Canonical Git is deterministic[1] [2]: it uses `branch.<HEAD>.remote` if configured, the sole remote when there is exactly one, otherwise the literal `"origin"`. The submodule helper[3] then joins the relative URL onto that remote's URL.

Mirror that rule in an unexported `defaultRemote` and call it from the relative-URL branch. Tests cover the rule directly and the end-to-end submodule flow; the integration test loops to catch any future regression that re-introduces map-order dependence.

[1]: https://github.com/git/git/blob/v2.54.0/remote.c#L1801-L1809
[2]: https://github.com/git/git/blob/v2.54.0/remote.c#L655-L669
[3]: https://github.com/git/git/blob/v2.54.0/builtin/submodule--helper.c#L53-L75

### DIFF
--- a/internal/url/url.go
+++ b/internal/url/url.go
@@ -13,7 +13,7 @@ import (
 var (
 	isSchemeRegExp = regexp.MustCompile(`^[^:]+://`)
 
-	// Ref: https://github.com/git/git/blob/master/Documentation/urls.txt#L37
+	// Ref: https://github.com/git/git/blob/v2.54.0/Documentation/urls.adoc#L41-L48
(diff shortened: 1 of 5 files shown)


pjbgf (approved):
@hiddeco thanks for working on this! 🙇

PR #2072, opened 2026-05-07
git: submodule, Add relative URL regression tests

hiddeco:
Reported by @N0zoM1z0, thank you! 🙇 

---

Mirror the v5 fix from PR #2070 (54a32347) to v6 as regression coverage. The v5 fix corrected `Submodule.Repository()` to resolve relative submodule URLs such as `../X.git` against the parent remote URL per git-submodule(1)[1] and canonical Git, rather than absolutizing them against the process CWD.

The v5 root cause — `transport.NewEndpoint` running scheme-less inputs through `parseFile`, which called `filepath.Abs` — does not exist on v6: `internal/url.ParseFile` stores the input verbatim, so the relative-resolution branch in
`Submodule.Repository()` already fires correctly. These tests lock that invariant in so a future refactor reintroducing eager normalization in `ParseFile` fails loudly instead of silently breaking submodule resolution.

Five tests cover relative URLs against HTTPS and SSH parents, deep traversal (`../../`), absolute local URL preservation, and the no-parent-remote case. The last asserts an explicit error; v6 still had the unguarded `remotes[0]` access that PR #2070 replaced, so port the matching guard alongside the tests.

[1]: https://git-scm.com/docs/git-submodule

### DIFF
--- a/submodule_test.go
+++ b/submodule_test.go
@@ -556,3 +556,96 @@ func primaryFixtureSubmoduleName(f *fixtures.Fixture) string {
 
 	return "basic"
 }
+
+// newSubmoduleForRelativeURL constructs an in-memory Repository with
+// the given parent remote URL configured as origin, plus a Submodule
+// whose configured URL is the given submoduleURL. Pass parentRemoteURL
+// = "" to omit the origin remote entirely.
+func newSubmoduleForRelativeURL(t *testing.T, parentRemoteURL, submoduleName, submoduleURL string) *Submodule {
+	t.Helper()
+
+	repo := &Repository{
+		Storer: memory.NewStorage(),
+		wt:     memfs.New(),
+	}
+	if parentRemoteURL != "" {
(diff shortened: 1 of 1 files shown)


submodule_test.go
```diff
@@ -365,3 +365,107 @@ func primaryFixtureSubmoduleName(f *fixtures.Fixture) string {
...
+		},
+		w: worktree,
+	}
+}
+
+func TestSubmoduleRepositoryRelativeURLHTTPSParent(t *testing.T) {
+	t.Parallel()
+
+	sm := newSubmoduleForRelativeURL(t,
+		"https://example.invalid/group/proj.git", "basic", "../X.git")
```
pjbgf: Let's group this 5 tests in a table driven manner, so that we reduce bloat.

pjbgf (dismissed):

pjbgf (approved):
@hiddeco Thanks for working on this. 🙇

PR #821, opened 2023-07-28
git: worktree, reset ignored files that are part of the worktree: Fixes #819

daolis:
do not exclude ignored filese from .gitignore in diffStagingWithWorktree when resetting worktree.

PR #2075, opened 2026-05-07
worktree: skip ignored directories during Reset walk

Soph:
Reset (HardReset, KeepReset, MergeReset) walks the working tree via diffStagingWithWorktree to find tracked files that are missing or stale on disk. Without the IgnoreMatcher optimization landed in #2048, that walk descends into every gitignored subtree before the loop body discards the resulting Delete actions, so every file under directories like node_modules is lstat'd for nothing.

The infrastructure to avoid this already exists: passing excludeIgnoredChanges=true to diffStagingWithWorktree builds a gitignore.Matcher and threads it into filesystem.Options.IgnoreMatcher, where the noder's shouldSkipIgnored prunes ignored, untracked subtrees at enumeration time while keeping tracked entries (or directories with tracked descendants) visible. Flip the flag to true at the two Reset call sites — resetWorktreeToTree and resetWorktree — to engage it.

The observable result is unchanged. resetWorktreeToTree's loop already skips Delete actions explicitly to preserve untracked files (the fix from 0c7da9764), and resetWorktree filters changes by the supplied files list, which never names an untracked-and-ignored path. The matcher only avoids the cost of walking those entries.

Benchmark on a tree with 100 tracked source files and N untracked files in a gitignored directory (Apple M5, -benchtime=10x):

  Untracked | before   | after  | speedup
  --------- | -------- | ------ | -------
  0         |  2.2 ms  | 2.6 ms | -
  1,000     |  7.7 ms  | 3.8 ms |  2.0x
  5,000     | 16.2 ms  | 5.2 ms |  3.1x
  20,000    | 51.8 ms  | 4.6 ms | 11.3x

The small overhead in the 0-untracked case is the same matcher build cost noted in #2048; the gain comes from the diff walk no longer descending into ignored subtrees. The numbers track BenchmarkStatusIgnoredDir closely because both operations share diffStagingWithWorktree.

Assisted-by: Claude Opus 4.7

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -690,7 +690,14 @@ func (w *Worktree) resetWorktreeToTree(fromTree, toTree *object.Tree, files []st
 	// Delete means "on disk, not in index" = untracked → always skip.
 	// Note: SkipWorktree entries are invisible to the merkletrie diff;
 	// they are cleaned up in step 3 below.
-	worktreeChanges, err := w.diffStagingWithWorktree(true, false)
+	//
+	// excludeIgnoredChanges=true engages the filesystem walker's
+	// IgnoreMatcher so untracked entries inside gitignored directories are
+	// pruned at enumeration time rather than walked and then dropped as
+	// Delete actions. The observable result is unchanged because Delete
+	// actions are skipped by the loop below; the matcher only avoids the
+	// pointless lstat of every file under directories like node_modules.
+	worktreeChanges, err := w.diffStagingWithWorktree(true, true)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


worktree.go
```diff
@@ -691,7 +691,14 @@ func (w *Worktree) resetWorktreeToTree(fromTree, toTree *object.Tree, files []st
...
 	// they are cleaned up in step 3 below.
-	worktreeChanges, err := w.diffStagingWithWorktree(true, false)
+	//
+	// excludeIgnoredChanges=true engages the filesystem walker's
+	// IgnoreMatcher so untracked entries inside gitignored directories are
+	// pruned at enumeration time rather than walked and then dropped as
+	// Delete actions. The observable result is unchanged because Delete
+	// actions are skipped by the loop below; the matcher only avoids the
+	// pointless lstat of every file under directories like node_modules.
+	worktreeChanges, err := w.diffStagingWithWorktree(true, true)
```
onee-only, line 700: All the callers of `diffStagingWithWorktree` now pass `excludeIgnoredChanges = true`. I think it's fair to remove the parameter and make this the only behavior.

Soph: Yeah I was also wondering why you wouldn't want this behavior. Will change !

Soph: ok, changed, also added some more tests after chatting with @pjbgf about #821 to make sure we are not changing the behavior.

onee-only: I didn't notice that. Thanks for letting me know!

pjbgf (approved):

PR #2078, opened 2026-05-07
v5: git: submodule, error on remote without URLs

hiddeco:
`defaultRemote`, introduced on this branch in a675ed42, returns a `*RemoteConfig` that `submodule.Repository()` then dereferences as `base.URLs[0]` to join a relative submodule URL. `RemoteConfig.Validate()` rejects an empty `URLs` slice on `SetConfig`, but a config that bypasses validation — for example a hand-edited `[remote "origin"]` section with no `url =` entry loaded from disk — could still produce a remote with no URLs and panic the caller with `index out of range [0] with length 0`.

Promote the invariant into `lookupRemote` and route the single-remote branch through it as well, so any code path that asks `defaultRemote` for the resolution base gets a usable remote or a clear `remote %q has no configured URL` error.

Backport of #2076.

### DIFF
--- a/submodule.go
+++ b/submodule.go
@@ -197,8 +197,8 @@ func defaultRemote(r *Repository) (*config.RemoteConfig, error) {
 	}
 
 	if len(cfg.Remotes) == 1 {
-		for _, rc := range cfg.Remotes {
-			return rc, nil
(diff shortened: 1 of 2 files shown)


pjbgf (approved):

PR #2081, opened 2026-05-07
Add `worktreeFilesystem` wrapper for worktree

pjbgf:
Adds a new abstraction for Worktree Filesystem, which removes from the worktree itself the need to know or enforce rules.
The new filesystem also implements:
- [core.protectHFS](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreprotectHFS)
- [core.protectNTFS](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreprotectNTFS)

This is a breaking change in v6, as it will make `Worktree.Filesystem` a func as opposed to a field as per `v5`. This is an important API UX change, as users shouldn't be able to modified a worktree fs mid-operation.


worktree_fs.go
```diff
@@ -0,0 +1,279 @@
...
+			i++
+		}
+		if i >= len(runes) {
+			return false
+		}
+		r := runes[i]
+		if r > 127 {
+			return false
+		}
+		if strings.ToLower(string(r)) != string(expected) {
```
hiddeco:
```suggestion
		if unicode.ToLower(r) != unicode.ToLower(expected) {
```

potentially?

worktree_fs.go
```diff
@@ -0,0 +1,279 @@
...
+	// must be at end of component
+	return i == len(runes)
+}
+
+// windowsPathReplacer defines the chars that need to be replaced
+// as part of windowsValidPath.
+var windowsPathReplacer *strings.Replacer
+
+func init() {
+	windowsPathReplacer = strings.NewReplacer(" ", "", ".", "")
```
hiddeco: Why is this wrapped in an `init()`?

worktree_fs.go
```diff
@@ -0,0 +1,279 @@
...
+func defaultProtectHFS() bool {
+	return runtime.GOOS == "darwin"
+}
+
+// hfsIgnoredCodepoints contains Unicode code points that HFS+ ignores
+// during path normalization. A path containing these characters between
+// the characters of ".git" will be treated as ".git" by HFS+.
+//
+// See upstream Git utf8.c next_hfs_char() for the full list.
+var hfsIgnoredCodepoints = map[rune]bool{
```
hiddeco: Can potentially just be a `map[rune]struct{}`?

worktree_fs.go
```diff
@@ -0,0 +1,279 @@
...
+	billy.Filesystem
+	protectNTFS bool
+	protectHFS  bool
+}
+
+func newWorktreeFilesystem(fs billy.Filesystem, protectNTFS, protectHFS bool) *worktreeFilesystem {
+	return &worktreeFilesystem{Filesystem: fs, protectNTFS: protectNTFS, protectHFS: protectHFS}
+}
+
+func (sfs *worktreeFilesystem) Create(filename string) (billy.File, error) {
```
hiddeco, line 28: Should `Chroot` and/or `TempFile` be wrapped as well? Additionally, do we want to publicly expose _unsafe_ methods publicly (through `wfs.Filesystem`)?

worktree_fs.go
```diff
@@ -0,0 +1,279 @@
...
+		return err
+	}
+	return sfs.Filesystem.Rename(from, to)
+}
+
+func (sfs *worktreeFilesystem) Symlink(target, link string) error {
+	if err := sfs.validPath(link); err != nil {
+		return err
+	}
+	return sfs.Filesystem.Symlink(target, link)
```
hiddeco, line 88: Do we need to validate the target? (Not doing this appears to be related to CVE-2024-32002).

worktree_fs.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+// validPath checks whether paths are valid.
+// The rules around invalid paths could differ from upstream based on how
+// filesystems are managed within go-git, but they are largely the same.
+//
+// For upstream rules:
+// https://github.com/git/git/blob/564d0252ca632e0264ed670534a51d18a689ef5d/read-cache.c#L946
+// https://github.com/git/git/blob/564d0252ca632e0264ed670534a51d18a689ef5d/path.c#L1383
+func (sfs *worktreeFilesystem) validPath(paths ...string) error {
```
hiddeco, line 150: We could catch NUL bytes and control characters here as well to reject as early as possible.

worktree_fs.go
pjbgf: Removed from init

worktree_fs.go
pjbgf: I marked `TempFile` as unsupported and wrapped `Chroot`. PTAL

hiddeco (approved):
Nice one! Thanks @pjbgf 🙇

PR #2082, opened 2026-05-07
v5: config: validate submodule names

hiddeco:
`Submodule.Validate()` validated `Path` against `dotdotPath` but left `Name` unchecked, so submodule names with unusual shapes (empty, bare `.`/`..`, names containing `..` components, NUL bytes, or leading or trailing separators) were accepted at parse time and propagated to `DotGit.Module(name)`, which joined them into the on-disk modules path without verifying the result stayed under `modules/`.

The parser now mirrors canonical Git's `check_submodule_name` from `submodule-config.c` [1], extended with a few defensive checks for go-git-specific edge cases. The storage layer adds an independent path-containment check in `DotGit.Module(name)`, so any caller that constructs a `Submodule` programmatically — including the transactional storage wrapper — is also protected, and a regression test exercises the storage-layer defence end-to-end. Nested but contained names like `lib/foo` continue to work; only path-traversal shapes are refused.

[1]: https://github.com/git/git/blob/v2.54.0/submodule-config.c#L214-L237

Backport of #2079.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -399,7 +399,8 @@ func unmarshalSubmodules(fc *format.Config, submodules map[string]*Submodule) {
 		m := &Submodule{}
 		m.unmarshal(sub)
 
-		if m.Validate() == ErrModuleBadPath {
+		if err := m.Validate(); errors.Is(err, ErrModuleBadPath) ||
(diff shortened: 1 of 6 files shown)


pjbgf (approved):

PR #2083, opened 2026-05-07
plumbing: format/idxfile, Validate offset64 indices

hiddeco:
When a 32-bit offset entry has its MSB set, the lower 31 bits index into the `Offset64` table. Neither `MemoryIndex.getOffset` nor `LazyIndex.offset` validated that the resulting position fell inside the decoded table; an idx with an MSB-set entry whose lower bits exceed the number of decoded 64-bit entries would index past the end of the table.

Bounds-check both lookups against the decoded table size and return `ErrMalformedIdxFile` when the index is out of range. `MemoryIndex`'s unexported `getOffset` now returns `(uint64, error)`; the change propagates through `FindOffset`, `genOffsetHash`, and the `Entries` iterator. `LazyIndex` caches the 64-bit-entry count discovered during init and uses it for the same check before issuing the `ReadAt`.

### DIFF
--- a/plumbing/format/idxfile/fuzz_helpers.go
+++ b/plumbing/format/idxfile/fuzz_helpers.go
@@ -2,7 +2,10 @@ package idxfile
 
 import (
 	"bytes"
+	"crypto/sha1"
 	"encoding/binary"
+
+	"github.com/go-git/go-git/v6/plumbing"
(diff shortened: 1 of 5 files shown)


pjbgf (approved):

PR #2084, opened 2026-05-07
v5: plumbing: format/idxfile, Validate offset64 indices

hiddeco:
When a 32-bit offset entry has its MSB set, the lower 31 bits index into the `Offset64` table. `MemoryIndex.getOffset` did not validate that the resulting position fell inside the decoded table; an idx with an MSB-set entry whose lower bits exceed the number of decoded 64-bit entries would index past the end of the table.

Bounds-check the lookup against the decoded table size and return `ErrMalformedIdxFile` when the index is out of range. The unexported `getOffset` now returns `(uint64, error)`; the change propagates through `FindOffset`, `genOffsetHash`, and the `Entries` iterator.

Backport of #2083.

### DIFF
--- a/plumbing/format/idxfile/idxfile.go
+++ b/plumbing/format/idxfile/idxfile.go
@@ -2,6 +2,7 @@ package idxfile
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"sort"
 	"sync"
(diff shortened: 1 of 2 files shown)


pjbgf (approved):

PR #2086, opened 2026-05-07
plumbing: format/packfile, Test patchDeltaWriter cap

hiddeco:
`patchDeltaWriter`'s preemptive `Buffer.Grow` is already clamped at `maxPatchPreemptionSize` since 5b89cb5a, but no test covered that path end-to-end. Add a regression test that constructs a delta whose header advertises a `math.MaxInt64` `targetSz` and asserts the call returns an error rather than driving the allocation.

Also restate the rationale for the clamp in the comment, naming both the input-validation goal and the symmetry with `patchDelta`.

### DIFF
--- a/plumbing/format/packfile/delta_test.go
+++ b/plumbing/format/packfile/delta_test.go
@@ -3,12 +3,14 @@ package packfile
 import (
 	"bytes"
 	"io"
+	"math"
 	"math/rand"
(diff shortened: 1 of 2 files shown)


pjbgf (dismissed):

pjbgf (approved):

PR #2090, opened 2026-05-07
*: Reject malformed variable-length integers

hiddeco:
Each delta operation's declared size is validated against the size still remaining in the target buffer, not the total target size. The previous check against the total accepted a sequence of operations whose cumulative output would exceed the declared target, leaving the unsigned remaining counter unable to reach zero and the loop free to keep writing past the declared target until the delta payload was consumed.

Three sibling code paths shared the flaw and are updated together: `patchDelta` (used by `PatchDelta` and `ApplyDelta`), `ReaderFromDelta`, and `patchDeltaWriter` (used by the packfile parser).

The loop structure is rewritten as `for remainingTargetSz > 0` so the invariant is explicit, and a post-loop check rejects deltas that leave bytes unconsumed -- matching the `data != top` sanity check in `patch-delta.c`[1]. Empty-target deltas (header-only, no operations) are now accepted, also matching upstream.

Several incidental defects on the same call paths are fixed: `patchDelta`'s copy-from-src failure used `break`, which only exited the switch and let the loop keep consuming delta bytes; `patchDeltaWriter` returned `(0, ZeroHash, nil)` on invalid input, silently reporting success; and `ReaderFromDelta`'s copy-from-src failure closed the writer without an error, silently truncating the consumer stream. Each now propagates `ErrInvalidDelta` (or `ErrDeltaCmd`) consistently.

`packfile.getMemoryObject` was re-inflating delta payloads into the buffer the scanner had already populated, appending a duplicate copy that previously went unnoticed because the loop silently ignored anything past the target. The fix matches the cached-content pattern already used in `Scanner.WriteObject`.

[1]: https://github.com/git/git/blob/c2c5f6b1e479f2c38e0e01345350620944e3527f/patch-delta.c

### DIFF
--- a/plumbing/format/packfile/patch_delta.go
+++ b/plumbing/format/packfile/patch_delta.go
@@ -240,12 +240,18 @@ func patchDelta(dst *bytes.Buffer, src, delta []byte) error {
 		return ErrInvalidDelta
 	}
 
-	srcSz, delta := packutil.DecodeLEB128(delta)
+	srcSz, delta, err := packutil.DecodeLEB128(delta)
+	if err != nil {
+		return fmt.Errorf("%w: %w", ErrInvalidDelta, err)
+	}
 	if srcSz != uint(len(src)) {
 		return ErrInvalidDelta
 	}
 
-	targetSz, delta := packutil.DecodeLEB128(delta)
+	targetSz, delta, err := packutil.DecodeLEB128(delta)
+	if err != nil {
+		return fmt.Errorf("%w: %w", ErrInvalidDelta, err)
+	}
 	remainingTargetSz := targetSz
 
 	var cmd byte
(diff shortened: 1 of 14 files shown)


plumbing/format/packfile/util/util.go
```diff
@@ -61,16 +73,23 @@ func ObjectType(b byte) plumbing.ObjectType {
 }
 
 // DecodeLEB128 decodes a number encoded as an unsigned LEB128 at the
-// start of some binary data and returns the decoded number and the rest
-// of the bytes.
-func DecodeLEB128(input []byte) (uint, []byte) {
+// start of some binary data and returns the decoded number, the rest
+// of the bytes, and an error if the encoded value does not fit in a
+// uint.
+func DecodeLEB128(input []byte) (uint, []byte, error) {
```
hiddeco, line 79: ⚠️ Signature change, but I believe this is still acceptable for v6?

pjbgf: This package was actually introduced in v6, so never release exposed on a stable release.

plumbing/format/packfile/patch_delta.go
```diff
@@ -240,12 +240,18 @@ func patchDelta(dst *bytes.Buffer, src, delta []byte) error {
...
+		return ErrInvalidDelta
+	}
 	if srcSz != uint(len(src)) {
 		return ErrInvalidDelta
 	}
 
-	targetSz, delta := packutil.DecodeLEB128(delta)
+	targetSz, delta, err := packutil.DecodeLEB128(delta)
+	if err != nil {
+		return ErrInvalidDelta
```
pjbgf: Let's wrap `ErrInvalidDelta` and return `err` as well.

pjbgf (approved):

PR #2092, opened 2026-05-07
v5: *: Reject malformed variable-length integers

hiddeco:
Variable-length integer decoders accepted continuation chains long enough to push the running shift past the bit-width of their accumulator, producing out-of-range values that propagated into buffer allocation hints and on-disk size fields.

Mirror the bound from canonical Git's `unpack_object_header_buffer`[1]: reject any continuation byte that would shift past the type's capacity. The same rule applies to `Scanner.readLength`, the package-local `decodeLEB128`/`decodeLEB128ByteReader`, and `ReadVariableWidthInt`; `decodeLEB128` now returns an error so callers surface invalid input rather than silently zeroing the result.

Add defense-in-depth at every call site that allocates from a parsed length: cap `bytes.Buffer.Grow` hints in the parser at 1 GiB, clamp the slice and maps initialized from `Parser.count` at a 4 Mi-entry hint so a hostile object count cannot request a multi-gigabyte allocation, and validate `OFS-delta` back-references in the scanner (must be `> 0` and `<= h.Offset`).

Cover the new behavior with unit, parser-level regression and fuzz tests (`FuzzReadLength`, `FuzzDecodeLEB128`,
`FuzzDecodeLEB128ByteReader`, `FuzzParser`).

[1]: https://github.com/git/git/blob/v2.54.0/packfile.c#L1135-L1158

Backport of #2090.

### DIFF
--- a/plumbing/format/packfile/patch_delta.go
+++ b/plumbing/format/packfile/patch_delta.go
@@ -35,6 +35,9 @@ const (
 	minDeltaSize = 4
 )
 
+// uintBits is the bit width of uint on the current platform (32 or 64).
+const uintBits = 32 << (^uint(0) >> 63)
+
 type offset struct {
 	mask  byte
 	shift uint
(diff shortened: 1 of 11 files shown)


utils/binary/read.go
```diff
@@ -92,6 +99,14 @@ func ReadVariableWidthInt(r io.Reader) (int64, error) {
 
 	var v = int64(c & maskLength)
 	for c&maskContinue > 0 {
+		// Reject input that, after the v++ and shift below, would not
+		// fit in an int64. The bound (math.MaxInt64-127)>>7 ensures
+		// both the increment and the (v << 7) + (c & 0x7F) stay
+		// non-negative.
+		if v > (math.MaxInt64-int64(maskLength))>>lengthBits {
```
pjbgf:
```suggestion
		if v >= (math.MaxInt64-int64(maskLength))>>lengthBits {
```

plumbing/format/packfile/patch_delta.go
```diff
@@ -241,12 +244,18 @@ func patchDelta(dst *bytes.Buffer, src, delta []byte) error {
 		return ErrInvalidDelta
 	}
 
-	srcSz, delta := decodeLEB128(delta)
+	srcSz, delta, err := decodeLEB128(delta)
+	if err != nil {
+		return ErrInvalidDelta
```
pjbgf: Let's wrap this and return `err`. Same for the next decode call.

pjbgf (approved):

PR #2094, opened 2026-05-08
*: Fix pack file race conditions and checksum mismatches

AriehSchneier:
Fix multiple concurrency issues in pack file handling that caused intermittent "malformed pack file: checksum mismatch" errors in CI.

## Root Causes

1. **Missing fsync before close**: Pack files were closed and renamed without flushing OS buffers, causing readers to see stale/incomplete data when accessing the file immediately after creation.

2. **Unprotected index map access**: The s.index map in ObjectStorage was accessed without proper mutex protection in multiple locations, leading to race conditions.

3. **DotGit incoming directory race**: hasIncomingObjects() had concurrent read/write races on incomingChecked and incomingDirName fields.

4. **FSObject shared file descriptor race**: FSObject.Reader() had races when multiple goroutines called Reader() concurrently on the same object due to shared pack file descriptor.

## Changes

**storage/filesystem/dotgit/writers.go**
- Add Sync() call before Close() in PackWriter to ensure data is flushed to disk before the pack file is renamed

**storage/filesystem/object.go**
- Reindex(): Add mutex protection around s.index = nil
- PackfileWriter(): Add mutex protection in Notify callback when writing to s.index[h]
- EncodedObject(): Add read lock when checking if s.index != nil
- HashesWithPrefix(): Copy index values to slice while holding lock before iteration to avoid races (more efficient than copying map)
- buildPackfileIters(): Add read lock when accessing s.index[h] in closure

**storage/filesystem/dotgit/dotgit.go**
- hasIncomingObjects(): Use sync.Once for race-free lazy initialization of incoming directory scan

**plumbing/format/packfile/fsobject.go**
- FSObject.Reader(): Hybrid approach for race-free concurrent reads
  - Primary: Try to open new file descriptor by path (enables true concurrent reads)
  - Fallback: Use existing o.pack with mutex if path open fails (handles test edge cases)
- Performance: ~400ns overhead per uncached read when opening new FD, but 37% speedup in concurrent workloads

**storage/filesystem/object_race_test.go** (new)
- Add TestConcurrentIndexAccess to reproduce and verify fixes for s.index races

**storage/filesystem/object_bench_test.go** (updated)
- Add benchmarks for FSObject.Reader() performance with new implementation
- Preserved existing BenchmarkAlternatesObjectLookup

## Performance

Benchmarks show FSObject.Reader() hybrid approach:
- File open cost: ~400ns vs ~1.5ns for seek (270x slower per operation)
- Concurrent reads: 37% faster due to true parallelism (no mutex contention)
- Most reads hit cache: ~19ns (no impact)
- Fallback path: Uses mutex for test scenarios where pack and fs don't match
- Trade-off: Small per-operation cost for correctness + concurrency benefits

## Test Results

- All tests pass including -race detector
- TestRepackObjectsWithNoDelete: consistently passes (was flaky)
- TestSmartMultiRoundFetch: consistently passes (was flaky)
- TestConcurrentIndexAccess: passes with race detector (reproduces races without fixes)
- TestPackfileEmbedFS: passes (uses fallback path for mismatched fs)

## Testing

Ran tests multiple times with `-race` detector - no failures or race conditions detected.

### DIFF
--- a/plumbing/format/packfile/fsobject.go
+++ b/plumbing/format/packfile/fsobject.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"io"
 	"os"
+	"sync"
 
 	billy "github.com/go-git/go-billy/v6"
 
@@ -12,7 +13,7 @@ import (
 	"github.com/go-git/go-git/v6/plumbing/cache"
 	"github.com/go-git/go-git/v6/plumbing/format/idxfile"
 	"github.com/go-git/go-git/v6/utils/ioutil"
-	"github.com/go-git/go-git/v6/utils/sync"
+	utilsync "github.com/go-git/go-git/v6/utils/sync"
 )
 
 // FSObject is an object from the packfile on the filesystem.
@@ -26,6 +27,7 @@ type FSObject struct {
 	pack     billy.File
 	packPath string
 	cache    cache.Object
+	mu       sync.Mutex // Protects pack access when used as fallback
 }
 
 // NewFSObject creates a new filesystem object.
@@ -65,41 +67,57 @@ func (o *FSObject) Reader() (io.ReadCloser, error) {
 		return reader, nil
 	}
 
-	var file io.Closer
-	_, err := o.pack.Seek(o.offset, io.SeekStart)
-	// fsobject aims to reuse an existing file descriptor to the packfile.
-	// In some cases that descriptor would already be closed, in such cases,
-	// open the packfile again and close it when the reader is closed.
-	if err != nil && errors.Is(err, os.ErrClosed) {
-		o.pack, err = o.fs.Open(o.packPath)
+	// Prefer opening a new file descriptor to enable concurrent reads without
+	// mutex contention. If the file doesn't exist (e.g., in test scenarios where
+	// the pack file is from a different filesystem), fall back to using o.pack.
+	file, err := o.fs.Open(o.packPath)
+	if err == nil {
+		// Successfully opened new file descriptor - use it (no mutex needed)
+		_, err = file.Seek(o.offset, io.SeekStart)
 		if err != nil {
+			_ = file.Close()
 			return nil, err
 		}
-		file = o.pack
-		_, err = o.pack.Seek(o.offset, io.SeekStart)
+
+		br := utilsync.GetBufioReader(file)
+		zr, err := utilsync.GetZlibReader(br)
+		if err != nil {
+			utilsync.PutBufioReader(br)
+			_ = file.Close()
+			return nil, err
+		}
+
+		return &zlibReadCloser{r: zr, f: file, rbuf: br}, nil
+	}
+
+	// Fallback: use the existing pack file descriptor with mutex protection.
+	// This handles cases where the packfile was deleted or is on a different
+	// filesystem (e.g., test scenarios).
+	o.mu.Lock()
+	defer o.mu.Unlock()
+
+	_, err = o.pack.Seek(o.offset, io.SeekStart)
+	if err != nil && errors.Is(err, os.ErrClosed) {
+		// Both paths failed - pack is closed and can't open new file
+		return nil, err
 	}
 	if err != nil {
-		if file != nil {
-			_ = file.Close()
-		}
 		return nil, err
 	}
 
-	br := sync.GetBufioReader(o.pack)
-
-	zr, err := sync.GetZlibReader(br)
+	br := utilsync.GetBufioReader(o.pack)
+	zr, err := utilsync.GetZlibReader(br)
 	if err != nil {
-		sync.PutBufioReader(br)
-		if file != nil {
-			_ = file.Close()
-		}
+		utilsync.PutBufioReader(br)
 		return nil, err
 	}
-	return &zlibReadCloser{r: zr, f: file, rbuf: br}, nil
+
+	// Don't close o.pack - it's shared and will be closed elsewhere
+	return &zlibReadCloser{r: zr, f: nil, rbuf: br}, nil
 }
 
 type zlibReadCloser struct {
-	r      *sync.ZLibReader
+	r      *utilsync.ZLibReader
 	f      io.Closer
 	rbuf   *bufio.Reader
 	closed bool
@@ -120,9 +138,9 @@ func (r *zlibReadCloser) Close() (err error) {
 		defer ioutil.CheckClose(r.f, &err)
 	}
 
-	defer sync.PutBufioReader(r.rbuf)
+	defer utilsync.PutBufioReader(r.rbuf)
 
-	defer sync.PutZlibReader(r.r)
+	defer utilsync.PutZlibReader(r.r)
 	return r.r.Close()
 }
 
--- a/storage/filesystem/dotgit/dotgit.go
+++ b/storage/filesystem/dotgit/dotgit.go
@@ -16,6 +16,7 @@ import (
 	"runtime"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/go-git/go-billy/v6"
@@ -115,8 +116,9 @@ type DotGit struct {
 	fs      billy.Filesystem
 
 	// incoming object directory information
-	incomingChecked bool
+	// incomingOnce ensures directory scan happens only once
 	incomingDirName string
+	incomingOnce    sync.Once
 
 	objectList []plumbing.Hash // sorted
 	objectMap  map[plumbing.Hash]struct{}
@@ -728,7 +730,7 @@ func (d *DotGit) incomingObjectPath(h plumbing.Hash) string {
 // hasIncomingObjects searches for an incoming directory and keeps its name
 // so it doesn't have to be found each time an object is accessed.
 func (d *DotGit) hasIncomingObjects() bool {
-	if !d.incomingChecked {
+	d.incomingOnce.Do(func() {
 		directoryContents, err := d.fs.ReadDir(objectsPath)
 		if err == nil {
 			for _, file := range directoryContents {
@@ -739,9 +741,7 @@ func (d *DotGit) hasIncomingObjects() bool {
 				}
 			}
 		}
-
-		d.incomingChecked = true
-	}
+	})
 
 	return d.incomingDirName != ""
 }
--- a/storage/filesystem/dotgit/writers.go
+++ b/storage/filesystem/dotgit/writers.go
@@ -122,6 +122,16 @@ func (w *PackWriter) Close() error {
 		return err
 	}
 
+	// Sync the file to ensure buffered data is written to stable storage.
+	// While Close() should make data visible, explicitly syncing before close
+	// ensures any OS-buffered writes are flushed, which may help prevent
+	// checksum mismatches when pack files are immediately read after creation.
+	if syncer, ok := w.fw.(interface{ Sync() error }); ok {
+		if err := syncer.Sync(); err != nil {
+			return err
+		}
+	}
+
 	if err := w.fw.Close(); err != nil {
 		return err
 	}
--- a/storage/filesystem/object.go
+++ b/storage/filesystem/object.go
@@ -32,14 +32,17 @@ type ObjectStorage struct {
 	// loaded loose objects.
 	objectCache cache.Object
 
-	dir   *dotgit.DotGit
+	dir *dotgit.DotGit
+
+	// index maps pack file hashes to their decoded indices.
 	index map[plumbing.Hash]idxfile.Index
+	muI   sync.RWMutex // Protects index map
 
 	packList    []plumbing.Hash
 	packListIdx int
-	packfiles   map[plumbing.Hash]*packfile.Packfile
-	muI         sync.RWMutex
-	muP         sync.RWMutex
+	// packfiles caches open packfile handles when KeepDescriptors is enabled.
+	packfiles map[plumbing.Hash]*packfile.Packfile
+	muP       sync.RWMutex // Protects packfiles map and packList
 
 	oh *plumbing.ObjectHasher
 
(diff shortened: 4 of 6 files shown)


storage/filesystem/object.go
```diff
@@ -753,7 +761,16 @@ func (s *ObjectStorage) HashesWithPrefix(prefix []byte) ([]plumbing.Hash, error)
 	if err := s.requireIndex(); err != nil {
 		return nil, err
 	}
-	for _, index := range s.index {
+
+	// Copy index values into slice while holding lock to avoid races during iteration
```
AriehSchneier, line 768: I asked Claude if we would be better off just holding the lock than copying, and this was the response:
```
You raise a good point about write contention being rare. However, there's a critical issue: the iteration does I/O.

Looking at the code:


for _, index := range indices {
    ei, err := index.Entries()  // <- LazyIndex.Entries() can open file (I/O)
    for {
        e, err := ei.Next()     // <- Reads from file (I/O)
        if e.Hash.HasPrefix(prefix) {
            // process...
        }
    }
}
For LazyIndex.Entries(), it calls s.idx.acquire() which may open the .idx file. Then Next() reads from that file. In a large repo with many pack files and objects, this iteration could take seconds or longer.

If we held RLock during this entire operation:

✅ Other HashesWithPrefix() calls wouldn't block (multiple readers OK)
❌ Reindex() would block for seconds
❌ PackfileWriter() (creating new packs) would block for seconds
❌ We're holding a lock during I/O (generally bad practice)
So I think the copy approach is correct - we just needed the slice optimization you suggested (which saves the hashing overhead). We:

Hold RLock briefly
Copy index references to slice (fast, no allocation pressure)
Release RLock
Do slow iteration with I/O without holding lock
```

AriehSchneier:
@pjbgf ready for review

storage/filesystem/dotgit/writers.go
```diff
@@ -122,6 +122,15 @@ func (w *PackWriter) Close() error {
 		return err
 	}
 
+	// Sync the file to disk before closing to ensure all data is flushed.
+	// This prevents checksum mismatches when the pack file is immediately
+	// read after creation, particularly on macOS and under high load.
```
hanwen-flow: this is not how Sync works. Ensures that the data is on persistent storage, which is useful if the machine is powered off unexpectedly. Syncing data can be useful for data reliability, but in itself it will not fix checksum mismatches under load.

storage/filesystem/object.go
```diff
@@ -207,7 +207,9 @@ func (s *ObjectStorage) requireIndex() error {
 
 // Reindex indexes again all packfiles. Useful if git changed packfiles externally
 func (s *ObjectStorage) Reindex() {
+	s.muI.Lock()
```
hanwen-flow, line 213: neither muI or muP is actually documented, so it is hard to judge whether this is correct. Maybe you could start with creating a test that reproduces the race condition (eg. by simulating the load you saw fail in a test and running that test with -race)

hanwen-flow (changes requested):

AriehSchneier:
Thank you for the review! I've addressed both comments:

## 1. Mutex Documentation (object.go)

Added documentation for `muI` and `muP`:
- `muI` protects the `index` map
- `muP` protects the `packfiles` map and `packList`

## 2. Race Condition Test (object_race_test.go)

Added `TestConcurrentIndexAccess` that reproduces the `s.index` race conditions:

**Without the mutex fixes:**
- Race detected at `object.go:210` (Reindex writing to s.index)
- Race detected at `object.go:545` (EncodedObject reading s.index)
- Race detected at `object.go:193` (requireIndex writing to s.index)

**With the mutex fixes:**
- All `s.index` races are eliminated ✅

Run with: `go test -race -run TestConcurrentIndexAccess ./storage/filesystem/`

## 3. Re: Sync() Comment

You're right that `Sync()` is primarily for data persistence (power loss scenarios). I've updated the comment to be more accurate about what it does. 

After investigating, I believe the checksum mismatches may have actually been caused by the `s.index` race conditions (which are now fixed), not the lack of `Sync()`. The `Sync()` call is defensive but may not be strictly necessary. I've kept it as a low-cost precaution, but I'm open to removing it if you think it's not appropriate.

## Unrelated Races Found

Note: `TestConcurrentIndexAccess` does detect some unrelated races:
- `DotGit.hasIncomingObjects()` - concurrent writes to `incomingChecked`/`incomingDirName` fields
- `FSObject.Reader()` - concurrent Reader() calls
- memfs (external billy library)

These are pre-existing issues unrelated to the s.index fixes in this PR. I can address them in a follow-up PR if desired.

---

All tests now pass including `-race` detector. The flaky tests (`TestRepackObjectsWithNoDelete`, `TestSmartMultiRoundFetch`) consistently pass.

AriehSchneier:
## Additional Race Conditions Fixed

CI revealed two more race conditions when running the new `TestConcurrentIndexAccess` test. Both are now fixed in the latest commits:

### 1. DotGit.hasIncomingObjects() Race
**Problem:** Concurrent reads/writes on `incomingChecked` and `incomingDirName` fields.

**Solution:** Use `sync.Once` for idiomatic lazy initialization (cleaner than double-checked locking pattern).

### 2. FSObject.Reader() Race  
**Problem:** Multiple goroutines calling `Reader()` on the same FSObject would race on the shared `o.pack` file descriptor.

**Solution:** Always open a new file descriptor for each `Reader()` call instead of reusing the shared one.

**Performance impact:**
- Opening new FD: ~400ns vs ~1.5ns for Seek (270x slower per operation)
- Concurrent reads: **37% faster** due to true parallelism (12.25ns vs 19.43ns)
- Most reads hit the object cache anyway (~19ns), so impact is minimal
- Trade-off is worthwhile: small cost for race-safety + concurrency benefits

All tests now pass cleanly with the race detector. ✅

AriehSchneier:
## FSObject.Reader() Update - Hybrid Approach

After investigation, updated the FSObject.Reader() fix to use a hybrid approach that handles both production and test scenarios:

### The Issue
Tests like `TestPackfileEmbedFS` pass:
- An open pack file from fixture location  
- A **different** filesystem (empty temp dir) via `WithFs()`

When FSObject tried `fs.Open(packPath)`, the file didn't exist → CI failures.

### The Fix
Hybrid approach in FSObject.Reader():

1. **Primary path**: Try to open new file descriptor by path
   - Enables race-free concurrent reads
   - ~400ns overhead but 37% better parallel throughput
   - No mutex needed

2. **Fallback path**: If open fails, use existing `o.pack` with mutex
   - Handles test scenarios with mismatched filesystems
   - Handles cases where pack was deleted but FD still valid
   - Mutex-protected to avoid races

### Results
- ✅ All tests pass including `TestPackfileEmbedFS`
- ✅ No data races detected
- ✅ Production code gets optimal concurrent performance
- ✅ Test edge cases handled gracefully

The benchmark file also now includes both the original `BenchmarkAlternatesObjectLookup` and the new FSObject performance benchmarks.

storage/filesystem/object.go
```diff
@@ -32,14 +32,19 @@ type ObjectStorage struct {
...
 
 	packList    []plumbing.Hash
 	packListIdx int
-	packfiles   map[plumbing.Hash]*packfile.Packfile
-	muI         sync.RWMutex
-	muP         sync.RWMutex
+	// packfiles caches open packfile handles when KeepDescriptors is enabled.
+	// Protected by muP for concurrent access.
+	packfiles map[plumbing.Hash]*packfile.Packfile
+	muP       sync.RWMutex // Protects packfiles map and packList
```
hanwen, line 45: same

storage/filesystem/object.go
```diff
@@ -32,14 +32,19 @@ type ObjectStorage struct {
 	// loaded loose objects.
 	objectCache cache.Object
 
-	dir   *dotgit.DotGit
+	dir *dotgit.DotGit
+
+	// index maps pack file hashes to their decoded indices.
+	// Protected by muI for concurrent access.
 	index map[plumbing.Hash]idxfile.Index
+	muI   sync.RWMutex // Protects index map
```
hanwen, line 39: comment dup from 2 lines above

If this class is threadsafe, the comment should say so.

storage/filesystem/dotgit/dotgit.go
```diff
@@ -115,8 +116,9 @@ type DotGit struct {
 	fs      billy.Filesystem
 
 	// incoming object directory information
-	incomingChecked bool
+	// incomingOnce ensures directory scan happens only once
 	incomingDirName string
+	incomingOnce    sync.Once
 
 	objectList []plumbing.Hash // sorted
```
hanwen, line 123: none of the data below is thread-safe is it? For example ObjectPacks calls genPackList, which writes into the packMap without protection.

A type should either be completely thread safe or not at all.

Your PR is addressing two types at the same time. I recommend doing one at a time instead.

TBH this class is confusing to me: the .git directory itself is mutable (you can write a new pack), but any precomputation based on sync.Once will only happen once. Won't that be invalidated under the right circumstances? It appears this code assumes nobody ever shells out to the git cli while also using go-git? That is very impractical for as long as go-git has no feature parity with git, and should be documented.

hanwen (changes requested):
this talks about DotGit, but DotGit is not thread-safe is it?

> ## 3. Re: Sync() Comment
> 
> You're right that `Sync()` is primarily for data persistence (power loss scenarios). I've updated the comment to be more accurate about what it does. 
> 
> After investigating, I believe the checksum mismatches may have actually been caused by the `s.index` race conditions (which are now fixed), not the lack of `Sync()`. The `Sync()` call is defensive but may not be strictly necessary. I've kept it as a low-cost precaution, but I'm open to removing it if you think it's not appropriate.

you should remove it. Sync has a performance cost, so you shoudl avoid
it if you want to go faster. If you want to guarantee durability, you
have to do it across the board, ie. across a meaningful subset of all
the write functionality, and document it as such. In any event, it is
unrelated to the current work, so it should go in a separate PR.

> ## Unrelated Races Found
> 
> Note: `TestConcurrentIndexAccess` does detect some unrelated races:
> - `DotGit.hasIncomingObjects()` - concurrent writes to `incomingChecked`/`incomingDirName` fields
> - `FSObject.Reader()` - concurrent Reader() calls
> - memfs (external billy library)

Are any of these documented or intended to be threadsafe?

If you make a type threadsafe, it should be safe across all its API
surface.

storage/filesystem/dotgit/dotgit.go
AriehSchneier: This PR is not about addressing all thread safety (see #773 for that), this PR was just about fixing the race conditions that were hit in the tests.
This specific change is just switching a raw bool which was trying to ensure we do something once only with a sync.Once, its a trivial change, very little performance impact but fixes the race condition that was encountered in the tests.

AriehSchneier:
@pjbgf do you think I should leave or remove the `Sync()`?

storage/filesystem/dotgit/dotgit.go
hanwen: >this PR was just about fixing the race conditions that were hit in the tests.

if the tests do concurrency on structures that do not support concurrency, the tests should be fixed or removed.

plumbing/format/packfile/fsobject.go
```diff
@@ -65,41 +67,57 @@ func (o *FSObject) Reader() (io.ReadCloser, error) {
...
-	_, err := o.pack.Seek(o.offset, io.SeekStart)
-	// fsobject aims to reuse an existing file descriptor to the packfile.
-	// In some cases that descriptor would already be closed, in such cases,
-	// open the packfile again and close it when the reader is closed.
-	if err != nil && errors.Is(err, os.ErrClosed) {
-		o.pack, err = o.fs.Open(o.packPath)
+	// Prefer opening a new file descriptor to enable concurrent reads without
+	// mutex contention. If the file doesn't exist (e.g., in test scenarios where
+	// the pack file is from a different filesystem), fall back to using o.pack.
+	file, err := o.fs.Open(o.packPath)
```
pjbgf, line 73: This fix has been superseded by https://github.com/go-git/go-git/pull/1998, as opening multiple FDs doesn't scale very well.

storage/filesystem/dotgit/dotgit.go
pjbgf: Thread safety is certainly something we need to improve across the board, I'm not sure we will be able to resolve all the existing concerns in one go though.

This change has been superseded by #2131. And as per my [comment](https://github.com/go-git/go-git/pull/2131#discussion_r3254081720) there, we should instead look into having a list of `incomingDirNames` and use `golang.org/x/sync/singleflight` to ensure each one is processed at most once.

storage/filesystem/dotgit/writers.go
```diff
@@ -122,6 +122,16 @@ func (w *PackWriter) Close() error {
 		return err
 	}
 
+	// Sync the file to ensure buffered data is written to stable storage.
+	// While Close() should make data visible, explicitly syncing before close
+	// ensures any OS-buffered writes are flushed, which may help prevent
+	// checksum mismatches when pack files are immediately read after creation.
+	if syncer, ok := w.fw.(interface{ Sync() error }); ok {
```
pjbgf, line 129: I'm not sure how this would help the specific problem we are trying to resolve here. As per upstream atomic creation of `.pack`, `.idx` and `.rev`, they all follow the sequence: `open(O_CREAT|O_RDWR|O_EXCL)` -> `write()` -> `close()` -> `rename()`.

My take is that if we are just talking about concurrent processes accessing **the final** file, post-rename, they should experience a consistent read.

That being said, upstream git has a [core.fsync](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corefsync) logic that provides a wide range of options on when/where `fsync` is called. This should be an issue by itself so that we could discuss what exactly would make sense to implement in go-git. Note that by default `pack` and `pack-metadata` do not call `fsync`.

AriehSchneier:
Closing in favor of #2139 which focuses on just the index map race fixes. The other fixes in this PR have been superseded by:
- #1998 (FSObject.Reader concurrency - uses ReadAt)  
- #2131 (DotGit.hasIncomingObjects - uses singleflight)

The Sync() change requires further discussion per @pjbgf's feedback about core.fsync configuration.

PR #2097, opened 2026-05-08
git: worktree path hardening

hiddeco:


ntfs.go
```diff
@@ -1,78 +1,10 @@
...
-import (
-	"runtime"
-	"strings"
-)
+import "runtime"
 
 // defaultProtectNTFS returns the default value for core.protectNTFS
 // when not explicitly configured. Matches upstream Git behaviour:
 // enabled by default on Windows.
 func defaultProtectNTFS() bool {
```
pjbgf: Let's move this around as well, it is not worth having a new file just for this.

hfs.go
```diff
@@ -1,91 +1,10 @@
...
-import (
-	"runtime"
-	"unicode"
-)
+import "runtime"
 
 // defaultProtectHFS returns the default value for core.protectHFS
 // when not explicitly configured. Matches upstream Git behaviour:
 // enabled by default on macOS.
 func defaultProtectHFS() bool {
```
pjbgf: Same for this

plumbing/object/tree.go
pjbgf, line 132: This would bypass validation via the Tree iterator, or anything going through the entries directly. Let's call here for ValidTreePath as well.

worktree_status.go
```diff
@@ -580,6 +581,19 @@ func (w *Worktree) fillEncodedObjectFromSymlink(dst io.Writer, path string, _ os
...
+	// mirroring the FindEntry-side gate that catches tree-derived names
+	// on the way out. Strict (always, regardless of protectHFS /
+	// protectNTFS): the index feeds future trees, and a tree must never
+	// carry .git, its HFS+/NTFS variants, or Windows reserved names —
+	// even when produced on a platform whose flag would otherwise
+	// tolerate them. The wrapper layer's tolerant validPath remains
+	// in place for legitimate submodule-cleanup reads/removes.
+	if err := pathutil.ValidTreePath(filename); err != nil {
+		return err
+	}
```
pjbgf: Let's move this away from here and directly into `index.Index`. Combined `index.Index` and `object.Tree` should be able to restrict the points needed to enforce this across the board.

worktree_commit.go
```diff
@@ -154,6 +155,9 @@ func (w *Worktree) CherryPick(commitOpts *CommitOptions, ortStrategyOption OrtMe
 
 			switch action {
 			case merkletrie.Delete:
+				if err := pathutil.ValidTreePath(change.From.Name); err != nil {
+					return err
+				}
```
pjbgf: This could potentially be delegated at `merkletrie/change.go`. WDYT?

worktree.go
```diff
@@ -806,14 +807,21 @@ func (w *Worktree) checkoutChange(ch merkletrie.Change, t *object.Tree, idx *ind
...
 			return err
 		}
 
 		isSubmodule = e.Mode == filemode.Submodule
 	case merkletrie.Delete:
-		return rmFileAndDirsIfEmpty(w.filesystem, ch.From.String())
+		from := ch.From.String()
+		if err := pathutil.ValidTreePath(from); err != nil {
+			return err
+		}
```
pjbgf: If check happens at Change, we remove the call here.

ntfs_test.go
pjbgf: Let's move these tests elsewhere so we don' need to keep this nor the `hfs_test.go` at root.

plumbing/object/tree.go
hiddeco: Done. `TreeWalker.Next` and `(*Tree).TreeEntryFile` both call `pathutil.ValidTreePath` now. That covers the iterator (which feeds `transformChildren` → `merkletrie.DiffTree`, `FileIter`, and the archive writers) and the `*File` boundary. Direct `Tree.Entries` access stays raw on purpose, this helps facilitate enumeration of entries of a known malformed tree.

worktree_status.go
hiddeco: `Index.Add(path)` now validates and returns `(*Entry, error)`. As every add / move flow that records a path in the index funnels through here, the worktree-side chokepoint at `addOrUpdateFileToIndex` is now gone. Combined with the tree-side gates (`FindEntry` / `TreeWalker.Next` / `TreeEntryFile`), `Index` and `Tree` own this end-to-end.

⚠️ The signature change is a breaking change.

worktree_commit.go
hiddeco: I went one level up: `TreeWalker.Next` validates each leaf, so the `Change` reaching `CherryPick` is name-validated transitively.

Kept the gate at the walker since that's where tree entries surface. A malformed entry stops the walk before merkletrie ever sees it. Putting it in `merkletrie/change.go` would push tree-specific knowledge into the generic differ.

worktree.go
hiddeco: Removed. `Modify`/`Insert` is covered by `FindEntry` one line down. `Delete` returns early via `rmFileAndDirsIfEmpty` against the tolerant wrapper, reached only from `resetWorktree`'s filesystem-vs-index diff, where we want cleanup of legitimately-tracked shapes like `submodule/.git` to succeed rather than abort the whole reset.

pjbgf (dismissed):

internal/pathutil/tree.go
```diff
@@ -0,0 +1,69 @@
...
+	for _, part := range parts {
+		if part == "." || part == ".." {
+			return fmt.Errorf("%w %q: cannot use %q", ErrInvalidPath, p, part)
+		}
+
+		if IsDotGitName(part) || IsHFSDotGit(part) {
+			return fmt.Errorf("%w component: %q", ErrInvalidPath, p)
+		}
+
+		if !WindowsValidPath(part) {
```
pjbgf: This seems to be more strict than upstream, as we are always running Windows checks even when not in Windows. In other places we run this only when `core.protectNTFS` is enabled.

pjbgf (approved):

PR #2101, opened 2026-05-09
git: Expose AlternatesFS on PlainOpenOptions

### LINKED ISSUE #2085: Expose AlternatesFS on PlainOpenOptions
`filesystem.Options.AlternatesFS` lets go-git follow `objects/info/alternates` paths that point outside the `.git` chroot, but `PlainOpenOptions` doesn't expose it. To set it you have to call `PlainOpen` for discovery, type-assert the storer, rebuild storage with `NewStorageWithOptions`, and reopen.

#1002 and #1006 both ended with users doing this by hand. I just did the same in git-pkgs.

Would you take a PR adding an `AlternatesFS billy.Filesystem` field to `PlainOpenOptions` and threading it through to `NewStorageWithOptions`? Default stays nil so the #953 hardening is unchanged unless the caller opts in.

andrew:
`filesystem.Options.AlternatesFS` lets go-git resolve `objects/info/alternates` paths that point outside the `.git` filesystem, but `PlainOpenWithOptions` had no way to pass it through. Callers currently have to open the repo, type-assert the storer, rebuild storage with `NewStorageWithOptions`, and reopen.

Add `AlternatesFS` to `PlainOpenOptions` and thread it through to `NewStorageWithOptions`. A nil value keeps current behavior so the #953 hardening is unchanged unless the caller opts in.

Fixes #2085

### DIFF
--- a/options.go
+++ b/options.go
@@ -813,6 +813,10 @@ type PlainOpenOptions struct {
 	// DetectDotGit defines whether parent directories should be
 	// walked until a .git directory or file is found.
 	DetectDotGit bool
+	// AlternatesFS provides the billy filesystem to be used for Git Alternates.
+	// If none is provided, it falls back to using the underlying instance used for
+	// DotGit.
+	AlternatesFS billy.Filesystem
 }
 
 // Validate validates the fields and sets the default values.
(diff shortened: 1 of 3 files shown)


pjbgf (approved):
@andrew thanks for working on this. 🙇

PR #2105, opened 2026-05-11
plumbing: object, tree entry validation

hiddeco:
Closes the encode/decode boundary that #2081 and #2097 left tolerant. `Tree.Encode` previously accepted entry names like `.git`, `..`, control characters, or HFS+/NTFS `.git` disguises, and never checked for duplicates; `Tree.Decode` did not validate names at all.

`Tree.Encode` now runs each name through `pathutil.ValidTreePath` and rejects duplicate names — including the file/dir collision where `foo` and `foo/` share a bare name. Callers that need to emit non-conformant bytes write them via `plumbing.EncodedObject` directly.

`Tree.Decode` stays permissive — upstream Git makes the same split between `parse_tree_buffer` and `fsck_tree`[1], and the permissive parse keeps inspection and recovery use cases working. The new `Tree.Validate` is the opt-in read-side counterpart, mirroring `fsck_tree`'s structural rules: null hashes, slashes, empty/`.`/`..`/`.git` names, bad file modes, duplicates, mis-sort, oversized names, and the four `*_SYMLINK` rules for `.gitmodules`/`.gitattributes`/`.gitignore`/`.mailmap`. Violations come back as `errors.Join` of rule errors wrapping `ErrInvalidTree`.

Two `fsck_tree` warnings need the raw octal mode string — zero-padded and non-canonical `0100664` modes — which `canonicalTreeMode` drops at parse. Those are wire-only and best handled by a future buffer-level validator.

[1]: https://github.com/git/git/blob/v2.54.0/fsck.c#L616-L800

### DIFF
--- a/internal/pathutil/hfs.go
+++ b/internal/pathutil/hfs.go
@@ -97,3 +97,15 @@ func IsHFSDotGit(part string) bool { return IsHFSDot(part, "git") }
 // ".gitmodules", catching attempts to plant the file via Unicode
 // code points that HFS+ would strip during normalisation.
 func IsHFSDotGitmodules(part string) bool { return IsHFSDot(part, "gitmodules") }
+
+// IsHFSDotGitattributes reports whether part is an HFS+ equivalent of
+// ".gitattributes".
+func IsHFSDotGitattributes(part string) bool { return IsHFSDot(part, "gitattributes") }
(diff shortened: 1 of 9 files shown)


pjbgf (approved):

PR #2117, opened 2026-05-12
v5: git: Allow MkdirAll on worktree-root paths

hiddeco:
The worktree filesystem wrapper rejects `""`, `"."`, and `"/"` in `MkdirAll` via `validPath`, even though `validReadPath` was already relaxed to treat those paths as legitimate references to the worktree root. The asymmetry was deliberate for genuine mutations — mutating the root itself is not a sensible operation for `Create` or `Remove` — but `MkdirAll` on an existing directory is not a mutation: it asks the filesystem to ensure a directory exists, against one that always does.

Short-circuit `MkdirAll` to a no-op for root-equivalents before `validPath` runs. `validPath` itself stays strict, so the other mutating operations continue to reject the root and the existing `TestValidPath` table is untouched. A new
`TestWorktreeFilesystemMkdirAllRootIsNoop` locks in the no-op contract against the wrapper.

Backport of #2115.

### DIFF
--- a/worktree_fs.go
+++ b/worktree_fs.go
@@ -123,6 +123,14 @@ func (sfs *worktreeFilesystem) Readlink(link string) (string, error) {
 }
 
 func (sfs *worktreeFilesystem) MkdirAll(path string, perm os.FileMode) error {
+	// MkdirAll on the worktree root is a no-op: the root always exists,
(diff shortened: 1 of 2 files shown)


pjbgf (approved):

PR #2119, opened 2026-05-13
plumbing: format decoder input bounds and contracts

hiddeco:
Four changes align `plumbing/format/*` decoder behavior with canonical Git invariants and clarify two internal API contracts.

`idxfile.Decoder.Decode` ports the size-consistency formula from `packfile.c:load_idx`[1]: an idx v2 file with `nr` objects (the last fanout entry) and hash size `hashsz` must have an on-disk length in `[minSize, maxSize]` derived from `nr` via overflow-checked multiplication. The decoder now buffers the input at the top of `Decode` and rejects inputs whose declared object count is inconsistent with the file size. All current callers in `storage/filesystem` and the HTTP dumb transport pass file-backed readers (the dumb transport opens the .idx after it has been downloaded to local storage), so the buffer cost is at most one allocation the size of the .idx file; in practice that is bounded by the size of the underlying pack file.

`packfile.Scanner` enforces the structural invariant that a packfile object's inflated length does not exceed the size declared in its header, at every site that consumes the inflated stream. The streaming `inflateContent` and forward-scan `objectEntry` paths wrap their write sinks in `boundedWriter`, which returns `ErrInflatedSizeMismatch` on overrun. The lazy `FSObject.Reader` and mmap `ondemandObject.Reader` wrap their returned `zlibReadCloser` in `BoundedReadCloser`, so an `io.ReadAll` on an opened object cannot stream past the declared size either. The call sites in `parser.go` and `packfile.go` thread the declared size through, including the delta paths where the size refers to the delta instruction stream rather than the resolved result.

`gitignore.wildmatch` is replaced with a port of the recursive `dowild` function in canonical Git's `wildmatch.c`[2] at tag v2.54.0. The previous Go implementation was an iterative backtracker; the port preserves the upstream algorithm exactly — including the `WM_ABORT_TO_STARSTAR` and `WM_ABORT_ALL` shortcuts, the bracket-expression state machine, and POSIX character-class evaluation — while taking an idiomatic Go shape: string slicing in place of `[]byte` conversions, `strings.IndexByte` instead of a hand-rolled helper, and a regular switch in place of the C `goto`-via-bool pattern. The `t3070-wildmatch.sh`-anchored conformance suite passes unchanged.

`objfile.Reader.Read` and `Reader.Hash` now require a successful `Header` call before they will return data, returning the new `ErrHeaderNotRead` sentinel (and a zero `plumbing.Hash{}` respectively) otherwise. The two-phase contract — `Header`, then `Read` — is unchanged; this commit makes it explicit at the API surface.

[1]: https://github.com/git/git/blob/v2.54.0/packfile.c#L232-L258
[2]: https://github.com/git/git/blob/v2.54.0/wildmatch.c#L59-L283

### DIFF
--- a/plumbing/format/objfile/reader.go
+++ b/plumbing/format/objfile/reader.go
@@ -13,9 +13,10 @@ import (
 
 // Errors returned by the objfile package.
 var (
-	ErrClosed       = errors.New("objfile: already closed")
-	ErrHeader       = errors.New("objfile: invalid header")
-	ErrNegativeSize = errors.New("objfile: negative object size")
+	ErrClosed        = errors.New("objfile: already closed")
+	ErrHeader        = errors.New("objfile: invalid header")
+	ErrHeaderNotRead = errors.New("objfile: Header must be called before Read")
+	ErrNegativeSize  = errors.New("objfile: negative object size")
 )
 
 // Reader reads and decodes compressed objfile data from a provided io.Reader.
@@ -104,12 +105,26 @@ func (r *Reader) prepareForRead(t plumbing.ObjectType, size int64) {
 //
 // If Read encounters the end of the data stream it will return err == io.EOF,
 // either in the current call if n > 0 or in a subsequent call.
+//
+// Read returns ErrHeaderNotRead if Header has not been called successfully.
 func (r *Reader) Read(p []byte) (n int, err error) {
+	if r.multi == nil {
+		return 0, ErrHeaderNotRead
+	}
 	return r.multi.Read(p)
 }
 
 // Hash returns the hash of the object data stream that has been read so far.
+// It returns a zero plumbing.Hash carrying the Reader's configured object
+// format if Header has not been called successfully — the format matters
+// because [plumbing.Hash] encodes it internally and the result feeds
+// serialisers that emit a format-sized byte slice.
 func (r *Reader) Hash() plumbing.Hash {
+	if r.multi == nil {
+		var h plumbing.Hash
+		h.ResetBySize(r.objectFormat.Size())
+		return h
+	}
 	return r.hasher.Sum()
 }
 
--- a/plumbing/format/packfile/scanner.go
+++ b/plumbing/format/packfile/scanner.go
@@ -34,7 +34,96 @@ var (
 	ErrUnsupportedVersion = NewError("unsupported packfile version")
 	// ErrSeekNotSupported returned if seek is not support.
 	ErrSeekNotSupported = NewError("not seek support")
+	// ErrInflatedSizeMismatch is returned when a packfile object inflates to
+	// more bytes than the size declared in its object header. A well-formed
+	// packfile never produces more data than the declared size; exceeding it
+	// indicates a structurally invalid entry.
+	ErrInflatedSizeMismatch = errors.New("packfile: inflated object exceeds declared size")
 )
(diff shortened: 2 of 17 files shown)


plumbing/format/objfile/reader.go
```diff
@@ -104,12 +105,21 @@ func (r *Reader) prepareForRead(t plumbing.ObjectType, size int64) {
...
+		return 0, ErrHeaderNotRead
+	}
 	return r.multi.Read(p)
 }
 
 // Hash returns the hash of the object data stream that has been read so far.
+// It returns a zero plumbing.Hash if Header has not been called successfully.
 func (r *Reader) Hash() plumbing.Hash {
+	if r.multi == nil {
+		return plumbing.Hash{}
```
pjbgf: This needs to be objectformat aware, else it could lead to malformed results like a ZeroHash but for the specific `r.objectFormat`.

plumbing/format/packfile/scanner.go
```diff
@@ -34,8 +34,86 @@ var (
...
+	closer  io.Closer
+	overrun bool
+}
+
+// NewBoundedReadCloser wraps rc so that the cumulative bytes returned from
+// Read never exceed limit. The first call that would have returned a byte
+// past limit instead returns ErrInflatedSizeMismatch; subsequent calls
+// keep returning the same error.
+func NewBoundedReadCloser(rc io.ReadCloser, limit int64) *BoundedReadCloser {
+	return &BoundedReadCloser{
```
pjbgf, line 103: Let's check for `limit` not being negative.

plumbing/format/packfile/scanner.go
```diff
@@ -439,6 +518,13 @@ func objectEntry(r *Scanner) (stateFn, error) {
 		mw = io.MultiWriter(mw, oh.content)
 	}
 
+	// Bound the inflated stream by the size declared in the object header.
+	// A well-formed packfile never produces more inflated bytes than that
+	// value, so any overrun signals a malformed entry. For delta entries
+	// the declared size is the size of the delta instruction stream, not
+	// the resolved object — same property, different semantic.
```
pjbgf:
```suggestion
	// Bind the inflated stream by the size declared in the object header.
	// A well-formed packfile never produces more inflated bytes than that
	// value, so any overrun signals a malformed entry. For delta entries
	// the declared size is the size of the delta instruction stream, not
	// the resolved object.
```

plumbing/format/packfile/scanner.go
```diff
@@ -34,8 +34,86 @@ var (
...
+
+// Write forwards p to the underlying writer while keeping the running total
+// at or below limit. On overrun it forwards the legal prefix and reports
+// the number of bytes actually consumed alongside ErrInflatedSizeMismatch,
+// matching the contract in io.Writer.
+func (b *boundedWriter) Write(p []byte) (int, error) {
+	if b.n+int64(len(p)) > b.limit {
+		remain := int(b.limit - b.n)
+		if remain > 0 {
+			_, _ = b.w.Write(p[:remain])
```
pjbgf: Instead of ignoring the error from `Write`, let's join/wrap it with `ErrInflatedSizeMismatch`, to avoid hard to debug issues.

pjbgf (approved):

PR #2121, opened 2026-05-13
plumbing: fuzz packp and capability decoders

hiddeco:


pjbgf (approved):

PR #2122, opened 2026-05-13
plumbing: object, Isolate tests from host git config.

cedric-appdirect:
Tests that create in-memory repositories and commit via the top-level git.Init/Worktree.Commit API inherit the host global git config through the NewAuto() ConfigLoader plugin. When a developer has commit.gpgSign=true, these tests fail because no ObjectSigner plugin is registered.

Add TestMain to plumbing/object and x/plumbing/worktree that replaces NewAuto() with a static config, mirroring the existing pattern in the root git package.

Assisted-by:i Cursor with Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/common_test.go
+++ b/common_test.go
@@ -14,15 +14,13 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
 
-	"github.com/go-git/go-git/v6/config"
+	testconfig "github.com/go-git/go-git/v6/internal/test/config"
 	"github.com/go-git/go-git/v6/plumbing"
 	"github.com/go-git/go-git/v6/plumbing/cache"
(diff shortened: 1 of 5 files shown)


AriehSchneier:
Love this idea, I currently have to turn off gpgSign every time I run CI. My only comment would be, this could be necessary for more folders and we will be duplicating the code, maybe create a util somewhere so that each `TestMain` is just a call to the single util?

pjbgf (changes requested):
I'd agree with @AriehSchneier. We actually already call this in a few locations, let's create a new `internal/test/config` and add this logic there instead.

pjbgf (approved):
@cedric-appdirect thanks for improving our test harness! 🙇

PR #2127, opened 2026-05-15
git: worktree, fix successive branch checkouts from same commit. Fixes #2124

### LINKED ISSUE #2124: Successive Branches Retain Commits from Previous Iterations
### Bug Description

After cloning a repository into memory (go-billy memfs), procedurally creating and checking out a new local branch, committing changes to the working tree, and repeating this workflow multiple times against a predetermined HEAD hash - each successive iteration includes all commits and changes from prior iterations.

Each branch created against the same hash reference should be isolated, excluding any changes made in a prior iteration. Instead, even when the hash reference explicitly points to a commit prior to these changes, they are still carried forward into each subsequent branch.

### go-git Version

v6.0.0-alpha.1

### Steps to Reproduce

1. Clone a repository using go-billy memfs.
2. Record the HEAD hash of the default branch before making any changes.
3. Create a new local branch based on the recorded hash.
4. Add a file to the working tree, stage it to the index, and create a commit.
5. Repeat steps 3–4 and observe that all files staged in prior iterations are included in each successive branch.

Code Example:
```golang
type client struct {
  repo *git.Repository
  hash plumbing.Hash
}

func NewClient() (*client, error) {
  store := memory.NewStorage()
  fs := memfs.New()
  repo, _ := git.Clone(store, fs, &git.CloneOptions{
    SingleBranch: false,
    URL: url,
  })
  head, _ := repo.Head()
  return &client{
    repo: repo,
    hash: head.Hash(),
  }, nil
}

func (c *client) Do(branch, file, content, msg string) {
  name := plumbing.NewBranchReferenceName(branch)
  ref := plumbing.NewHashReference(name, c.hash)
  c.repo.Storer.SetReference(ref)
  wt, _ := c.repo.Worktree()
  wt.Checkout(&git.CheckoutOptions{
    Branch: name,
    Force: true,
  })
  fileRef, _ := wt.Filesystem.Create(file)
  fileRef.Write([]byte(content))
  fileRef.Close()
  wt.Add(file)
  wt.Commit(msg, &git.CommitOptions{
    Author: &object.Signature{
      Name: "Nick",
      Email: "nick@test.com",
      When: time.Now(),
    },
  })
}

func (c *client) Files() []string {
  wt, _ := c.repo.Worktree()
  var files []string
  util.Walk(wt.Filesystem, ".", func(path string, info os.FileInfo, err error) error {
    files = append(files, path)
    return nil
  })
  return files
}

func Test(t *testing.T) {
  var branches = map[string]struct {
    file string
    content string
    msg string
  }{
    "branch1": {
      file: "test1.txt",
      content: "test1.txt",
      msg: "commit1",
    },
    "branch2": {
      file: "test2.txt",
      content: "test2.txt",
      msg: "commit2",
    },
    "branch3": {
      file: "test3.txt",
      content: "test3.txt",
      msg: "commit3",
    },
  }

  client, _ := NewClient()
  for branch, data := range branches {
    t.Run(branch, func(t *testing.T) {
      client.Do(branch, data.file, data.content, data.msg)
      t.Logf("%s Files: %v", branch, client.Files())
    })
  }
}
```

### Additional Information

This behaviour is identical to that described in a previously raised issue, #1439. That issue was resolved by pre-creating all branches upfront, as the hash reference was found to be acting as a moving pointer between iterations. However, in this case, the same behaviour is observed even after explicitly setting an identical commit hash on each successive branch creation.

AriehSchneier:
## Summary

Fixes #2124 - When checking out multiple branches from the same base commit with `Force:true`, files from previous iterations were incorrectly preserved in the worktree.

## Root Cause

`Checkout()` updated HEAD before calling `Reset()`, so when `Reset()` captured the "previous" tree via `headTree()`, it got the tree from the NEW commit (just set by Checkout) rather than the OLD commit (where the worktree actually was). This caused `resetWorktreeToTree` to diff the same tree against itself, resulting in no changes.

## Solution

Capture the current tree BEFORE updating HEAD in `Checkout()`, and pass it to `Reset()` via a new internal field in `ResetOptions`. This ensures `HardReset` and `KeepReset` properly diff from the actual previous state.

## Changes

- Added internal `fromTree` field to `ResetOptions`
- Modified `Checkout()` to capture tree before updating HEAD
- Modified `Reset()` to use provided `fromTree` when available
- Added comprehensive regression test `TestCheckoutSuccessiveBranchesFromSameCommit`

## Testing

- ✅ All existing tests pass (100+ worktree and reset tests)
- ✅ New regression test reproduces and validates the fix for issue #2124
- ✅ Verified behavior matches git's handling of successive branch checkouts

## AI Disclosure

This contribution was assisted by Claude Sonnet 4.5 as documented in the `Assisted-by:` commit trailer, in accordance with the [AI Contribution Policy](https://github.com/go-git/go-git/blob/main/AI_POLICY.md).

### DIFF
--- a/options.go
+++ b/options.go
@@ -460,6 +460,12 @@ type ResetOptions struct {
 
 	// SkipSparseDirValidation will skip the validation for SparseDirs.
 	SkipSparseDirValidation bool
+
+	// fromTree is used internally by Checkout to pass the tree that the
+	// worktree is currently at, before HEAD is updated. This ensures that
+	// HardReset and KeepReset properly diff from the actual previous state
+	// rather than the new HEAD.
+	fromTree *object.Tree
 }
 
 // Validate validates the fields and sets the default values.
--- a/worktree.go
+++ b/worktree.go
@@ -215,6 +215,16 @@ func (w *Worktree) Checkout(opts *CheckoutOptions) error {
 		ro.Mode = SoftReset
 	}
 
+	// For HardReset and KeepReset, capture the current tree BEFORE updating
+	// HEAD. This ensures resetWorktreeToTree correctly diffs from where we
+	// actually are, not from where HEAD will point after the update.
+	if ro.Mode == HardReset || ro.Mode == KeepReset {
+		ro.fromTree, err = w.headTree()
+		if err != nil {
+			return err
+		}
+	}
+
 	if !opts.Hash.IsZero() && !opts.Create {
 		err = w.setHEADToCommit(opts.Hash)
 	} else {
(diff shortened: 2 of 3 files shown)


pjbgf (approved):

PR #2129, opened 2026-05-16
internal: server/git, Fix TestGitServer_Timeout flakiness on Windows CI

aymanbagabas:
Fixes a flaky `TestGitServer_Timeout` reported by @AriehSchneier in #2053.

The test compares an elapsed duration measured on the client goroutine against the server's idle timeout, but the server arms its deadline on a separate goroutine. On loaded Windows runners the server can arm the deadline before the client captures `start := time.Now()`, causing elapsed to fall short.

This change anchors the two clocks by sending a 1-byte probe before timing: pkt-line decoding needs 4 bytes for the length header, so the server consumes the byte and re-enters `Read`, arming a fresh idle deadline at approximately the same wall-clock moment as the client's `start`. Recording `start` before the `Write` keeps any scheduling latency on the elapsed side, making the lower-bound assertion robust.

Note: the server-side proposal in #2053 (deferring the initial `updateDeadline()` to first `Read`) does not eliminate the race because `serverConn.Read` already calls `updateDeadline()` at the start of every read.

cc @AriehSchneier

### DIFF
--- a/internal/server/git/git_test.go
+++ b/internal/server/git/git_test.go
@@ -176,20 +176,24 @@ func TestGitServer_Timeout(t *testing.T) {
 	require.NoError(t, err)
 	defer conn.Close()
 
-	// Stay idle past the configured Timeout; the server must close
-	// the connection, surfacing as a Read error.
 	require.NoError(t, conn.SetReadDeadline(time.Now().Add(5*time.Second)))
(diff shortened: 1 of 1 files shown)


AriehSchneier (approved):

pjbgf (approved):

PR #2131, opened 2026-05-16
storage: dotgit, fix hasIncomingObjects race

hiddeco:


storage/filesystem/dotgit/dotgit.go
```diff
@@ -726,9 +729,11 @@ func (d *DotGit) incomingObjectPath(h plumbing.Hash) string {
 }
 
 // hasIncomingObjects searches for an incoming directory and keeps its name
-// so it doesn't have to be found each time an object is accessed.
+// so it doesn't have to be found each time an object is accessed. The
+// lazy initialisation runs under sync.Once so concurrent callers cannot
+// race on the cached fields.
 func (d *DotGit) hasIncomingObjects() bool {
-	if !d.incomingChecked {
+	d.incomingOnce.Do(func() {
```
pjbgf, line 736: The existing code seems to imply that incoming dirs would happen at most once within the lifetime of the dotgit instance, in which case the use of `sync.Once` here is appropriate.

My take is that that is not the case, instead `incomingDirName` should be an array instead, and to ensure we process each dir at most once, we could use `golang.org/x/sync/singleflight` instead.

pjbgf: The existing assumptions would be largely correct for CLI's, but maybe would not be the case for long running processes. I'll go ahead and merge this as fix the current race condition. But the long term of this feature may need to change as a follow-up.

pjbgf (approved):

AriehSchneier:
@pjbgf That was part of this PR, which also has some other race fixes:
https://github.com/go-git/go-git/pull/2094

PR #2132, opened 2026-05-16
storage: pool .pack file descriptors via internal/packhandle

hiddeco:
Introduces `internal/packhandle`, a per-pack file-descriptor pool that opens lazily, shares one descriptor across concurrent readers, and grace-closes on idle. `storage/filesystem` routes its read paths through the pool via a new `DotGit.OpenPackForReading`.

`KeepDescriptors` and `MaxOpenDescriptors` become silently honoured no-ops.

## Performance

`BenchmarkObjectStorage_PackHandle`, basic fixture, Apple M4 Pro, `darwin/arm64`, `n=5`, `benchstat` vs `origin/main`:

| sub-benchmark | main | branch | Δ |
|---|---|---|---|
| `EncodedObject/G=1/fd` | 44.83 µs | 4.91 µs | −89.06% |
| `EncodedObject/G=NumCPU/fd` | 42.12 µs | 16.03 µs | −61.95% |
| `EncodedObject/G=1/mmap` | 46.31 µs | 1.47 µs | −96.82% |
| `EncodedObject/G=NumCPU/mmap` | 44.70 µs | 1.70 µs | −96.20% |
| `HasEncodedObject/G=1/fd` | 21.12 µs | 20.72 µs | −1.89% |
| `HasEncodedObject/G=NumCPU/fd` | 19.45 µs | 19.87 µs | +2.19% |
| `HasEncodedObject/G=1/mmap` | 20.34 µs | 20.31 µs | ~ |
| `HasEncodedObject/G=NumCPU/mmap` | 20.02 µs | 20.05 µs | ~ |
| **geomean** | 29.99 µs | 8.71 µs | **−70.97%** |

PR #2133, opened 2026-05-16
plumbing, storage: add CloseIdleDescriptors soft-close primitive

hiddeco:
Adds an opt-in `CloseIdleDescriptors` surface that drops idle pack file descriptors without invalidating storage state. For long-running processes (servers, daemons) that keep a `Storage` open across idle periods, this enables bounded FD usage on demand without losing the warm `*PackHandle` cache or paying the terminal `Close`-and-reopen cost.

The primitive lives on the unexported `sharedFile` as `releaseNow`: closes the FD inline when no readers are active, otherwise latches "close on next quiescence" so in-flight reads finish normally and the FD drops the instant the last release brings `refs` to zero. The next reader transparently reopens via the source opener. The public surface fans out across three FD owners — `LazyIndex.CloseIdleDescriptors` (idx/rev), `PackHandle.CloseIdleDescriptors` (pack + the cached `LazyIndex` inside the handle), and `ObjectStorage.CloseIdleDescriptors` (`DotGit`'s `PackHandle` catalog, the per-pack `LazyIndex` cache distinct from the one inside the handle, and cached alternates) — with `Storage.CloseIdleDescriptors` promoted from the embedded `ObjectStorage`. The `storer.IdleReleaser` interface exposes this up the stack.

Builds on the PackHandle refcounted FD lifecycle (#2132); prerequisite for the pack hot-path + LRU FD pool work (#2134).

### DIFF
--- a/plumbing/format/packfile/fsobject.go
+++ b/plumbing/format/packfile/fsobject.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"math"
 	"os"
+	stdsync "sync"
 
 	billy "github.com/go-git/go-billy/v6"
 
@@ -16,6 +17,46 @@ import (
 	"github.com/go-git/go-git/v6/utils/ioutil"
 	"github.com/go-git/go-git/v6/utils/sync"
 )
+
+// probeSize is the byte count for the closed-FD probe. A one-byte
+// ReadAt distinguishes a live descriptor from a closed one without
+// mutating the file's seek cursor; a zero-length read is unusable
+// because some implementations (e.g. [os.File]) return (0, nil) on
+// a closed file.
+const probeSize = 1
+
+// probeBufPool returns the per-call backing array for [probePack].
+// Pooling keeps the read path allocation-free on what is a very hot
+// code path.
+var probeBufPool = stdsync.Pool{
+	New: func() any {
+		var buf [probeSize]byte
+		return &buf
+	},
+}
+
+// probePack tests whether pack is still readable at offset by
+// issuing a [probeSize]-byte [io.ReaderAt.ReadAt]. The error is
+// returned verbatim so any wrapping context (path, syscall) is
+// preserved for the caller; classification is left to
+// [errors.Is]:
+//
+//   - nil means the descriptor is live; the caller may keep using
+//     pack.
+//   - an error matching [os.ErrClosed] means the descriptor has
+//     been closed and the caller should reopen the file.
+//   - any other error is propagated, matching the canonical Git
+//     behaviour in `packfile.c:use_pack`, which does not retry on
+//     transient I/O errors.
+//
+// [io.EOF] indicates the offset is at or past end-of-file, which
+// implies a truncated pack — propagate rather than masking.
+func probePack(pack io.ReaderAt, offset int64) error {
+	buf := probeBufPool.Get().(*[probeSize]byte)
+	defer probeBufPool.Put(buf)
+	_, err := pack.ReadAt(buf[:], offset)
+	return err
+}
 
 // FSObject is an object from the packfile on the filesystem.
 type FSObject struct {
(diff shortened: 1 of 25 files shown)


storage/filesystem/storage.go
```diff
@@ -31,16 +32,29 @@ type Storage struct {
...
-	// MaxOpenDescriptors is the max number of file descriptors to keep
-	// open. If KeepDescriptors is true, all file descriptors will remain open.
+	// MaxOpenDescriptors used to cap the number of .pack file
+	// descriptors kept open in an LRU ring.
+	//
+	// Deprecated: Pack file descriptors auto-release after a
+	// 1-second idle grace period, so an explicit cap is no longer
+	// meaningful. Setting this field has no effect. Will be removed
+	// in v7.
 	MaxOpenDescriptors int
```
pjbgf: Given that we are ignoring these, there could be a behaviour impact on existing consumers. Due to that, I rather we remove and make it a breaking change in `v6`.

storage/filesystem/dotgit/dotgit.go
```diff
@@ -89,8 +93,12 @@ type Options struct {
...
 	ExclusiveAccess bool
-	// KeepDescriptors makes the file descriptors to be reused but they will
-	// need to be manually closed calling Close().
+	// KeepDescriptors used to make pack file descriptors stay open
+	// across calls to ObjectPack.
+	//
+	// Deprecated: descriptor lifecycle is now owned by
+	// packhandle.PackHandle (acquired via DotGit.PackHandle); this
+	// field has no effect. Will be removed in v7.
 	KeepDescriptors bool
```
pjbgf: Same for this, given that the contract with the users is broken, we may as well formalise it by removing it. The benefits of the new approach out-weight the loss of control this provided.

```suggestion
```

plumbing/format/packhandle/packhandle.go
```diff
@@ -0,0 +1,303 @@
+package packhandle
```
pjbgf: Let's move this within `x/plumbing`.

plumbing/format/idxfile/shared_file.go
```diff
@@ -5,52 +5,68 @@ import (
...
+//
+// The opener MUST return a billy.ReaderAtCloser whose ReadAt is
+// concurrent-safe across goroutines (the standard contract for
+// io.ReaderAt and the documented contract for billy.ReaderAtFS).
+// SharedFile does not synchronise concurrent ReadAt calls itself.
 //
 // All methods are safe for concurrent use.
-type sharedFile struct {
-	opener      openFileFunc
+type SharedFile struct {
```
pjbgf: potentially not needed longer term, so I'd rather we don't export it or export it within `internal`.

storage/filesystem/storage.go
```diff
@@ -31,6 +32,13 @@ type Storage struct {
 	ReflogStorage
 }
 
+// CloseIdleDescriptors is promoted from the embedded
+// [ObjectStorage] via Go's method-set promotion rules. The
+// compile-time assertion pins the relationship to
+// [storer.IdleReleaser] so a future rename or signature change
+// on either side breaks the build immediately.
+var _ storer.IdleReleaser = (*Storage)(nil)
+
```
pjbgf, line 45: Let's assert the same for `DotGit`:

```
	_ storer.IdleReleaser = (*dotgit.DotGit)(nil)
```

plumbing/format/packfile/fsobject.go
```diff
@@ -17,6 +18,50 @@ import (
...
+// implies a truncated pack — propagate rather than masking.
+func probePack(pack io.ReaderAt, offset int64) error {
+	buf := probeBufPool.Get().(*[probeSize]byte)
+	defer probeBufPool.Put(buf)
+	_, err := pack.ReadAt(buf[:], offset)
+	if err == nil {
+		return nil
+	}
+	if errors.Is(err, os.ErrClosed) {
+		return os.ErrClosed
```
pjbgf: To avoid losing any additional wrapped info:

```suggestion
		return err
```

pjbgf (approved):

PR #2134, opened 2026-05-16
storage: pack read optimisations and storage-wide LRU FD pool

hiddeco:
This PR primarily addresses resource hygiene and concurrent correctness: bounded FDs across a Storage (and across multiple Storages via `Options.Pool`), `*fdpool.Pool.Stats()` for observability, and four `s.index` race fixes. Read-path performance changes are minimal — measured against `main` with `benchstat (-count=6, -benchtime=1s)`:

- Allocations reduced 14% geomean, up to 47% on hot read paths (ParallelReads, AlternatesObjectLookup/EncodedObject). Lower GC pressure.
- Alternates lookup latency -59% (with variance caveats; replication recommended).
- Flat throughput on the primary read path (ParallelReads).
- Single-pack ReindexThenRead is +14% slower (cost of singleflight ceremony when there's no parallelisation benefit to capture).

### DIFF
--- a/storage/filesystem/dotgit/dotgit.go
+++ b/storage/filesystem/dotgit/dotgit.go
@@ -31,6 +31,7 @@ import (
 	plumbhash "github.com/go-git/go-git/v6/plumbing/hash"
 	"github.com/go-git/go-git/v6/storage"
 	"github.com/go-git/go-git/v6/utils/ioutil"
+	"github.com/go-git/go-git/v6/x/fdpool"
 )
 
 const (
@@ -107,6 +108,11 @@ type Options struct {
 	// WriteReverseIndex controls whether .rev files are written when
 	// creating new packfiles. Defaults to true.
 	WriteReverseIndex bool
+	// Pool, when non-nil, governs the LRU eviction of file
+	// descriptors held by [packhandle.PackHandle] instances
+	// constructed by this DotGit. Nil disables pooling for the
+	// pack-FD lifecycle (grace-period close on quiescence).
+	Pool *fdpool.Pool
 }
 
 // The DotGit type represents a local git repository on disk. This
@@ -511,7 +517,7 @@ func (d *DotGit) packHandle(hash plumbing.Hash) (*packhandle.PackHandle, error) 
 		},
 	}
 
-	ph, err := packhandle.New(sources, hash)
+	ph, err := packhandle.NewWithPool(sources, hash, d.options.Pool)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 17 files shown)


plumbing/format/idxfile/idxfile.go
```diff
@@ -38,6 +39,19 @@ type Index interface {
...
+	// returns all entries (equivalent to Entries). The returned
+	// iterator must be Closed by the caller to release any held
+	// resources.
+	EntriesWithPrefix(prefix []byte) (EntryIter, error)
+	// MayContain reports whether the index might contain h. A false
+	// return is authoritative ("h is definitely not in this pack")
+	// based on the idx fanout table; true means the caller should
+	// call Contains or FindOffset for a definitive answer. The
+	// check is O(1) and does not touch the underlying file.
+	MayContain(h plumbing.Hash) bool
```
pjbgf: Let's handle this as a separate change.

```suggestion
```

plumbing/format/idxfile/shared_file.go
```diff
@@ -4,60 +4,103 @@ import (
...
 )
 
-const defaultCloseGracePeriod = time.Second
+// PoolMember is the minimal contract for participation in a Pool.
+// *SharedFile satisfies it via ReleaseNow.
+//
+// The fdpool package defines a structurally identical contract;
+// keeping the type here lets the import dependency run one way
+// (fdpool → idxfile), avoiding a cycle.
+type PoolMember interface {
```
pjbgf: Let's move it outside `idxfile` ideally in `x/` or somewhere `internal`.

plumbing/format/packfile/fsobject.go
```diff
@@ -2,100 +2,80 @@ package packfile
...
-	fs       billy.Filesystem
-	pack     billy.File
-	packPath string
-	cache    cache.Object
+	hash   plumbing.Hash
+	offset int64
+	size   int64
+	typ    plumbing.ObjectType
+	index  idxfile.Index
+	handle *packhandle.PackHandle
```
pjbgf: Not sure this is the correct link to object store, but this can be revisited later on the separate use cases we discussed.

plumbing/format/idxfile/idxfile.go
```diff
@@ -359,6 +400,58 @@ func (i *idxfileEntryIter) Close() error {
 	return nil
 }
 
+// idxfilePrefixIter walks a single fanout bucket of a MemoryIndex,
+// yielding entries whose hash starts with prefix. The bucket is
+// sorted by hash, so once a name is read whose first bytes do not
+// match prefix the iterator stops.
+type idxfilePrefixIter struct {
+	idx    *MemoryIndex
```
pjbgf: It would be better if we don't need to keep ref to `MemoryIndex` as that is rather heavy. A cursor + source should suffice for iterating through an `idx`.

storage/filesystem/dotgit/dotgit.go
```diff
@@ -1463,7 +1541,15 @@ func (d *DotGit) Alternates() ([]*DotGit, error) {
 		if err != nil {
 			return nil, fmt.Errorf("cannot chroot %q: %w", path, err)
 		}
-		alternates = append(alternates, New(afs))
+		// Inherit the parent DotGit's FD pool so the storage-wide
+		// budget covers alternate packs and idxes too. Other
+		// Options keep their defaults from New.
+		alternates = append(alternates, NewWithOptions(afs, Options{
+			ObjectFormat:      formatcfg.DefaultObjectFormat,
```
pjbgf: This must align with the repository to avoid data corruption:

```suggestion
			ObjectFormat:      d.options.ObjectFormat,
```

storage/filesystem/object.go
```diff
@@ -191,54 +201,126 @@ func (s *ObjectStorage) requireIndex() error {
 	}
 	s.muI.RUnlock()
 
-	s.muI.Lock()
-	defer s.muI.Unlock()
+	_, err, _ := s.indexSF.Do("populate", func() (any, error) {
```
pjbgf: Let's make `populate` a `const`.

pjbgf (approved):

PR #2140, opened 2026-05-19
plumbing: protocol/packp, Handle shallow-clone no-op push in update-request decoder

shanzhubb:
## What
Fix `UpdateRequests.Decode` to gracefully handle a shallow-clone no-op push
where the client sends shallow lines followed immediately by a flush packet
with no command lines.

## Why
When a standard Git client (git version 2.50.1) pushes from a shallow clone and has no changes to
push, it still writes the update-request body as:
```
shallow \n 0000 (flush)
```

This is valid behaviour from the reference `git` implementation — the client
declares its shallow boundary, then discovers there are no commands to send and
emits a flush. You can reproduce this with:

```bash
git clone --depth=1 <repo-url> shallow-copy
cd shallow-copy
git push origin main   # nothing new to push
```

### DIFF
--- a/plumbing/protocol/packp/updreq_decode.go
+++ b/plumbing/protocol/packp/updreq_decode.go
@@ -118,6 +118,12 @@ func (req *UpdateRequests) Decode(r io.Reader) error {
 		}
 	}
 
+	// A shallow-clone no-op push sends shallow lines followed directly by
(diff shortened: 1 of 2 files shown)


pjbgf:
@shanzhubb thanks for looking into this. I'm closing this PR as it has already been fixed and merged via 8168978c and 62ea9135.

PR #2142, opened 2026-05-20
plumbing: format/packfile, allow injecting a custom ObjectSelector

Soph:
## Summary

Adds two related extension points to `packfile.Encoder` so callers can take over delta selection while still using the encoder for the write phase:

- `DeltaSelector` and `NewDeltaSelector` are exported (formerly the unexported `deltaSelector` / `newDeltaSelector`). Behaviour and method set unchanged; this only widens visibility.
- `ObjectSelector` is a new interface (just `ObjectsToPack`), satisfied by `*DeltaSelector` by default.
- `WithObjectSelector` is an `EncoderOption` that overrides which `ObjectSelector` `Encode` calls for the selection step. The encoder's internal `*DeltaSelector` is still used for write-phase recovery (`restoreOriginal` on cyclic chains), so the storer passed to `NewEncoder` must remain valid.
- `PrecomputedSelector` wraps a precomputed `[]*ObjectToPack` as an `ObjectSelector` for the common "run selection ahead, then encode" pattern.
- `NewEncoder` now takes variadic `EncoderOption` values. The existing three-argument call sites are unaffected.

## Motivation

When an `Encoder` writes to a slow or strict transport — e.g. an HTTP request body to a CDN-fronted `git-receive-pack` endpoint — `Encode`'s synchronous selection-then-write shape produces a multi-second window during which no bytes flow. Some CDN edges (we observed this against Cloudflare's git frontend) interpret that as an idle upload and close the connection mid-stream.

With this change a caller can run `DeltaSelector.ObjectsToPack` ahead of time, then invoke `Encode` behind a `PrecomputedSelector` so the write phase starts immediately and streams continuously:

```go
sel := packfile.NewDeltaSelector(store)
objects, _ := sel.ObjectsToPack(hashes, packWindow)  // slow phase, before HTTP body opens

enc := packfile.NewEncoder(w, store, useRefDeltas,
    packfile.WithObjectSelector(packfile.PrecomputedSelector(objects)))
enc.Encode(hashes, packWindow)  // selection step returns instantly; pack bytes stream
```

The same hook is also useful for tests that want to mock or assert on the object set, and for callers exploring alternative selection strategies.

## Discussion

This approach was suggested by @pjbgf in Discord as a cleaner alternative to splitting `Encode` into two new public methods — keeping the `Encoder` API surface unchanged and making the selector the injection point. Filing as a PR per that conversation; happy to iterate.

## Test plan
- [x] `go test ./plumbing/format/packfile/...` — existing tests still pass
- [x] `go test ./...` — full suite passes
- [x] `go vet ./...` and `gofmt -l` clean
- [x] New `TestWithObjectSelectorMatchesDefault` asserts the precomputed path produces byte-identical output to the default path
- [x] New `TestWithObjectSelectorNilPreservesDefault` asserts `WithObjectSelector(nil)` is a no-op

## AI disclosure

This change was prepared with assistance from Claude Opus 4.7 — design discussion, code drafting, and tests. Every line was reviewed by me; the `Assisted-by:` trailer is on the commit per AI_POLICY.md.

### DIFF
--- a/plumbing/format/packfile/encoder.go
+++ b/plumbing/format/packfile/encoder.go
@@ -16,21 +16,64 @@ import (
 	"github.com/go-git/go-git/v6/utils/sync"
 )
 
+// ObjectSelector decides which objects go into a pack and in what
+// order, including any delta relationships. The default selector is
+// *DeltaSelector.
+type ObjectSelector interface {
+	ObjectsToPack(hashes []plumbing.Hash, packWindow uint) ([]*ObjectToPack, error)
+}
+
 // Encoder gets the data from the storage and write it into the writer in PACK
-// format
+// format.
+//
+// The encoder has two selector fields: deltaSelector is the
+// encoder's own *DeltaSelector, used internally for write-phase
+// recovery (e.g. restoreOriginal on cyclic chains). objectSelector is
+// what Encode calls to obtain the object list — by default the same
+// *DeltaSelector, but a caller can override it via WithObjectSelector.
 type Encoder struct {
-	selector *deltaSelector
-	w        *offsetWriter
-	zw       sync.ZlibWriter
-	hasher   hash.Hash
+	deltaSelector  *DeltaSelector
+	objectSelector ObjectSelector
+	w              *offsetWriter
+	zw             sync.ZlibWriter
+	hasher         hash.Hash
 
 	useRefDeltas bool
+}
+
+// EncoderOption configures an Encoder at construction time.
+type EncoderOption func(*Encoder)
+
+// WithObjectSelector overrides the ObjectSelector used by Encode to
+// produce the object list. The default is the encoder's own
+// *DeltaSelector, which runs delta selection synchronously when
+// Encode is called.
+//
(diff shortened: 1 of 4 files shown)


plumbing/format/packfile/encoder.go
```diff
@@ -16,21 +17,112 @@ import (
 	"github.com/go-git/go-git/v6/utils/sync"
 )
 
+// ObjectSelector decides which objects go into a pack and in what
+// order, including any delta relationships. The default selector is
+// *DeltaSelector; callers can supply a different one via
+// WithObjectSelector — for example a passthrough wrapping objects
+// that were produced ahead of time, so Encoder.Encode can stream
+// output without an internal selection delay.
```
pjbgf: If this interface is used in other locations, this part of the comment may be out of place. The current comment in the `Encoder` seems good enough to cover `WithObjectSelector`.

```suggestion
// *DeltaSelector.
```

pjbgf (approved):
@Soph Thanks for working on this. :bow:

PR #2147, opened 2026-05-22
*: speed up multi-pack reads via Index fanout probes

hiddeco:
Third in the pack-hotpath stack. Builds on #2133 (FD lifecycle) and #2134 (storage-wide FD budget) to land the read-path wins.

### Changes

- `idxfile.Index.MayContain` — O(1) fanout-bucket short-circuit; `findObjectInPackfile` consults it before `FindOffset` and skips packs whose first-byte bucket is empty.
- MRU pack ordering — sticky `lastHitPackIdx` hint in `ObjectStorage.findObjectInPackfile` resolves repeated reads on the same pack in O(1).
- `idxfile.Index.EntriesWithPrefix` — fanout-bounded prefix iteration with in-bucket bsearch positioning (mirrors upstream Git's `for_each_prefixed_object_in_pack`). Replaces the per-entry prefix scan in `ObjectStorage.HashesWithPrefix`.
- Storage-level reads route through `PackHandle.Index()` so `.idx`/`.rev` descriptors share the FD-pool budget.
- Race test for `cleanPackList`-vs-Acquire; multi-byte-prefix iteration tests covering the mid-bucket match case.

### Benchmarks

Apple M4 Pro, `-benchtime=1s -count=6`, benchstat vs `pack-hotpath-fd-pool`:

| Benchmark | sec/op | B/op | allocs/op |
|---|---:|---:|---:|
| `FindObjectInPackfile_FanoutMiss` | 13.66µ → 10.28µ (**−24.7%**) | 4.90Ki → 2.81Ki (**−42.7%**) | 89 → 59 (**−33.7%**) |
| `FindObjectInPackfile_MRU_Hit` | 12.04µ → 8.87µ (**−26.3%**) | 4.91Ki → 2.73Ki (**−44.5%**) | 89 → 56 (**−37.1%**) |
| `HashesWithPrefix_TwoByte` | 1093.5µ → 32.05µ (**−97.1%**) | 111.5Ki → 6.44Ki (**−94.2%**) | 4219 → 95 (**−97.8%**) |

Cumulative across the four-PR stack (vs `main`): `FanoutMiss` −81%, `MRU_Hit` −83%, `HashesWithPrefix_TwoByte` −97%, `AlternatesObjectLookup` (3 variants) −46–67%, `ParallelReads` −78%.

### DIFF
--- a/storage/filesystem/object.go
+++ b/storage/filesystem/object.go
@@ -191,15 +191,6 @@ func findInAlternates[T any](s *ObjectStorage, fn func(*ObjectStorage) (T, error
 	return foundVal, nil
 }
 
-// indexLoaded reports whether s.index has been populated. Safe for
-// concurrent use: reads s.index under the read half of s.muI.
-func (s *ObjectStorage) indexLoaded() bool {
-	s.muI.RLock()
-	loaded := s.index != nil
-	s.muI.RUnlock()
-	return loaded
-}
-
 // requireIndex ensures s.index is populated, performing a cold-load
 // on first access. Concurrent first-readers are coalesced via
 // singleflight so populateIndex runs once per thundering herd; the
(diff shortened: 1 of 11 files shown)


storage/filesystem/object.go
```diff
@@ -535,56 +569,45 @@ func (s *ObjectStorage) packfile(idx idxfile.Index, pack plumbing.Hash) (*packfi
...
-
-	size, err = s.encodedObjectSizeFromPackfile(h)
-	if err == nil {
-		return size, nil
 	}
 
-	if !errors.Is(err, plumbing.ErrObjectNotFound) {
-		return 0, err
+	// Pack-membership-first: a single O(1) probe (Phase 2's
+	// MayContain + Phase 5's MRU hint) decides where to route
```
pjbgf: Comments around "Phases" feel largely redundant. It may be worth dropping them or have them in a higher level documentation instead.

storage/filesystem/object.go
```diff
@@ -535,56 +569,45 @@ func (s *ObjectStorage) packfile(idx idxfile.Index, pack plumbing.Hash) (*packfi
...
+	// without any syscall. If h lives in one of our packs we skip
```
pjbgf: The claim of syscall-free check may be out of date, this is likely incorrect given the use of `LazyIndex.FindHash`.

storage/filesystem/object.go
```diff
@@ -535,56 +569,45 @@ func (s *ObjectStorage) packfile(idx idxfile.Index, pack plumbing.Hash) (*packfi
...
+	// the loose Stat outright; if it doesn't, we go straight to
+	// loose. Strictly cheaper than either packs-first (paid a
+	// per-pack scan + loose fallback) or loose-first (paid a
+	// wasted Stat on every packed read, including cache hits).
+	if pack, idx, offset := s.findObjectInPackfile(h); pack != plumbing.ZeroHash {
+		if hash, hashErr := idx.FindHash(offset); hashErr == nil {
```
pjbgf: EncodedObjectSize receives a hash `h`. We are using that to lookup the offset, to then get the object hash again. We can probably drop the initial two calls:

```suggestion
```

storage/filesystem/object.go
```diff
@@ -476,27 +511,26 @@ func (s *ObjectStorage) LazyWriter() (w io.WriteCloser, wh func(typ plumbing.Obj
 // HasEncodedObject returns nil if the object exists, without actually
 // reading the object data from storage.
 func (s *ObjectStorage) HasEncodedObject(h plumbing.Hash) (err error) {
```
pjbgf, line 513: This may be adding a regression when an object exists in loose object store and the only idx file in the object store is corrupted. The previous version would return the object. Now we are returning an error due to the `requireIndex` below.

I haven't tested this with `git`, but my assumption is that the scenario above would not lead to a hard failure upstream.

pjbgf (changes requested):

pjbgf (approved):

PR #2153, opened 2026-05-23
demo: mmap'ed reads

hanwen:
before:

time /tmp/merge-base  ~/vc/linux/.git refs/tags/v7.0 refs/tags/v6.9 a38297e3fb012ddfa7ce0321a7e5a8daeb1872b6

real	0m10.530s
user	0m8.033s
sys	0m3.195s

after:

$ time /tmp/merge-base  ~/vc/linux/.git refs/tags/v7.0 refs/tags/v6.9 a38297e3fb012ddfa7ce0321a7e5a8daeb1872b6

real	0m6.716s
user	0m7.314s
sys	0m0.161s

this change, but osfs.New() with `KeepDescriptors: true`

$ time /tmp/merge-base  ~/vc/linux/.git refs/tags/v7.0 refs/tags/v6.9
a38297e3fb012ddfa7ce0321a7e5a8daeb1872b6

real	0m11.368s
user	0m8.909s
sys	0m3.257s

CGit:

time git  --git-dir ~/vc/linux/.git/ merge-base refs/tags/v6.9 refs/tags/v7.0 a38297e3fb012ddfa7ce0321a7e5a8daeb1872b6

real	0m2.948s
user	0m2.701s
sys	0m0.230s

After packhandle cache (#2132)

time /tmp/merge-base  ~/vc/linux/.git refs/tags/v7.0 refs/tags/v6.9
a38297e3fb012ddfa7ce0321a7e5a8daeb1872b6

real	0m10.596s
user	0m9.272s
sys	0m2.489s

PR #2154, opened 2026-05-25
storage: filesystem, pool FSObject reads via PackHandle

hiddeco:
Materialising an `FSObject` returned by `EncodedObject` previously fell back to a direct `o.fs.Open(packPath)` on every `Reader()` call, bypassing the per-pack file-descriptor pool introduced in `billy.Filesystem` that pays an mmap/munmap on each Open, the regression amplified to ~400x more pack opens over 500 reads on the basic fixture[1].

The cure threads the dotgit-cached `*packhandle.PackHandle` into the `*packfile.Packfile` via a new `WithPackHandle` option taking a `PackHandleResolver`. `objectFromHeader` captures the resolver in the `FSObject.acquireRandom` closure, so every read acquires a fresh cursor against the pool-owned `SharedFile`. The resolver re-runs on every `Reader` call, which lets cached FSObjects survive `cleanPackList` invalidations such as the one `NewObjectPack` triggers before a repack.

The internal `*packhandle.PackHandle` stays internal — the public `packfile.PackHandle` interface returns `io.ReadSeekCloser` and `RandomReader`, and `DotGit.PackHandle` returns the interface through an unexported adapter so callers can't accidentally call `Close()` on a catalog-owned handle. `Packfile.Close` joins the scanner-close and owned-handle close errors via `errors.Join`.

[1]: https://github.com/go-git/go-git/pull/2153#issuecomment-4526461972


plumbing/format/packfile/internal_test.go
```diff
@@ -246,3 +249,125 @@ func writeTestPackFile(t *testing.T, pack []byte) *os.File {
...
+	stub := stubPackHandle{}
+
+	resolver := PackHandleResolver(func() (PackHandle, error) {
+		resolverCalls.Add(1)
+		return stub, nil
+	})
+
+	p := NewPackfile(nil, WithPackHandle(resolver))
+	// Close without init — exercises the WithPackHandle path
+	// directly. p.h is still nil (init hasn't run); p.closeOwnedH
```
pjbgf:
```suggestion
	// directly. p.handle is still nil (init hasn't run); p.closeHandle
```

storage/filesystem/object_bench_test.go
```diff
@@ -266,3 +266,181 @@ func benchWarmPack(b *testing.B, fs billy.Filesystem) {
 		return // warm the first pack only
 	}
 }
+
+// BenchmarkObjectStorage_FSObjectReader_OpenCount reports the
```
pjbgf: We probably better off converting this into unit tests, so that we can block future regressions at PR checks.

pjbgf (approved):

PR #2158, opened 2026-05-27
plumbing: format/index, sort entries by Name and Stage

### LINKED ISSUE #2157: Index Entry Sorting Order
### Bug Description

Hi,

I notice in git's documentation: index entries should be sorted by Name, then Stage (if the names are the same): https://git-scm.com/docs/index-format#_index_entry

The current go-git implementation is to sort [byName](https://github.com/go-git/go-git/blob/main/plumbing/format/index/encoder.go#L88), [byName.Less](https://github.com/go-git/go-git/blob/main/plumbing/format/index/encoder.go#L266-L270).

Should the Less() be updated to take the Stage field into account?

### go-git Version

v5.19.1

### Steps to Reproduce

_No response_

### Additional Information

_No response_

wacky6:
Per Git index format specification, index entries should be sorted by name, then stage: https://git-scm.com/docs/index-format#_index_entry

Fixes: https://github.com/go-git/go-git/issues/2157

### DIFF
--- a/plumbing/format/index/encoder.go
+++ b/plumbing/format/index/encoder.go
@@ -85,7 +85,7 @@ func (e *Encoder) encodeHeader(idx *Index) error {
 }
 
 func (e *Encoder) encodeEntries(idx *Index) error {
-	sort.Sort(byName(idx.Entries))
+	sort.Sort(byNameAndStage(idx.Entries))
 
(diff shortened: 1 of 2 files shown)


pjbgf (approved):
@wacky6 thanks for working on this. :bow:

PR #2162, opened 2026-05-28
*: harden format parsers against unbounded loops

hiddeco:


plumbing/format/objfile/writer_fuzz_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+		r, err := NewReader(&buf, format.SHA1)
+		if err != nil {
+			t.Fatalf("NewReader after successful Write: %v", err)
+		}
+		t.Cleanup(func() { _ = r.Close() })
+
+		gotType, gotSize, err := r.Header()
+		if err != nil {
+			t.Fatalf("Header after successful Write: %v", err)
```
pjbgf: This one likely we want to ignore.

```suggestion
			return
```

plumbing/format/objfile/writer_fuzz_test.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		if gotType != ot {
+			t.Fatalf("type round-trip: want %v, got %v", ot, gotType)
+		}
+		if gotSize != size {
+			t.Fatalf("size round-trip: want %d, got %d", size, gotSize)
```
pjbgf: Same for these two:
```suggestion
			return
		}
		if gotSize != size {
			return
```

plumbing/format/commitgraph/encoder.go
```diff
@@ -112,6 +112,9 @@ func (e *Encoder) prepare(idx Index, hashes []plumbing.Hash) (hashToIndex map[pl
 }
 
 func (e *Encoder) encodeFileHeader(chunkCount int) (err error) {
+	if chunkCount < 0 || chunkCount > 255 {
```
pjbgf: Potentially `< 0` could be handled separately, or the error be made a bit more generic:
```suggestion
	if chunkCount > 255 {
```

plumbing/format/commitgraph/encoder_test.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+func TestEncodeFileHeaderRejectsTooManyChunks(t *testing.T) {
+	t.Parallel()
+	// The on-disk format stores num_chunks as a uint8, so any encoder
+	// path that tries to emit more than 255 chunk types would silently
+	// truncate. The encoder must reject the configuration at write time.
+	e := commitgraph.NewEncoder(&bytes.Buffer{})
+
+	err := commitgraph.EncodeFileHeaderForTest(e, 256)
```
pjbgf: Let's change the package name to `commitgraph` and test the private func directly.

```suggestion
	err := encodeFileHeader(e, 256)
```

Then we can remove `plumbing/format/commitgraph/export_test.go`.

pjbgf (approved):

PR #2164, opened 2026-05-29
plumbing: assorted memory optimisations

kokes:
I noticed there were needless memory allocations in a few places. I only touched paths affected by `git grep` as I was working with that.

<details>

<summary>benchmark function</summary>

```go
package git

import (
	"regexp"
	"testing"
)

func BenchmarkWorktreeGrepRepository(b *testing.B) {
	b.StopTimer()

	r, err := PlainOpen(".")
	if err != nil {
		b.Fatal(err)
	}
	defer func() { _ = r.Close() }()

	w, err := r.Worktree()
	if err != nil {
		b.Fatal(err)
	}

	opts := &GrepOptions{
		Patterns: []*regexp.Regexp{regexp.MustCompile("Gr.*p"), regexp.MustCompile("Grep")},
	}

	b.StartTimer()

	for b.Loop() {
		results, err := w.Grep(opts)
		if err != nil {
			b.Fatal(err)
		}
		if len(results) == 0 {
			b.Fatal("expected grep matches in repository")
		}
	}
}

```

</details>

Benchmark results:

```
$ benchstat bench_before.txt bench_after.txt
goos: darwin
goarch: arm64
cpu: Apple M4 Pro
                       │ bench_before.txt │          bench_after.txt           │
                       │      sec/op      │   sec/op     vs base               │
WorktreeGrepRepository        92.38m ± 4%   88.73m ± 1%  -3.95% (p=0.000 n=10)

                       │ bench_before.txt │           bench_after.txt            │
                       │       B/op       │     B/op      vs base                │
WorktreeGrepRepository      20.310Mi ± 0%   6.374Mi ± 1%  -68.62% (p=0.000 n=10)

                       │ bench_before.txt │          bench_after.txt           │
                       │    allocs/op     │  allocs/op   vs base               │
WorktreeGrepRepository        71.29k ± 0%   67.00k ± 0%  -6.01% (p=0.000 n=10)
```

So runtime largely unaffected, but allocation volume down in a pretty big way.

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -1,6 +1,7 @@
 package git
 
 import (
+	"bufio"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


worktree.go
```diff
@@ -1383,23 +1384,25 @@ func findMatchInFiles(fileiter *object.FileIter, treeName string, opts *GrepOpti
...
-func findMatchInFile(file *object.File, treeName string, opts *GrepOptions) ([]GrepResult, error) {
-	var grepResults []GrepResult
-
-	content, err := file.Contents()
+func findMatchInFile(file *object.File, treeName string, opts *GrepOptions) (grepResults []GrepResult, err error) {
+	reader, err := file.Reader()
 	if err != nil {
 		return grepResults, err
 	}
+	defer ioutil.CheckClose(reader, &err)
```
AriehSchneier, line 1392: I think this is the wrong `err`, you have shadowed the return value

kokes: I tried reproducing this minimally and I see a defer call to an error pointer to be assigning to the right error variable. But maybe I'm missing something, I have been bitten by shadowing a bunch.

returning err: https://go.dev/play/p/ANI66-Tda50
explicitly returning nil to depend on the defer assignment: https://go.dev/play/p/7DeOpO_uqxA

pjbgf (dismissed):
@kokes thanks for working on this. 🙇

worktree.go
AriehSchneier: (TLDR the code is correct as is)
Seems my understanding of `:=` was wrong, it will reuse an existing variable if defined in the same scope (and named returned values are scoped to the body of the function), where `:=` can create a shadow is if you are in a different scope, eg: https://go.dev/play/p/M-Wq7fJt9gG

pjbgf:
@kokes once the lint errors are fixed we can go ahead with merging these changes.

kokes:
- fixed linting issue
- squashed and reworded commit to align with this repo's format
- signed off the commit

PR #2166, opened 2026-05-31
DRAFT: feat: add ssh-config 

janderssonse:
A PR suggestion to solve the issue discussed in https://github.com/go-git/go-git/discussions/1207:

The problem a some missing go-git ssh-config-wireup to support the common enterprise case of needing to  Proxy SSH through an HTTP CONNECT/SOCKS proxy. A common setup when wanting to use ssh to github over their https url, when ssh is totally blocked, and enterprise also has an mandatory proxy for the https-calls.

So, I think we don't need to support GIT_SSH_COMMAND as I used it the discussion examples for 1207, as basically is just a caller to the underlying ssh impl anyway. This PR is much simpler, and more in line with existing go-git idioms.

A small ssh-config and as the ProxyCommand (in the 1207-discussion) is just "tunnel SSH through an HTTP CONNECT proxy to reach
  ssh.github.com:443." go-git already supports that,
  existing client.WithProxyURL (HTTP CONNECT + SOCKS via golang.org/x/net/proxy:

 Also one extra commit in this, with an explicit HostKeyCallback (e.g. InsecureIgnoreHostKey)
  and no known_hosts file present,  Transport.connect now falls back to default negotiation, instead of erroring. If this papercut is wrong/out of scope i'll just remove it.

### DIFF
--- a/internal/transport/test/sshserver.go
+++ b/internal/transport/test/sshserver.go
@@ -1,0 +1,55 @@
+package test
+
+import (
+	"io"
+	"net"
+	"os/exec"
+	"sync"
+	"testing"
+
+	glssh "github.com/gliderlabs/ssh"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/crypto/ssh/testdata"
+)
+
+// StartGitSSHServer starts an SSH server that serves git by running the
+// requested git command over the session, accepting any authentication. It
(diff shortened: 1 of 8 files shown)


janderssonse:
I really think I can do this already with supported framework. closing this draft for now.

PR #2167, opened 2026-05-31
config: read repositoryformatversion in unmarshalCore

### LINKED ISSUE #2155: `config.Unmarshal` never reads `core.repositoryformatversion`
### Bug Description

`config.Config.unmarshalCore` reads `bare`, `worktree`, and `commentChar` from the parsed config but does **not** read `repositoryformatversion`. As a result, `cfg.Core.RepositoryFormatVersion` is always `""` after loading a repo, regardless of what is on disk.

This is a silent data-loss bug on the *read* path. The *write* path uses the field correctly, so a roundtrip (load → save) erases the version. It also defeats `verifyExtensions`, which then forces every repo down the V0/`""` branch even when the file clearly says `repositoryformatversion = 1`.

### go-git Version

v5.18.0

### Steps to Reproduce

**Affected versions:** at least `v5.17.0` through `v5.18.0` (likely older — the read path has always been missing; it only became user-visible when `verifyExtensions` was added in v5.17).

## Repro

```go
package main

import (
	"fmt"
	"os"
	"os/exec"

	"github.com/go-git/go-git/v5"
)

func main() {
	dir, _ := os.MkdirTemp("", "repro")
	defer os.RemoveAll(dir)

	run := func(args ...string) {
		c := exec.Command("git", args...)
		c.Dir = dir
		_ = c.Run()
	}
	run("init", "-q")
	run("config", "core.repositoryformatversion", "1")
	// noop-v1 is in `builtinExtensions` but NOT in `extensionsValidForV0`,
	// so it is *only* accepted on the V1 branch of verifyExtensions.
	run("config", "extensions.noop-v1", "true")

	_, err := git.PlainOpen(dir)
	fmt.Println("PlainOpen:", err)
}
```

Output:

```
PlainOpen: core.repositoryformatversion does not support extension: noop-v1
```

`noop-v1` exists specifically to test V1 compatibility (per the comment next to its declaration), so this repro is a direct contradiction: a `repositoryformatversion = 1` repo using the V1-test extension fails because go-git thinks the version is V0.

`git config core.repositoryformatversion` against the same repo returns `1`.

## Root cause

`config/config.go` defines the key and uses it on the write path:

```go
// config/config.go (~line 266)
repositoryFormatVersionKey = "repositoryformatversion"

// config/config.go (~line 432, marshal)
if string(c.Core.RepositoryFormatVersion) != "" {
    s.SetOption(repositoryFormatVersionKey, string(c.Core.RepositoryFormatVersion))
}
```

But `unmarshalCore` never populates the field.

## Suggested fix

```go
func (c *Config) unmarshalCore() {
    s := c.Raw.Section(coreSection)
    if s.Options.Get(bareKey) == "true" {
        c.Core.IsBare = true
    }

    c.Core.Worktree = s.Options.Get(worktreeKey)
    c.Core.CommentChar = s.Options.Get(commentCharKey)
    c.Core.RepositoryFormatVersion = format.RepositoryFormatVersion(
        s.Options.Get(repositoryFormatVersionKey),
    )
}
```

`objectformat` (the other field documented for the `[core]` section) appears to have the same problem and probably wants the same treatment in this fix.

c-tonneslan:
\`marshalCore\` writes \`core.repositoryformatversion\` when it isn't empty, but \`unmarshalCore\` never set \`Core.RepositoryFormatVersion\` from the parsed section. So on a repo with explicit \`repositoryformatversion=1\` the field loaded as the zero value and \`verifyExtensions\` treated the repo as version 0, which then rejected V1 extensions like \`worktreeConfig\` that the on-disk config clearly enabled.

Mirror of the V6 fix on \`main\` (the read is already there in commit 6923824). Test fails on \`releases/v5.x\` without the change.

fixes #2155

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -329,6 +329,10 @@ func (c *Config) unmarshalCore() {
 	if parsed := parseConfigBool(s.Options.Get(protectHFSKey)); parsed.IsSet() {
 		c.Core.ProtectHFS = parsed
 	}
+
+	if s.Options.Get(repositoryFormatVersionKey) == string(format.Version_1) {
+		c.Core.RepositoryFormatVersion = format.Version_1
+	}
 }
 
 func (c *Config) unmarshalUser() {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/go-git/go-billy/v5/osfs"
 	"github.com/go-git/go-billy/v5/util"
 	"github.com/go-git/go-git/v5/plumbing"
+	format "github.com/go-git/go-git/v5/plumbing/format/config"
 	. "gopkg.in/check.v1"
 )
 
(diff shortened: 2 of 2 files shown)


pjbgf (approved):
@c-tonneslan thanks for working on this. 🙇

PR #2171, opened 2026-06-02
git: repository_extensions, Fix case-mismatched extension whitelist and accept standard extensions

### LINKED ISSUE #2156: `verifyExtensions` rejects standard extensions due to case-mismatched whitelist
### Bug Description

`verifyExtensions` lowercases extension names on read but compares them against whitelists whose keys preserve upstream-canonical camelCase spellings. As a result, the standard git extensions `worktreeConfig`, `partialClone`, and `preciousObjects` are rejected on V0 repos even though they appear in `extensionsValidForV0` with the comment "These are the only extensions that can be enabled while core.repositoryformatversion is unset or set to 0."

`worktreeConfig` is enabled automatically by `git worktree` tooling (notably anything driving per-worktree config), so this breaks `PlainOpen` for a fairly common repo layout.

### go-git Version

v5.18.0

### Steps to Reproduce

```go
package main

import (
	"fmt"
	"os"
	"os/exec"

	"github.com/go-git/go-git/v5"
)

func main() {
	dir, _ := os.MkdirTemp("", "repro")
	defer os.RemoveAll(dir)

	run := func(args ...string) {
		c := exec.Command("git", args...)
		c.Dir = dir
		_ = c.Run()
	}
	run("init", "-q")
	run("config", "core.repositoryformatversion", "0") // explicit V0
	run("config", "extensions.worktreeConfig", "true")

	_, err := git.PlainOpen(dir)
	fmt.Println("PlainOpen:", err)
}
```

Output:

```
PlainOpen: core.repositoryformatversion does not support extension: worktreeconfig
```

`worktreeConfig` is documented as valid for V0, so the error is self-contradictory.

### Additional Information

`repository_extensions.go` lowercases extension names on read:

```go
// ~line 72
for _, opt := range section.Options {
    out = append(out, extension{
        name:  strings.ToLower(opt.Key),
        value: strings.ToLower(opt.Value),
    })
}
```

…but both whitelist maps preserve the upstream-canonical casing:

```go
// ~line 32
builtinExtensions = map[string]struct{}{
    "noop":    {},
    "noop-v1": {},
}

// ~line 48
extensionsValidForV0 = map[string]struct{}{
    "noop":            {},
    "partialClone":    {},  // never matches: lookup uses "partialclone"
    "preciousObjects": {},  // never matches: lookup uses "preciousobjects"
    "worktreeConfig":  {},  // never matches: lookup uses "worktreeconfig"
}
```

So in the V0 code path, only `noop` actually works — and `noop` only because it happens to be lowercase already. The other three entries in `extensionsValidForV0` are unreachable.

There is also a related gap on the V1 path: even if the case mismatch is fixed for V0, a V1 repo with `extensions.worktreeConfig` still falls through to the `builtinExtensions` check (which only contains `noop`/`noop-v1`) and is rejected with `ErrUnknownExtension`. These three extensions are standard git features and should be accepted on V1 too, since go-git doesn't have to change behavior for any of them — it just needs to not reject them. (`partialClone` may be worth a separate discussion if the storer is expected to behave differently.)

toller892:
## Description

Fixes #2156

`verifyExtensions` rejects standard Git extensions (`worktreeConfig`, `partialClone`, `preciousObjects`) on V0 repos due to a case mismatch: the `extensionsValidForV0` map used camelCase keys, but `extensions()` normalises all names to lowercase with `strings.ToLower`. The lookup always missed, making these extensions unreachable — only `noop` (already lowercase) worked.

This also affected V1 repos: even with the case fix, these extensions would fail the `builtinExtensions` check since they weren't listed there.

### Changes

- **`repository_extensions.go`**: Lowercase all keys in `extensionsValidForV0` to match the output of `extensions()`. Add `worktreeconfig`, `partialclone`, `preciousobjects` to `builtinExtensions` so they're accepted at any format version.
- **`repository_extensions_test.go`**: Add regression tests that verify camelCase extension input (`worktreeConfig`, `partialClone`, `preciousObjects`) is accepted at V0, and that `worktreeConfig` is accepted at V1.

### Why these extensions are safe to accept

go-git does not need to change behaviour for any of these extensions — it just needs to not reject them:

- **`worktreeConfig`**: Enables per-worktree configuration. Automatically set by `git worktree` tooling. Breaking `PlainOpen` for any repo that has used `git worktree`.
- **`partialClone`**: Enables partial clone (blob filter). go-git can accept the extension without implementing the partial clone protocol.
- **`preciousObjects`**: Marks the repo as a promisor remote object store. No behaviour change needed in go-git.

### Cross-reference

The `main` branch (v6) already has the lowercase keys in `extensionsValidForV0`. This PR backports the same fix plus the `builtinExtensions` additions to `releases/v5.x`.

Assisted-by: Hermes Agent <noreply@hermes.agent>

### DIFF
--- a/repository_extensions.go
+++ b/repository_extensions.go
@@ -30,26 +30,43 @@ var (
 	// Some extensions are storage-specific, those are defined by the Storers
 	// themselves by implementing the ExtensionChecker interface.
 	builtinExtensions = map[string]struct{}{
-		// noop does not change git’s behavior at all.
+		// noop does not change git's behavior at all.
 		// It is useful only for testing format-1 compatibility.
 		//
 		// This extension is respected regardless of the
 		// core.repositoryFormatVersion setting.
 		"noop": {},
 
-		// noop-v1 does not change git’s behavior at all.
+		// noop-v1 does not change git's behavior at all.
 		// It is useful only for testing format-1 compatibility.
 		"noop-v1": {},
+
+		// worktreeConfig enables per-worktree configuration,
+		// automatically set by git worktree. go-git does not
+		// change behavior for this extension but must not
+		// reject it.
+		"worktreeconfig": {},
+
+		// partialClone enables partial clone (filter by blob
+		// size, etc.). go-git accepts but does not implement
+		// the partial clone protocol.
+		"partialclone": {},
+
+		// preciousObjects marks the repo as a promisor remote
+		// object store that must not be deleted by gc.
+		"preciousobjects": {},
 	}
 
 	// Some Git extensions were supported upstream before the introduction
 	// of repositoryformatversion. These are the only extensions that can be
 	// enabled while core.repositoryformatversion is unset or set to 0.
+	// Keys must be lowercase to match the output of extensions(), which
+	// normalises extension names with strings.ToLower.
 	extensionsValidForV0 = map[string]struct{}{
 		"noop":            {},
-		"partialClone":    {},
-		"preciousObjects": {},
-		"worktreeConfig":  {},
+		"partialclone":    {},
+		"preciousobjects": {},
+		"worktreeconfig":  {},
 	}
 )
 
--- a/repository_extensions_test.go
+++ b/repository_extensions_test.go
@@ -58,6 +58,42 @@ func TestVerifyExtensions(t *testing.T) {
 			},
 		},
 		{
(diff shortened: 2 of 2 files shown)


clundquist-stripe:
👋 this would be helpful to merge as v6.x doesn't have a release version yet

PR #2183, opened 2026-06-09
plumbing: commitgraph, lightweight traversal Commit and upstream format conformance

pjbgf:
## Summary

Two related improvements to the commit-graph packages:

  1. **A lightweight `Commit` for graph traversal** (`plumbing/object/commitgraph`),
     replacing the full `*object.Commit` that `CommitNode.Commit()` previously
     returned, plus allocation tuning of its decoder.
  2. **Four commit-graph file-format conformance fixes** (`plumbing/format/commitgraph`),
     aligning the reader/encoder with upstream C Git, including SHA-256 support.

  This does **not** yet wire the commit-graph into `Log`/merge-base/negotiation, it just hardens and slims the existing machinery.

  ## Changes

  ### Lightweight traversal commit
  - New `commitgraph.Commit` carrying only base traversal fields, with a trimmed
    decoder that stops after the committer header (so it never parses gpgsig,
    mergetag or the message body).
  - `CommitNode.Commit()` now returns `*commitgraph.Commit`; `objectCommitNode`
    holds the storer to resolve trees and decodes on demand.
  - Decoder allocation tuning: read header lines in place via `ReadSlice` (with an
    `ErrBufferFull` fallback), decode OIDs straight into a stack array via
    `hex.Decode` + `FromBytes`, and compare header keys as `[]byte` without string
    allocation.

  ### Format conformance (vs upstream commit-graph.c v2.54.0)
  - Reject **non-monotonic OID fanout** (`graph_read_oid_fanout`), which also keeps
    `GetIndexByHash`'s binary-search bound inside the OID lookup chunk.
  - Parse a **chain file whose final line lacks a trailing newline** instead of
    silently dropping the newest graph.
  - Validate the header's **base commit-graph count** against the actual chain depth.
  - **SHA-256 commit-graph support**: derive OID width from the header hash-version
    byte and size every OID read accordingly; the encoder infers SHA-256 from the
    index's hashes and emits a consistent header/widths/chunk sizes.

  ## Performance

  Object-backed commit walking (decode path), `benchstat`, n=6:

  - Per-commit `Get`: **51 → 30 allocs/op (-41%)** cumulatively
    (lightweight struct -30% allocs, decoder tuning a further ~13 allocs).
  - Object-backed walks: ~-39% allocs/op, ~-16% B/op.
  - Pure commit-graph-file walks (CTime/Date/Topo) are unchanged — they never
    decode an object.

### DIFF
--- a/_examples/ls/main.go
+++ b/_examples/ls/main.go
@@ -74,7 +74,11 @@ func main() {
 
 	revs, err := getLastCommitForPaths(commitNode, treePath, paths)
 	CheckIfError(err)
-	for path, rev := range revs {
+	for path, node := range revs {
+		// The commit node only carries traversal fields; load the full
+		// commit object to print its message.
+		rev, err := r.CommitObject(node.ID())
+		CheckIfError(err)
 		// Print one line per file (name hash message)
 		hash := rev.Hash.String()
 		line := strings.Split(rev.Message, "\n")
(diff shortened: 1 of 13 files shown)


onee-only:
> This does not yet wire the commit-graph into Log/merge-base/negotiation, it just hardens and slims the existing machinery.

Since this PR is open, I'd like to know if there are any plans for this. It'll be awesome to see this feature getting implemented.

PR #2187, opened 2026-06-10
config: url, Add pushInsteadOf support and bring push API closer to Git. Fixes #865.

### LINKED ISSUE #865: support pushInsteadOf in config
https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtpushInsteadOf

jyn514:
Parse `url.*.pushInsteadOf` rules and use them when selecting the default push URL. Preserve explicit remote `pushurl` precedence and apply `pushInsteadOf` before normal `insteadOf` rewrites, matching Git behavior.

Assisted-by: Codex 5.5 <noreply@openai.com>
Assisted-by: Claude Opus 4.7 <noreply@anthropic.com>
Signed-off-by: jyn <github@jyn.dev>

This commit is LLM-generated, except that I added the comments. I self-reviewed the tests and they looked ok. I haven't looked too hard at the code.

The original motivation for this PR was a bug in `git-bug` where it would fail when I had `pushInsteadOf` set but not `insteadOf`. It would get an empty string in insteadOf, try to apply the first rewrite I found in the config, and fail because I didn't have any repository forked at that remote. I'm fixing that in a second, but it needs this upstream fix first.

---

New API:
```go
// config/url.go

type URL struct {  // existing
  PushInsteadOfs []string  // new
}

func (u *URL) ApplyPushInsteadOf(url string); // new

// config/config.go

type RemoteConfig struct {  // existing
  PushURLs []string  // new
}

func (c *RemoteConfig) ResolvedPushURL() string;  // new
func (c *RemoteConfig) ResolvedPushURLs() []string; // new
```

Behavior changes:
- `Remote.Push()` now pushes to *all* remotes, not just the first remote, matching git.
- `Remote.String()` now reports `PushURL()` instead of the last URL. Previously these were *usually* the same, but they could diverge if there were multiple `url` fields for the same remote: before, go-git would return the last `url`, now it returns the first. If there is an explicit `pushurl`, the behavior is the same as before.
- `URL.ApplyInsteadOf` now uses the longest prefix instead of the first matching prefix. Longest prefix is Go's behavior, documented here: https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtinsteadOf

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1025,6 +1025,8 @@ type RemoteConfig struct {
 	// URLs the URLs of a remote repository. It must be non-empty. Fetch will
 	// always use the first URL, while push will use all of them.
 	URLs []string
+	// PushURLs are the explicit push URLs of a remote repository.
+	PushURLs []string
 	// Mirror indicates that the repository is a mirror of remote.
 	Mirror bool
 
@@ -1032,6 +1034,10 @@ type RemoteConfig struct {
 	insteadOfRulesApplied bool
 	// originalURLs are the urls before applying insteadOf rules
 	originalURLs []string
+	// originalPushURLs are the explicit push URLs before applying URL rules.
+	originalPushURLs []string
+	// pushInsteadOfURLs are the push URLs derived from pushInsteadOf rules.
+	pushInsteadOfURLs []string
 
 	// Fetch the default set of "refspec" for fetch operation
 	Fetch []RefSpec
@@ -1079,7 +1085,8 @@ func (c *RemoteConfig) unmarshal(s *format.Subsection) error {
 
 	c.Name = c.raw.Name
 	c.URLs = append([]string(nil), c.raw.Options.GetAll(urlKey)...)
-	c.URLs = append(c.URLs, c.raw.Options.GetAll(pushurlKey)...)
+	c.PushURLs = append([]string(nil), c.raw.Options.GetAll(pushurlKey)...)
+	c.URLs = append(c.URLs, c.PushURLs...)
 	c.Fetch = fetch
 	c.Mirror = c.raw.Options.Get(mirrorKey) == "true"
 
@@ -1092,15 +1099,17 @@ func (c *RemoteConfig) marshal() *format.Subsection {
 	}
 
 	c.raw.Name = c.Name
-	if len(c.URLs) == 0 {
+	urls, pushURLs := c.urlsAndPushURLsForMarshal()
+	if len(urls) == 0 {
 		c.raw.RemoveOption(urlKey)
 	} else {
-		urls := c.URLs
-		if c.insteadOfRulesApplied {
-			urls = c.originalURLs
-		}
-
 		c.raw.SetOption(urlKey, urls...)
+	}
+
+	if len(pushURLs) == 0 {
+		c.raw.RemoveOption(pushurlKey)
+	} else {
+		c.raw.SetOption(pushurlKey, pushURLs...)
 	}
 
 	if len(c.Fetch) == 0 {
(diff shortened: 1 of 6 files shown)


jyn514:
(codex found some more bugs, i'm fixing those too)

jyn514:
some of these bug fixes bring go-git closer to git's behavior, but are breaking changes for existing users. not sure what the policy is on those. i'll write them up in a second.

jyn514:
ok, this should be ready for review.

config/config.go
```diff
@@ -1126,19 +1141,117 @@ func (c *RemoteConfig) IsFirstURLLocal() bool {
...
+
+	rewriteInsteadOf := func(url string) (string, bool) {
+		match, prefix := findLongestURLMatch(url, urlRules, func(u *URL) []string {
+			return u.InsteadOfs
+		})
+		if match == nil {
+			return url, false
+		}
+		return match.Name + url[len(prefix):], true
+	}
```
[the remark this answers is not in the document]
jyn514: applyURLRules is private so i'm not sure this is a big deal.

config/url.go
```diff
@@ -50,40 +55,57 @@ func (u *URL) marshal() *format.Subsection {
...
 				longestMatch = u
-				longestMatchLength = lengthCurrentInsteadOf
+				longestPrefix = currentPrefix
+				longestMatchLength = lengthCurrentPrefix
 			}
 		}
 	}
 
-	return longestMatch
+	return longestMatch, longestPrefix
```
AriehSchneier: Nothing ever uses longestPrefix other than to len(longestPrefix), why not just return longestMatchLength here?
Even better, it looks like every time findLongestURLMatch is called it is just used to replace the start of remoteURL if it found a match, why not just do the replace here and return it directly?
PS I just happened to [find that bug](https://github.com/go-git/go-git/pull/2189) at the same time!

config/url.go
```diff
@@ -50,40 +55,57 @@ func (u *URL) marshal() *format.Subsection {
...
-			return u.Name + url[len(j):]
+	return u.apply(url, u.InsteadOfs)
+}
+
+// ApplyPushInsteadOf applies the push URL rewrite rules to the given URL.
+func (u *URL) ApplyPushInsteadOf(url string) string {
+	return u.apply(url, u.PushInsteadOfs)
+}
+
+func (u *URL) apply(url string, prefixes []string) string {
```
AriehSchneier, line 102: This could either call findLongestURLMatch or be replaced by it entirely (by making the other do the replacement)

jyn514: seems good, done :) replaced `findLongest` with `rewriteLongest`.

config/url.go
jyn514: nice catch 😄 

i think this comment is no longer relevant now that this is rewriteLongest instead of findLongest.

config/url.go
```diff
@@ -50,40 +55,53 @@ func (u *URL) marshal() *format.Subsection {
...
 			}
 		}
 	}
 
-	return longestMatch
+	if longestMatch == nil {
+		return remoteURL, false
+	}
+
+	return longestMatch.Name + remoteURL[len(longestPrefix):], true
```
AriehSchneier:
```suggestion
	return longestMatch.Name + remoteURL[longestMatchLength:], true
```

config/url.go
```diff
@@ -50,40 +55,53 @@ func (u *URL) marshal() *format.Subsection {
...
 	u.raw.SetOption(insteadOfKey, u.InsteadOfs...)
+	u.raw.SetOption(pushInsteadOfKey, u.PushInsteadOfs...)
 
 	return u.raw
 }
 
-func findLongestInsteadOfMatch(remoteURL string, urls map[string]*URL) *URL {
+func rewriteLongestURLMatch(remoteURL string, urls map[string]*URL, prefixes func(*URL) []string) (string, bool) {
 	var longestMatch *URL
+	var longestPrefix string
```
AriehSchneier, line 65: longestPrefix is now unused

PR #2189, opened 2026-06-10
Fix URL insteadOf to use longest match when multiple values exist

AriehSchneier:
## Summary

Fixed a bug in `URL.ApplyInsteadOf()` where it would use the first matching `insteadOf` value instead of the longest one when a single `[url]` section contained multiple `insteadOf` entries that both matched the URL.

This violated git's specification which states: _"When more than one insteadOf strings match a given URL, the longest match is used."_

Additionally fixed a second issue where when multiple URLs had the same `insteadOf` value (tie in length), go-git used random map iteration order instead of config file order like git does.

### Bug 1: Wrong match within single URL section

When a single `[url]` section has multiple `insteadOf` values that both match a URL, the **longest** match should be used (per git's spec), but the old code used the **first** match:

**Example configuration:**
```gitconfig
[url "ssh://git@github.com/"]
    insteadOf = https://github.com/
    insteadOf = https://github.com/enterprise/
```

**Before (❌ broken):**
```
Input:  https://github.com/enterprise/user/repo.git
Output: ssh://git@github.com/enterprise/user/repo.git
        (incorrectly used first match: https://github.com/)
```

**After (✅ fixed):**
```
Input:  https://github.com/enterprise/user/repo.git  
Output: ssh://git@github.com/user/repo.git
        (correctly used longest match: https://github.com/enterprise/)
```

### Bug 2: Non-deterministic tie-breaking

When multiple URLs have the same `insteadOf` value (tie in length), git uses **config file order** (first wins), but go-git used **random map iteration order**:

**Example configuration:**
```gitconfig
[url "ssh://git@github.com/"]
    insteadOf = https://github.com/
[url "git@github.com:"]
    insteadOf = https://github.com/
```

**Before (❌ non-deterministic):**
```
Result: ssh://git@github.com/user/repo.git  OR  git@github.com:user/repo.git
        (random - depends on map iteration order)
```

**After (✅ deterministic):**
```
Result: ssh://git@github.com/user/repo.git
        (always uses first URL in config file, matching git)
```

## Implementation Changes

- Modified `ApplyInsteadOf()` to find the longest matching `insteadOf` value instead of using the first match
- Optimized `findLongestInsteadOfMatch()` to apply the rewrite in a single pass instead of two, and renamed it to `applyLongestInsteadOfMatch()` to better reflect its behavior
- **Changed `Config.URLs` from `map[string]*URL` to `[]*URL`** to preserve config file order for deterministic tie-breaking (matching git's behavior)
- When multiple URLs have identical `insteadOf` lengths, use config file order (first wins), matching git's behavior exactly
- Updated `x/plugin/config` static plugin's `cloneURLs` to work with slice instead of map
- Added comprehensive tests for these edge cases

## Testing

- ✅ All existing config tests pass
- ✅ Added `TestApplyInsteadOfLongestMatchWithinSameURL` - unit test for bug #1
- ✅ Added `TestApplyInsteadOfLongestMatchIntegration` - integration test with remotes for bug #1
- ✅ Added `TestApplyInsteadOfDuplicateInsteadOfValues` - test for bug #2 (deterministic ordering)
- ✅ Updated `TestApplyLongestInsteadOfMatch` to verify single-pass behavior
- ✅ Verified behavior matches git exactly in all cases

## Related

This fixes edge cases that were not covered by the existing multiple `insteadOf` support added in PR #1410.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -210,8 +210,8 @@ type Config struct {
 	// equal Branch.Name
 	Branches map[string]*Branch
 	// URLs list of url rewrite rules, if repo url starts with URL.InsteadOf value, it will be replaced with the
-	// key instead.
-	URLs map[string]*URL
+	// URL.Name instead. Ordered by appearance in config file.
+	URLs []*URL
 	// Raw contains the raw information of a config file. The main goal is
 	// preserve the parsed information from the original format, to avoid
 	// dropping unsupported fields.
@@ -326,7 +326,7 @@ func NewConfig() *Config {
 		Remotes:    make(map[string]*RemoteConfig),
 		Submodules: make(map[string]*Submodule),
 		Branches:   make(map[string]*Branch),
-		URLs:       make(map[string]*URL),
+		URLs:       make([]*URL, 0),
 		Raw:        format.New(),
 	}
 
@@ -666,13 +666,14 @@ func (c *Config) unmarshalRemotes() error {
 
 func (c *Config) unmarshalURLs() error {
 	s := c.Raw.Section(urlSection)
+	c.URLs = make([]*URL, 0, len(s.Subsections))
 	for _, sub := range s.Subsections {
 		r := &URL{}
 		if err := r.unmarshal(sub); err != nil {
 			return err
 		}
 
-		c.URLs[r.Name] = r
+		c.URLs = append(c.URLs, r)
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


pjbgf:
To be handled/documented as part of #910.

pjbgf (approved):

PR #2194, opened 2026-06-17
storage: filesystem, Clean up empty parent directories on ref removal

### LINKED ISSUE #1822: RemoveReference should clean up empty directories

Abiggj:
When a reference or reflog is removed from the filesystem, empty parent directories are now recursively deleted. This prevents name conflicts where a directory (e.g., 'refs/heads/feature/') would prevent the creation of a reference with the same name.

A protection mechanism ensures that core Git directory scaffolding (like 'refs/heads', 'refs/tags', and 'objects/info') is preserved even when empty, maintaining compatibility with Git standards and repository initialization.

Key changes:
- Implemented `removeEmptyParentDirs` and `isProtectedPath` in `storage/filesystem/dotgit/dotgit.go`.
- Updated `RemoveRef` and `DeleteReflog` to utilize the new cleanup logic.
- Added regression test `TestRemoveRefAndParentDirs` in `storage/filesystem/dotgit/dotgit_test.go`.

This PR fixes #1822

### DIFF
--- a/storage/filesystem/dotgit/dotgit.go
+++ b/storage/filesystem/dotgit/dotgit.go
@@ -293,6 +293,10 @@ func (d *DotGit) ReflogWriter(name plumbing.ReferenceName) (billy.File, error) {
 func (d *DotGit) DeleteReflog(name plumbing.ReferenceName) error {
 	p := d.fs.Join(logsPath, string(name))
 	err := d.fs.Remove(p)
+	if err == nil {
+		_ = d.removeEmptyParentDirs(p)
+	}
+
 	if os.IsNotExist(err) {
 		return nil
 	}
@@ -1147,6 +1151,9 @@ func (d *DotGit) RemoveRef(name plumbing.ReferenceName) error {
 	_, err := d.fs.Stat(path)
 	if err == nil {
 		err = d.fs.Remove(path)
+		if err == nil {
+			_ = d.removeEmptyParentDirs(path)
+		}
 		// Drop down to remove it from the packed refs file, too.
 	}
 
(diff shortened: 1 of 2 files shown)


storage/filesystem/dotgit/dotgit.go
```diff
@@ -1157,6 +1164,62 @@ func (d *DotGit) RemoveRef(name plumbing.ReferenceName) error {
...
+
+		if err := d.fs.Remove(path); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (d *DotGit) isProtectedPath(p string) bool {
```
pjbgf, line 1195: Upstream has a [slightly different approach](https://github.com/git/git/blob/95e20213faefeb95df29277c58ac1980ab68f701/refs/files-backend.c#L1301), whereby it will always keep the first two components. That way we don't need an arbitrary list here. Based on the other suggested change, this func could be removed.

storage/filesystem/dotgit/dotgit.go
```diff
@@ -1157,6 +1164,62 @@ func (d *DotGit) RemoveRef(name plumbing.ReferenceName) error {
...
+			break
+		}
+
+		if err := d.fs.Remove(path); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
pjbgf, line 1193: For closer alignment with upstream (without hard-coding paths):

```suggestion
// protectedRefComponents is the number of leading reference-name components
// that are never pruned (e.g. refs/heads, refs/remotes). This mirrors git's
// try_remove_empty_parents, which preserves the first two components of the
// reference name; see refs/files-backend.c.
const protectedRefComponents = 2

// removeEmptyParentDirs prunes now-empty parent directories of a removed ref or
// reflog file, walking upward and stopping at the protected prefix or at the
// first non-empty directory.
//
// The floor is derived from name, which is always relative and slash-separated,
// so the number of prunable levels is independent of path. path is only used
// for the filesystem operations and may be absolute; it is walked with
// filepath.Dir, so its leading directories (the .git prefix) never affect the
// floor. Cleanup is best-effort: callers discard the returned error.
func (d *DotGit) removeEmptyParentDirs(path string, name plumbing.ReferenceName) error {
	prune := len(strings.Split(string(name), "/")) - protectedRefComponents - 1

	dir := path
	for i := 0; i < prune; i++ {
		dir = filepath.Dir(dir)

		files, err := d.fs.ReadDir(dir)
		if err != nil {
			return err
		}

		if len(files) > 0 {
			break
		}

		if err := d.fs.Remove(dir); err != nil {
			return err
		}
	}

	return nil
}
```

pjbgf (changes requested):
@Abiggj thanks for working on this. Here's some feedback. Please remember to sign-off the commit and reword the commit message so that it aligns with the contributing guidelines.

Issue #2112, opened 2026-05-12
go-git ignores files inside un-ignored directory with wildcard pattern


aktau:
### Bug Description

When using a `.gitignore` file with a wildcard ignore `*` followed by a directory-only inclusion like `!dir/`, `go-git` fails to ignore files inside `dir/`, whereas standard Git ignores them.

In `plumbing/format/gitignore/pattern.go`, the `simpleNameMatch` function iterates over all components of a path. If any component matches the pattern, it returns `true` (matched), unless it's the last component and it's checking `dirOnly`.

For a path `["dir", "file.txt"]` and pattern `!dir/`:
 
1. Component `dir` matches `dir`.
2. It is not the last component (`i == 0`, `len(path)-1 == 1`).
3. The function returns `true` (matched).
4. This causes the file to be included, overriding the `*` pattern.

Effectively, `!dir/` acts as `!dir/**` in `go-git`.

(I did use AI to generate a bug report with identifiers that avoid the non-public parts of my project, I edited it myself and ran the reproducer myself.)

### go-git Version

488798e28afdf2a31fc6ce32fd3ec8023025f449

### Steps to Reproduce

To reproduce:

Standard Git:

```bash
mkdir git-repro
cd git-repro
git init
echo '*' > .gitignore
echo '!my-dir/' >> .gitignore
mkdir -p my-dir/sub
echo 'test' > my-dir/sub/file.txt
git status
```

This would say: `nothing to commit` (the file inside `my-dir` is ignored).

git based on `go-git`:

Run these commands to set up the environment and run the reproducer:

```bash
mkdir gogit-repro
cd gogit-repro
# Save the Go code below as repro.go
go mod init repro
go get github.com/go-git/go-git/v5
go run repro.go
```

```go
// repro.go
package main

import (
	"fmt"
	"os"

	"github.com/go-git/go-git/v5"
)

func main() {
	// Setup environment
	os.MkdirAll("my-dir/sub", 0755)
	os.WriteFile(".gitignore", []byte("*\n!my-dir/\n"), 0644)
	os.WriteFile("my-dir/sub/file.txt", []byte("test"), 0644)

	r, err := git.PlainInit(".", false)
	if err != nil {
		panic(err)
	}
	w, err := r.Worktree()
	if err != nil {
		panic(err)
	}

	status, err := w.Status()
	if err != nil {
		panic(err)
	}

	fmt.Println("Status:")
	for path, s := range status {
[...]

PR #2195, opened 2026-06-18
plumbing: format/ignore, Ensure inclusion patterns match full paths

Abiggj:
This PR fixes issue #2112 where `.gitignore` inclusion patterns (e.g., `!dir/`) incorrectly matched files inside those directories when a parent directory match was found. 
 
### Changes
- **Modified `simpleNameMatch`**: Updated logic to require that inclusion patterns match the final component of the path. Parent directory matches are now ignored for inclusions, preventing them from overriding wildcard ignores for nested files.
- **Modified `globMatch`**: Enforced full path consumption (or trailing `**` matches) for inclusion patterns to be considered a match.
- **Updated `TestSimpleMatch_inclusion`**: Adjusted existing tests to reflect standard Git behavior, where an inclusion pattern for a directory does not automatically include its contents. 

These changes ensure that configurations like:
`  *`
 ` !my-dir/`
correctly ignore files inside `my-dir/`, matching the behavior of standard Git.

### DIFF
--- a/plumbing/format/gitignore/conformance_test.go
+++ b/plumbing/format/gitignore/conformance_test.go
@@ -105,7 +105,7 @@ func (s *ConformanceSuite) gitOracle(patterns []string, path string, isDir bool)
 	if path == "." || path == ".." || strings.Contains(path, "//") {
 		return false, false
 	}
-	if skipOnLegacyGit(patterns) {
+	if skipOnLegacyGit(patterns) || skipNonSlashAdjacentDoubleStar(patterns) {
 		return false, false
 	}
 	if err := cleanScratch(root); err != nil {
@@ -119,9 +119,6 @@ func (s *ConformanceSuite) gitOracle(patterns []string, path string, isDir bool)
 		return false, false
 	}
 	arg := path
-	if isDir && !strings.HasSuffix(arg, "/") {
-		arg += "/"
-	}
 	// `check-ignore` (without --no-index) stats the path so it can distinguish
 	// dirs from files, which is the only way it produces the same answer as
 	// `git status` for re-include patterns over directories. The platform's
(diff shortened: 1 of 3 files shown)


pjbgf (changes requested):
@Abiggj thanks for looking into this. Please re-phrase the commit so that it matches our [contributing guide](https://github.com/go-git/go-git/blob/main/CONTRIBUTING.md#format-of-the-commit-message): `plumbing: format/ignore, Ensure inclusion patterns match full paths`.

Add the tests below into `plumbing/format/gitignore/conformance_test.go` so that we avoid future regression:

```go

func (s *ConformanceSuite) TestIgnoreNegation_ancestorMatch() {
	tests := []struct {
		patterns []string
		path     string
		isDir    bool
		ignored  bool
		desc     string
	}{
		{
			[]string{"*.log", "!important"},
			"dir/important/app.log", false, true,
			"!important matches ancestor dir, must not re-include app.log",
		},
		{
			[]string{"*.log", "!important"},
			"important/app.log", false, true,
			"!important is ancestor-only; *.log still ignores the file",
		},
		{
			[]string{"*.log", "!important"},
			"important", true, false,
			"!important re-includes the dir itself (basename match)",
		},
		{
			[]string{"*", "!value/vul?ano"},
			"value/vulkano/tail", false, true,
			"!value/vul?ano matches an ancestor only; * keeps the file ignored",
		},
	}
	for _, tt := range tests {
		m := s.createMatcher(tt.patterns)
		s.assertIgnore(m, tt.patterns, tt.path, tt.isDir, tt.ignored, tt.desc)
	}
}
```

AriehSchneier:
> Add the tests below into `plumbing/format/gitignore/conformance_test.go` so that we avoid future regression:
> ...

I would add a further 2 more:
```
{
    []string{"*", "!dir/"},
    "dir/file.txt", false, true,
    "!dir/ re-includes only the dir itself; * still ignores files inside",
},
{
    []string{"*", "!dir/"},
    "dir", true, false,
    "!dir/ re-includes the directory itself",
},

```

Abiggj:
Updated the commit message to follow the contributing guide and added the requested regression tests (including the extra !dir/ cases). All tests are passing.

Issue #2198, opened 2026-06-18
Add support for `feature.manyFiles=true`


pjbgf:
### Feature Description

Enable config options that optimize for repos with many files in the working directory. 

- [ ] `index.skipHash=true` speeds up index writes by not computing a trailing checksum. Note that this will cause Git versions earlier than 2.13.0 to refuse to parse the index and Git versions earlier than 2.40.0 will report a corrupted index during git fsck. Tracked by https://github.com/go-git/go-git/issues/2196.

- [ ] `index.version=4` enables path-prefix compression in the index.

- [ ] `core.untrackedCache=true` enables the untracked cache. This setting assumes that mtime is working on your machine.

Ref:
- https://git-scm.com/docs/git-config#Documentation/git-config.txt-featuremanyFiles

Issue #2197, opened 2026-06-18
Add support for `feature.experimental=true`


pjbgf:
### Feature Description

As creation time, this results in the following features being enabled:

- [ ] `fetch.negotiationAlgorithm=skipping` may improve fetch negotiation times by skipping more commits at a time, reducing the number of round trips.

- [ ] `pack.useBitmapBoundaryTraversal=true` may improve bitmap traversal times by walking fewer objects.

- [ ] `pack.allowPackReuse=multi` may improve the time it takes to create a pack by reusing objects from multiple packs instead of just one.

- [ ] `pack.usePathWalk` may speed up packfile creation and make the packfiles be significantly smaller in the presence of certain filename collisions with Git’s default name-hash.

- [ ] `init.defaultRefFormat=reftable` causes newly initialized repositories to use the reftable format for storing references. This new format solves issues with case-insensitive filesystems, compresses better and performs significantly better with many use cases. Refer to Documentation/technical/reftable.adoc for more information on this new storage format. Depends on #1827.

This relates to v3 alignment: #1831.

Ref:
- https://git-scm.com/docs/git-config#Documentation/git-config.txt-featureexperimental

PR #2199, opened 2026-06-18
config: expand feature.manyFiles and feature.experimental aliases for index.skipHash

### LINKED ISSUE #2196: index: feature.manyFiles / feature.experimental not expanded, so index.skipHash is ignored (invalid checksum on read)
### Bug Description

When a repository's `.git/config` enables `index.skipHash` indirectly via Git's
`feature.manyFiles = true` (or `feature.experimental = true`) meta-setting,
go-git fails to read the index with:

```
index decoder: invalid checksum
```

Git writes the index trailer as 20 (or 32 for SHA-256) zero bytes when
`index.skipHash` is in effect. go-git supports `index.skipHash` when the key is
present **literally** under `[index]`, but it does not expand Git's
`feature.*` aliases, so `Config.Index.SkipHash` stays unset and the decoder
verifies the checksum against the all-zero trailer and rejects the index.

### go-git Version

v6.0.0-alpha

### Steps to Reproduce

```sh
git init repo && cd repo
git config feature.manyFiles true   # implies index.skipHash=true in Git >= 2.40
echo hi > a.txt && git add a.txt && git commit -m init
```

```go
package main

import (
    "fmt"

)

func main() {
    repo, err := git.PlainOpen("repo")
    if err != nil {
        panic(err)
    }
    wt, err := repo.Worktree()
    if err != nil {
        panic(err)
    }
    _, err = wt.Status()
    fmt.Println("status err:", err) // => index decoder: invalid checksum
}
```

`git status` on the same repo works fine. Adding `git config index.skipHash true`
explicitly also makes go-git succeed, confirming the gap is alias expansion, not
checksum handling.

### Expected behavior

go-git should honor Git's `feature.manyFiles` / `feature.experimental`
meta-settings, which expand to (per Git docs):

- `feature.manyFiles=true` → `index.skipHash=true`, `index.version=4`, `core.untrackedCache=true`
- `feature.experimental=true` → `index.skipHash=true` (among others)

so that reading such an index does not error. At minimum, `index.skipHash`
should be inferred from `feature.manyFiles` / `feature.experimental` when the
literal `index.skipHash` key is absent.

### Notes

- `feature.manyFiles` is increasingly common: some Git GUIs/tooling and large
  monorepos enable it for performance.
- A repo with this config also typically carries `UNTR` (untracked cache) and
  `FSMN` (fsmonitor) index extensions; those decode fine as optional extensions —
  the only failure is the zero-checksum verification.
- Possibly related: #1123 (index version 4 read failure).

jfjrh2014:
## Summary

Git's `feature.manyFiles` and `feature.experimental` meta-settings implicitly enable `index.skipHash` (among other settings). go-git only honored the literal `[index] skipHash` key, so repositories created with these meta-settings had their index read fail with an `invalid checksum` error because `skipHash` stayed false and the decoder verified the all-zero trailer.

This expands the aliases so `index.skipHash` is inferred correctly, matching the behaviour of the reference Git implementation. The literal `index.skipHash` key takes precedence over the aliases.

Fixes #2196
Refs #2197, #2198

## References

- `git-config(1)`: feature.manyFiles → index.skipHash=true, index.version=4, core.untrackedCache=true; feature.experimental → index.skipHash=true
- PR #2181 fixed the decoder side (accept empty trailing checksum); this PR fixes the config side (expand the aliases)
- Verified against the reference Git implementation:

```sh
$ git init repo && cd repo
$ git config feature.manyFiles true
$ git config --get index.skipHash   # empty (alias not expanded in config output)
$ echo hi > a.txt && git add a.txt && git commit -m init
$ git ls-files --debug | head -3     # but index internally has skipHash in effect
```

## Changes

- `unmarshalIndex()` now checks `feature.manyFiles` and `feature.experimental` when the literal `index.skipHash` key is absent
- Extracted `parseIndexSkipHash()` helper for clarity
- 4 new test cases covering: manyFiles=true, experimental=true, manyFiles=false, explicit override

## Checklist

- [x] Tests added (4 new cases in `TestUnmarshalIndexSkipHash`)
- [x] `go test ./config/` passes (7/7 cases)
- [x] `go vet ./config/` clean
- [x] Signed off

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -488,11 +488,6 @@ const (
 	uploadArchiveSection       = "uploadArchive"
 	allowUnreachableKey        = "allowUnreachable"
 
-	// featureSection is a Git meta-setting that expands to multiple
-	// other config keys. See git-config(1).
-	featureSection = "feature"
-	manyFilesKey   = "manyFiles"
-
 	// DefaultPackWindow holds the number of previous objects used to
 	// generate deltas. The value 10 is the same used by git command.
 	DefaultPackWindow = uint(10)
@@ -739,8 +734,8 @@ func (c *Config) unmarshalIndex() {
 	// (among other settings). Infer the alias when the literal key is absent,
 	// so that reading an index from a repository using this feature does not
 	// fail with a checksum error. See git-config(1) and #2196.
-	feat := c.Raw.Section(featureSection)
-	if strings.EqualFold(feat.Options.Get(manyFilesKey), "true") {
+	feat := c.Raw.Section("feature")
+	if strings.EqualFold(feat.Options.Get("manyFiles"), "true") {
 		c.Index.SkipHash = OptBoolTrue
 	}
 }
--- a/config/config.go
+++ b/config/config.go
@@ -488,6 +488,12 @@ const (
 	uploadArchiveSection       = "uploadArchive"
 	allowUnreachableKey        = "allowUnreachable"
 
+	// featureSection is a Git meta-setting that expands to multiple
+	// other config keys. See git-config(1).
+	featureSection  = "feature"
+	manyFilesKey    = "manyFiles"
+	experimentalKey = "experimental"
+
 	// DefaultPackWindow holds the number of previous objects used to
 	// generate deltas. The value 10 is the same used by git command.
 	DefaultPackWindow = uint(10)
@@ -726,11 +732,32 @@ func (c *Config) unmarshalProtocol() error {
 }
 
 func (c *Config) unmarshalIndex() {
+	// Literal index.skipHash takes precedence over the feature.* aliases.
+	if c.parseIndexSkipHash() {
+		return
+	}
+	// Git's feature.manyFiles and feature.experimental meta-settings
+	// implicitly enable index.skipHash (among other settings). Infer the
+	// alias when the literal key is absent, so that reading an index from
+	// a repository using these features does not fail with a checksum
+	// error. See git-config(1) and #2196.
+	feat := c.Raw.Section(featureSection)
+	if strings.EqualFold(feat.Options.Get(manyFilesKey), "true") ||
+		strings.EqualFold(feat.Options.Get(experimentalKey), "true") {
+		c.Index.SkipHash = OptBoolTrue
+	}
+}
+
+// parseIndexSkipHash reads the literal [index] skipHash option. It returns
+// true if the option was present and parsed (so callers can short-circuit).
+func (c *Config) parseIndexSkipHash() bool {
 	s := c.Raw.Section(indexSection)
 	v, err := strconv.ParseBool(s.Options.Get(skipHashKey))
-	if err == nil {
-		c.Index.SkipHash = NewOptBool(v)
+	if err != nil {
+		return false
 	}
+	c.Index.SkipHash = NewOptBool(v)
+	return true
 }
 
 func (c *Config) unmarshalInit() {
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -726,11 +732,32 @@ func (c *Config) unmarshalProtocol() error {
...
+		return
+	}
+	// Git's feature.manyFiles and feature.experimental meta-settings
+	// implicitly enable index.skipHash (among other settings). Infer the
+	// alias when the literal key is absent, so that reading an index from
+	// a repository using these features does not fail with a checksum
+	// error. See git-config(1) and #2196.
+	feat := c.Raw.Section(featureSection)
+	if strings.EqualFold(feat.Options.Get(manyFilesKey), "true") ||
+		strings.EqualFold(feat.Options.Get(experimentalKey), "true") {
```
pjbgf: > feature.experimental [does not expand/enable](https://github.com/git/git/blob/4621f8ce5e9b97aa2e8d0d9ffe9d25df2471074d/repo-settings.c#L53-L64) this upstream.

```suggestion
	if strings.EqualFold(feat.Options.Get(manyFilesKey), "true") {
```

config/config_test.go
```diff
@@ -754,6 +754,26 @@ func TestUnmarshalIndexSkipHash(t *testing.T) {
...
+		{
+			name:  "feature.manyFiles enables skipHash",
+			input: "[feature]\n\tmanyFiles = true\n",
+			want:  OptBoolTrue,
+		},
+		{
+			name:  "feature.experimental enables skipHash",
+			input: "[feature]\n\texperimental = true\n",
+			want:  OptBoolTrue,
+		},
```
pjbgf:
```suggestion
```

config/config_test.go
```diff
@@ -754,6 +754,26 @@ func TestUnmarshalIndexSkipHash(t *testing.T) {
...
+		{
+			name:  "feature.manyFiles=false does not enable skipHash",
+			input: "[feature]\n\tmanyFiles = false\n",
+			want:  OptBoolUnset,
+		},
+		{
+			name:  "explicit index.skipHash overrides feature.manyFiles",
```
pjbgf, line 768: Please add another that handles an explicit false for `index.skipHash` and a `true` on `feature.manyFiles`.

config/config.go
```diff
@@ -488,6 +488,12 @@ const (
 	uploadArchiveSection       = "uploadArchive"
 	allowUnreachableKey        = "allowUnreachable"
 
+	// featureSection is a Git meta-setting that expands to multiple
+	// other config keys. See git-config(1).
+	featureSection  = "feature"
+	manyFilesKey    = "manyFiles"
+	experimentalKey = "experimental"
```
pjbgf: This does not need to be added as part of this change:
```suggestion
```

pjbgf (changes requested):
@jfjrh2014 thanks for looking into this. Here are a few things we need to change:

jfjrh2014:
Thanks @pjbgf for the detailed review. Addressed all points in d2d6a03e:

1. Dropped `feature.experimental` entirely — confirmed upstream `repo-settings.c` only checks `feature.manyFiles` for `index.skipHash`, so the `||` branch was incorrect.
2. Removed `experimentalKey` constant and the obsolete test case.
3. Kept `strings.EqualFold` for the `manyFiles=true` check and the explicit-false / explicit-override test cases you asked for.

Ready for re-review when you have a moment.

config/config.go
jfjrh2014: Addressed in e58373b.

config/config_test.go
jfjrh2014: Addressed in e58373b.

PR #2216, opened 2026-06-24
plumbing: transport, Wire Protocol v2 across HTTP and share the client core (8/11)

aymanbagabas:
## Part 8: wire Protocol v2 across HTTP and share the client core

Builds on Part 7 (#2212). Parts 5 to 7 implemented v2 for the stream sessions (SSH, git://, file) inline. This part extracts that orchestration into one shared place and wires the HTTP transport to it, so the client speaks v2 over HTTP too.

### What changed

- New `internal/transport` package holds the v2 client orchestration: `ls-refs`, the `do_fetch_pack_v2`-style fetch negotiation with incremental have batching, `FetchSupports`, the capability echo, and caller-owned packfile streaming. It is internal, so no new public v2 API is added.
- `FetchRequest` moves into `internal/transport` and `plumbing/transport` aliases it, so the v0/v1 and v2 fetch paths use the exact same request type. The public `transport.Commander` interface is unchanged; the shared helpers take a plain `CommandFunc` to avoid an import cycle.
- Stream transports delegate their v2 `GetRemoteRefs` and `Fetch` to the shared helpers, passing their `FetchRequest` straight through.
- The smart HTTP session decodes the v2 capability advertisement, implements `Commander` over a single stateless POST (with the `Git-Protocol` header), and runs `GetRemoteRefs` and `Fetch` through the same helpers. Each fetch round owns its response body, which the helper closes between rounds.

### Tests

An end-to-end test drives the go-git HTTP client over v2 (handshake, ls-refs with HEAD symref and ref-prefix filtering, and fetch) against go-git's own v2 server. The existing stream tests run through the shared core unchanged.

### Reviewing just this part

https://github.com/aymanbagabas/go-git/compare/gitprotocol-v2-7-v2-fetch...gitprotocol-v2-9-http-shared

### Notes

Targets `main` (v6). Depends on Part 7 (#2212); merge that first. Server-side v2 cleanup and HTTP archive support follow in later parts.

### DIFF
--- a/internal/transport/transport.go
+++ b/internal/transport/transport.go
@@ -1,0 +1,39 @@
+// Package transport holds transport internals shared across go-git's transport
+// implementations (the public facade lives in plumbing/transport, which aliases
+// the exported types here). It is not part of go-git's public API.
+package transport
+
+import (
+	"github.com/go-git/go-git/v6/plumbing"
+	"github.com/go-git/go-git/v6/plumbing/protocol/packp"
+	"github.com/go-git/go-git/v6/plumbing/protocol/packp/sideband"
+)
+
+// FetchRequest describes a fetch. It is shared by the v0/v1 and v2 fetch paths;
+// plumbing/transport aliases it as transport.FetchRequest.
+type FetchRequest struct {
+	// Progress is the progress sideband.
+	Progress sideband.Progress
+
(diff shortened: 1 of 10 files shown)


aymanbagabas:
Superseded by #2231 (recreated as a same-repo stacked PR on origin).

PR #2221, opened 2026-06-25
plumbing: format/config, add canonical git value, key, accessor, and ConfigSet API

aymanbagabas:
## Summary

This makes `plumbing/format/config` follow canonical git more closely by adding a value, key, accessor, and multi-source read API on top of the existing `Section`/`Subsection`/`Option` model. It is purely additive: no existing behavior or API changes, and no consumers need updating.

This also addresses the inconsistent boolean parsing noted in https://github.com/go-git/go-git/pull/1995 by providing one git-spec-compliant value parser.

### Value parsing (`value.go`), mirroring git's `parse.c`
- `ParseBool` follows `git_parse_maybe_bool`: `true`/`yes`/`on` and `false`/`no`/`off` case-insensitively, any integer by truthiness, and the bare-key shorthand (an empty value is true, matching `[core] bare`).
- `ParseInt`, `ParseInt64`, `ParseUint`, `ParseUint64` follow `git_parse_signed`/`git_parse_unsigned`: base-0 parsing so hex and octal work, optional `k`/`m`/`g` (1024) unit suffixes, range checked.
- Sentinel errors: `ErrInvalidBool`, `ErrInvalidNumber`, `ErrValueOutOfRange`.

### Key handling (`key.go`), mirroring git's `do_parse_config_key` and `iskeychar`
- `SplitKey` splits `section[.subsection].variable` (first dot is the section, last dot is the variable, the middle is the subsection and may contain dots).
- `IsValidKey` / `ValidateKey` enforce that the section and variable names are alphanumeric or `-`, the variable name starts with a letter, and the subsection contains no newline.

### Accessors (`accessor.go`, `option.go`), mirroring git's `git_configset_get_*`
- Flat dotted-key API on `Config`: `Lookup` (value plus presence), `Get`, `GetAll` (multivar), `Has`, typed `Bool`/`Int`/`Int64`/`Uint`/`Uint64`/`String` with defaults, and key-validated `Set`/`Add`/`Unset`.
- Reads merge all blocks of a repeated `[section]` with the last value winning, exactly as git does.
- The same typed getters are also available on `Options`, so `Section` and `Subsection` get typed access via their `Options`.

### Multi-source reads (`interface.go`, `configset.go`)
- `Getter` is the minimal read primitive (`Lookup`, `GetAll`). A single parsed `*Config`, an effective `*ConfigSet`, and any caller-supplied source all satisfy it, so they compose freely.
- `ConfigSet` is a read-only view over multiple `Getter` sources in precedence order, the equivalent of git's `config_set` composing several files into one effective view:

  ```go
  set := config.NewConfigSet(local, global, system) // highest precedence first
  name := set.String("user.name", "")               // highest-precedence value
  all  := set.GetAll("safe.directory")              // accumulated across all scopes
  ```

  Because `ConfigSet` is itself a `Getter`, sets nest. It never mutates its sources, is not encodable, and has no `Set`/`Add`/`Unset`, so the writable one-file `Config` keeps accurate per-file `Has`/`GetAll` semantics while effective resolution stays an explicit, separate concept rather than a fallback baked into the low-level type.

## Testing

`go build ./...`, `go vet ./...`, package tests, and `golangci-lint run` are all green; gofmt clean. Value parsing and key handling have fuzz targets (`FuzzParseValue`, `FuzzKey`), each run to over a million executions with no crashes. The only failing tests in my environment are the pre-existing `safe.bareRepository` external-git failures that also fail on a clean `main`.

This change was produced with AI assistance (see the `Assisted-by` trailers on the commits).

### DIFF
--- a/plumbing/format/config/accessor.go
+++ b/plumbing/format/config/accessor.go
@@ -1,0 +1,140 @@
+package config
+
+// This file adds the canonical git "config set" access API on top of the
+// Section/Subsection/Option model: typed getters and setters keyed by a flat
+// "section[.subsection].variable" string, mirroring git's
+// git_configset_get_value, git_configset_get_bool, git_configset_get_int and
+// friends. Reads merge all blocks of a repeated section, with the last value
+// winning, exactly as git does.
+
+// allOptions collects, in file order, the options of every block of the named
+// section (and subsection, when given). git treats repeated [section] blocks as
+// a single logical section, so accessors must look across all of them.
+func (c *Config) allOptions(section, subsection string) Options {
+	var opts Options
+	for _, s := range c.Sections {
+		if !s.IsName(section) {
+			continue
+		}
+		if subsection == "" {
+			opts = append(opts, s.Options...)
+			continue
+		}
+		for _, ss := range s.Subsections {
+			if ss.IsName(subsection) {
+				opts = append(opts, ss.Options...)
+			}
+		}
+	}
+	return opts
+}
+
+// Lookup returns the last value set for the canonical key and whether it was
+// present at all. It is the equivalent of git's git_configset_get_value.
+func (c *Config) Lookup(key string) (string, bool) {
+	section, subsection, variable := SplitKey(key)
+	return c.allOptions(section, subsection).Lookup(variable)
+}
+
+// Get returns the last value set for the canonical key, or the empty string if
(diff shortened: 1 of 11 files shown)


aymanbagabas:
Part of the config redesign tracked in the RFC: https://github.com/go-git/go-git/issues/2224

PR #2223, opened 2026-06-25
x: config, add experimental global and system config loaders

aymanbagabas:
## Summary

Experimental follow-up to the canonical config work in https://github.com/go-git/go-git/pull/2221 (this PR stacks on it; the `Getter`/`ConfigSet` commit belongs to that PR and drops out once it merges).

It adds opt-in loaders for git's user- and system-level configuration files, the explicit pieces for *effective* (multi-source) config resolution that are deliberately kept out of the repository core. The repository core stays local-config only.

### `x/config` (package `xconfig`, experimental)

Loaders that read git's global and system config files and return each scope as a read-only `config.ConfigSet`, which callers compose for git-like effective resolution:

```go
globalSet, _ := xconfig.GlobalConfig() // ~/.gitconfig layered over the XDG file
systemSet, _ := xconfig.SystemConfig() // /etc/gitconfig (platform-aware)

eff := config.NewConfigSet(local, globalSet, systemSet)
window, _ := eff.Int("pack.window", 10 /* call-site default */)
```

Path resolution follows git, including the environment variables that override it, all verified against the reference `git` implementation:

- `GIT_CONFIG_GLOBAL` selects a single global file (replacing both `~/.gitconfig` and the XDG file), or disables global config when empty.
- `GIT_CONFIG_SYSTEM` selects a single system file, or disables system config when empty.
- `GIT_CONFIG_NOSYSTEM` skips system config entirely.
- `XDG_CONFIG_HOME` locates the XDG file, with platform fallbacks (Windows `APPDATA`/`PROGRAMFILES`).

The default global scope layers `~/.gitconfig` over the XDG file (both are read, `~/.gitconfig` wins), matching git rather than treating them as mutually exclusive. Missing files are skipped; a malformed file is a hard error. Includes (`include.path`/`includeIf`), command-line `-c`, and inline defaults are intentionally out of scope.

## Notes

- A `DefaultConfig()` source was intentionally left out. Git's defaults are call-site defaults (the typed getters already take a default value), and modelling them as a config source would make `Has` report unconfigured keys as present. Defaults stay behavior-local exported constants, like the existing `DefaultObjectFormat`.
- This is part of the foundation that will eventually replace `Repository.ConfigScoped`, which flattens scopes into one `*Config` and loses provenance. That migration is a separate change; this PR does not touch `ConfigScoped`.

This change was produced with AI assistance (see the `Assisted-by` trailer on the commit).

### DIFF
--- a/plumbing/format/config/accessor.go
+++ b/plumbing/format/config/accessor.go
@@ -1,0 +1,140 @@
+package config
+
+// This file adds the canonical git "config set" access API on top of the
+// Section/Subsection/Option model: typed getters and setters keyed by a flat
+// "section[.subsection].variable" string, mirroring git's
+// git_configset_get_value, git_configset_get_bool, git_configset_get_int and
+// friends. Reads merge all blocks of a repeated section, with the last value
+// winning, exactly as git does.
+
+// allOptions collects, in file order, the options of every block of the named
+// section (and subsection, when given). git treats repeated [section] blocks as
+// a single logical section, so accessors must look across all of them.
+func (c *Config) allOptions(section, subsection string) Options {
+	var opts Options
+	for _, s := range c.Sections {
+		if !s.IsName(section) {
+			continue
+		}
+		if subsection == "" {
+			opts = append(opts, s.Options...)
+			continue
+		}
+		for _, ss := range s.Subsections {
+			if ss.IsName(subsection) {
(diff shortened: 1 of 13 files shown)


aymanbagabas:
Part of the config redesign tracked in the RFC: https://github.com/go-git/go-git/issues/2224

PR #2188, opened 2026-06-10
plumbing: transport, Protocol v2 for upload-pack

blmayer:
I only did ls-refs and fetch command handling. Protocol defaults to v0 when no
Git-Protocol header is sent.

Not sure this is ready as is, we can work on a proper PR to meet the repo standards.

Thanks again.

PR #2226, opened 2026-06-25
plumbing: transport, Add options and result struct to GetRemoteRefs (3/11)

aymanbagabas:
## Part 3 of 11: GetRemoteRefs options and result struct

Third of the Protocol v2 series. Builds on Part 2 (#2225).

### What changes

`Session.GetRemoteRefs` now takes a nullable `*GetRemoteRefsOptions` and
returns a `*RemoteRefs`, following git's `transport_get_remote_refs`:

- `GetRemoteRefsOptions` carries only `RefPrefixes`, which map to the
  Protocol v2 ls-refs `ref-prefix` arguments. They are ignored for v0/v1,
  where the server advertises every reference.
- `RemoteRefs.References` holds the advertised refs and `RemoteRefs.Unborn`
  reports the symref target of an unborn HEAD.

`Unborn` is output only: git sends the ls-refs `unborn` argument whenever the
server supports it rather than gating it behind a request flag. An unborn HEAD
is modeled as a symbolic reference whose target has no corresponding hash
reference, so no new reference type is required.

Existing callers, including the #2188 HTTP backend test, are updated to the
new signature.

### Reviewing just this part

https://github.com/go-git/go-git/compare/gitprotocol-v2-2-transport...gitprotocol-v2-3-getremoterefs

### Notes

- Targets `main` (v6). Built and tested green. Squashed.

Depends on Part 2 (#2225); merge that first.

---
_Recreated as a same-repo stacked PR (supersedes #2208)._

### DIFF
--- a/internal/server/servers_test.go
+++ b/internal/server/servers_test.go
@@ -71,9 +71,9 @@ func (s *ServerSuite) TestUploadPack() {
 	s.Require().NoError(err)
 	s.T().Cleanup(func() { s.Require().NoError(sess.Close()) })
 
-	refs, err := sess.GetRemoteRefs(context.Background())
+	refs, err := sess.GetRemoteRefs(context.Background(), nil)
 	s.Require().NoError(err)
-	s.Greater(len(refs), 0, "server should advertise refs")
+	s.Greater(len(refs.References), 0, "server should advertise refs")
 }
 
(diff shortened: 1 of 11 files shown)


pjbgf (approved):
@aymanbagabas Thanks for working on this. :bow:

PR #2227, opened 2026-06-25
plumbing: packp, Protocol v2 wire types with caller-owned packfile streaming (4/11)

aymanbagabas:
## Part 4 of 11: Protocol v2 wire types

Fourth of the Protocol v2 series. Builds on Part 3 (#2226). Session wiring
follows in Parts 5-11.

This part adds the Protocol v2 wire encoding and decoding. It is types only:
nothing is wired into the transport sessions yet, so there is no behavior
change for existing callers.

### What changes

- `command.go`: the generic v2 command request envelope.
- `lsrefs.go`: the ls-refs command arguments and response (ref-prefix,
  symrefs, peel, unborn).
- `fetch.go`: the fetch command arguments and the fetch response.
- `capability_adv.go`: the v2 capability advertisement.
- `list.go`: v2 capability list encoding.
- A `Commander` session interface for running arbitrary v2 commands.

### Fetch response: caller-owned packfile streaming

`FetchOutput` decodes and encodes only the metadata sections
(`acknowledgments`, `shallow-info`, `wanted-refs`, `packfile-uris`). It does
not carry the packfile. A `Packfile bool` reports whether a packfile section
follows; when it does, the reader is left positioned at the first packfile
pkt-line so the caller streams it, and on encode the caller writes the packfile
data and the final flush-pkt. A negotiation round that ends with a flush-pkt
carries no packfile. This matches the two response shapes in gitprotocol-v2.

Fuzz targets are added for the fetch decoders.

### Reviewing just this part

https://github.com/go-git/go-git/compare/gitprotocol-v2-3-getremoterefs...gitprotocol-v2-4-v2types

### Notes

- Targets `main` (v6). Built and tested green, including a fuzz smoke of the
  fetch response decoder. Squashed.

Depends on Part 3 (#2226); merge that first.

---
_Recreated as a same-repo stacked PR (supersedes #2209)._

### DIFF
--- a/plumbing/protocol/packp/capability_adv.go
+++ b/plumbing/protocol/packp/capability_adv.go
@@ -1,0 +1,96 @@
+package packp
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/go-git/go-git/v6/plumbing/format/pktline"
+	"github.com/go-git/go-git/v6/plumbing/protocol"
+	"github.com/go-git/go-git/v6/plumbing/protocol/capability"
+)
+
+// CapabilityAdv represents a protocol v2 server capability advertisement.
+// It includes the version line and the capability lines that follow it.
+//
+// In protocol v2, the server sends:
+//
+//	version 2\n
+//	agent=git/2.45.0\n
+//	ls-refs=unborn\n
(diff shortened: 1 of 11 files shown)


pjbgf (approved):

PR #2228, opened 2026-06-25
plumbing: transport, Protocol v2 session handshake and Command (5/11)

aymanbagabas:
## Part 5: Protocol v2 session handshake and Command

First piece of the Protocol v2 client wiring, built on the v2 wire types from Part 4 (#2227). This part is implemented (no longer a scaffold).

### What changed

- Stream sessions (SSH, git://, file) now detect `protocol.V2` during the handshake and decode the v2 capability advertisement (`packp.CapabilityAdv`) instead of the v0/v1 `AdvRefs`. For v2, references are not read at handshake time; they come later via `ls-refs` (Part 6).
- The negotiated version and the server's v2 capabilities are stored on the session.
- `StreamSession` implements the `Commander` interface. `Command` takes the command-specific arguments (a `packp.CommandArgs` such as `LsRefsArgs` or `FetchArgs`) plus a response decoder, builds the v2 request envelope through `packp.CommandRequest` (command line, the capabilities collected during the handshake, delim-pkt, the arguments, flush-pkt), and decodes the response.
- The capabilities sent with each command are the agent and the server's `object-format` echoed back, so both sides agree on the hash algorithm.

### Tests

Real coverage replaces the earlier skipped scaffold: a v2 handshake decode, the `Command` envelope encoding, a non-v2 rejection, and an end-to-end `ls-refs` round over an in-memory pipe against the server's v2 path.

### Reviewing just this part

https://github.com/go-git/go-git/compare/gitprotocol-v2-4-v2types...gitprotocol-v2-5-v2-handshake

### Notes

Targets `main` (v6). Depends on Part 4 (#2227); merge that first.

---
_Recreated as a same-repo stacked PR (supersedes #2210)._

### DIFF
--- a/plumbing/transport/pack_session.go
+++ b/plumbing/transport/pack_session.go
@@ -14,12 +14,15 @@ import (
 // built-in Fetch and Push operations.
 //
 // Sessions that negotiate Protocol v2 (version 2) implement this interface.
-// The Command method executes a named v2 command, encoding the request
-// via req.Encode and decoding the response via resp.Decode. The transport
-// layer handles the v2 envelope (command name, capabilities, delim-pkt,
-// flush-pkt, and for HTTP, response-end).
+// The Command method executes a named v2 command: req carries the
+// command-specific arguments and is encoded into the request, while resp
+// decodes the response. For example, GetRemoteRefs runs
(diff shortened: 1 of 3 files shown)


pjbgf (approved):

PR #2229, opened 2026-06-25
plumbing: transport, ls-refs over Protocol v2 for GetRemoteRefs (6/11)

aymanbagabas:
## Part 6: ls-refs over Protocol v2 for GetRemoteRefs

Builds on Part 5 (#2228). When the session negotiated v2, `GetRemoteRefs` lists references through the `ls-refs` command instead of relying on the handshake advertisement (which v2 does not send).

### What changed

- `GetRemoteRefs` on the stream session maps its `GetRemoteRefsOptions` to a `packp.LsRefsArgs`: it always requests peeled tags and symref targets so `HEAD` resolves to its branch, passes through any ref-prefix filters, and requests unborn `HEAD` reporting when the server advertises `ls-refs=unborn`.
- The decoded `LsRefsOutput` is converted to `RemoteRefs`. An unborn `HEAD` (a symbolic `HEAD` whose target has no corresponding hash reference) is surfaced via `RemoteRefs.Unborn`.
- The v0/v1 path is unchanged: the server advertises every reference at handshake time and the options are ignored.

### Tests

End-to-end `ls-refs` against the server v2 path, ref-prefix filtering, and an unborn-`HEAD` case driven by a crafted advertisement.

### Reviewing just this part

https://github.com/go-git/go-git/compare/gitprotocol-v2-5-v2-handshake...gitprotocol-v2-6-v2-lsrefs

### Notes

Targets `main` (v6). Depends on Part 5 (#2228); merge that first.

---
_Recreated as a same-repo stacked PR (supersedes #2211)._

### DIFF
--- a/plumbing/transport/pack_stream.go
+++ b/plumbing/transport/pack_stream.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"io"
+	"slices"
 	"strings"
 
 	"github.com/go-git/go-git/v6/plumbing/protocol"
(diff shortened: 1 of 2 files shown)


pjbgf (approved):

PR #2230, opened 2026-06-25
plumbing: transport, fetch over Protocol v2 with caller-owned packfile (7/11)

aymanbagabas:
## Part 7: fetch over Protocol v2 with caller-owned packfile

Builds on Part 6 (#2229). When the session negotiated v2, `Fetch` runs the `fetch` command instead of the v0/v1 upload-request negotiation.

### What changed

- The `FetchRequest` is mapped to a `packp.FetchArgs` (wants, haves, ofs-delta, no-progress, include-tag, and filter or shallow only when the server advertises them).
- Negotiation follows git's `do_fetch_pack_v2`: each round sends the wants, the common commits acked so far, and a growing batch of haves, until the server reports `ready` (the packfile follows in the same response) or the client runs out of haves and sends `done`. This is a real multi-round loop with incremental have batching, reusing the existing `initialFlush`, `maxInVain`, and `nextFlush` pacing.
- Packfile streaming is owned by the transport, not the `FetchOutput` type. Once the server commits to a packfile, `FetchOutput.Decode` leaves the reader positioned at the first packfile pkt-line and `Fetch` demultiplexes the sideband-64k channel into storage, applying any shallow-info from the response.

### Tests

A clone against the server v2 path, a multi-round negotiation case, and a case that asserts the loop genuinely runs more than two rounds with incremental batching, plus rejection of filter and shallow when the server does not advertise them.

A file-transport integration test exercises the full stream-transport v2 client path (version negotiation via GIT_PROTOCOL, ls-refs, and fetch) end to end against the in-process upload-pack server. The stream transports share NewStreamSession, so this also covers git:// and ssh.

### Reviewing just this part

https://github.com/go-git/go-git/compare/gitprotocol-v2-6-v2-lsrefs...gitprotocol-v2-7-v2-fetch

### Notes

Targets `main` (v6). Depends on Part 6 (#2229); merge that first.

---
_Recreated as a same-repo stacked PR (supersedes #2212)._

### DIFF
--- a/plumbing/transport/file/file_integration_test.go
+++ b/plumbing/transport/file/file_integration_test.go
@@ -11,8 +11,10 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/go-git/go-git/v6/internal/transport/test"
+	"github.com/go-git/go-git/v6/plumbing"
 	"github.com/go-git/go-git/v6/plumbing/protocol"
 	"github.com/go-git/go-git/v6/plumbing/transport"
+	"github.com/go-git/go-git/v6/storage/memory"
 )
 
 func TestFileTransport_Integration(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pjbgf (approved):

PR #2232, opened 2026-06-25
plumbing: transport, Refactor server-side Protocol v2 onto packp types (9/11)

aymanbagabas:
## Part 9: refactor the server-side Protocol v2 onto packp types

Builds on Part 8 (#2231). The original v2 server (from the upload-pack v2 work) parsed and produced the wire by hand. This part separates the v2 advertisement from v0/v1 and routes the v2 commands through the shared packp types.

### What changed

- `AdvertiseRefs` is now v0/v1 only. A new `AdvertiseCapabilities` writes the v2 capability advertisement through `packp.CapabilityAdv` (no references; clients use ls-refs). `UploadPack` dispatches to it when the negotiated version is 2.
- The v2 advertisement does not emit the smart-HTTP `# service=...` line. git omits it for v2 even over HTTP (http-backend.c `get_info_refs`), so the response starts directly with the version packet. A test locks this in.
- The v2 command dispatcher peeks the command line, sets the matching argument decoder (`packp.LsRefsArgs` or `packp.FetchArgs`), and decodes the request through `packp.CommandRequest`, the same type the client encodes.
- The fetch response (acknowledgments, shallow-info, and the packfile section header) is emitted through `packp.FetchOutput`. The handler streams the packfile after the header, mirroring caller-owned streaming on the client.
- ls-refs lines are still encoded directly. A v2 HEAD line carries both a resolved object id and a symref-target, which a single `plumbing.Reference` (hash XOR symbolic) cannot represent, so `packp.LsRefsOutput` is not used on the server. A test locks in the resolved-HEAD-oid behavior.

### Tests

The existing server v2 suite passes against the refactored code, plus new tests asserting the v2 advertisement omits `# service=` and that HEAD is advertised with its resolved object id and symref-target. The git-CLI-over-v2 HTTP end-to-end test continues to exercise the server.

### Reviewing just this part

https://github.com/go-git/go-git/compare/gitprotocol-v2-9-http-shared...gitprotocol-v2-8-v2-server

### Notes

Targets `main` (v6). Depends on Part 8 (#2231); merge that first.

---
_Recreated as a same-repo stacked PR (supersedes #2213)._

### DIFF
--- a/plumbing/transport/serve.go
+++ b/plumbing/transport/serve.go
@@ -21,7 +21,13 @@ import (
 var ErrUpdateReference = errors.New("failed to update ref")
 
 // AdvertiseRefs is a server command that implements the reference
-// discovery phase of the Git transfer protocol.
+// discovery phase of the v0/v1 Git transfer protocol. Protocol v2 advertises
+// capabilities only, via [AdvertiseCapabilities]; the sole reason this function
+// accepts protocol.V2 is the receive-pack fallback: v2 has no push, so when a
+// client requests v2 for receive-pack git ignores it and serves a classic
+// advertisement (builtin/receive-pack.c), while http-backend still suppresses
+// the "# service=..." smart-reply line for the v2 request (http-backend.c
+// get_info_refs). Both behaviours are reproduced below.
 func AdvertiseRefs(
 	_ context.Context,
 	st storage.Storer,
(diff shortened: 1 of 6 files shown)


pjbgf (approved):

PR #2233, opened 2026-06-25
plumbing: transport, git archive over HTTP (10/11)

aymanbagabas:
## Part 10: git archive over HTTP

Builds on Part 9 (#2232). With the v2 HTTP session in place, this part adds git archive over HTTP, which git only supports over v2 (remote-curl.c `stateless_connect`).

### What changed

- The HTTP handshake discovers protocol support through the upload-pack info/refs endpoint when the command is upload-archive, and forces Protocol v2 for that discovery (git only attempts HTTP archive in v2 mode, so the caller does not need to set it). If the server does not speak v2, the handshake returns `ErrArchiveUnsupported`.
- The session that comes back implements `Archiver`. `smartPackSession.Archive` POSTs the archive request to the git-upload-archive endpoint and reads the ACK/NACK and the sideband-encoded archive stream from the response.
- The server backend routes POST git-upload-archive to the existing upload-archive handler.

This mirrors canonical git: `remote-curl.c` discovers via `git-upload-pack`, requires v2, and POSTs to `git-upload-archive` without resending the capability listing.

### Tests

An end-to-end test runs git archive over HTTP through the go-git client (without setting Protocol explicitly) against go-git's own backend and verifies the resulting tar.

### Reviewing just this part

https://github.com/go-git/go-git/compare/gitprotocol-v2-8-v2-server...gitprotocol-v2-10-http-archive

### Notes

Targets `main` (v6). Depends on Part 9 (#2232); merge that first.

---
_Recreated as a same-repo stacked PR (supersedes #2217)._

### DIFF
--- a/backend/http.go
+++ b/backend/http.go
@@ -32,6 +32,7 @@ var httpServices = []httpService{
 	{regexp.MustCompile(`(.*?)/objects/pack/pack-[0-9a-f]{40,64}\.idx$`), http.MethodGet, (*Backend).handleDumbIdxFile, ""},
 	{regexp.MustCompile("(.*?)/git-upload-pack$"), http.MethodPost, (*Backend).handleServiceRPC, transport.UploadPackService},
 	{regexp.MustCompile("(.*?)/git-receive-pack$"), http.MethodPost, (*Backend).handleServiceRPC, transport.ReceivePackService},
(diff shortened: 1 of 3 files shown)


pjbgf (approved):

PR #2234, opened 2026-06-25
plumbing: transport, Run the upload-pack suite across v0/v1/v2 (11/11)

aymanbagabas:
## Part 11: run the upload-pack suite across v0/v1/v2

Builds on Part 10 (#2233). The shared `UploadPackSuite` ran only the default (v0) path. This part parameterizes it by protocol version and runs it for v0, v1, and v2 on every pack transport (file, git, http, ssh), giving the v2 client end-to-end coverage against both go-git's own server (file) and the reference git server (http via git-http-backend, git:// daemon, ssh).

### What changed

- `UploadPackSuite` gains a `Protocol` field; each transport's entry point runs the suite once per version. Version-specific assertions are made version-aware, and the v2 case positively asserts the negotiated fetch/ls-refs capabilities, so a silent fallback to v0 fails the test rather than passing.
- The ssh test handler forwards the SSH session environment (`GIT_PROTOCOL`) to the spawned git, so ssh genuinely speaks the requested version.

### v2 client fixes surfaced by the suite

Running the suite against the reference git server surfaced two places where the v2 client diverged from the v0/v1 contract, fixed here:

- An ls-refs result with only a symbolic or unborn HEAD and no hash references is an empty repository, so `GetRemoteRefs` returns `ErrEmptyRemoteRepository` like v0/v1. `RemoteRefs.Unborn` is still surfaced when real references coexist.
- `FetchV2` short-circuits with `ErrNoChange` when every want is already local and no shallow change was requested, mirroring git's `everything_local`. The sentinel moves to `internal/transport` and `transport.ErrNoChange` aliases it.

### Reviewing just this part

https://github.com/go-git/go-git/compare/gitprotocol-v2-10-http-archive...gitprotocol-v2-11-suite-versions

### Notes

Targets `main` (v6). Depends on Part 10 (#2233); merge that first.

---
_Recreated as a same-repo stacked PR (supersedes #2218)._

### DIFF
--- a/internal/transport/test/common.go
+++ b/internal/transport/test/common.go
@@ -2,13 +2,54 @@ package test
 
 import (
 	"net"
+	"os/exec"
 	"path/filepath"
+	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/go-git/go-billy/v6"
 	fixtures "github.com/go-git/go-git-fixtures/v6"
 	"github.com/stretchr/testify/require"
+
+	"github.com/go-git/go-git/v6/plumbing/protocol"
 )
+
+// GitSupportsV2 reports whether the git binary in PATH supports protocol v2
+// (git >= 2.18). Transports backed by the reference git server use it to skip
+// the v2 suite run on older git, which would silently fall back to v0.
+func GitSupportsV2() bool {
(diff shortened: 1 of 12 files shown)


pjbgf (approved):

PR #2235, opened 2026-06-26
plumbing: object, add scheme-agnostic object verification API

pjbgf:
Introduce a plugin-based, scheme-agnostic API for verifying Git object signatures, replacing the OpenPGP-only `Commit.Verify`/`Tag.Verify`.

- Add plugin.Verifier, plugin.Verification and the plugin.ObjectVerifier() key (x/plugin), mirroring the existing Signer plugin.
- Commit.Verify / Tag.Verify now take a context and VerifyOption(s), using the verifier from WithVerifier or, by default, the registered ObjectVerifier. Core ships only the contract; concrete verifiers live off-tree in go-git/x.
- EncodeWithoutSignature returns a streaming io.Reader whose WriteTo feeds the verifier's hash directly, avoiding a full-payload buffer.
- Signature and SignatureSHA256 are now []byte, removing string<->[]byte conversions in the sign/verify paths.
- Multi-signature rejection moves to the OpenPGP verifier; core stays scheme-agnostic and no longer depends on ProtonMail/go-crypto.
- Move the gpg/git conformance suite to go-git/x.

This results in:
- By default, go-git has no default `Verifier` in-tree, which aligns with the approach taken for `Signer`.
- The `ObjectVerifier` implementations will be placed in `go-git/x` so that users of go-git that don't need the third-party dependencies won't have them in their projects.

#### Global verifier
```golang
func init() {
    plugin.Register(plugin.ObjectVerifier(), func() plugin.Verifier {
        v, _ := gpg.FromKeyRing(myKeyRing)
        return v
    })
}

v, err := commit.Verify(ctx)
```

#### Explicit verifier
```golang
verifier, _ := gpg.FromKeyRing(myKeyRing)
v, err := tag.Verify(ctx, object.WithVerifier(verifier))
```

Proposed implementation for gpg and ssh can be seen [here](https://github.com/go-git/x/pull/13).

Supersedes #1883.
Relates to https://github.com/go-git/go-git/issues/1869.


x/plugin/plugin_verifier.go
```diff
@@ -0,0 +1,59 @@
...
+// Verification is only returned when the signature is valid and trusted.
+type Verification struct {
+	// Signer is the verified identity in a scheme-neutral form, such as a
+	// key fingerprint, an SSH principal or an e-mail address.
+	Signer string
+	// Method is the signature scheme that was verified.
+	Method SignatureType
+	// Details carries scheme-specific data, for example an *openpgp.Entity
+	// for OpenPGP signatures. Callers type-assert on it based on Method.
+	Details any
```
pjbgf, line 38: @Bluebugs This should enable any additional scheme-specific information to be shared with users (as per separate thread).

Trust levels becomes an implementation detail of the verifier, as opposed to the API first class citizen. I haven't added them to the initial GPG implementation, but that is something we should be able to easily add as a follow-up.

cedric-appdirect: Nice. I haven't worked much on go-git lately as I don't want to overwhelm more the PR queue :-) But I might try to implement support for allowed_signers on top of your PR for go-git/x ( https://man7.org/linux/man-pages/man1/ssh-keygen.1.html#ALLOWED_SIGNERS ).

pjbgf:
A discussion point around the future of Tag/Commits and the introduction of "read-only" alternatives: 
```
BenchmarkReadCommitVerify/body=200-24         	 2338753	       459.2 ns/op	     304 B/op	       7 allocs/op
BenchmarkReadCommitVerify/body=4096-24        	  579508	      2097 ns/op	     304 B/op	       7 allocs/op
BenchmarkReadCommitVerify/body=262144-24      	   10000	    112914 ns/op	     304 B/op	       7 allocs/op
BenchmarkReadTagVerify/body=200-24            	 2180050	       582.5 ns/op	     376 B/op	       9 allocs/op
BenchmarkReadTagVerify/body=4096-24           	  497545	      2438 ns/op	     376 B/op	       9 allocs/op
BenchmarkReadTagVerify/body=262144-24         	   10000	    120918 ns/op	     377 B/op	       9 allocs/op


BenchmarkCommitVerify/body=200-24         	  662184	      1893 ns/op	 178.03 MB/s	    2156 B/op	      39 allocs/op
BenchmarkCommitVerify/body=4096-24        	  172166	      7959 ns/op	 531.82 MB/s	   20273 B/op	      41 allocs/op
BenchmarkCommitVerify/body=262144-24      	    3795	    354616 ns/op	 739.62 MB/s	 1079762 B/op	     110 allocs/op
BenchmarkTagVerify/body=200-24            	  516109	      1938 ns/op	 160.99 MB/s	    2316 B/op	      37 allocs/op
BenchmarkTagVerify/body=4096-24           	  175252	      6664 ns/op	 631.48 MB/s	   20017 B/op	      38 allocs/op
BenchmarkTagVerify/body=262144-24         	    3724	    377877 ns/op	 694.02 MB/s	 1079485 B/op	     107 allocs/op
```

tests/objectverify/tag_test.go
pjbgf, line 1: These tests are for the actual implementation, which are now moved to `go-git/x` as per https://github.com/go-git/x/pull/13.

Issue #2185, opened 2026-06-09
Git hooks


iamsoloma:
### Feature Description

Add the ability to catch pre-receive, update, and post-receive hooks from the server and determine further behavior with own golang code.

andrew:
Working on it here: https://github.com/go-git/go-git/pull/2236

pjbgf:
Closing as per #2236.

PR #2236, opened 2026-06-29
plumbing/transport: receive_pack, add ReceivePackHooks with PreReceive/PostReceive

andrew:
Refs #2185.

Adds an optional `Hooks` field to `ReceivePackRequest` so a server using `transport.ReceivePack` can run policy checks between unpack and ref update without reimplementing receive-pack on top of `packp.UpdateRequests` + `packfile.UpdateObjectStorage` + ref writes.

```go
type ReceivePackHooks struct {
    PreReceive  func(ctx context.Context, st storage.Storer, cmds []*packp.Command, progress io.Writer) error
    PostReceive func(ctx context.Context, st storage.Storer, cmds []*packp.Command)
}
```

`PreReceive` runs after the packfile is unpacked into the storer but before any ref is updated. Returning a non-nil error rejects every ref in the push with `err.Error()` as the report-status reason, mirroring git's pre-receive hook semantics. `PostReceive` runs after refs are updated.

`PreReceive` is given a progress writer. When the client negotiated `side-band` or `side-band-64k`, writes go to sideband band 2 so a hook can produce `remote:` lines for the client (rule names, principal lists, links to a policy page). Without sideband it is `io.Discard`.

This replaces the `// TODO: support hooks` marker. The per-ref `update` hook from #2185 is not included; the same shape can carry it later.

Motivation: this is what silo (a small forge that runs gittuf verification on the receive path) needs to drop its private receive-pack adapter. Branch protection, signed-commit checks, and any other policy that must inspect proposed updates after objects are available but before refs move now sits on `transport.ReceivePack` directly.

Tests cover nil hooks, allow path, reject path with `ng <ref> <reason>` in the report-status, that `PostReceive` does not run when `PreReceive` rejects, and that `PreReceive` progress writes are demultiplexable on sideband band 2.

AI disclosure: drafting and the test suite were assisted by Claude Opus 4.7. Every line is owned by me and was reviewed against the existing `transport.ReceivePack` flow before submission.

### DIFF
--- a/plumbing/transport/receive_pack.go
+++ b/plumbing/transport/receive_pack.go
@@ -23,10 +23,38 @@ type ReceivePackRequest struct {
 	GitProtocol   string
 	AdvertiseRefs bool
 	StatelessRPC  bool
+
+	// Hooks, if set, are invoked between packfile unpack and ref update
+	// (PreReceive) and after ref update (PostReceive). A non-nil PreReceive
+	// error rejects every ref in the push with that error as the
+	// report-status reason; refs are not updated and PostReceive is not run.
+	Hooks *ReceivePackHooks
+}
+
+// ReceivePackHooks holds server-side callbacks for ReceivePack.
+//
+// These are the in-process equivalent of git's pre-receive and post-receive
+// hooks. They run after the packfile has been unpacked into st but before
+// (PreReceive) and after (PostReceive) ref updates, so a server can enforce
+// branch protection, signed-commit checks, or other policy without
+// reimplementing receive-pack.
+type ReceivePackHooks struct {
+	// PreReceive runs after the packfile is unpacked into st but before any
+	// ref is updated. cmds are the proposed updates. progress writes to the
+	// sideband progress channel (band 2) when the client negotiated
+	// side-band/side-band-64k, or is discarded otherwise; write human-readable
+	// lines for the client's `remote:` output. Returning a non-nil error
+	// refuses every ref with err.Error() as the report-status reason.
+	PreReceive func(ctx context.Context, st storage.Storer, cmds []*packp.Command, progress io.Writer) error
+
+	// PostReceive runs after refs are updated. cmds are the same commands
+	// passed to PreReceive. The hook itself must handle any failure it cares
+	// about; nothing is surfaced to the client and the function does not
+	// return an error.
+	PostReceive func(ctx context.Context, st storage.Storer, cmds []*packp.Command)
 }
 
 // ReceivePack is a server command that serves the receive-pack service.
-// TODO: support hooks
 func ReceivePack(
 	ctx context.Context,
 	st storage.Storer,
@@ -128,13 +156,18 @@ func ReceivePack(
 	var (
 		useSideband bool
 		writer      io.Writer = w
+		progress              = io.Writer(io.Discard)
 	)
 	if !caps.Supports(capability.NoProgress) {
+		var mux *sideband.Muxer
 		if caps.Supports(capability.Sideband64k) {
-			writer = sideband.NewMuxer(sideband.Sideband64k, w)
-			useSideband = true
+			mux = sideband.NewMuxer(sideband.Sideband64k, w)
 		} else if caps.Supports(capability.Sideband) {
-			writer = sideband.NewMuxer(sideband.Sideband, w)
+			mux = sideband.NewMuxer(sideband.Sideband, w)
+		}
+		if mux != nil {
+			writer = mux
+			progress = sidebandProgress{mux}
 			useSideband = true
 		}
 	}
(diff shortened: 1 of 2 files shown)


plumbing/transport/receive_pack.go
```diff
@@ -23,10 +23,38 @@ type ReceivePackRequest struct {
 	GitProtocol   string
 	AdvertiseRefs bool
 	StatelessRPC  bool
+
+	// Hooks, if set, are invoked between packfile unpack and ref update
+	// (PreReceive) and after ref update (PostReceive). A non-nil PreReceive
+	// error rejects every ref in the push with that error as the
+	// report-status reason; refs are not updated and PostReceive is not run.
+	Hooks *ReceivePackHooks
```
pjbgf:
```suggestion
	Hooks ReceivePackHooks
```

plumbing/transport/receive_pack.go
```diff
@@ -23,10 +23,38 @@ type ReceivePackRequest struct {
 	GitProtocol   string
 	AdvertiseRefs bool
 	StatelessRPC  bool
+
+	// Hooks, if set, are invoked between packfile unpack and ref update
+	// (PreReceive) and after ref update (PostReceive). A non-nil PreReceive
+	// error rejects every ref in the push with that error as the
+	// report-status reason; refs are not updated and PostReceive is not run.
```
pjbgf: Let's provide docs on the hooks themselves.
```suggestion
	// Hooks are optional server-side callbacks. The zero value installs none.
```

plumbing/transport/receive_pack.go
```diff
@@ -23,10 +23,38 @@ type ReceivePackRequest struct {
...
+}
+
+// ReceivePackHooks holds server-side callbacks for ReceivePack.
+//
+// These are the in-process equivalent of git's pre-receive and post-receive
+// hooks. They run after the packfile has been unpacked into st but before
+// (PreReceive) and after (PostReceive) ref updates, so a server can enforce
+// branch protection, signed-commit checks, or other policy without
+// reimplementing receive-pack.
+type ReceivePackHooks struct {
```
pjbgf, line 38: In order to keep this type stable, let's define a struct for each:

```golang
	PreReceive func(context.Context, *PreReceiveInfo) error

	PostReceive func(context.Context, *PostReceiveInfo) error
```

This way, if we need to add new fields over time it won't break existing hooks. The types would then be:

```golang
// PreReceiveInfo carries the inputs to a PreReceive hook.
type PreReceiveInfo struct {
	// Storer reads the proposed new state: the objects from this push are
	// already present alongside the existing repository.
	Storer storage.Storer
	// Commands are the proposed ref updates. Treat as read-only.
	Commands []*packp.Command
	// PushOptions are the client's push options (empty if none).
	PushOptions []string
	// Progress writes to the client's sideband progress channel (band 2) when
	// negotiated, or is io.Discard otherwise. Valid only during the call.
	Progress io.Writer
}

// PostReceiveInfo carries the inputs to a PostReceive hook.
type PostReceiveInfo struct {
	// Storer reads the committed repository state.
	Storer storage.Storer
	// Commands are the ref updates that were applied successfully. Refs whose
	// update failed are omitted. Treat as read-only.
	Commands []*packp.Command
	// PushOptions are the client's push options (empty if none).
	PushOptions []string
	// Progress writes to the client's sideband progress channel (band 2) when
	// negotiated, or is io.Discard otherwise. Valid only during the call.
	Progress io.Writer
}

```

plumbing/transport/receive_pack.go
```diff
@@ -146,10 +179,37 @@ func ReceivePack(
 		return res
 	}
 
+	if opts.Hooks != nil && opts.Hooks.PreReceive != nil {
```
pjbgf: With the change above, `opts.Hooks` is no longer a ref:

```suggestion
	if opts.Hooks.PreReceive != nil {
```

plumbing/transport/receive_pack.go
```diff
@@ -146,10 +179,37 @@ func ReceivePack(
...
+			}
+			return hookErr
+		}
+	}
+
 	var firstErr error
 	cmdStatus := make(map[plumbing.ReferenceName]error)
 	updateReferences(st, updreq, cmdStatus, &firstErr)
 
+	if opts.Hooks != nil && opts.Hooks.PostReceive != nil {
```
pjbgf: With the change above, `opts.Hooks` is no longer a ref:

```suggestion
	if opts.Hooks.PostReceive != nil {
```

plumbing/transport/receive_pack_test.go
```diff
@@ -0,0 +1,236 @@
...
+	require.NoError(t, err)
+
+	assert.Contains(t, out.String(), "unpack ok")
+	assert.Contains(t, out.String(), "ok refs/heads/main")
+
+	_, err = st.Reference(ref)
+	assert.ErrorIs(t, err, plumbing.ErrReferenceNotFound)
+}
+
+func TestReceivePackPreReceiveAllowsUpdate(t *testing.T) {
```
pjbgf, line 82: Let's add another test case whereby multiple commands are being sent and they are partially allowed/rejected (e.g. one ref allowed and one rejected).

pjbgf (changes requested):
@andrew thanks for working on this. Above are some changes that we should do to make the API more stable for the long run.

andrew:
Thanks for the feedback @pjbgf, all applied in 1f2cde8e. Hooks is now a value, both callbacks take `*PreReceiveInfo` / `*PostReceiveInfo`, push options are threaded through, PostReceive only sees the refs that applied, and there's a partial-success test.

pjbgf (approved):
@andrew thanks for working on this. 🙇

PR #2237, opened 2026-06-30
plumbing: default to wire protocol v2 and close client/server conformance gaps

### LINKED ISSUE #264: support for git wire protocol version 2
Does go-git supports git wire protocol version 2? 

I try to use go-git with the recent git client (after 2.26) which uses git wire protocol version 2 by default, but it seems that go-git supports only version 1. Am I right?

Do you plan to add support for version 2 as well?

pjbgf:
Builds on the Protocol v2 series (#2228–#2234) to make v2 the **default** wire
protocol and to close the conformance gaps that only surfaced once the v2 paths
were exercised end-to-end. Validated against C Git v2.54 (`fetch-pack.c`,
`upload-pack.c`, `ls-refs.c`, `serve.c`).

### Default to v2 (client)

Wiring so the high-level API actually negotiates v2, with safe fallback to
v0/v1 when the server doesn't support it:

- `config`: `DefaultProtocolVersion` is now v2; marshalling omits the
  `[protocol]` section when it equals the default.
- `git`: `Remote.transportProtocol()` reads `protocol.version` and sets it on
  the handshake for fetch/ls-remote (push stays v0/v1 — v2 has no push).
- **exact-SHA1 fetch**: v2 accepts `want <oid>` without the server advertising
  `allow-*-sha1-in-want`, so the session surfaces those caps for
  `isSupportedRefSpec` (never sent on the wire).
- **object-format reconciliation**: a v2 fetch now adopts the server's hash
  algorithm on a fresh clone and errors with `mismatched algorithms` on a real
  mismatch, instead of failing later as an opaque checksum error.
- **detached HEAD**: a detached remote HEAD over v2 ls-refs resolves to a
  symbolic HEAD on clone (the hash→branch heuristic, shared with v0/v1).
- ref discovery is scoped with `ls-refs` `ref-prefix` hints derived from the
  refspecs (incl. `refs/heads/` for a HEAD source), matching canonical git.

### Server fetch conformance

- **stateful multi-round negotiation**: the server no longer closes the
  connection after an acknowledgments-only round, so a stream/SSH/file client
  can refine its haves across rounds (previously every fetch with common
  history failed with `read/write on closed pipe`).
- **shallow deepen**: implements `deepen-since`, `deepen-not`, and
  `deepen-relative` (rejecting `deepen` + since/not as upstream does), and
  bounds **deepening an already-shallow clone** via a two-view object
  difference (never under-sends).
- **include-tag**: auto-follows annotated tags whose target is in the pack.
- `no-progress` is honored by construction (the v2 server emits no progress
  band); locked with a test.

### Spec conformance (packp)

- `FetchOutput.Decode` is now a strict single-pass state machine: rejects
  repeated/out-of-order sections, unknown section headers, unrecognized lines,
  and terminators that contradict the response shape (typed
  `MalformedResponseError`).
- v2 `agent` capability is sent only when the server advertised it.
- `acknowledgments` ordering follows `send_acks` (NAK suppressed once ready).
- `ls-refs` `ref-prefix` arguments are validated before encoding (no empty /
  whitespace / control characters).
- the advertised protocol version is single-sourced from the handshake rather
  than parsed independently in `AdvRefs`.

### Hardening & tests

- defensive caps on decoder sections (ref-prefix → `TOO_MANY_PREFIXES`;
  want/have/shallow/ACK/wanted-refs bounded) against hostile input.
- new fuzz targets for `CommandRequest`, `CapabilityAdv`, `LsRefsArgs/Output`,
  and `parseLsRefsLine`, plus extended `FetchOutput` corpus.
- `internal/gitcli` helper to skip v2 integration tests on Git < 2.18.

### Compatibility

No breaking changes for existing callers: clients that can't reach v2 fall back
to v0/v1, and push is unaffected. The only behavioral change is that fetch /
ls-remote now prefer v2 by default (overridable via `protocol.version`).

Supersedes #2204.
Closes #264.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -31,7 +31,7 @@ const (
 	// should be marshalled or not.
 	// Note that this does not need to align with the default protocol
 	// version from plumbing/protocol.
-	DefaultProtocolVersion = protocol.V0 // go-git only supports V0 at the moment
+	DefaultProtocolVersion = protocol.V2
 )
 
 // ConfigStorer is a generic storage of Config object.
@@ -991,10 +991,23 @@ func (c *Config) marshalURLs() {
 }
 
 func (c *Config) marshalProtocol() {
-	// Only marshal protocol section if a version was set.
 	if c.Protocol.Version != DefaultProtocolVersion {
 		s := c.Raw.Section(protocolSection)
 		s.SetOption(versionKey, c.Protocol.Version.String())
+		return
+	}
+
+	// The struct holds the default version. Clear any stale protocol.version
+	// left over in the raw config so switching back to the default persists,
+	// and drop the section if it becomes empty. Guard on HasSection so a
+	// non-default round-trip does not introduce an empty [protocol].
+	if !c.Raw.HasSection(protocolSection) {
+		return
+	}
+	s := c.Raw.Section(protocolSection)
+	s.RemoveOption(versionKey)
+	if len(s.Options) == 0 && len(s.Subsections) == 0 {
+		c.Raw.RemoveSection(protocolSection)
 	}
 }
 
--- a/plumbing/protocol/packp/fetch.go
+++ b/plumbing/protocol/packp/fetch.go
@@ -12,6 +12,27 @@ import (
 	"github.com/go-git/go-git/v6/plumbing/format/pktline"
 )
 
+// maxSectionLines bounds how many entries a single fetch section (want, have,
+// shallow, ACK, wanted-ref, ...) may contribute on decode. It is a defensive
+// backstop against a hostile peer streaming unbounded lines into an in-memory
+// slice; it sits far above any legitimate request or response (a single
+// negotiation round carries at most a flush-batch of haves, and real repos have
+// far fewer than four million refs). It is a var only so tests can lower it.
+var maxSectionLines = 1 << 22
+
+// MalformedResponseError reports a server response that violates the
+// gitprotocol-v2 grammar: a malformed pkt-line, an unrecognized line within a
+// section, an unexpected/repeated/out-of-order section, or a section terminator
+// that contradicts the response shape. It mirrors the situations where upstream
+// fetch-pack.c calls die() on the response.
+type MalformedResponseError struct {
+	Reason string
+}
+
+func (e *MalformedResponseError) Error() string {
+	return "malformed v2 fetch response: " + e.Reason
+}
+
 // FetchArgs represents the arguments for the v2 fetch command.
 type FetchArgs struct {
 	// Wants is the list of object IDs the client wants.
@@ -61,7 +82,7 @@ func (r *FetchArgs) Encode(w io.Writer) error {
 	plumbing.HashesSort(wants)
 	for _, h := range wants {
 		if _, err := pktline.Writef(w, "want %s\n", h); err != nil {
-			return fmt.Errorf("encoding want %q: %s", h, err)
+			return fmt.Errorf("encoding want %q: %w", h, err)
 		}
 	}
 
@@ -69,37 +90,37 @@ func (r *FetchArgs) Encode(w io.Writer) error {
 	plumbing.HashesSort(haves)
 	for _, h := range haves {
 		if _, err := pktline.Writef(w, "have %s\n", h); err != nil {
-			return fmt.Errorf("encoding have %q: %s", h, err)
+			return fmt.Errorf("encoding have %q: %w", h, err)
 		}
 	}
 
 	if r.Done {
 		if _, err := pktline.WriteString(w, "done\n"); err != nil {
-			return fmt.Errorf("encoding done: %s", err)
+			return fmt.Errorf("encoding done: %w", err)
 		}
 	}
 
 	if r.ThinPack {
 		if _, err := pktline.WriteString(w, "thin-pack\n"); err != nil {
-			return fmt.Errorf("encoding thin-pack: %s", err)
+			return fmt.Errorf("encoding thin-pack: %w", err)
 		}
 	}
 
 	if r.NoProgress {
 		if _, err := pktline.WriteString(w, "no-progress\n"); err != nil {
-			return fmt.Errorf("encoding no-progress: %s", err)
+			return fmt.Errorf("encoding no-progress: %w", err)
 		}
 	}
 
 	if r.IncludeTag {
 		if _, err := pktline.WriteString(w, "include-tag\n"); err != nil {
-			return fmt.Errorf("encoding include-tag: %s", err)
+			return fmt.Errorf("encoding include-tag: %w", err)
 		}
 	}
 
 	if r.OFSDelta {
 		if _, err := pktline.WriteString(w, "ofs-delta\n"); err != nil {
-			return fmt.Errorf("encoding ofs-delta: %s", err)
+			return fmt.Errorf("encoding ofs-delta: %w", err)
 		}
 	}
 
@@ -107,13 +128,13 @@ func (r *FetchArgs) Encode(w io.Writer) error {
 	plumbing.HashesSort(shallows)
 	for _, h := range shallows {
 		if _, err := pktline.Writef(w, "shallow %s\n", h); err != nil {
-			return fmt.Errorf("encoding shallow %q: %s", h, err)
+			return fmt.Errorf("encoding shallow %q: %w", h, err)
 		}
 	}
 
 	if r.Deepen > 0 {
 		if _, err := pktline.Writef(w, "deepen %d\n", r.Deepen); err != nil {
-			return fmt.Errorf("encoding deepen %d: %s", r.Deepen, err)
+			return fmt.Errorf("encoding deepen %d: %w", r.Deepen, err)
 		}
 	}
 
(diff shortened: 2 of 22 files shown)


remote.go
```diff
@@ -335,6 +336,61 @@ func (r *Remote) FetchContext(ctx context.Context, o *FetchOptions) error {
...
+		if i := strings.IndexByte(src, '*'); i >= 0 {
+			src = src[:i]
+		}
+		// A leading wildcard trims to an empty prefix, which would emit an
+		// invalid "ref-prefix " argument. Treat it as non-convertible and
+		// request the full advertisement instead of under-scoping it.
+		if src == "" {
+			return nil
+		}
+		prefixes = append(prefixes, src)
```
aymanbagabas: **🔴 HIGH — unqualified refspecs under-scope the v2 `ref-prefix` (compat regression)**

`fetchRefPrefixes` handles `HEAD` (adds `refs/heads/`) and leading wildcards, but a plain short-name refspec like `master:foo` still emits only `ref-prefix master`. A spec-compliant v2 server matches `ref-prefix` against the **full** refname, so `refs/heads/master` is omitted from the advertisement and the fetch fails with `ErrRemoteRefNotFound`.

Because this PR makes v2 the default, this is a user-visible regression for existing short-name refspecs. Upstream git expands short names through `expand_ref_prefix` (all of `RefRevParseRules`).

**Fix:** for an unqualified, non-wildcard source, expand it via `plumbing.RefRevParseRules` (emit every candidate prefix), or return `nil` to request the full advertisement. Add a regression test using a short-name refspec against a strict v2 server.

plumbing/transport/upload_pack.go
```diff
@@ -616,49 +625,126 @@ func serveFetchV2(_ context.Context, st storage.Storer, w io.WriteCloser, args *
...
+	// newBoundary is the requested shallow boundary, or nil for a non-shallow
+	// fetch. It is reported to the client and bounds the packfile.
+	var newBoundary []plumbing.Hash
+	if depth > 0 || revList {
 		var shupd packp.ShallowUpdate
-		if err := getShallowCommits(st, wants, depth, &shupd); err != nil {
+		if revList {
+			err = getShallowCommitsByRevList(st, wants, since, notTips, &shupd)
+		} else {
+			err = getShallowCommits(st, wants, depth, &shupd)
```
aymanbagabas: **🔴 HIGH — `deepen-relative` is ignored for already-shallow clients**

The new shallow boundary is always computed from `wants` via `getShallowCommits(st, wants, depth, …)`, so for a client that is already shallow, `want HEAD + have HEAD + shallow HEAD + deepen 1 + deepen-relative` computes the same boundary as the existing tip and returns an empty/no-op deepen response. This breaks `git fetch --deepen=N` semantics.

Verified against upstream `upload-pack.c`, which passes `data->deepen_relative` into `get_shallow_commits` so the depth is counted **relative to the client's shallow commits**, not the wanted tips.

**Fix:** when `args.DeepenRelative && len(clientShallows) > 0`, compute the boundary from `clientShallows` with relative-depth semantics; otherwise reject the combination until implemented. Add a test for an existing shallow clone using `deepen-relative`.

_Single-model finding — worth a human cross-check against `git/git`._

plumbing/transport/upload_pack.go
```diff
@@ -616,49 +625,126 @@ func serveFetchV2(_ context.Context, st storage.Storer, w io.WriteCloser, args *
...
+		// boundary, so compute two views and send their difference:
+		//   newView    = objects reachable from the wants, grafted at the new
+		//                boundary (the client's deepened view).
+		//   clientView = objects the client already has, reachable from its haves
+		//                grafted at its existing shallow boundary.
+		// newView \ clientView is exactly what the client is missing. It never
+		// omits a needed object; at worst it re-sends one the client has, which
+		// is harmless. This is what bounds a deepen of an already-shallow clone.
+		boundary := newBoundary
+		if boundary == nil {
```
aymanbagabas: **🔴 HIGH — full unshallow transition is silently dropped**

`newBoundary == nil` is overloaded: it means both *"no deepen requested"* and *"deepen requested but the new boundary is empty"* (i.e. the deepen reaches full history). For an already-shallow client requesting a depth large enough to unshallow, this falls back to `boundary = clientShallows`, sends no ancestors, and omits the required `unshallow` lines. Upstream git still calls `send_unshallow` when the deepen removes the shallow boundary.

**Fix:** track an explicit `shallowRequested`/`newBoundaryComputed` bool separately from `len(newBoundary)`. For a requested deepen with an empty boundary, use an empty boundary for `newView`, emit `shallow-info` with the appropriate `unshallow` lines, and do not fall back to `clientShallows`.

_Single-model finding — verify against upstream `send_unshallow`._

plumbing/protocol/packp/fetch.go
aymanbagabas, line 1: **🟠 MEDIUM — `decodePackfileURIs` has no `maxSectionLines` cap (client-side DoS)**

Every sibling section this PR added is bounded (e.g. wanted-refs at line ~700 returns `MalformedResponseError` past `maxSectionLines`), but `packfile-uris` appends unbounded. A hostile server can stream unlimited lines here and exhaust client memory.

**Fix:** apply the same `maxSectionLines` cap used by the other sections, and add a fuzz seed / test exercising a large packfile-uris section.

aymanbagabas, line 1: **🟠 MEDIUM — `deepen-not` is unbounded in `FetchArgs.Decode` (server-side DoS)**

Server-side mirror of the packfile-uris issue: `want`/`have`/`shallow` are bounded, but `deepen-not` appends without a `maxSectionLines` cap, letting a hostile client exhaust server memory.

**Fix:** bound this section like its siblings and add a fuzz seed covering many `deepen-not` lines.

plumbing/transport/negotiate.go
```diff
@@ -324,3 +324,66 @@ func readShallows(
...
+	}
+
+	var serverFormat config.ObjectFormat
+	switch v := config.ObjectFormat(advertised[0]); v {
+	case config.SHA1, config.SHA256:
+		serverFormat = v
+	case config.UnsetObjectFormat:
+		// An empty value carries no algorithm; treat it as if the server did
+		// not advertise object-format at all.
+		return nil
```
aymanbagabas, line 367: **🟠 MEDIUM — sha256 client guard bypassed when the server advertises an empty object-format**

`ReconcileObjectFormatV2` returns `nil` unconditionally for an empty advertised `object-format`. This contradicts the sibling `len(advertised)==0` branch and the intent of commit a3e6202: a sha256 client can slip past the guard when the server advertises the key with an empty value.

**Fix:** apply the same sha1-only guard used when object-format is entirely absent (reject a sha256 client), rather than returning `nil`.

config/config.go
aymanbagabas, line 1: **🟠 MEDIUM — `marshalProtocol` leaves a stale `protocol.version` in `c.Raw`**

The guard only *writes* the key when the version differs from the default, but it never *clears* it. If the raw config already contains `protocol.version = 2` and the struct is set back to the default, the stale key survives marshalling, so switching back to the default can't be persisted.

**Fix:** in the `else` branch, remove the `version` option (and the `protocol` section if it becomes empty) so the marshalled config reflects the default.

plumbing/transport/http/handshake.go
aymanbagabas, line 1: **🟠 MEDIUM — HTTP v2 response body leaks on decode-error paths**

When `resp.Decode(r)` fails here (and similarly in the v2 fetch path ~line 362), the function returns before the drain/close block at lines 295–298, leaking the response body. This violates the repo's own resource-cleanup rule (every `Open()`/body should have a matching `defer Close()`).

**Fix:** `defer` the drain-and-close of `r.resp.Body` immediately after it's set, so all error paths release it.

internal/gitcli/version.go
```diff
@@ -0,0 +1,53 @@
...
+	"strconv"
+	"strings"
+	"testing"
+)
+
+// SkipIfProtocolV2Unsupported skips the test unless a git new enough to
+// advertise wire protocol v2 is available on PATH. Protocol v2 was introduced
+// in Git 2.18; older servers (the CI matrix still exercises Git 2.11) only
+// speak v0/v1, so tests asserting v2 negotiation cannot pass against them.
+func SkipIfProtocolV2Unsupported(t testing.TB) {
```
aymanbagabas: **🟠 MEDIUM — `SkipIfProtocolV2Unsupported` is dead code**

This helper has zero callers, so its CI-guarding purpose (skipping v2 tests against Git < 2.18 in the 2.11 matrix) is unrealized — the v2 tests it was meant to guard will still run and fail on old git.

**Fix:** either wire it into the gitcli v2 tests that require Git ≥ 2.18, or drop it. The version parsing itself is correct.

plumbing/protocol/packp/fetch.go
aymanbagabas, line 1: **🟡 LOW — truncated response silently returns `nil` instead of an error**

When `io.EOF` is hit while `expectPackfile == true` (a metadata section already committed the response to the packfile shape), the loop returns `nil`. A premature EOF at this point should be a `MalformedResponseError{Reason: "expected packfile section"}`, matching the flush/response-end handling a few lines below.

**Fix:** if `expectPackfile` is set when EOF is reached, return the malformed-response error rather than `nil`.

plumbing/transport/pack_stream.go
aymanbagabas, line 1: **🟡 LOW — stale doc comment: agent is no longer *always* advertised**

The comment says the client "always advertises the agent," but the PR now gates the agent capability on the server having advertised it (see `internal/transport/v2.go:56`). Update the comment to reflect the conditional behavior.

plumbing/protocol/packp/fuzz_test.go
aymanbagabas, line 1: **🟡 LOW — fuzz seeds don't exercise the unbounded sections**

The reframed fuzz seeds are correctly scoped and OSS-Fuzz-safe, but none exercise a large `packfile-uris` or `deepen-not` section — exactly the two decoders flagged as missing `maxSectionLines` bounds. Adding seeds with many lines in those sections would give the new bounds direct fuzz coverage.

pjbgf:
All the points above are now fixed and the commits collapsed into 3. PTAL @aymanbagabas

aymanbagabas (approved):
All 12 review comments from the earlier pass have been addressed in the consolidated branch — including the three HIGH shallow/ref-prefix correctness issues (verified against upstream git), the DoS-bound and resource-leak fixes, and the config/marshal and fuzz-seed follow-ups. Nice work. LGTM.

PR #2238, opened 2026-07-02
plumbing/format/diff: use git's funcname heuristic for hunk context

### LINKED ISSUE #2193: ctxPrefix of hunk is wrong
### Bug Description

In git cli when I have a c file with this func and add a simple comment, the context line contains the line where the func is defined:
like `@@ -2,0 +3 @@ int main(int argc, char *argv[])`

example output:
```bash
$ git diff -U0
diff --git a/main.c b/main.c
index f9faa8d..3675fff 100644
--- a/main.c
+++ b/main.c
@@ -2,0 +3 @@ int main(int argc, char *argv[])
+       // hello
```

In go-git implementation the `ctxPrefix`, it just uses one line above the change -> `ctxPrefix` is just `{` (when contextLines is set to 0)
example output:
```bash
$ go run . -U0
diff --git a/main.c b/main.c
index f9faa8d..3675fff 100644
--- a/main.c
+++ b/main.c
@@ -2,0 +3 @@ {
+       // hello
```

### go-git Version

v6

### Steps to Reproduce

1. add `main.c`
```c
int main(int argc, char *argv[])
{
  return EXIT_SUCCESS;
}
```
2. commit `main.c`

3. add comment inside main into `main.c`
```c
int main(int argc, char *argv[])
{
  // hello
  return EXIT_SUCCESS;
}
```
4. commit `main.c`

5. diff both commits -> see `int main(int argc, char *argv[])` as ctxPrefix
```bash
$ git diff f9faa8d..3675fff -U0
diff --git a/main.c b/main.c
index f9faa8d..3675fff 100644
--- a/main.c
+++ b/main.c
@@ -2,0 +3 @@ int main(int argc, char *argv[])
+       // hello
```

6. some go code:
```go
	repo, _ := git.Clone(....)
	head, _ := repo.Head()
	commit, _ := repo.CommitObject(head.Hash())

	parent, _ := commit.Parent(0)
	patch, _ := parent.Patch(commit)

	diffEncoder := diff.NewUnifiedEncoder(os.Stdout, 0) // diff without context lines
	diffEncoder.SetColor(diff.NewColorConfig())      // use default colors
	if err := diffEncoder.Encode(patch); err != nil {
		panic(err)
	}
```

7. see that the diff shows the wrong ctxPrefix

### Additional Information

_No response_

SillyZir:
Fixes #2193.

go-git's unified diff used the line immediately before a change as the hunk header context (`@@ ... @@ {`), whereas git shows the enclosing function or section line (`@@ ... @@ int main(int argc, char *argv[])`).

This searches backward for a line matching git's default funcname heuristic (first character a letter, `_`, or `$`) and carries the last match across hunks, matching git's output.

Affected package tests pass locally.

## AI Assistance Disclosure

Per go-git's [AI Contribution Policy](https://github.com/go-git/go-git/blob/master/AI_POLICY.md): this patch was developed with AI assistance. I reviewed and tested the change before submission (affected package tests pass locally, and the output was verified against reference `git` behaviour), and I take responsibility for it. The commit carries an `Assisted-by:` trailer.

### DIFF
--- a/plumbing/format/diff/unified_encoder.go
+++ b/plumbing/format/diff/unified_encoder.go
@@ -179,6 +179,7 @@ type hunksGenerator struct {
 	current                     *hunk
 	hunks                       []*hunk
 	beforeContext, afterContext []string
+	funcName                    string
 }
 
 func newHunksGenerator(chunks []Chunk, ctxLines int) *hunksGenerator {
@@ -228,15 +229,14 @@ func (g *hunksGenerator) processHunk(i int, op Operation) {
 		return
 	}
 
-	var ctxPrefix string
 	linesBefore := len(g.beforeContext)
 	if linesBefore > g.ctxLines {
-		ctxPrefix = g.beforeContext[linesBefore-g.ctxLines-1]
+		g.funcName = g.findFuncName(linesBefore-g.ctxLines-1, g.funcName)
 		g.beforeContext = g.beforeContext[linesBefore-g.ctxLines:]
 		linesBefore = g.ctxLines
 	}
 
-	g.current = &hunk{ctxPrefix: strings.TrimSuffix(ctxPrefix, "\n")}
+	g.current = &hunk{ctxPrefix: strings.TrimSuffix(g.funcName, "\n")}
 	g.current.AddOp(Equal, g.beforeContext...)
 
 	switch op {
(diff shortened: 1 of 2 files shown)


SillyZir:
Rebased onto `main` — the stray merge commit is gone — and added a regression test.

The existing encoder fixtures all use single-letter context lines, and every one of those satisfies the funcname heuristic, so the suite passed identically with and without this change. The new fixture deletes two lines from a C function whose opening brace sits on its own line: without the patch the encoder emits `@@ -3,5 +3,4 @@ {` and `@@ -9,5 +8,4 @@ <tab>int e = 6;`; with it both hunks read `int main(void)`, which is what `git diff -U2` produces on the same content. The second hunk is preceded only by indented statements, so it also covers the carry-over of the previous match.

CI hasn't run on this at all yet — as a first-time contributor it looks like it needs a maintainer to approve the workflow run. Happy to adjust anything if you'd prefer a different approach.

PR #2243, opened 2026-07-07
plumbing: object, include submodules in patches

### LINKED ISSUE #1387: object.Patch is incorrect for commits which contains submodules
Code:
```go
package main

import (
	"fmt"

	"github.com/go-git/go-git/v5"
	"github.com/go-git/go-git/v5/plumbing"
	"github.com/go-git/go-git/v5/plumbing/object"
)

func main() {
	repo, _ := git.Clone(memory.NewStorage(), nil, &git.CloneOptions{
		URL: "https://github.com/git-fixtures/submodule.git",
	})
	commitHash := plumbing.NewHash("c7431b5bc9d45fb64a87d4a895ce3d1073c898d2")
	commit, _ := repo.CommitObject(commitHash)
	commitTree, _ := commit.Tree()
	emptyTree := &object.Tree{}
	patch, _ := emptyTree.Patch(commitTree)
	fmt.Println(patch.String())
}

```
Actual:
```diff
diff --git a/.gitmodules b/.gitmodules
new file mode 100644
index 0000000000000000000000000000000000000000..97b013ecd2cc7f572960509f659d8068798d59ca
--- /dev/null
+++ b/.gitmodules
@@ -0,0 +1,3 @@
+[submodule "basic"]
+       path = basic
+       url = https://github.com/git-fixtures/basic.git
```
Expected (as via `git`):
```diff
diff --git a/.gitmodules b/.gitmodules
new file mode 100644
index 0000000..97b013e
--- /dev/null
+++ b/.gitmodules
@@ -0,0 +1,3 @@
+[submodule "basic"]
+       path = basic
+       url = https://github.com/git-fixtures/basic.git
diff --git a/basic b/basic
new file mode 160000
index 0000000..6ecf0ef
--- /dev/null
+++ b/basic
@@ -0,0 +1 @@
+Subproject commit 6ecf0ef2c2dffb796033e5a02219af86ec6584e5
```

upuddu:
`Change.Patch` / `Tree.Patch` silently drop submodules. When a commit adds, removes or updates a gitlink, the resulting patch contains no entry for it at all — as reported in #1387, diffing an empty tree against a tree with submodules omits the submodule sections entirely.

The cause is that submodules aren't blob objects, so `Change.Files()` returns `nil` for them and the changed entry is treated as empty when the file-patch header is built, so the encoder skips it.

Git renders a submodule in a diff as a single `Subproject commit <hash>` line. This builds that content synthetically for gitlink changes (add / remove / commit update, reusing the existing line diff) and stops treating submodule entries as empty in the header, so the `diff --git`, `new file mode 160000` and `index` lines are emitted.

Before, `emptyTree.Patch(commitTree)` on `git-fixtures/submodule` produced only the `.gitmodules` and `README.md` sections. Now it also produces:

```diff
diff --git a/basic b/basic
new file mode 160000
index 0000000000000000000000000000000000000000..6ecf0ef2c2dffb796033e5a02219af86ec6584e5
--- /dev/null
+++ b/basic
@@ -0,0 +1 @@
+Subproject commit 6ecf0ef2c2dffb796033e5a02219af86ec6584e5
```

Added `TestPatchWithSubmodule` covering both submodule addition and a submodule commit update.

Fixes #1387

### DIFF
--- a/plumbing/object/patch.go
+++ b/plumbing/object/patch.go
@@ -50,6 +50,14 @@ func getPatchContext(ctx context.Context, message string, changes ...*Change) (*
 }
 
 func filePatchWithContext(ctx context.Context, c *Change) (fdiff.FilePatch, error) {
+	// Submodules (gitlinks) are not blob objects, so their contents cannot be
+	// read as files. Git represents them in a diff by a single
+	// "Subproject commit <hash>" line, so build that content synthetically
+	// instead of dropping the change from the patch.
+	if isSubmodule(c.From) || isSubmodule(c.To) {
+		return submoduleFilePatch(ctx, c)
+	}
+
 	from, to, err := c.Files()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


plumbing/object/patch.go
```diff
@@ -98,6 +106,57 @@ func filePatchWithContext(ctx context.Context, c *Change) (fdiff.FilePatch, erro
...
+
+// submoduleContent returns the textual representation git uses for a submodule
+// (gitlink) in a diff: a single "Subproject commit <hash>" line. It returns an
+// empty string when the entry does not point to a submodule.
+func submoduleContent(e ChangeEntry) string {
+	if !isSubmodule(e) {
+		return ""
+	}
+
+	return fmt.Sprintf("Subproject commit %s\n", e.TreeEntry.Hash)
```
pjbgf, line 121: As per upstream: https://github.com/git/git/blob/55526a18268bbc1ddaf8a6b7850c33d984eac9e9/diff.c#L4479

pjbgf (approved):
Grazie @upuddu! 🙇

PR #2247, opened 2026-07-11
storage: dotgit, reject path traversal in reference names

Sahana2524:
**Reference names weren't contained to the ref storage**

A loose reference is written to `.git/<name>` with the name used verbatim as a path, so a name such as `refs/heads/../../config` (which a malicious remote can advertise, and which survives refspec mapping to `refs/remotes/origin/../../config`) resolves to unrelated metadata like `.git/config` or `.git/HEAD`. The check lives at the dotgit storage entry points, the same choke point as the existing submodule-name containment in `Module`, so every path that stores, reads, or removes a ref is covered and benign names such as `HEAD` and `refs/heads/*` are unaffected.

### DIFF
--- a/storage/filesystem/dotgit/dotgit.go
+++ b/storage/filesystem/dotgit/dotgit.go
@@ -24,6 +24,7 @@ import (
 	"github.com/go-git/go-billy/v6"
 
 	"github.com/go-git/go-git/v6/internal/packhandle"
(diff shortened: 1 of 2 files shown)


pjbgf (approved):

ℹ️ I added some additional checks that target Windows/Darwin to align with upstream git.

PR #2248, opened 2026-07-13
plumbing: gitignore, Clarify pattern and matcher parameters

### LINKED ISSUE #494: Documentation on gitignore package
Hi, I was looking into the [gitignore package](https://pkg.go.dev/github.com/go-git/go-git/v5/plumbing/format/gitignore), but It's not quite clear to me how this should be used correctly.  

I have really no idea what the `domain` argument is for in the `ParsePattern`  function. I did look at some tests, but I still couldn't figure it out. From some of my own tests, it seems to work just fine when this parameter is nil.

The second thing that's a bit confusing is  that the `path` parameter on `Matcher.Match` method is a slice. I believe this should be the full path of a file split on some path separator. I.e. forward slash on Linux.

`isDir` on the `Matcher.Match` method is also a bit confusing. My guess is that this flag indicates whether the passed path is a file or directory.

It would be great if this package was better documented, or at least had comments in the code explaining the parameters.

ahmojo:
## Summary

Clarifies the exported comments for `Pattern.Match`, `ParsePattern`,
`Matcher.Match`, and `NewMatcher`.

## What changed

- Defines `domain` and `path` as ordered logical path components and explains how each domain scopes matching.
- Documents the endpoint meaning of `isDir`, descendant behavior for patterns ending in `/`, and matcher priority.

The two-file patch contains exported comment changes. It does not change runtime behavior or add dependencies.

## Why

The existing comments do not explain how callers should construct `domain` and
`path` or how `isDir` affects directory-only patterns. The new wording follows
the current implementation, `ReadPatterns` traversal, and focused matcher tests.

## Tests

- [x] `go test ./plumbing/format/gitignore`
- [x] `go doc github.com/go-git/go-git/v6/plumbing/format/gitignore`
- [x] `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run ./plumbing/format/gitignore`
- [x] `git diff --check`

Closes #494

## AI assistance

OpenAI GPT-5.6 Sol (xhigh) via Codex assisted with repository research and
drafting the documentation update. I reviewed each statement against the
current implementation and tests and ran the listed checks.

### DIFF
--- a/plumbing/format/gitignore/matcher.go
+++ b/plumbing/format/gitignore/matcher.go
@@ -1,16 +1,21 @@
 package gitignore
 
-// Matcher defines a global multi-pattern matcher for gitignore patterns
+// Matcher defines a global multi-pattern matcher for gitignore patterns.
 type Matcher interface {
-	// Match matches patterns in the order of priorities. As soon as an inclusion or
-	// exclusion is found, not further matching is performed.
+	// Match reports whether path is excluded by the highest-priority matching
+	// pattern. Path is an ordered sequence of logical path components. Patterns
+	// created with ParsePattern match only paths beginning with their domain.
+	// isDir reports whether the final path component is a directory. For a
+	// pattern ending in "/", isDir only restricts a match at the candidate
+	// endpoint; descendants of a matched directory may still match.
 	Match(path []string, isDir bool) bool
 }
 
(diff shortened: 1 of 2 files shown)


pjbgf (approved):
@ahmojo thanks for working on this! 🙇‍♂️

PR #2250, opened 2026-07-13
git: fetch, refuse to clobber existing tags

Sahana2524:
A remote can silently repoint an existing local tag on a normal fetch or pull:
- buildFetchedTags is the sole path that writes auto-followed tags (a default fetch runs under AllTags, pull under TagFollowing), and it wrote the server's value with old set to nil, so an existing tag was overwritten unconditionally
- a hostile or on-path remote can advertise refs/tags/<pinned> at an object it also serves in the pack, moving a tag out from under a user or CI that later builds or verifies it
- git refuses the same case with "would clobber existing tag"; the guard here creates a tag that is missing and leaves an existing one pointing where it was

Force updates and the AllTags refspec are unaffected, since tags reach storage only through this helper.

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -788,6 +788,9 @@ func (r *Remote) addObject(rs config.RefSpec,
 		return nil
 	}
 	if !rs.IsForceUpdate() {
+		if err := checkTagUpdate(cmd); err != nil {
+			return err
+		}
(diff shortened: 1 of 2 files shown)


pjbgf (approved):

I've expanded the tests around this area and also the code so that force push is also considered. On the back of that, `AllTags` no longer use the force semantics, which better aligns with upstream

PR #2252, opened 2026-07-14
plumbing: transport/http, redact credentials in error messages

Sahana2524:
On any non-2xx response the error carries the full request URL, and that URL keeps whatever userinfo came from the clone URL:

    unexpected requesting "https://user:s3cr3t@example.com/repo.git/info/refs?service=git-upload-pack" status code: 500

`Err.Error` formats `resp.Request.URL` with `%q` and `httpRequester.doPost` formats the joined service URL with `%s`. Both errors surface to `Clone`/`Fetch` callers that commonly log them, so a password passed as `https://user:token@host/repo` lands in logs in cleartext.

The package already has `redactedURL`, used for the trace output and the redirect errors, so I routed these two paths through it too. Only the password becomes `REDACTED`; the rest of the URL stays so the message is still useful. Added a regression test on the `checkError` path.

### DIFF
--- a/plumbing/transport/http/common.go
+++ b/plumbing/transport/http/common.go
@@ -24,9 +24,9 @@ func (e *Err) StatusCode() int { return e.Status }
 func (e *Err) Error() string {
 	format := "unexpected requesting %q status code: %d"
 	if e.Reason != "" {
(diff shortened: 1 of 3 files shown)


pjbgf (approved):
@Sahana2524 thanks for working on this. :bow:

PR #2256, opened 2026-07-15
git: submodule, skip .gitmodules entries missing from the index

### LINKED ISSUE #2255: Clone with RecurseSubmodules fails with bare "entry not found" on a stale .gitmodules entry (divergence from git)
### Summary

Cloning with `RecurseSubmodules` (and any `Submodules().Update()`) aborts with a bare `entry not found` when `.gitmodules` declares a submodule whose `path` has no corresponding gitlink (mode `160000`) entry in the index. The git CLI handles the same repository fine, so this is a behaviour divergence, and the unwrapped error makes it very hard for users to diagnose (see downstream impact below).

This is the same defect as #145, which was closed as fixed — but the fix referenced there (#266) addressed relative submodule URLs, a different problem raised later in that thread's comments. The stale-entry failure is still present on `v5.19.1` and on `main`.

### Reproduction (no network, no auth)

A "ghost" entry like this occurs in the wild whenever someone removes a submodule with `git rm` / `git rm --cached` but forgets to delete its `.gitmodules` section — git never complains about it afterwards, so it tends to stay unnoticed for years:

```sh
git init -b main demo && cd demo
printf '[submodule "ghost"]\n\tpath = ghost\n\turl = https://example.com/ghost.git\n' > .gitmodules
git add . && git commit -m "stale .gitmodules entry"
cd .. && git clone --recurse-submodules demo clone-cli   # works fine, entry silently ignored
```

```go
_, err := git.PlainClone("clone-gogit", false, &git.CloneOptions{
	URL:               "<path-or-url-to>/demo",
	RecurseSubmodules: git.DefaultSubmoduleRecursionDepth,
})
// err: entry not found
```

Verified failing identically on v5.11.0, v5.19.1 and current main (da9f7d8). The failure happens before any submodule network access — the ghost URL is never dereferenced — and also reproduces when the repository has valid submodules alongside the stale entry, and with a `path = ./sub` spelling of an existing submodule (any `.gitmodules` path that is not byte-identical to an index entry).

### Analysis

go-git enumerates the sections of `.gitmodules` and requires each declared path to exist in the index: clone → checkout → [`updateSubmodules`](https://github.com/go-git/go-git/blob/da9f7d8a0e98b475600177348d6ece384a370f36/worktree.go#L154-L161) → `Submodule.update`, where the [`idx.Entry(s.c.Path)`](https://github.com/go-git/go-git/blob/da9f7d8a0e98b475600177348d6ece384a370f36/submodule.go#L268-L271) lookup error — `index.ErrEntryNotFound`, the bare `entry not found` — is returned as-is and aborts the whole clone.

git iterates in the opposite direction: it enumerates gitlink entries in the index and looks up their metadata in `.gitmodules` (a `.gitmodules`-only entry is simply not a submodule from its point of view, for `clone --recurse-submodules`, `submodule update --init --recursive` and `submodule status` alike).

Notably [`Submodule.status`](https://github.com/go-git/go-git/blob/da9f7d8a0e98b475600177348d6ece384a370f36/submodule.go#L75-L76) already tolerates `ErrEntryNotFound` for exactly this case — only `update` does not.

### Proposed fix

Two small changes, matching git's behaviour where it matters and keeping the error informative where it doesn't:

1. `Submodules.UpdateContext` (the path used by clone/pull recursion): skip submodules whose path is not present in the index, instead of aborting the whole operation.
2. `Submodule.update`: wrap the lookup error with the submodule name, so the single-submodule path fails with `submodule "ghost": entry not found` instead of a bare `entry not found`.

I have this implemented with tests and can send the PR right away (plus a backport to `releases/v5.x` — this is where the downstream impact currently lives).

### Downstream impact

fluxcd/flux2#4564 — Flux's source-controller surfaces this as `unable to clone '<url>': entry not found` on any `GitRepository` with `spec.recurseSubmodules: true` whose repo carries a stale `.gitmodules` entry, and the message gives users no hint that a submodule (let alone which one) is involved; full write-up in https://github.com/fluxcd/flux2/issues/4564#issuecomment-4982371307.

gecube:
Fixes #2255.

A `.gitmodules` section whose `path` has no gitlink entry in the index — typically left behind when a submodule is removed with `git rm` but its `.gitmodules` section is not deleted — caused `Submodules.Update` to abort the whole operation with a bare `entry not found`. Since clone and pull recursion go through this path, any repository carrying such a stale entry could not be cloned with `RecurseSubmodules` at all, while the git CLI handles it fine (it enumerates gitlink entries in the index and looks up their metadata in `.gitmodules`, so `.gitmodules`-only entries are invisible to clone, update and status alike).

Changes:

* `Submodules.UpdateContext` now skips entries missing from the index, matching the git CLI. The check runs before `Init`, so a skipped entry is not recorded into `.git/config` either (git does not initialize such entries).
* A direct `Update` on a single stale `Submodule` still fails, but the error is now wrapped with the submodule name (`submodule "ghost": entry not found`) instead of surfacing as a bare index lookup failure — the unwrapped message is what made this so hard to diagnose downstream (fluxcd/flux2#4564).
* New `Submodule.indexEntry` helper shared by both paths.

Tests: three self-contained offline tests (list update skips + no config pollution, single update error wrapping, end-to-end `PlainClone` with `RecurseSubmodules`) and one fixture-based test verifying that real submodules alongside a stale entry still update correctly. All three offline tests fail with `entry not found` before the fix.

Happy to follow up with a backport to `releases/v5.x` — that is where the downstream impact currently lives.

### DIFF
--- a/submodule.go
+++ b/submodule.go
@@ -247,6 +247,16 @@ func (s *Submodule) UpdateContext(ctx context.Context, o *SubmoduleUpdateOptions
 	return s.update(ctx, o, plumbing.ZeroHash)
 }
 
+// indexEntry returns the index entry recorded for the submodule path.
+func (s *Submodule) indexEntry() (*index.Entry, error) {
+	idx, err := s.w.r.Storer.Index()
+	if err != nil {
+		return nil, err
+	}
+
+	return idx.Entry(s.c.Path)
+}
+
 func (s *Submodule) update(ctx context.Context, o *SubmoduleUpdateOptions, forceHash plumbing.Hash) error {
 	if !s.initialized && !o.Init {
 		return ErrSubmoduleNotInitialized
@@ -258,16 +268,11 @@ func (s *Submodule) update(ctx context.Context, o *SubmoduleUpdateOptions, force
 		}
 	}
 
-	idx, err := s.w.r.Storer.Index()
-	if err != nil {
-		return err
-	}
-
 	hash := forceHash
 	if hash.IsZero() {
-		e, err := idx.Entry(s.c.Path)
+		e, err := s.indexEntry()
 		if err != nil {
-			return err
+			return fmt.Errorf("submodule %q: %w", s.c.Name, err)
 		}
 
 		hash = e.Hash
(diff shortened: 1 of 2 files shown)


pjbgf (changes requested):
@gecube thanks for looking into this. The changes LGTM, once the lint issues are fixed we can go ahead with the merge.

Please note that the commit message needs to be updated to align with the PR message.

pjbgf (approved):
@gecube thanks for working on this. 🙇 

I'm cutting a release soon, so if you get a back-port PR for this we may be able to get it in.

PR #2258, opened 2026-07-15
git: worktree, stat checked-out files via the open fd.

cedric-appdirect:
checkoutFile returns the file FileInfo obtained from fstat on the open descriptor before it is closed, and addIndexFromFile consumes it directly instead of issuing a separate lstat per checked-out file. This removes a path lookup from the checkout hot path and, for BoundOS, an extra os.Root open/close per file. Symlinks still lstat since there is no fd to fstat.

```
Benchmark results (darwin/arm64, Apple M4 Pro, benchstat):
  BenchmarkCloneLargeRepo                       537.6ms -> 380.1ms  -29.30% (p=0.001, n=7)
  BenchmarkCloneDeepRepo                       1283.4ms -> 969.8ms  -24.43% (p=0.001, n=7)
  BenchmarkResetHardIgnoredDir/BaselineNoIgnoredFiles
                                               11.74ms -> 11.80ms        ~ (p=0.683, n=15)
```

The reset benchmark resets to HEAD, so no checkout occurs and it is unaffected by this change, as expected.

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -993,11 +993,12 @@ func (w *Worktree) checkoutChangeRegularFile(cfg *config.Config,
 			return err
 		}
 
-		if err := w.checkoutFile(cfg, fs, f); err != nil {
+		fi, err := w.checkoutFile(cfg, fs, f)
+		if err != nil {
 			return err
 		}
 
-		return w.addIndexFromFile(fs, name, e.Hash, idx)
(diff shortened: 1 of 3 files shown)


pjbgf:
@cedric-appdirect LGTM. Please rebase and fix the conflict above.

PR #2260, opened 2026-07-15
git: worktree, checkout decompress AutoCRLF blobs once.

cedric-appdirect:
copyObjectToWorktree opened the blob reader twice when AutoCRLF was enabled - once to gather conversion stats and again to copy the content - decompressing each object twice. Buffer the single decompression into a pooled bytes.Buffer, stat the in-memory bytes, then write, so the object is read exactly once.

Performance impact measured on provided benchmark:

| | main (before) | branch (after) | Change | Statistics | Units |
|---|---|---|---|---|---|
|CheckoutAutoCRLF-12 | 356.3m ± 1% | 324.0m ± 3% | -9.07% | (p=0.001 n=7) | sec/op |
|CheckoutAutoCRLF-12 | 228.6Mi ± 0% | 228.6Mi ± 0% | -0.01% | (p=0.038 n=7) | B/op |
|CheckoutAutoCRLF-12 | 1.203M ± 0% | 1.203M ± 0% | -0.00% | (p=0.001 n=7) | allocs/op |

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -1030,6 +1030,31 @@ func (w *Worktree) copyObjectToWorktree(cfg *config.Config, object *object.File,
 	defer ioutil.CheckClose(src, &err)
 
 	if cfg.Core.AutoCRLF == "true" {
+		const maxBufferedBlob = 1 << 20
+
+		if object.Size > maxBufferedBlob {
+			br := sync.GetBufioReader(src)
+			defer sync.PutBufioReader(br)
+
+			stat, err := convert.GetStat(br)
+			if err != nil {
+				return err
+			}
+
+			src, err = object.Reader()
+			if err != nil {
+				return err
+			}
+			defer ioutil.CheckClose(src, &err)
+
+			if !stat.IsBinary() {
+				dst = convert.NewCRLFWriter(dst)
+			}
+
+			_, err = ioutil.CopyBufferPool(dst, src)
+			return err
+		}
+
 		buf := sync.GetBytesBuffer()
 		defer sync.PutBytesBuffer(buf)
 
--- a/worktree.go
+++ b/worktree.go
@@ -1,6 +1,7 @@
 package git
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


worktree.go
```diff
@@ -996,23 +997,24 @@ func (w *Worktree) copyObjectToWorktree(cfg *config.Config, object *object.File,
...
+		defer sync.PutBytesBuffer(buf)
 
-		stat, err := convert.GetStat(br)
-		if err != nil {
+		if _, err := io.Copy(buf, src); err != nil {
 			return err
 		}
 
-		src, err = object.Reader()
+		stat, err := convert.GetStat(bytes.NewReader(buf.Bytes()))
```
pjbgf, line 1065: This makes memory churn directly related to the blob size, as we materialise it to memory. Slightly amending the benchmark to handle  larger objects:

```
┌──────────────────┬───────────────┬────────────────┬──────────────┐
│      Metric      │ main (before) │ branch (after) │    Change    │
├──────────────────┼───────────────┼────────────────┼──────────────┤
│ sec/op           │ 664.9m        │ 572.4m         │ -14%         │
├──────────────────┼───────────────┼────────────────┼──────────────┤
│ peak-extra-bytes │ 18.7M         │ 152.8M         │ +717% (8.2×) │
├──────────────────┼───────────────┼────────────────┼──────────────┤
│ B/op             │ 6.25Mi        │ 48.6Mi         │ +678%        │
├──────────────────┼───────────────┼────────────────┼──────────────┤
│ allocs/op        │ 3.052M        │ 3.052M         │ ~0%          │
└──────────────────┴───────────────┴────────────────┴──────────────┘
```

pjbgf (changes requested):
I'm not sure this is a path worth pursuing as per benchmarks below.

worktree.go
cedric-appdirect: Indeed, it makes sense. Just a small variation with a threshold for small/large file. When a git repository is mostly with text and code, the double decompress is a noticeable penalty. Now this increase the code complexity, so not really sure. I will try to explore if other approach work. I tried using TeeReader first, but couldn't.

cedric-appdirect:
This is for item 1 of https://github.com/go-git/go-git/issues/1956

PR #2263, opened 2026-07-17
plumbing: protocol/packp, reject control bytes in daemon request

Sahana2524:
An attacker-controlled `git://` URL like `git://host/repo%00host=evil%00%00version=2` reaches the daemon-request encoder with a `Pathname` that still holds the decoded NUL bytes (`net/url` unescapes `%00`, and the git transport passes `req.URL.Path` straight through). `Encode` wrote every field verbatim into the single NUL-framed pkt-line:

    git-upload-pack /repo\x00host=evil\x00\x00version=2\x00host=realhost:9418\x00

So the request the client sends carries an attacker-chosen `host=` selector and extra parameters in front of the ones the transport set from the connection, which is enough to steer a daemon's virtual-host routing on a path-only injection (a malicious `.gitmodules` submodule url is one way in). Canonical git never hits this because the path is a C string a NUL truncates.

`validate()`, which `Encode` always calls first, now refuses control bytes in the command, pathname, host, and extra params before anything is written. It is the same rejection `ls-refs` already applies to ref-prefixes. The other NUL-delimited pkt-line writers take storage-validated ref names or fixed capabilities, so the daemon request is the one field reachable directly from a URL, and I kept the change to it.

### DIFF
--- a/plumbing/protocol/packp/gitproto.go
+++ b/plumbing/protocol/packp/gitproto.go
@@ -33,6 +33,42 @@ func (g *GitProtoRequest) validate() error {
 		return fmt.Errorf("%w: empty request command", ErrInvalidGitProtoRequest)
 	}
 
+	// The request is a single pkt-line whose fields are separated by NUL
+	// bytes ("<command> <pathname>\x00host=<host>\x00\x00<params>..."). A
+	// control byte in any field breaks that framing or splices in extra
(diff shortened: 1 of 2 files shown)


pjbgf (approved):

ℹ️ I've expanded your fix so that it runs as part of the `Decode` path as well.

PR #2264, opened 2026-07-19
plumbing: protocol/packp, compare zero object ids by value in Action

Sahana2524:
**Zero id check in Command.Action is object-format sensitive**

`Command.Action` decides create/delete/update by comparing `Old`/`New` to `plumbing.ZeroHash` with `==`, and that comparison also covers the id's object-format field. A zero id parsed from a push request carries the negotiated format ("sha256" for a 64-hex id) while `plumbing.ZeroHash` is format-unset, so a 64-hex zero never matches: on the receive-pack side a real delete is read as an update and the ref gets rewritten to the all-zero id instead of removed, and a both-zero command that should be rejected as invalid passes `validate`. The client controls the id text, so this is reachable on sha1 repos too. Moved the three checks to `Hash.IsZero`, which looks only at the bytes like Git's `is_null_oid`, so the sentinel no longer depends on the id's format.

### DIFF
--- a/options.go
+++ b/options.go
@@ -464,7 +464,7 @@ type ResetOptions struct {
 
 // Validate validates the fields and sets the default values.
 func (o *ResetOptions) Validate(r *Repository) error {
-	if o.Commit == plumbing.ZeroHash {
+	if o.Commit.IsZero() {
 		ref, err := r.Head()
 		if err != nil {
 			return err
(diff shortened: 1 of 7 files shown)


pjbgf (approved):
@Sahana2524 thank you for working on this! 🙇 

ℹ️ I've expanded your fix to other areas that should be object format-agnostic.

PR #2265, opened 2026-07-20
[WIP] Add `ctx` argument to `Storer` interface

LukeSheard:
I have found that it's incredibly useful to be able to pass a `ctx` into the `Storer` interface to enable cancelling backend operations associated with any store work. E.g. passing metadata / logging or cancelling DB transactions proactively / downstream writes.

It's a fairly sizable refactor of the interfaces, but ahead of v6 I thought now would be a good time to propose it.

### DIFF
--- a/_examples/blame/main.go
+++ b/_examples/blame/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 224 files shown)


pjbgf (changes requested):
@LukeSheard Thanks for working on this, expanding `context` support to go-git would be great, specially ahead of cutting v6.

To decrease the migration impact for existing users, we will need to make some trade-offs to add support in v6 and potentially clean up the API (without the `Context` suffix) on v7. Here's my take on getting this over the line:

1. Skip changes to `Plain*` funcs that don't already have a Context suffixed method for v6, instead we add context to their counterparts (e.g. `Init` as opposed to `PlainInit`). 
2. Split this into at least two PRs, to help fast track the review/merge. One adding the user-facing Context aware funcs:
```
git.BlameContext
git.Repository.GrepContext
git.Repository.LogContext
git.Worktree.GrepContext
git.Worktree.StatusContext
git.Worktree.StatusWithOptionsContext
git.Submodule.StatusContext
git.Submodules.StatusContext
git.Worktree.AddContext
git.Worktree.AddGlobContext
git.Worktree.AddWithOptionsContext
git.Worktree.CheckoutContext
git.Worktree.CherryPickContext
git.Worktree.CleanContext
git.Worktree.CommitContext
git.Worktree.MoveContext
git.Worktree.RemoveContext
git.Worktree.RemoveGlobContext
git.Worktree.ResetContext
git.Worktree.RestoreContext
git.Repository.MergeContext
git.Repository.PruneContext
git.Repository.RepackObjectsContext
git.Repository.DeleteObjectContext
```

And another to define a `WithContext` stand-alone interface that can return a context wrapper around `Storer`/`Storage`. That way we can add support in-tree as well as providing a primitive that off-tree implementations can also rely on.

It may be easier to start from the second PR to agree on the API, and once that is done, the user-facing changes and E2E wiring up should be fairly straight-forward.

Then at `v7` we may review the approach and drop all the existing `Context` suffixed funcs and review the Storer/Storage interfaces to have context awareness as a first class citizen.

PR #2266, opened 2026-07-21
*: reject NTFS and HFS disguised ".." in tree and submodule paths

Sahana2524:
**Disguised ".." slips past the tree and submodule path checks**

`ValidTreePath` and `Submodule.Validate` reject a literal `..` component but not the NTFS/HFS+ spellings a filesystem folds back to it (trailing spaces or dots, an Alternate Data Stream like `..::$INDEX_ALLOCATION`, or an ignorable code point such as `.<U+200C>.`), so a crafted tree entry name or `.gitmodules` path can escape the worktree root on those systems during checkout or submodule init. The name side already closes this (`validSubmoduleName`, and `validReferenceName` for refs) via `pathutil.IsHFSDot`/`IsNTFSDot` with a "." needle, and `ValidTreePath` already rejects a disguised `.git`, so this applies the same per-component check to `..`.

### DIFF
--- a/config/modules.go
+++ b/config/modules.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"regexp"
 	"strings"
 
 	"github.com/go-git/go-git/v6/internal/pathutil"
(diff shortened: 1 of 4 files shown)


Sahana2524:
gentle ping

hiddeco:
While reviewing this I found a couple of other related edge cases. I kept your (slightly modified) original commit in place, but this will be superseded by #2376.

Thanks a lot for your work on this, and your patience! :bow:

PR #2269, opened 2026-07-23
internal: v2, skip closeReader on a cancelled context

georg:
FetchV2's round loop called `closeReader(packReader)` unconditionally right after `streamPackfile` returned. `streamPackfile` wraps `packReader` in `ioutil.NewContextReader`, whose background goroutine can still be blocked in the underlying Read after the `<-ctx.Done()` branch returns (see its doc comment) -- closeReader's own drain-then-close then reads and closes that same packReader concurrently with the still-in-flight orphaned goroutine.
Caught by -race under context cancellation during a v2 fetch: DATA RACE between NewContextReader's orphaned background goroutine reading packReader and closeReader's own drain-then-close of that same packReader.

This is the same race class plumbing/transport/http's Fetch and Push guard against for the v0/v1 path (commits 37ef6522, e519c7a5): skip the close only when the read itself was a cancellation, since a non-cancellation error or
success means the last Read already returned via the result channel and the goroutine is quiescent, so closing is safe and necessary -- otherwise the response/connection leaks. On cancellation, a packReader backed by a
context-bound request (fetchV2's round callback builds its request with `http.NewRequestWithContext(ctx, ...)`, same as Fetch/Push) gets torn down by that request's own context handling instead, so nothing is leaked by
skipping the close here.

Classify against streamErr itself via errors.Is, not a fresh `ctx.Err()` check: ctx can turn Err() non-nil an instant after streamPackfile already returned with its read fully quiescent, and checking ctx.Err() again at that later, independent point would incorrectly skip the close and leak packReader.

Adds a test that FetchV2 skips draining/closing packReader on a cancelled context (verified failing against the prior unconditional close, which hangs trying to drain a reader that's blocked forever exactly like the hung-peer scenario this guards against) and still closes it on a non-cancellation error.

Assisted-by: Claude Sonnet 5

### DIFF
--- a/internal/transport/v2.go
+++ b/internal/transport/v2.go
@@ -224,7 +224,26 @@ func FetchV2(ctx context.Context, st storage.Storer, req *FetchRequest, round Fe
 
 		if out.Packfile {
 			streamErr := streamPackfile(ctx, st, packReader, req.Progress)
-			closeReader(packReader)
+			// Skip draining/closing on cancellation: streamPackfile wraps
+			// packReader in a NewContextReader, whose background goroutine
+			// can still be blocked in the underlying Read after the
+			// <-ctx.Done() branch returns (see its doc comment). Closing
+			// here would race that goroutine's in-flight Read -- the same
+			// class of race fixed in plumbing/transport/http's Fetch/Push
+			// for the v0/v1 path. On a non-cancellation error or success,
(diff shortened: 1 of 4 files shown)


pjbgf (approved):
@georg thanks for working on this. 🙇 

I've expanded the fix to `handshake.go` so that it is now covered across the board.

PR #2279, opened 2026-07-25
plumbing: format/gitignore, fix trailing ** matching unrelated paths

### LINKED ISSUE #2278: gitignore: a trailing `**` matches unrelated directories, hiding untracked files from `Worktree.Status`
## Bug Description

A gitignore pattern whose last segment is `**` can match paths that the earlier
segments of the same pattern already ruled out.

The most damaging shape is the common `**/<dir>/**` form. Given a `.gitignore`
containing only:

```
**/node_modules/**
```

go-git reports **every directory in the repository** as ignored — including
directories that contain no `node_modules` component anywhere in their path, and
even when the repository has no `node_modules` directory at all.

Because `Worktree.Status` stops descending into a directory the ignore matcher
excludes, this silently hides every untracked file below the repository root.
`Status()` reports a near-empty result and callers that rely on it to detect
uncommitted work see a clean tree.

A second, narrower case comes from the same code path: a trailing `**` reached
when the candidate path has already been fully consumed keeps the preceding
segment's result instead of computing its own, so `a/**/*/**` matches `a/f.txt`.

Reference git disagrees in both cases. Per `gitignore(5)`, a trailing `/**`
"matches everything inside" the directory it follows — it does not license a
match for paths that never contained the preceding segment.

## go-git Version

`main` @ ffa616afd85ba272fe85d43156b069a8aa762e4c (v6). The affected code
(`globMatch` in `plumbing/format/gitignore/pattern.go`) is also present in v5.

## Steps to Reproduce

Matcher-level minimal working example:

```golang
package main

import (
	"fmt"
	"strings"

	"github.com/go-git/go-git/v6/plumbing/format/gitignore"
)

func main() {
	m := gitignore.NewMatcher([]gitignore.Pattern{
		gitignore.ParsePattern("**/node_modules/**", nil),
	})

	for _, tc := range []struct {
		path  string
		isDir bool
	}{
		{"src", true},
		{"src/pkg", true},
		{"src/pkg/app.go", false},
		{"node_modules", true},
		{"node_modules/react/index.js", false},
	} {
		got := m.Match(strings.Split(tc.path, "/"), tc.isDir)
		fmt.Printf("%-28s isDir=%-5v ignored=%v\n", tc.path, tc.isDir, got)
	}
}
```

Actual output:

```
src                          isDir=true  ignored=true    <-- wrong
src/pkg                      isDir=true  ignored=true    <-- wrong
src/pkg/app.go               isDir=false ignored=false
node_modules                 isDir=true  ignored=true
node_modules/react/index.js  isDir=false ignored=true
```

Expected output, matching `git check-ignore` (git 2.47.3) over the same
`.gitignore`:

```
src/                         ignored=false
src/pkg/                     ignored=false
src/pkg/app.go               ignored=false
node_modules/                ignored=true
node_modules/react/index.js  ignored=true
```

### End-to-end effect on `Worktree.Status`

```sh
mkdir repro && cd repro && git init -q
printf '**/node_modules/**\n' > .gitignore
mkdir -p src/pkg && echo 'package pkg' > src/pkg/app.go && echo hi > README.md
git add .gitignore && git commit -qm init
```

Reference `git status --porcelain`:

```
?? README.md
?? src/
```

`Worktree.Status()` on the same repository:

```
?? README.md
```

`src/pkg/app.go` is missing entirely — `src` was reported as an ignored
directory, so the walk never descended into it.

## Additional Information

Root cause is in `globMatch`
(`plumbing/format/gitignore/pattern.go`), which treats its `matched` variable as
a flag that only ever moves to `true`:

- A `**/<segment>` traversal that runs out of path sets `matched = false` and
  falls through. The next loop iteration is the trailing `**`, and its branch is
  evaluated *before* the `if len(path) == 0 { return false }` guard, so it sets
  `matched` back to `true` for any directory candidate.
- The trailing `**` branch only ever raises `matched`, so when its own condition
  is not satisfied it silently inherits whatever the previous segment left
  behind.

A randomised differential run against `git check-ignore` (296 patterns ×
66 paths) shows five pattern families over-matching on current `main`:
`**/*.txt/**`, `**/a/**`, `**/baz/**/**`, `?/**/?/**/` and `a/**/*/**`.

I have a fix with regression tests cross-checked against the `git` binary via
the existing `ConformanceSuite` oracle, and will open a PR referencing this
issue.

glitch-ux:
Fixes #2278

## Why

A `.gitignore` containing `**/node_modules/**` makes go-git report *every*
directory in the repository as ignored — including directories with no
`node_modules` component in their path, and even when no `node_modules` exists.

That hides files rather than just mislabelling them: `Worktree.Status` stops
descending into a directory the matcher excludes, so untracked work below the
repository root never shows up. `Status()` looks clean when it isn't.

The cause is in `globMatch`, which treats `matched` as a flag that only ever
moves to `true`. When the earlier segments of a pattern rule a path out, the
trailing `**` runs afterwards and flips it back on. A second symptom shares the
cause: a trailing `**` reached with the path already consumed inherits the
previous segment's answer, so `a/**/*/**` matches `a/f.txt`.

[gitignore(5)](https://git-scm.com/docs/gitignore) defines the trailing `/**`
form as matching everything *inside* the directory it follows — not paths that
never contained the preceding segment. Confirmed against `git check-ignore`
(git 2.47.3).

## What changed

Eight lines in `plumbing/format/gitignore/pattern.go`: a `**` that never finds
the segment following it now returns a definitive non-match, and the trailing
`**` assigns its own result instead of only raising the flag.

## Verification

- Regression tests added to `ConformanceSuite`, so every expectation is
  cross-checked against the real `git` binary. All fail on `main`, pass here.
- Randomised differential run against `git check-ignore` (296 patterns × 66
  paths): five over-matching pattern families on `main`, zero after the fix,
  no new divergence.
- `golangci-lint` (v2.11.4, `make validate-lint`): 0 issues. `go test ./...`
  shows the same failures before and after.

### DIFF
--- a/plumbing/format/gitignore/conformance_test.go
+++ b/plumbing/format/gitignore/conformance_test.go
@@ -311,6 +311,14 @@ func (s *ConformanceSuite) TestWildmatchSlashSemantics() {
 		{"**/bar/**", "deep/foo/bar/baz/", true, "double star with double star continuation"},
 		{"**/bar/*", "deep/foo/bar", false, "double star continuation needs path"},
 		{"**/bar/**", "deep/foo/bar/", true, "double star double star with directory"},
+		{"**/bar/**", "deep/", false, "double star suffix needs the named segment"},
+		{"**/bar/**", "deep/foo/", false, "double star suffix needs the named segment when nested"},
+		{"**/bar/**", "deep/foo/baz/", false, "double star suffix rejects a sibling directory"},
+		{"**/bar/**", "deep/foo/file.txt", false, "double star suffix rejects a file outside the named segment"},
+		{"a/**/*/**", "a/f.txt", false, "double star suffix needs something below the consumed path"},
+		{"a/**/*/**", "a/b/c/f.txt", true, "double star suffix with a path left to cover"},
+		{"a/**/*/**", "a/b/", true, "double star suffix with a directory left to cover"},
+		{"**/*.txt/**", "a/f.txt", false, "double star suffix does not match the segment it follows"},
 		{"**/bar**", "foo/bar/baz", true, "double star with adjacent stars"},
 		{"*/bar/**", "foo/bar/baz/x", true, "mixed single double star"},
 		{"*/bar/**", "deep/foo/bar/baz/x", false, "mixed stars wrong depth"},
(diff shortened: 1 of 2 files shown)


pjbgf (approved):
@glitch-ux thanks for working on this. 🙇

glitch-ux:
@pjbgf My pleasure! :)

PR #2281, opened 2026-07-28
plumbing: transport, reject want lines for unadvertised objects

Sahana2524:
The server advertises one tip; a client asks for a different object and gets it:

    advertised:
      7f80bcc93a8d1226ac4f7eac0299ca27ada0a95a HEAD
      7f80bcc93a8d1226ac4f7eac0299ca27ada0a95a refs/heads/master

    want cb94db795b1542f008a5868b24d42d2abcc57e8e   (no ref points at it)
    -> PACK  cb94db79 c51d7a8d 58e7f96b

`UploadPack` passes `upreq.Wants` straight to revlist, and `serveFetchV2` does the same with `args.Wants`, so anything still in the object store gets packed on request: commits orphaned by a force-push before gc runs, refs an operator kept out of the advertisement, and, where a fork network shares one object store, another repository's objects. Canonical git declines these in `check_non_tip` unless one of the `uploadpack.allow*SHA1InWant` knobs is set, and the upload-archive path here already takes the same line through `archive.ResolveTreeish(..., allowUnreachable=false)`.

The check sits in front of both server entry points, so the advertisement stays the only thing a client can ask for. Symrefs are resolved and annotated tag chains walked, since a tag is advertised alongside its peeled value.

One thing worth flagging for review: a want that is reachable but not itself advertised is now refused. That matches git's default, though it is a change for anyone who was relying on the old behavior, and it is why `TestUploadPackV2FetchCommonButNotReady` now gives its ancestor commit a ref of its own. I left the `allowTipSHA1InWant` / `allowReachableSHA1InWant` / `allowAnySHA1InWant` knobs out rather than guess at how you would want them configured, but happy to add them if you would rather have the escape hatch in the same change.

### DIFF
--- a/plumbing/transport/upload_pack.go
+++ b/plumbing/transport/upload_pack.go
@@ -126,6 +126,10 @@ func UploadPack(
 			wants = upreq.Wants
 			caps = upreq.Capabilities
 
+			if err := checkWants(st, wants); err != nil {
+				return err
+			}
+
 			if err := r.Close(); err != nil {
 				return fmt.Errorf("closing reader: %w", err)
 			}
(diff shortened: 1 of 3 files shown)


Sahana2524:
any update?

PR #2284, opened 2026-07-29
plumbing: format/gitignore, Avoid blind ignore-file opens in ReadPatterns

cedric-appdirect:
## Why

`Worktree.Status()` collects ignore rules via `gitignore.ReadPatterns`, which blindly attempts to open `<dir>/.gitignore` **and** `<dir>/.git/info/exclude` at every directory it recurses into — even though the `ReadDir` listing it already holds answers whether those files exist. On `BoundOS` filesystems each failed open pays a full path-traversal cost; in the #181 investigation profile these failed opens dominated `ReadPatterns`, which itself accounted for ~28% of `Status()` CPU on a home-directory-style worktree.

Reference git reads a directory's `.gitignore` only when it exists, and consults `$GIT_DIR/info/exclude` only of the repository being walked ([`dir.c` `setup_standard_excludes()`](https://github.com/git/git/blob/master/dir.c)).

## What

- `ReadPatterns` now runs `ReadDir` first and opens `.gitignore` only when it appears in the listing. Directories without ignore files — the overwhelming common case — cost a single `ReadDir`, with zero open attempts.
- `.git/info/exclude` is read only at the root of the walked filesystem, and only when `.git` is present there. This also fixes a semantic divergence: a nested `<dir>/.git/info/exclude` (e.g. an embedded repository) previously contributed patterns to the outer walk; reference git never does that.
- Public API unchanged. One subtlety: on `ReadDir` error the function now returns `nil` patterns instead of partially-read ones; the sole caller (`collectIgnorePatterns`) discards patterns on error, so behavior is unchanged.

This is an independent follow-up to #2282 (stacks on `main`, not on that branch); the larger "lazy/incremental pattern walk" idea mentioned there remains future work.

## Benchmarks

New `BenchmarkReadPatterns` over 1,100 directories with no ignore files in subdirectories (home-directory-like worst case), `osfs` with `BoundOS` on an Apple M4 Pro. A/B of `main` vs this branch, 7 runs each (`-benchtime=3x`), via `benchstat`:

```
                              │ before (main) │  after (this PR)  │
                              │    sec/op     │  sec/op  vs base  │
ReadPatterns/NoIgnoreFiles-12    645.2m ± 79%  206.4m ± 8%  -68.01% (p=0.001 n=7)
ReadPatterns/RootIgnoreFiles-12  438.3m ± 19%  203.8m ± 10% -53.51% (p=0.001 n=7)
```

(The high variance on `main` reflects the syscall-bound failed-open cost; the after side is stable.)

## Test plan

- [x] New tests: nested `.git/info/exclude` is not read; zero blind opens when no ignore files exist; present ignore files are opened exactly once (open-counting billy filesystem wrapper)
- [x] Existing `plumbing/format/gitignore` suite: 105 passed
- [x] `go test .`: 754 passed, 0 failed
- [x] `go test -race ./...`: green; the 4 `plumbing/transport/http` backend E2E failures also occur on pristine `main` — they are caused by a local global git `pre-push` hook rejecting the tests' pushes to `main`, and pass with `core.hooksPath` disabled
- [x] `golangci-lint run` clean on changed packages

## AI disclosure

This contribution was produced with AI assistance (Kimi k3), including profiling analysis of #181, implementation and tests. Commits carry an `Assisted-by` trailer per AI_POLICY.md. The behavior was cross-checked against reference `git` semantics (`.git/info/exclude` scope).

### DIFF
--- a/plumbing/format/gitignore/dir.go
+++ b/plumbing/format/gitignore/dir.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"bytes"
 	"io"
-	gofs "io/fs"
 	"os"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


pjbgf (approved):

pjbgf:
@cedric-appdirect thanks for working on this. :bow:

PR #2286, opened 2026-07-30
plumbing: transport/http, drop credentials on cross-host redirect

Sahana2524:
With the endpoint carrying `testuser:testpass` and an `Options.Authorizer` setting `PRIVATE-TOKEN`, a 301 on the `/info/refs` GET puts both on the redirect target:

    redirect target received
      Authorization: Basic dGVzdHVzZXI6dGVzdHBhc3M=
      Private-Token: super-secret

`checkRedirect` left this to `http.Client`, which only withholds `Authorization` when the target sits outside the initial host's registered domain, and that test ignores the port. Anything else, including whatever an `Authorizer` sets, crosses unconditionally. libcurl with `CURLOPT_UNRESTRICTED_AUTH` off, which is what git relies on, compares host, port and scheme exactly.

Nothing that worked before stops working: `Handshake` already drops `URL.User` and the `Authorizer` once the redirect lands on a different host, so every request after discovery was unauthenticated there anyway. I kept the allowlist to the headers the transport sets itself, since a header it did not add seems hard to assume safe for a host the credentials were not issued for.

`TestRedirectStripsCredentials` covers the session side of this and still passes; the new test checks what the target actually received.

### DIFF
--- a/plumbing/transport/http/http.go
+++ b/plumbing/transport/http/http.go
@@ -130,11 +130,11 @@ func wrapCheckRedirect(policy RedirectPolicy, next func(*http.Request, []*http.R
 // default policy is "initial", where only the GET /info/refs discovery
 // request is allowed to follow redirects.
 //
-// Credential handling on redirect is left to Go's http.Client, which
-// already strips the Authorization header when a redirect crosses to a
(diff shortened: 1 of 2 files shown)


Sahana2524:
gentle ping

PR #2289, opened 2026-08-01
*: enable gosec linter

ferhatelmas:
related to #1753


ferhatelmas:
#1804 got outdated, reopened minimally (at least it can stop new violations)

PR #2294, opened 2026-08-03
plumbing: format/packfile, reject objects shorter than declared size

Sahana2524:
Feeding a single-entry pack whose blob header declares 32 bytes but whose
zlib stream inflates to 8 through `packfile.NewParser` with an
`idxfile.Writer` observer:

    pack trailer checksum:  valid
    parser:                 no error
    .idx entry:             b35dd6ca7fb627a36cee1f0bd8a2446b27447d9b
    sha1 of its content:    a96198c3fc9f41d213050dec8f2fdd63ef0213bf

The declared size is what the object id gets hashed over, so the entry
is indexed under an id that does not hash its own content.

`boundedWriter` rejects an entry that inflates past `oh.Size`, but the
count returned by the copy in `objectEntry` and `inflateContent` is
discarded, so a stream that stops short of the declared size is accepted.
`get_data` in index-pack.c only leaves its inflate loop on
`stream.total_out == size`, so comparing the bytes actually produced
against the header lines the scanner up with that.

This originally also covered short copy ops in the streaming delta
appliers. bb3d462 and f6f7fea have since fixed those on main, so I
rebased and dropped that half. What's left is the inflate side only.

I kept the two error values separate, `ErrInflatedSizeMismatch` for the
overrun and `ErrInflatedSizeShort` for the underrun, since they signal
different malformations. Folding them into one would also be reasonable
if you'd rather keep the exported surface smaller.

### DIFF
--- a/plumbing/format/packfile/parser_test.go
+++ b/plumbing/format/packfile/parser_test.go
@@ -549,6 +549,34 @@ func TestParserRejectsOverflowingObjectHeader(t *testing.T) {
 	require.ErrorContains(t, err, "malformed pack")
 }
 
+func TestParserRejectsShortInflatedObject(t *testing.T) {
+	t.Parallel()
+
+	// A single blob entry whose header declares 32 bytes but whose zlib
+	// stream inflates to 8. The pack is otherwise well formed, trailer
+	// included, so nothing else flags it. The declared size is what the
+	// object id is computed over, so accepting it would index the entry
+	// under an id that does not hash its own content.
+	var body bytes.Buffer
+	h := sha1.New()
+	w := io.MultiWriter(&body, h)
+
+	_, _ = w.Write([]byte("PACK"))
+	_ = binary.Write(w, binary.BigEndian, uint32(2))
+	_ = binary.Write(w, binary.BigEndian, uint32(1))
+
+	writePackObjectHeader(t, w, plumbing.BlobObject, 32)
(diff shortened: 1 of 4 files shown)


Sahana2524:
The three red test legs are all `TestRemoteSuite` hitting its 10s deadline
on live network calls, not something from this change:

    TestListPeeling (windows, stable + oldstable)
      http transport: Post "https://github.com/git-fixtures/tags.git/git-upload-pack":
      context deadline exceeded
    TestList (ubuntu, stable)
      context deadline exceeded

Both go through `Remote.List`, which stops at the ref advertisement and
never reaches the packfile parser. The same commit passed the full test
job on ubuntu/oldstable, ubuntu with git master and v2.11.0, and both
macOS legs. Locally `go test -race ./...` is green, golangci-lint reports
only the pre-existing revive warning in `util_fuzz_test.go`, and
FuzzParser, FuzzPatchDelta and FuzzScanner each run clean. Happy to
rebase if you'd like a fresh run.

Sahana2524:
Rebased onto main. bb3d462 and f6f7fea already cover the delta half, so I dropped those hunks and their tests. This is now just the inflate-side check in the scanner plus its parser test, and the description is updated to match.

The one red leg (oldstable, ubuntu) is `TestReceivePackSuite/TestFullSendPackOnNonEmpty` in `plumbing/transport/git` failing with `context canceled before the port is connectable` while starting `git daemon`, so it doesn't reach the packfile code. The other five test legs pass on the same commit. I can't re-run it from the fork.

PR #2299, opened 2026-08-06
worktree: honor global ignore files in Status

### LINKED ISSUE #436: Worktree.Status() and git status differ on newlines
I'm calling `Worktree.Status().IsClean()` and `Worktree.Status().String()` on a repository which tell me pretty much all files have changed, however `$ git status` tells me the worktree is clean. The repo in question is this https://github.com/dsccommunity/ActiveDirectoryDsc in which files end in CRLF `\r\n` instead of LF `\n`.

I believe the relevant part of my `~/.gitconfig`:
```
[core]
	autocrlf = input
```

Is there a way to configure `go-git` to have like a regular `git status` in this case?

### LINKED ISSUE #795: Global gitignore not read by default
### Discussed in https://github.com/go-git/go-git/discussions/793

<div type='discussions-op-text'>

<sup>Originally posted by **romainreignier** June 15, 2023</sup>
In my `~/.gitconfig`, I have the following section:

```
[core]
        excludesfile = /home/user/.gitignore
```

But go-git does not seem to read it with `worktree.Status()` and `status.IsClean()`.

I have found the function [`gitignore.LoadGlobalPatterns()`](https://github.com/go-git/go-git/blob/35f7e6770361a2c16c9b6c44acdc38ae04c75bd3/plumbing/format/gitignore/dir.go#L140) which is supposed to help. But it seems that this function is never called and I don't understand how to use it.

Does someone know how to use it?

Thanks in advance</div>

When using a global gitignore file defined in `~/.gitconfig`, `go-git` does not read it  it with `worktree.Status()` and `status.IsClean()`.

In order for this `gitignore` to be read, the [`gitignore.LoadGlobalPatterns()`](https://github.com/go-git/go-git/blob/35f7e6770361a2c16c9b6c44acdc38ae04c75bd3/plumbing/format/gitignore/dir.go#L140) needs to be used, as done in this code I have found with a Github search:

https://github.com/apache/skywalking-eyes/blob/5dfa68f93380a5e57259faaf95088b7f133b5778/pkg/header/check.go#L84-L94

aryansk:
Summary

This addresses #795 by making Worktree.Status load Git global ignore patterns before evaluating untracked and modified paths.

Changes

- Load global patterns from an OS-backed filesystem so absolute paths such as the configured core.excludesFile are readable.
- Keep repository patterns and explicit Worktree.Excludes behavior, with explicit worktree excludes taking precedence.
- Limit the global-file lookup to the native OS filesystem so in-memory and non-OS filesystems keep their existing behavior.

Validation

- git diff --check passed.
- Go tests were not run locally because the Go toolchain is unavailable in this environment; CI should exercise the go-git test suite.

AI assistance: ChatGPT assisted with code exploration and drafting. I reviewed the resulting patch and remain responsible for it.

Fixes #795

### DIFF
--- a/worktree_status.go
+++ b/worktree_status.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-git/go-billy/v6/osfs"
 	"github.com/go-git/go-billy/v6/util"
 
 	"github.com/go-git/go-git/v6/config"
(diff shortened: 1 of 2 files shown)


Soph:
Hey, thanks for your contribution! I think your PR addresses #795 and not #436, right? And I know it's still in Draft but while I comment: please add tests especially regression tests.

aryansk:
Thanks for catching that. You are right that this change targets issue 795, so I corrected the PR body from 436 to 795. I also added a regression test in commit 29329c2 that configures a global excludes file and verifies Worktree.Status hides matching untracked files while still reporting a visible file. git diff --check passes; I could not run Go tests because the Go toolchain is unavailable in this checkout, so CI should validate the test and implementation.

aryansk:
Thanks for checking. You’re right: this PR addresses #795, not #436, and the body is aligned with #795. The branch already includes regression coverage in `worktree_status_test.go`; this environment does not have the Go toolchain installed, so I haven’t claimed a local Go test run. If you prefer a different issue linkage or title, I can adjust it.

PR #2302, opened 2026-08-06
plumbing: format/gitignore, Inherit patterns through the ReadPatterns walk.

Soph:
**Draft: seeking design input, not review.** The pruning below is correct and a large win, but it exposes that the exported `ReadPatterns` + `Matcher` pair cannot represent "a parent directory is excluded". I don't think that is fixable at acceptable cost in the current design — numbers below.

## The defect

`ReadPatterns` recursed without passing the patterns it had collected, so every directory rebuilt its set from its own ignore files. A rule could therefore only prune a *direct child* of the `.gitignore` declaring it. With root `outer/ignored/`: `Match(["outer"])` fails (correctly), the recursion into `outer/` starts with an empty set, and the subtree is walked in full.

`readPatterns` now threads the in-scope patterns down the recursion. Sibling-subtree patterns stay out of that scope — they cannot apply, and including them made the walk order-dependent.

This is what git does. `prep_exclude` keeps a prefix-keyed `exclude_stack` and aborts before reading a directory's `.gitignore` when that directory is already excluded, unless the matching pattern is a negation ([dir.c](https://github.com/git/git/blob/master/dir.c)).

## Numbers

`BenchmarkReadPatterns/PrunedPattern`, 1,100 directories under one excluded directory: rule naming a direct child 79.4µs → 44µs, rule naming a grandchild **56.5ms → 64µs**.

At `Status()` level, via a new `BenchmarkStatusNestedIgnoredDir` (the existing `BenchmarkStatusIgnoredDir` cannot show this — its rule names a direct child, which `main` already prunes, and its ignored tree is 21 directories wide):

| ignored dirs | `main` | this PR |
| --- | --- | --- |
| 200 | 19.6ms | 3.0ms (-84.7%) |
| 2000 | **252.1ms** | **4.0ms (-98.4%)** |

`main` does not merely lose here, it fails to scale: ten times the directories costs it thirteen times the wall clock, because pattern collection walks the whole ignored tree that the diff walk then skips.

### Where this PR does not help

Also added, `BenchmarkStatusManyIgnoreFiles` — every directory declares its own `.gitignore` and nothing is excluded, so there is nothing to prune:

| ignore files | `main` | this PR |
| --- | --- | --- |
| 50 | 8.36ms | 8.62ms (~, p=0.818) |
| 500 | 88.2ms | 94.2ms (+6.9%, p=0.015) |

No benefit, and possibly a small cost — though +6.9% sits inside that arm's own ±21% spread, so "no measurable change" is the fair reading. `BenchmarkStatusLarge` (5,000 files, no `.gitignore` at all) likewise shows nothing.

This is the honest shape of the change: a large win on one layout, nothing on the others. It is the same boundary the correctness problem sits on — both come from collecting every pattern in the repository up front.

## The problem

```
.gitignore:                 outer/ignored/
                            !outer/ignored/keep.txt
outer/ignored/.gitignore:   keep.txt
```

`git check-ignore -v outer/ignored/keep.txt` → ignored, via `.gitignore:1:outer/ignored/`.

| case | `main` | this PR | + ancestor-aware `Match` | git |
| --- | --- | --- | --- | --- |
| **A** root negation + nested re-exclude | true | **false** | true | true |
| **B** root negation, no nested file | false | false | true | true |
| **C** nested `!keep.txt` under an excluded dir | false | true | true | true |

Case A is a regression in this PR, caught in review. But B is the same layout without the nested file and is wrong before *and* after: `Matcher` is a flat pattern list with no notion of an excluded ancestor, so it has never matched git for this class. `main` only gets A right by accident — it descends into the excluded directory and collects `keep.txt`, which sorts last. C is the common user mistake, and `main` gets it wrong.

### Ancestor-aware matching is correct but too expensive

Checking each ancestor prefix before the normal scan fixes A, B and C, and keeps the conformance suite's ~900 `check-ignore` oracle assertions green. Cost:

```
Matcher.Match (n=6)                geomean 433n → 1.64µs   +277%
  patterns=100/depth=10                   3.22µs → 19.3µs
BenchmarkStatusLarge/Clean (n=5)          19.38m → 22.16m   +14.4% (p=0.016)
```

n=5 is below benchstat's threshold, so treat the end-to-end figure as directional.

In `Status()` that work is redundant: the noder already prunes excluded directories top-down, so no ancestor is ever excluded by the time it asks about a file, and the loop runs to completion returning false every time. Git avoids this by carrying the excluded ancestor in the walk (`dir->internal.pattern`) instead of deciding per query. The state belongs in the traversal, not the predicate.

## Options

1. Land as-is, documenting A as a known limitation of the flat API (B is pre-existing).
2. Land with ancestor-aware `Match` — correct everywhere, ~14% on `Status()`.
3. Park this and build the scoped walk #2284 deferred.

I would prefer 3, because it fixes correctness and cost together: re-inclusion under an excluded parent becomes impossible, per-entry matching drops from O(all patterns repo-wide) to O(ancestor chain), and `Status()` stops traversing the worktree twice — once in `ReadPatterns`, once in the merkletrie diff. `readPatterns` already computes the per-directory scope such a matcher needs and discards it.

I have a working prototype, measured on the same three benchmarks (all vs `main`):

| | this PR | scoped walk |
| --- | --- | --- |
| `StatusNestedIgnoredDir` 2000 dirs | -98.4% | **-99.0%** |
| `StatusManyIgnoreFiles` 500 files | ~0 / +6.9% | **-40.4%** |
| `StatusLarge` (no `.gitignore`) | ~0 | **-41.9%** |

It also passes cases A, B and C. Ratios are robust; absolute figures are not precise to three digits, since several arms ran at ±20% or worse.

The blocker is API shape: `Options.IgnoreMatcher` is the one-method `gitignore.Matcher`, so a push/pop scoped matcher needs a new exported type or a breaking change. That may warrant an RFC; happy to write one.

## Tests

`ReadDir`-counting assertions over four rule placements (*grandchild* and *unanchored-at-depth* fail without this change; *direct child* and *declared in a subdirectory* are controls), a hermetic memfs test for case C, and an oracle test cross-checking six placements against `git ls-files --others --ignored`, gated on `skipOnLegacyGit` for the git 2.11.0 CI leg. No test covers case A yet — which behaviour is correct depends on the option chosen.

`golangci-lint` v2.11.4 reports 0 issues. `TestConformanceSuite/TestIgnoreDoubleStarPrefix` fails on `main` too (git 2.50.1) and is unrelated.

Per [AI_POLICY.md](https://github.com/go-git/go-git/blob/main/AI_POLICY.md): produced with AI assistance (Claude Opus 5), disclosed with an `Assisted-by:` trailer on the commit.

### DIFF
--- a/plumbing/format/gitignore/dir.go
+++ b/plumbing/format/gitignore/dir.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"io"
 	"os"
+	"slices"
 	"strings"
 
 	"github.com/go-git/go-billy/v6"
@@ -55,7 +56,22 @@ func readIgnoreFile(fs billy.Filesystem, path []string, ignoreFile string) (ps [
 // matching reference git which reads $GIT_DIR/info/exclude of the
 // repository being walked. Ignore files are opened only when present in
 // the directory listing, so directories without them cost a single ReadDir.
+//
+// Excluded directories are not descended into, at any depth below the ignore
+// file declaring the rule, so patterns declared inside an excluded directory
+// are not read and do not appear in the result. See the package documentation
+// for why gitignore(5) makes that equivalent to reading them.
 func ReadPatterns(fs billy.Filesystem, path []string) (ps []Pattern, err error) {
+	return readPatterns(fs, path, nil)
+}
+
+// readPatterns collects the ignore patterns declared at path and below.
+//
+// inherited holds the patterns already in effect from ancestor directories in
+// ascending order of priority. They take part in the decision to descend into
+// a subdirectory, but are not returned, so each pattern reaches the caller
+// exactly once.
+func readPatterns(fs billy.Filesystem, path []string, inherited []Pattern) (ps []Pattern, err error) {
 	fis, err := fs.ReadDir(fs.Join(path...))
 	if err != nil {
 		return nil, err
@@ -80,25 +96,34 @@ func ReadPatterns(fs billy.Filesystem, path []string) (ps []Pattern, err error) 
 		ps = append(ps, subps...)
 	}
 
+	// Ancestors first, so the deeper file wins. This deliberately excludes the
+	// patterns that the loop below collects from sibling subtrees: they cannot
+	// apply here, and including them would make the walk order-dependent.
+	inScope := inherited
+	if len(ps) > 0 {
+		inScope = slices.Concat(inherited, ps)
+	}
+	m := NewMatcher(inScope)
+
 	for _, fi := range fis {
-		if fi.IsDir() && fi.Name() != gitDir {
-			if NewMatcher(ps).Match(append(path, fi.Name()), true) {
-				continue
-			}
-
-			var subps []Pattern
-			subps, err = ReadPatterns(fs, append(path, fi.Name()))
-			if err != nil {
-				return ps, err
-			}
-
-			if len(subps) > 0 {
-				ps = append(ps, subps...)
-			}
+		if !fi.IsDir() || fi.Name() == gitDir {
+			continue
 		}
+
+		sub := slices.Concat(path, []string{fi.Name()})
+		if m.Match(sub, true) {
+			continue
+		}
+
+		subps, err := readPatterns(fs, sub, inScope)
+		if err != nil {
+			return ps, err
+		}
+
+		ps = append(ps, subps...)
 	}
 
-	return ps, err
+	return ps, nil
 }
 
 func loadPatterns(fs billy.Filesystem, path string) (ps []Pattern, err error) {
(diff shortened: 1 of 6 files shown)


Soph:
Superseded by #2309, which takes option 3 from the analysis above.

The measurements decided it. This PR fixed the pruning defect but left the exported `ReadPatterns` + `Matcher` pair unable to express an excluded parent, and the obvious repair — checking ancestors inside `Match` — cost +277% geomean on `Match` and ~14% on `StatusLarge`, because it recomputes per entry what the walk already knows.

Evaluating rules per directory as the walk descends fixes cases A, B and C by construction and is faster rather than slower, versus `main`:

| | this PR | #2309 |
| --- | --- | --- |
| `StatusNestedIgnoredDir` 2000 dirs | -98.4% | **-99.0%** |
| `StatusManyIgnoreFiles` 500 files | ~0 / +6.9% | **-40.4%** |
| `StatusLarge` no `.gitignore` | ~0 | **-41.9%** |

The two `Status` benchmarks added here carry over to #2309 unchanged, and `ReadPatterns` is left exactly as it is on `main`, deprecated in favour of the new `gitignore.Scope`, so no behaviour regresses for anyone still calling it.

Closing to keep the queue honest; the analysis stays readable here.

PR #2303, opened 2026-08-07
config, storage/filesystem: honor core.sharedRepository for object files

### LINKED ISSUE #843: ObjectWriter creates files as mode 0600, even if core.sharedRepository = "group"
Per https://git-scm.com/docs/git-config, if `core.sharedRepository` = group, then "the repository is made shareable between several users in a group (making sure all the files and objects are group-writable)". This is useful when sharing an on-disk repo between two unix users (e.g. a web server, and a normal user accessing the repo via git push/pull over SSH).

go-git seems to unconditionally create object files as 0600, meaning they are unaccessible to members of the group.

### Steps to reproduce

1. Create a repository with `git init --shared` (this sets `core.sharedRepository` to `group`)
2. Commit via https://github.com/go-git/go-git/blob/master/_examples/commit/main.go

### Expected
Files created in `git/objects/` are at least group-readable, e.g. 0660.

### Actual
Files created in `.git/objects/` are 0600.

### What's happening

1. ObjectWriter uses billy.Filesystem's TempFile: https://github.com/go-git/go-git/blob/809f9df1b76258a311a20c76d346e86aca0a08f8/storage/filesystem/dotgit/writers.go#L256
2. billy.Filesystem's TempFile uses ioutil.TempFIle: https://github.com/src-d/go-billy/blob/780403cfc1bc95ff4d07e7b26db40a6186c5326e/osfs/os.go#L99
3. ioutil.TempFile uses 0600 https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/os/tempfile.go;l=44

MsfPablo:
Parse `core.sharedRepository` in Config, propagate it through `dotgit.Options`, and apply group/all writable permissions to loose objects (and their parent directories) created by `ObjectWriter` and `PackWriter`.

- Adds `SharedRepository` type and constants to the config package.
- `ObjectWriter` now creates files as 0660 for `group` and 0664 for `all`.
- Parent directories are chmod'd to 2770 / 2775 on Unix so group ownership is inherited.
- The default (`umask` / unset) keeps the existing read-only object behavior.

Fixes #843.

AI disclosure: this PR was prepared with assistance from Claude.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -96,6 +96,10 @@ type Config struct {
 		// directional characters that HFS+ would normalize away).
 		// When unset, defaults to true on macOS.
 		ProtectHFS OptBool
+		// SharedRepository controls permissions applied to repository files.
+		// Valid values are "umask" (default), "group", "all" or an octal
+		// mode such as "0660". See git-config documentation for core.sharedRepository.
+		SharedRepository SharedRepository
 	}
 
 	User user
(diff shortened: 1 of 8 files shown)


flynnjustin24 (approved):
https://github.com/go-git/go-git/issues/843

flynnjustin24 (approved):

PR #2305, opened 2026-08-08
[v5] plumbing: revlist, Skip path validation in the object walk

### LINKED ISSUE #2251: Push fails with "invalid path: contains control character" when history (already on remote) contains such a path, even though the pushed commit does not
### Bug Description

When pushing a commit, `Repository.Push` computes the set of objects to send via `revlist.Objects`, which walks the reachable trees through `iterateCommitTrees` → `object.TreeWalker.Next`. Since v5.19.1, `TreeWalker.Next` calls `pathutil.ValidTreePath` on every entry it enumerates (added in #2105), and the revlist `TreeWalker` is created with the default `skipPathValidation = false`.

As a result, if **any** reachable commit in history contains a tree entry whose name has a control character even one that was deleted years ago, and even one whose objects are **already present on the remote** and would therefore never be sent  the entire push fails with:

    invalid path "\x1b\x1b\x1b": contains control character

**Expected:** the push succeeds. The offending object is not part of the new commit, is already on the remote, and is never materialised to a filesystem. Upstream Git pushes such a repository without complaint.

**Actual:** the push aborts before any network transfer, because the revlist tree walk validates historical entry names it is only enumerating to compute the object set.

This is the same class of issue as #2222 ("skip path validation in tree diff walk"), which already established the principle that inspection-only walks that never materialise names to disk should not run `ValidTreePath`. The revlist walk performed for push is exactly such an inspection-only walk: it computes which objects to send, it writes nothing to the working tree. It looks like #2222 addressed the diff walk (`treenoder.transformChildren`) but not the revlist walk (`plumbing/revlist.iterateCommitTrees`).

### go-git Version

v5.19.1

### Steps to Reproduce

1. Create a repo, commit a tree entry whose name contains a control character (a byte < 0x20), then delete it in a later commit. Upstream Git allows this — only `/` and NUL are forbidden in a path component.
2. Push all history to a bare remote.
3. Add a normal commit on top (that does **not** touch the bad path) and push it again.
4. Observe the push fail with `invalid path "...": contains control character`, even though the bad object is already on the remote and is not part of the new commit.

Code example:
```golang
// setup: a repo whose history contains a tree entry with a control-char name,
// already pushed to `remote`. HEAD is a later, clean commit.

repo, _ := git.PlainOpen("repo-with-bad-path-in-history")

// make an unrelated clean commit
wt, _ := repo.Worktree()
os.WriteFile("repo-with-bad-path-in-history/foo.txt", []byte("hi"), 0o644)
wt.Add("foo.txt")
wt.Commit("clean commit", &git.CommitOptions{Author: sig})

// this fails, even though the control-char object is already on the remote
// and the new commit does not reference it:
err := repo.Push(&git.PushOptions{RemoteName: "origin"})
// err: invalid path "\x1b\x1b\x1b": contains control character

### Additional Information

**Call path that triggers it:**

**Why upstream Git does not reject this:** `verify_path` forbids only `/` and NUL inside a path component; control characters and newlines are valid pathname bytes on disk. go-git's `ValidTreePath` is stricter, which is correct at *materialisation* boundaries (checkout, archive) but not for a read-only revlist walk that only computes the object set to transfer.

**Suggested fix:** set `skipPathValidation = true` on the `TreeWalker` created in `iterateCommitTrees` (revlist), mirroring what #2222 did for `treenoder.transformChildren` (diff). The revlist walk enumerates objects for transfer and never funnels entry names into the filesystem, so the materialisation-safety validation added in #2105 is not needed there. Materialising callers (`archive`, `FileIter`, `FindEntry`, `TreeEntryFile`) keep validating as before.

**Real-world impact:** a control-char filename committed in 2021 and deleted the next day still lives in history and is cloned by everyone. Every push of a version tag now fails, on every branch that descends from that commit, with no change to the user's code only a go-git bump triggered it. The only user-side remedy today is rewriting history (`git filter-repo`) across all affected branches.

### PR DESCRIPTION
**Pull request by r0h1tb**
# [v5] plumbing: revlist, Skip path validation in the object walk

Fixes #2251

Targets `releases/v5.x`. `main` is unaffected — the v6 object walk was rewritten
to read `Tree.Entries` directly and no longer goes through `TreeWalker`.

## Problem

`Push` computes the set of objects to send with `revlist.Objects`. If **any**
reachable tree in history contains an entry whose name holds a control
character, the whole push aborts before any network transfer:

```
invalid path "\x1b\x1b\x1b": contains control character
```

The offending entry does not have to be part of the commit being pushed. In the
reported case it was deleted years earlier and its objects were already on the
remote, so they would never have been sent. Upstream Git pushes such a
repository without complaint: `verify_path` forbids only `/` and NUL inside a
path component.

## Root cause

`iterateCommitTrees` enumerates entries through `TreeWalker.Next`, which since
#2105 validates every name:

```go
// plumbing/object/tree.go
if err := pathutil.ValidTreePath(entry.Name); err != nil {
    return name, entry, err
}
```

`ValidTreePath` exists so that callers **materialising** entries into a worktree
or archive stay safe. The revlist walk materialises nothing — it only collects
hashes to decide what to send — but a validator error propagates out of
`iterateCommitTrees` and fails the entire walk.

This is the same defect #2222 fixed for the tree diff walk, at a call site that
fix did not cover.

## Fix

`TreeWalker` gains an opt-out, and revlist takes it:

```go
treeWalker := object.NewTreeWalker(tree, true, seen)
treeWalker.SkipPathValidation()
```

Validation stays **on by default**; only this inspection-only walk opts out.
Path safety for callers that do materialise names is untouched and still
enforced at `FindEntry`, `TreeEntryFile`, `archive` and `FileIter`.

**Why an exported method rather than the unexported field #2222 used.** On
`main` the only opt-out caller is `treeNoder`, which lives in package `object`
and can set the field directly. `revlist` is a separate package, so v5 needs an
exported entry point. `SkipPathValidation()` is the smallest such surface I
could find — happy to reshape it into an exported field, a `NewTreeWalker`
variant, or an `internal/` accessor if you'd prefer not to grow the v5 API.

## A subtlety worth flagging

`Next` checks its `seen` set **before** validating, so an entry whose blob is
still reachable from a newer tree is skipped before the validator sees it. My
first attempt at a regression test reused one blob across both trees and
**passed against the unfixed branch**. Content that was genuinely deleted has a
blob nothing else references, which is exactly why the reported case involves
history. Both tests now use distinct blobs.

## What I deliberately left out

- **No change to `ValidTreePath` itself**, and no relaxation of any
  materialisation boundary. Only the walk that decides which objects to send.
- **No v6 change** — the bug does not exist there.
- **No backport of the `archive`/`FileIter` call sites.** Those do materialise
  names, so validation belongs there.

## The second commit

`plumbing: object, Skip path validation in tree diff walk` backports #2222,
which landed on `main` but never on `releases/v5.x`. I found it by checking the
sibling call site. On this branch `DiffTree` still fails:

```
from: invalid path "\x1b\x1b\x1b": contains control character
```

It is a separate commit precisely so you can drop it if you'd rather track the
v5 backport on its own — the first commit stands alone.

## Tests

Two new files, both failing on the unfixed branch with the reported error:

```
$ git stash                       # revert source, keep tests
$ go test ./plumbing/revlist/ -check.f ControlCharacter

FAIL: revlist_pathvalidation_test.go:80: RevListSuite.TestRevListObjects_ControlCharacterInHistory
... ("invalid path \"\\x1b\\x1b\\x1b\": contains control character")

FAIL: revlist_pathvalidation_test.go:121: RevListSuite.TestRevListObjects_ControlCharacterInPushedCommit
... ("invalid path \"\\x1b\\x1b\\x1b\": contains control character")

OOPS: 0 passed, 2 FAILED
```

```
$ git stash -- plumbing/object/treenoder.go
$ go test ./plumbing/object/ -check.f DiffTreePathValidation

FAIL: difftree_pathvalidation_test.go:54: DiffTreePathValidationSuite.TestDiffTree_ControlCharacterEntry
... ("from: invalid path \"\\x1b\\x1b\\x1b\": contains control character")

OOPS: 0 passed, 1 FAILED
```

With the fixes applied, all three pass.

The trees are written through `Tree.Encode`, which rejects only NUL — the same
constraint upstream Git applies — so the fixtures hold names Git itself would
accept, rather than anything synthetic.

## Verification

`go test -count=1 ./...` on macOS 26.1 arm64, Go 1.26.0:

| | ok | FAIL | no test files |
|---|---|---|---|
| baseline (`releases/v5.x` @ 42852fda) | 53 | 0 | 7 |
| with both commits | 53 | 0 | 7 |

Package-level results are byte-identical to the baseline. No pre-existing
failures on this branch in my environment, with one caveat worth naming: on the
very first cold run `plumbing/transport/git` failed 12 tests with
`dial tcp [::1]:PORT: connect: connection refused`. It passed on every
subsequent run including the recorded cold baseline, so I treated it as a
local port-binding flake in that suite's `git daemon` setup, not a real result.

`gofmt -l` is clean on all five changed files, and `go vet ./...` is clean.
(`gofmt -l .` reports 30 pre-existing files, e.g. `plumbing/format/commitgraph/*`,
none of them touched here.)

## Reshaping

Happy to split the two commits into separate PRs, drop the second, rename or
restructure `SkipPathValidation`, or move the regression tests into the existing
`revlist_test.go` / `difftree_test.go` if you'd rather not have new files.

### DIFF
--- a/plumbing/object/difftree_pathvalidation_test.go
+++ b/plumbing/object/difftree_pathvalidation_test.go
@@ -1,0 +1,74 @@
+package object
+
+import (
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/filemode"
+	"github.com/go-git/go-git/v5/storage/memory"
+
+	. "gopkg.in/check.v1"
+)
+
+type DiffTreePathValidationSuite struct{}
+
+var _ = Suite(&DiffTreePathValidationSuite{})
+
+func (s *DiffTreePathValidationSuite) storeBlob(c *C, sto *memory.Storage, content string) plumbing.Hash {
+	obj := sto.NewEncodedObject()
+	obj.SetType(plumbing.BlobObject)
+
+	w, err := obj.Writer()
+	c.Assert(err, IsNil)
+	_, err = w.Write([]byte(content))
+	c.Assert(err, IsNil)
+	c.Assert(w.Close(), IsNil)
+
+	h, err := sto.SetEncodedObject(obj)
+	c.Assert(err, IsNil)
+	return h
+}
+
+// storeTree writes a tree verbatim. Tree.Encode rejects only NUL, matching
+// upstream Git, which forbids just '/' and NUL inside a path component.
+func (s *DiffTreePathValidationSuite) storeTree(c *C, sto *memory.Storage, entries []TreeEntry) *Tree {
+	tree := &Tree{Entries: entries}
+
+	obj := sto.NewEncodedObject()
+	c.Assert(tree.Encode(obj), IsNil)
+
+	h, err := sto.SetEncodedObject(obj)
+	c.Assert(err, IsNil)
+
+	stored, err := GetTree(sto, h)
+	c.Assert(err, IsNil)
+	return stored
+}
+
+// TestDiffTree_ControlCharacterEntry checks that a tree entry name upstream
+// Git accepts does not abort a diff. DiffTree computes changes in memory and
+// never writes a name to disk, so the worktree-safety validation in
+// TreeWalker.Next does not apply to it.
+//
+// Before the fix this failed with:
+//
+//	from: invalid path "\x1b\x1b\x1b": contains control character
+func (s *DiffTreePathValidationSuite) TestDiffTree_ControlCharacterEntry(c *C) {
+	sto := memory.NewStorage()
+
+	before := s.storeBlob(c, sto, "one")
+	after := s.storeBlob(c, sto, "two")
+
+	from := s.storeTree(c, sto, []TreeEntry{
+		{Name: "\x1b\x1b\x1b", Mode: filemode.Regular, Hash: before},
+	})
+	to := s.storeTree(c, sto, []TreeEntry{
+		{Name: "\x1b\x1b\x1b", Mode: filemode.Regular, Hash: after},
+	})
+
+	changes, err := DiffTree(from, to)
+	c.Assert(err, IsNil)
+	c.Assert(changes, HasLen, 1)
+
+	action, err := changes[0].Action()
+	c.Assert(err, IsNil)
+	c.Assert(action.String(), Equals, "Modify")
+}
--- a/plumbing/object/tree.go
+++ b/plumbing/object/tree.go
@@ -474,6 +474,14 @@ type TreeWalker struct {
 	recursive bool
 	seen      map[plumbing.Hash]bool
 
+	// skipPathValidation disables the pathutil.ValidTreePath check in Next.
+	// It is set by inspection-only callers (e.g. the revlist object walk)
+	// that never funnel entry names into the filesystem and must enumerate
+	// trees faithfully, including entries with names upstream Git accepts
+	// but that are unsafe to materialise (control characters, `.git`-shaped
+	// names).
+	skipPathValidation bool
+
 	s storer.EncodedObjectStorer
 	t *Tree
 }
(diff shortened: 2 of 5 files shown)


r0h1tb:
Behavioural delta: `Push` no longer aborts when a reachable tree in history holds an entry name upstream Git accepts. Nothing else changes — validation stays on by default everywhere it already was.

Reproducing the failure without the fix, from a checkout of this branch:

```console
$ git stash                                              # revert source, keep tests
$ go test -count=1 ./plumbing/revlist/ -check.f ControlCharacter
FAIL: RevListSuite.TestRevListObjects_ControlCharacterInHistory
... ("invalid path \"\\x1b\\x1b\\x1b\": contains control character")
FAIL: RevListSuite.TestRevListObjects_ControlCharacterInPushedCommit
... ("invalid path \"\\x1b\\x1b\\x1b\": contains control character")
OOPS: 0 passed, 2 FAILED

$ git stash pop && git stash -- plumbing/object/treenoder.go
$ go test -count=1 ./plumbing/object/ -check.f DiffTreePathValidation
FAIL: DiffTreePathValidationSuite.TestDiffTree_ControlCharacterEntry
... ("from: invalid path \"\\x1b\\x1b\\x1b\": contains control character")
OOPS: 0 passed, 1 FAILED
```

Three things I would rather flag myself than have you find:

1. **This adds exported API to a stable line.** `TreeWalker.SkipPathValidation()` exists only because `revlist` is outside package `object` and cannot set the field the way #2222 did. If growing the v5 surface is unwelcome, an exported field or an `internal/` accessor works just as well — say which and I will redo it.

2. **My first regression test passed against the unfixed branch.** `Next` consults its `seen` set before validating, so reusing one blob across both trees meant the bad entry was skipped before the validator saw it. Both tests now use distinct blobs, which is what makes the historical case reach the validator at all.

3. **The second commit is scope beyond the issue.** It backports #2222, which never reached this branch, after checking the sibling call site. Drop it if you would rather track that separately — the first commit stands alone.

I also confirmed this does not affect `main`: the v6 object walk reads `Tree.Entries` directly and never goes through `TreeWalker`, so only `releases/v5.x` needs it.

pjbgf (approved):
@r0h1tb thanks for working on this. 🙇

PR #2309, opened 2026-08-10
plumbing: format/gitignore, Add Scope for per-directory ignore evaluation.

Soph:
> **Draft.** This adds and removes public API, which CONTRIBUTING lists under the [RFC process](https://github.com/go-git/go-git/tree/main/rfcs). Opening as a draft so the design and the measurements are reviewable; happy to write it up as an RFC instead if you would prefer that first.

Supersedes #2302, which fixed the same defect inside the existing flat-pattern design. Its own body argued for this approach once the numbers were in.

## API surface

| | |
| --- | --- |
| Added | `gitignore.Scope`, `NewScope`, `DirPatterns`, `RootPatterns`, `IgnoreFile`, `filesystem.Options.IgnoreScope` |
| Removed | `merkletrie/filesystem.Options.IgnoreMatcher` — present in `v6.0.0-alpha.5` but never in a stable release, so not a compatibility break ([#2048](https://github.com/go-git/go-git/pull/2048)) |
| Deprecated | `gitignore.ReadPatterns` (unchanged behaviour) |

## The problem

`Status` collected every ignore pattern in the worktree before diffing, then matched every entry against that one flat list. Two consequences.

**It cannot express an excluded parent.** gitignore(5): a file cannot be re-included once a parent directory of it is excluded. A flat list has nowhere to record that, so a negation anywhere in the repository can re-include a path git ignores:

```
.gitignore:  outer/ignored/
             !outer/ignored/keep.txt

$ git check-ignore -v outer/ignored/keep.txt
.gitignore:1:outer/ignored/	outer/ignored/keep.txt
```

git attributes the verdict to the rule that excluded the parent and ignores the negation. `main` reports `keep.txt` as not ignored.

**It walks the worktree twice** — once to gather patterns, once to diff — and the first walk descended into ignored trees the second one then skipped.

## The design

`gitignore.Scope` carries the patterns in effect for one directory *together with whether that directory is excluded*. `Descend` derives a subdirectory's scope; once excluded it stays excluded and `Match` answers without consulting a pattern, so a negation below cannot win.

`Descend` takes the subdirectory's own patterns as a function, not a slice, and does not call it when the subdirectory is excluded — the decision not to read an ignore file below an excluded directory lives in one place where a caller cannot get it wrong.

The noder resolves each directory's scope from the `ReadDir` listing it already takes, so `.gitignore` is opened only in directories actually visited and never below an excluded one, and the separate collection pass is gone.

This is git's structure. `prep_exclude` keeps a prefix-keyed `exclude_stack`, aborts before reading a directory's `.gitignore` when that directory already matched unless the match was a negation, and `last_matching_pattern` short-circuits on the excluded ancestor rather than recomputing it per query ([dir.c](https://github.com/git/git/blob/master/dir.c)).

## Numbers

`Status`, versus `main`:

| | `main` | this PR |
| --- | --- | --- |
| `StatusNestedIgnoredDir` 2000 ignored dirs | 252.1ms | **2.4ms** (-99.0%) |
| `StatusManyIgnoreFiles` 500 ignore files | 88.2ms | **52.5ms** (-40.4%) |
| `StatusLarge` 5000 files, no `.gitignore` | 23.1ms | **13.4ms** (-41.9%) |

`main` fails to scale on the first: ten times the directories cost it thirteen times the wall clock. The other two are shapes where pruning alone cannot help, since nothing is excluded wholesale — the saving there is the removed traversal and the smaller pattern set each entry is tested against.

Two of these benchmarks are new; the pre-existing `BenchmarkStatusIgnoredDir` could not show any of it, because its rule names a direct child of the root and its ignored tree is 21 directories wide.

`tests/gitcompare` measures the same shapes against the git binary. Read against its `Baseline/EmptyRepo` case, which is what a git *process* costs before it looks at a file (~11ms on macOS, a few on Linux):

| | git wall | git less baseline | go-git |
| --- | --- | --- | --- |
| `NestedIgnoredDir` | 12.1ms | ~1.2ms | 1.9ms |
| `ManyIgnoreFiles` | 33.1ms | ~22ms | 72.5ms |
| `LargeNoIgnoreFiles` | 18.8ms | ~8ms | 17.7ms |

## What this does not fix

`ManyIgnoreFiles` is still ~3x git's compute. Profiling says that is not the ignore machinery — reading the 500 `.gitignore` files is 22%, and git reads them too. The rest is elsewhere:

- **37% is `ObjectStorage.EncodedObject`**, one tree object per directory for the HEAD↔index diff. It scales with directory count, not file count: 1.4% at 20 directories, 37% at 500. go-git decodes the index's cache-tree extension into `idx.Cache` (`index/decoder.go:320`) but never reads it, so it redoes work git's cache-tree lets it skip.
- **95% of total time is syscalls**, `os.doInRoot` alone ~46%: every open goes through `BoundOS`/`os.Root`, which revalidates each path component. That is the deliberate symlink-safety boundary, but it means several syscalls where git pays one.

Both look worth their own issues; neither belongs here.

## Tests

- `gitignore`: `Scope` unit tests covering excluded-parent precedence, that `Descend` never reads an ignore file below an excluded directory, ordering, and the degenerate roots.
- `merkletrie/filesystem`: the nine existing ignore tests move to `IgnoreScope`, plus two the flat matcher could not express — a rule in a subdirectory's `.gitignore`, and a tracked entry forcing descent into an excluded directory where a negation must still not re-include a sibling. I checked the latter fails if excluded-parent precedence is removed.
- root: `Status` compared against `git ls-files --others --exclude-standard` over six layouts. **This found a real bug** — the root scope was omitted when the root declared no patterns, so a worktree whose only rules live below the root reported ignored files as untracked. Fixed in the same series.

Only pre-existing failure is `TestConformanceSuite/TestIgnoreDoubleStarPrefix`, which fails on `main` too (git 2.50.1). `golangci-lint` v2.11.4 reports 0 issues repo-wide, and every commit builds and tests independently.

## AI assistance disclosure

Per [AI_POLICY.md](https://github.com/go-git/go-git/blob/main/AI_POLICY.md), produced with AI assistance (Claude Opus 5), disclosed with an `Assisted-by:` trailer on each commit. Behaviour was verified against reference git two ways: oracle tests comparing `Status` and pattern verdicts against the git binary, and by reading `prep_exclude` in git's `dir.c` to confirm the design matches.

### DIFF
--- a/plumbing/format/gitignore/scope.go
+++ b/plumbing/format/gitignore/scope.go
@@ -1,0 +1,146 @@
+package gitignore
+
+import (
+	"os"
+	"slices"
+
+	"github.com/go-git/go-billy/v6"
+)
+
+// Scope is the set of ignore patterns in effect for the entries of a single
+// directory: those inherited from its ancestors, in ascending order of
+// priority, followed by those the directory declares itself.
+//
+// A Scope also records whether the directory it belongs to is excluded. That
+// is what a flat []Pattern cannot express. gitignore(5) states that a file
+// cannot be re-included once a parent directory of it is excluded, so once a
+// directory is excluded every entry below it is ignored whatever the patterns
+// there say. Reference git tracks the same state in its walk rather than
+// recomputing it per query: prep_exclude stops at the excluded directory and
+// last_matching_pattern returns its pattern directly (dir.c).
+//
+// A Scope is immutable and safe for concurrent use. Descend derives the Scope
+// for a subdirectory; the zero Scope matches nothing and is a valid root for a
+// walk with no patterns at all.
+type Scope struct {
+	// patterns is ancestors-then-own, in ascending order of priority.
+	patterns []Pattern
+	matcher  Matcher
+	// excluded reports whether this directory, or an ancestor of it, matched
+	// an exclude rule.
+	excluded bool
+}
+
+// NewScope returns the root Scope for a walk. base holds the patterns that
+// apply to the whole tree before any .gitignore is read: typically
+// .git/info/exclude, the core.excludesfile patterns, and any supplied by the
+// caller. They are ordered by ascending priority, as for NewMatcher.
+func NewScope(base []Pattern) *Scope {
+	// Cloned so the Scope really is immutable, as documented above: without it
+	// a caller mutating its own slice afterwards would change this Scope and
+	// every Scope derived from it, including concurrently. Descend already
+	// copies, so this is the only place the caller's memory could be shared.
+	base = slices.Clone(base)
+	return &Scope{patterns: base, matcher: NewMatcher(base)}
+}
+
+// Descend returns the Scope for the subdirectory of s at dir. dir is the full
+// path of the subdirectory from the root of the walk, so that patterns keep
+// matching against complete paths.
+//
+// readOwn supplies the patterns the subdirectory declares itself. It is called
+// only when the subdirectory is not excluded, because git does not read ignore
+// files below an excluded directory and nothing in one could change an
+// outcome. Passing it as a function rather than a slice keeps that decision
+// here, so callers cannot pay for a read that is then discarded. It may be nil
+// when the subdirectory has no ignore file, which callers already know from
+// the directory listing they hold.
+func (s *Scope) Descend(dir []string, readOwn func() ([]Pattern, error)) (*Scope, error) {
+	if s.excluded || s.matches(dir, true) {
+		// Nothing below an excluded directory can change an outcome, so the
+		// pattern set is frozen here and readOwn is never called.
+		return &Scope{patterns: s.patterns, matcher: s.matcher, excluded: true}, nil
+	}
+
+	if readOwn == nil {
+		return s, nil
+	}
+
+	own, err := readOwn()
+	if err != nil {
+		return nil, err
+	}
+	if len(own) == 0 {
+		return s, nil
+	}
+
+	patterns := slices.Concat(s.patterns, own)
+	return &Scope{patterns: patterns, matcher: NewMatcher(patterns)}, nil
+}
+
+// Excluded reports whether the directory this Scope belongs to, or an ancestor
+// of it, is excluded. Every entry below such a directory is ignored.
+func (s *Scope) Excluded() bool {
+	return s.excluded
+}
+
+// Match reports whether path is excluded by the patterns in effect. path is an
+// ordered sequence of the components of a complete path from the root of the
+// walk, and isDir reports whether its final component is a directory.
+//
+// Unlike a Matcher built from a flat pattern list, Match honours excluded
+// ancestors: below an excluded directory it reports true without consulting
+// any pattern, so a negation there cannot re-include the entry.
+func (s *Scope) Match(path []string, isDir bool) bool {
+	if s.excluded {
+		return true
+	}
+	return s.matches(path, isDir)
+}
+
+// Patterns returns the patterns in effect, ancestors first. The result must
+// not be modified.
+func (s *Scope) Patterns() []Pattern {
+	return s.patterns
+}
+
+func (s *Scope) matches(path []string, isDir bool) bool {
+	if s.matcher == nil {
+		return false
+	}
+	return s.matcher.Match(path, isDir)
+}
+
+// DirPatterns returns the patterns declared by the .gitignore of a single
+// directory, without recursing. A missing file is not an error and yields no
+// patterns. Use it to build the readOwn argument of Scope.Descend while walking
+// a tree; ReadPatterns is the eager whole-tree equivalent.
+func DirPatterns(fs billy.Filesystem, path []string) ([]Pattern, error) {
+	ps, err := readIgnoreFile(fs, path, gitignoreFile)
+	if err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
+	return ps, nil
+}
+
+// RootPatterns returns the patterns that apply to a whole worktree before any
+// .gitignore is consulted: .git/info/exclude followed by the .gitignore at the
+// root, in ascending order of priority. Missing files are not an error.
+func RootPatterns(fs billy.Filesystem) ([]Pattern, error) {
+	// Both files are opened directly rather than confirmed against a listing of
+	// the root. Two opens cost less than enumerating a directory that may hold
+	// thousands of entries, and the walk lists the root for its own purposes
+	// anyway, so a listing here would be the second of two.
+	//
+	// Errors from the exclude file are dropped, as ReadPatterns drops them: it
+	// is optional, and a worktree filesystem may refuse to open a path inside
+	// .git outright rather than reporting it as missing.
+	ps, _ := readIgnoreFile(fs, nil, infoExcludeFile)
+
+	root, err := DirPatterns(fs, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return append(ps, root...), nil
+}
--- a/plumbing/format/gitignore/scope_test.go
+++ b/plumbing/format/gitignore/scope_test.go
@@ -1,0 +1,262 @@
+package gitignore
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/go-git/go-billy/v6"
+	"github.com/go-git/go-billy/v6/memfs"
+	"github.com/go-git/go-billy/v6/util"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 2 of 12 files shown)


plumbing/format/gitignore/scope.go
```diff
@@ -0,0 +1,162 @@
...
+		return false
+	}
+	return s.matcher.Match(path, isDir)
+}
+
+// ReadDirPatterns reads the patterns declared by the .gitignore of a single
+// directory, without recursing. A missing file is not an error and yields no
+// patterns. Use it to build the own argument of Scope.Descend while walking a
+// tree; ReadPatterns is the eager whole-tree equivalent.
+func ReadDirPatterns(fs billy.Filesystem, path []string) ([]Pattern, error) {
```
pjbgf: From a API perspective I'm leaning more towards: 

```suggestion
func DirPatterns(fs billy.Filesystem, path []string) ([]Pattern, error) {
```

plumbing/format/gitignore/scope.go
```diff
@@ -0,0 +1,162 @@
...
+	return ps, nil
+}
+
+// ReadRootPatterns reads the patterns that apply to a whole worktree before
+// any .gitignore is consulted: .git/info/exclude followed by the .gitignore at
+// the root, in ascending order of priority. Missing files are not an error.
+//
+// Both are read only when the root listing shows them, so a worktree with
+// neither costs a single ReadDir, as in ReadPatterns.
+func ReadRootPatterns(fs billy.Filesystem) ([]Pattern, error) {
```
pjbgf: Same here:

```suggestion
func RootPatterns(fs billy.Filesystem) ([]Pattern, error) {
```

We are returning the root patterns, if we are reading it on demand or from a cache, feels like an implementation detail.

plumbing/format/gitignore/scope.go
```diff
@@ -0,0 +1,162 @@
...
+	entries, err := fs.ReadDir("")
```
pjbgf: On large dirs this can be costly. Let's use [util.Walk](https://github.com/go-git/go-billy/blob/main/util/walk.go#L60), so that we walk one entry at a time instead of loading all of them into memory.

utils/merkletrie/filesystem/node.go
```diff
@@ -35,14 +36,22 @@ type Options struct {
...
 	Index *index.Index
 
-	// IgnoreMatcher, if non-nil, is consulted while walking the tree.
-	// Untracked entries (files or directories) that match the matcher are
-	// excluded from the walk so callers do not have to descend into large
-	// gitignored directories like node_modules. Tracked entries are always
-	// walked even if they match, so modifications to them are still
-	// reported. Requires Index to be set: without an index there is no way
-	// to identify tracked entries, so the matcher is treated as a no-op.
-	IgnoreMatcher gitignore.Matcher
```
pjbgf, line 45: Non-breaking as never released on an official version: https://github.com/go-git/go-git/pull/2048

pjbgf (changes requested):
Thanks for working on this @Soph. A few nits below, otherwise LGTM

plumbing/format/gitignore/scope.go
Soph: Done — renamed to `DirPatterns`.

plumbing/format/gitignore/scope.go
Soph: Done — renamed to `RootPatterns`. Agreed on the reasoning, and it applies to `DirPatterns` too: what the caller wants is the patterns, not a statement about how they were obtained.

utils/merkletrie/filesystem/node.go
Soph: Good — thanks for confirming. I have dropped the "breaking" framing from the PR description accordingly; #2048 landed after `v6.0.0-alpha.5`, so no released version ever exposed the field.

Soph: Correction to my previous reply — I had the ordering backwards. #2048 merged 2026-05-04, `v6.0.0-alpha.5` was published 2026-07-29, so the field *is* present in that alpha rather than unreleased.

Your point still holds on the reading that matters: no stable release ever exposed it, so removing it in v6 is fair game. Worth noting for anyone on the alpha that only `merkletrie/filesystem.Options` is affected — `storage/filesystem.Options` is a different type and untouched.

plumbing/format/gitignore/scope.go
Soph: The concern is right and I have taken it, but not via `util.Walk` — I do not think that helps here, for two reasons.

`util.Walk` still materialises the whole directory. It calls `readdirnames`, which does `fs.ReadDir(dir)` and then allocates a second slice of just the names ([glob.go:99](https://github.com/go-git/go-billy/blob/main/util/glob.go#L99)), and it adds an `Lstat` per entry on top. Its own doc says as much: *"requires Walk to read an entire directory into memory before proceeding to walk that directory."* `billy.Filesystem` has no streaming readdir, so there is nowhere for one-entry-at-a-time to come from.

It is also recursive, and this function only needs the root, so it would descend the entire worktree.

What the concern points at is that the listing should not be there at all. `RootPatterns` was reading the whole root just to answer "do `.git` and `.gitignore` exist", and the walk lists the root again for its own purposes straight afterwards — so it was the first of two listings. It now opens the two files directly:

```go
func RootPatterns(fs billy.Filesystem) ([]Pattern, error) {
	ps, _ := readIgnoreFile(fs, nil, infoExcludeFile)

	root, err := DirPatterns(fs, nil)
	if err != nil {
		return nil, err
	}

	return append(ps, root...), nil
}
```

Two opens, no enumeration, so a root with thousands of entries costs the same as an empty one. The dropped error is deliberate and matches what `ReadPatterns` already does: the exclude file is optional, and a worktree filesystem may refuse to open a path inside `.git` outright rather than reporting it missing — I hit exactly that on `BoundOS` while building this.

Note this is only the root. Per-directory rules are read in the walk via `DirPatterns`, and there the noder passes the listing it already took, so #2284's no-blind-opens property is preserved without any extra enumeration.

plumbing/format/gitignore/scope_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+		entries, err := fs.ReadDir(joined)
+		require.NoError(t, err)
+
+		hasIgnoreFile := map[string]bool{}
+		for _, e := range entries {
+			if e.IsDir() {
+				hasIgnoreFile[e.Name()] = false
+			}
+		}
```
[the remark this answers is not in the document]
Soph: Removed — it was dead: written to and never read.

One correction to the reasoning though: this would not have failed compilation. Go only rejects a declared variable that is never *used*, and indexing a map to assign counts as a use, so the code built fine (CI was green on it). Dead rather than illegal — still worth deleting.

plumbing/format/gitignore/scope.go
```diff
@@ -0,0 +1,141 @@
...
+	excluded bool
+}
+
+// NewScope returns the root Scope for a walk. base holds the patterns that
+// apply to the whole tree before any .gitignore is read: typically
+// .git/info/exclude, the core.excludesfile patterns, and any supplied by the
+// caller. They are ordered by ascending priority, as for NewMatcher.
+func NewScope(base []Pattern) *Scope {
+	return &Scope{patterns: base, matcher: NewMatcher(base)}
+}
```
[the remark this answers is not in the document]
Soph: Good catch, this one was a real defect: the doc promised immutability and concurrency safety that `NewScope` did not deliver.

Fixed by cloning. `Descend` already copies via `slices.Concat`, so the root's `base` was the only place caller memory was shared:

```go
base = slices.Clone(base)
return &Scope{patterns: base, matcher: NewMatcher(base)}
```

Added `TestNewScopeCopiesBase` to pin it, and confirmed it fails without the clone rather than passing vacuously.

utils/merkletrie/filesystem/node.go
```diff
@@ -230,18 +255,60 @@ func (n *node) calculateChildren() error {
 // name should be skipped because it matches the configured ignore matcher
 // AND has no entry in the index. Tracked entries are never skipped so
 // modifications to them are still reported.
```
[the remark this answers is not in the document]
Soph: Right, and thanks — the comment was orphaned when `resolveScope` was inserted above `shouldSkipIgnored`, so it was documenting the wrong function.

Moved back to `shouldSkipIgnored` and reworded. It now says "ignore scope in effect" in the commit that removes `IgnoreMatcher`, and the intermediate commit that still has both fields says "ignore rules in effect", so the wording is accurate at every point in the series rather than only at the tip.

pjbgf (approved):

PR #2310, opened 2026-08-11
gitignore: dirOnly re-include patterns must not cascade to descendants

### LINKED ISSUE #2112: go-git ignores files inside un-ignored directory with wildcard pattern

thesahibnanda-max:
Fixes #2112

## Bug

A `.gitignore` re-include pattern like `!dir/` is meant to restore only the directory entry itself, not everything nested inside it. `simpleNameMatch` let a `dirOnly` inclusion pattern that matched an *ancestor* path component stand in as a match for the full descendant path, so:
```
*
!my-dir/
```
wrongly reported files under `my-dir/` (e.g. `my-dir/sub/file.txt`) as untracked in `Worktree.Status()` instead of ignored, even though real `git` keeps them ignored.

## Fix

In `plumbing/format/gitignore/pattern.go`, `simpleNameMatch` now skips (rather than matches) when a `dirOnly` inclusion pattern hits a non-final path component — it only applies at the exact matched entry, not its descendants. Exclusion patterns are unaffected.

## Testing
- New unit tests in `pattern_test.go` covering the pattern in isolation.
- New case in `conformance_test.go`, cross-checked against the real `git` binary via the existing oracle harness.
- New end-to-end regression test in `worktree_status_test.go` reproducing the original report via `Worktree.Status()`.
- `go test ./plumbing/format/gitignore/...` and `go test .` pass (aside from one pre-existing, unrelated oracle mismatch confirmed present on unmodified `main` too).

### DIFF
--- a/plumbing/format/gitignore/conformance_test.go
+++ b/plumbing/format/gitignore/conformance_test.go
@@ -742,6 +742,37 @@ func (s *ConformanceSuite) TestIgnoreDoubleStarReinclude() {
 	}
 }
 
+// TestIgnoreReincludedDirDoesNotCascade verifies that a dirOnly re-include
+// pattern (e.g. "!dir/") does not extend a match to the directory's
+// descendants. Files and subdirectories nested inside a directory named by
+// such a pattern remain subject to the earlier "*" exclusion, matching
+// Git's documented behavior that re-including a directory does not
+// implicitly re-include its contents.
+//
+// This does not assert on the directory entry itself ("my-dir"): Git has a
+// separate, narrower quirk where a bare "*" specifically resists being
+// overridden by a later re-include pattern for the exact excluded entry,
+// even with no descendants involved. That is unrelated to the cascading
+// bug fixed here (go-git#2112, which is about nested files) and is left
+// alone.
+func (s *ConformanceSuite) TestIgnoreReincludedDirDoesNotCascade() {
+	patterns := []string{"*", "!my-dir/"}
+	m := s.createMatcher(patterns)
+
+	tests := []struct {
+		path    string
+		isDir   bool
+		ignored bool
+		desc    string
+	}{
+		{"my-dir/sub", true, true, "a subdirectory is not re-included"},
+		{"my-dir/sub/file.txt", false, true, "a nested file remains ignored"},
+	}
+	for _, tt := range tests {
+		s.assertIgnore(m, patterns, tt.path, tt.isDir, tt.ignored, tt.desc)
+	}
+}
+
 // TestIgnoreDoubleStarPrefix verifies that foo**/bar matches foo/bar (and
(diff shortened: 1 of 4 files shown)


thesahibnanda-max:
Closing in favor of #2311 — same change, rebased onto a fresh branch with a corrected commit message (package-prefixed per CONTRIBUTING.md) and DCO sign-off, which this PR's CI checks were failing on.

PR #2311, opened 2026-08-11
plumbing: gitignore, dirOnly re-include patterns must not cascade to descendants. Fixes #2112

### LINKED ISSUE #2112: go-git ignores files inside un-ignored directory with wildcard pattern

thesahibnanda-max:
Fixes #2112

(Supersedes #2310, closed — that PR's commit didn't follow the repo's commit-message/DCO conventions; this one carries the same change with a corrected, signed-off commit.)

## Problem

Given a `.gitignore` containing:
```
*
!my-dir/
```
real `git` keeps everything under `my-dir/` ignored except the directory entry itself is un-excluded for traversal purposes — but files nested inside it (e.g. `my-dir/sub/file.txt`) remain subject to the `*` exclusion and stay ignored. go-git instead reported those nested files as **untracked**, so `Worktree.Status()` surfaced files that real `git status` would never show.

## Root cause

`plumbing/format/gitignore/pattern.go`, `simpleNameMatch`:

```go
func (p *pattern) simpleNameMatch(path []string, isDir bool) bool {
	for i, name := range path {
		if !wildmatch(p.pattern[0], name) {
			continue
		}
		if p.dirOnly && !isDir && i == len(path)-1 {
			return false
		}
		return true
	}
	return false
}
```

This loop lets a bare (non-glob) pattern match a name at *any* depth in the path component list — which is correct and necessary for patterns like `node_modules` matching `a/node_modules/file.txt` at any nesting level. The bug: that "ancestor match stands in for the whole descendant path" behavior was applied uniformly, but it's only valid for **exclusion** patterns (excluding an ancestor directory correctly excludes everything below it too). For a `dirOnly` **inclusion** pattern (`!dir/`), a match against an *ancestor* component (not the final path element) was also being treated as a match for the whole descendant path — incorrectly resurrecting files nested arbitrarily deep under a re-included directory name, even when nothing else in the pattern list re-includes them specifically.

## Fix

`simpleNameMatch` now treats a `dirOnly` inclusion pattern's match against a non-final (ancestor) path component as no match for that call, instead of a match for the whole path:

```go
last := i == len(path)-1
if p.dirOnly && p.inclusion && !last {
	// A dirOnly re-include pattern restores only the directory entry
	// itself; it must not cascade to the directory's descendants.
	continue
}
```

Exclusion patterns (`p.inclusion == false`) are completely unaffected — an excluded ancestor still correctly excludes its descendants, matching existing behavior and existing tests (e.g. `TestSimpleMatch_inTheMiddle_dirWanted`).

### Scope note

I verified this fix against the real `git` binary (this package already has a `conformance_test.go` harness that cross-checks against `git check-ignore`/`git status` when available). While investigating, I found real `git` has a separate, narrower quirk: a bare `*` pattern specifically resists being overridden by *any* later negation for the exact directory entry it matched (i.e. even `!my-dir/` alone, with **no** nested descendants involved, doesn't un-ignore `my-dir` itself when preceded by `*` — confirmed with `git status --porcelain --ignored`). Other exclude patterns (a literal name, or a non-`*` wildcard like `m*`) don't have this quirk; negation works normally against them. That's unrelated to the cascading-into-descendants bug this issue reports, so I deliberately left it alone and didn't assert on it in the new tests, to keep this PR minimal and precisely scoped to #2112.

## Tests added

- `plumbing/format/gitignore/pattern_test.go`: unit-level coverage of the pattern in isolation (ancestor match no longer cascades; the directory itself, matched directly, still returns `Include`).
- `plumbing/format/gitignore/conformance_test.go`: a new case cross-checked against the real `git` binary via the existing oracle harness.
- `worktree_status_test.go`: an end-to-end regression test reproducing the original report through `Worktree.Status()`.

## Testing performed

- `go test ./plumbing/format/gitignore/...` — passes (one pre-existing, unrelated `TestIgnoreDoubleStarPrefix` oracle mismatch confirmed present on unmodified `main` too, via `git stash`).
- `go test .` — no new failures versus a baseline run on unmodified `main` (same pre-existing, environment-specific failures on both, verified by diffing the failing-test lists).
- `go vet ./...` — clean (one pre-existing, unrelated warning in Windows SSH agent code, not part of this change).

### DIFF
--- a/plumbing/format/gitignore/conformance_test.go
+++ b/plumbing/format/gitignore/conformance_test.go
@@ -742,6 +742,37 @@ func (s *ConformanceSuite) TestIgnoreDoubleStarReinclude() {
 	}
 }
 
+// TestIgnoreReincludedDirDoesNotCascade verifies that a dirOnly re-include
+// pattern (e.g. "!dir/") does not extend a match to the directory's
+// descendants. Files and subdirectories nested inside a directory named by
+// such a pattern remain subject to the earlier "*" exclusion, matching
+// Git's documented behavior that re-including a directory does not
+// implicitly re-include its contents.
+//
+// This does not assert on the directory entry itself ("my-dir"): Git has a
+// separate, narrower quirk where a bare "*" specifically resists being
+// overridden by a later re-include pattern for the exact excluded entry,
+// even with no descendants involved. That is unrelated to the cascading
+// bug fixed here (go-git#2112, which is about nested files) and is left
+// alone.
+func (s *ConformanceSuite) TestIgnoreReincludedDirDoesNotCascade() {
+	patterns := []string{"*", "!my-dir/"}
+	m := s.createMatcher(patterns)
+
+	tests := []struct {
+		path    string
+		isDir   bool
+		ignored bool
+		desc    string
+	}{
+		{"my-dir/sub", true, true, "a subdirectory is not re-included"},
+		{"my-dir/sub/file.txt", false, true, "a nested file remains ignored"},
+	}
+	for _, tt := range tests {
+		s.assertIgnore(m, patterns, tt.path, tt.isDir, tt.ignored, tt.desc)
+	}
+}
+
 // TestIgnoreDoubleStarPrefix verifies that foo**/bar matches foo/bar (and
 // foo<anything>/bar) but does not match foobar — i.e. ** does not collapse
 // across the slash boundary into a prefix.
--- a/plumbing/format/gitignore/pattern.go
+++ b/plumbing/format/gitignore/pattern.go
@@ -483,7 +483,15 @@ func (p *pattern) simpleNameMatch(path []string, isDir bool) bool {
 		if !wildmatch(p.pattern[0], name) {
 			continue
 		}
-		if p.dirOnly && !isDir && i == len(path)-1 {
+		last := i == len(path)-1
+		if p.dirOnly && p.inclusion && !last {
+			// A dirOnly re-include pattern (e.g. "!dir/") restores only the
+			// directory entry itself, not everything nested inside it, so a
+			// match against an ancestor component must not be treated as a
+			// match for the full descendant path.
+			continue
+		}
+		if p.dirOnly && !isDir && last {
 			return false
 		}
 		return true
--- a/plumbing/format/gitignore/pattern_test.go
+++ b/plumbing/format/gitignore/pattern_test.go
@@ -21,6 +21,18 @@ func (s *PatternSuite) TestSimpleMatch_inclusion() {
 	s.Equal(Include, r)
 }
 
+func (s *PatternSuite) TestSimpleMatch_inclusion_dirOnly_atStart() {
+	p := ParsePattern("!dir/", nil)
+	r := p.Match([]string{"dir"}, true)
+	s.Equal(Include, r)
+}
+
+func (s *PatternSuite) TestSimpleMatch_inclusion_dirOnly_doesNotCascadeToDescendants() {
+	p := ParsePattern("!dir/", nil)
+	r := p.Match([]string{"dir", "sub", "file.txt"}, false)
+	s.Equal(NoMatch, r)
+}
+
 func (s *PatternSuite) TestMatch_domainLonger_mismatch() {
 	p := ParsePattern("value", []string{"head", "middle", "tail"})
 	r := p.Match([]string{"head", "middle"}, false)
--- a/plumbing/format/gitignore/scope.go
+++ b/plumbing/format/gitignore/scope.go
@@ -62,20 +62,67 @@ func (s *Scope) Descend(dir []string, readOwn func() ([]Pattern, error)) (*Scope
 		return &Scope{patterns: s.patterns, matcher: s.matcher, excluded: true}, nil
 	}
 
-	if readOwn == nil {
-		return s, nil
+	patterns := dropExhausted(s.patterns, dir)
+
+	var own []Pattern
+	if readOwn != nil {
+		var err error
+		own, err = readOwn()
+		if err != nil {
+			return nil, err
+		}
 	}
 
-	own, err := readOwn()
-	if err != nil {
-		return nil, err
-	}
 	if len(own) == 0 {
-		return s, nil
+		if len(patterns) == len(s.patterns) {
+			return s, nil
+		}
+		return &Scope{patterns: patterns, matcher: NewMatcher(patterns)}, nil
 	}
 
-	patterns := slices.Concat(s.patterns, own)
+	patterns = slices.Concat(patterns, own)
 	return &Scope{patterns: patterns, matcher: NewMatcher(patterns)}, nil
+}
+
+// dropExhausted removes patterns from patterns that can no longer
+// distinguish dir from anything below it. A glob pattern (one containing
+// "/") is anchored to a fixed number of path components; once dir's depth
+// reaches or exceeds domain-plus-pattern length, globMatch has nothing left
+// to consume, and its "leftover suffix still counts as matched" behavior —
+// needed so an excluded directory excludes its descendants — would
+// otherwise let the pattern wrongly reassert dir's old verdict past a
+// boundary a more specific, later-checked pattern has since overridden.
+// See TestScopeOverriddenGlobDoesNotReclaimDescendants.
+//
+// A pattern using "**" is kept regardless of depth: it has unbounded reach
+// by design. A non-glob pattern is also always kept: matching at any depth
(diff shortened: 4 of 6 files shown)


thesahibnanda-max:
The `test` CI jobs on the first commit were failing for a real reason, not flakiness: `TestWorktreeSuite/TestStatusIgnored` regressed.

That test covers a case real `git` genuinely supports — a root-anchored glob (`vendor/g*/`) excludes two sibling directories, and a nested `vendor/.gitignore`'s `!github.com/` re-includes one of them specifically. My first commit's fix for #2112 (stop a `dirOnly` inclusion pattern's ancestor match from cascading to descendants) was correct for the `*` + `!my-dir/` case in the issue, but too broad: it also blocked this legitimate, different kind of override.

The second commit adds the missing piece. `globMatch`'s "a leftover, unconsumed path suffix still counts as matched" behavior is what makes an excluded directory correctly exclude everything below it — but nothing stopped a directory-anchored glob pattern from reasserting that same old verdict on a directory's contents even after a more specific, nested pattern had already overridden that exact directory. A single `Pattern` can't know this on its own; only `Scope` has the cross-pattern, cross-directory context needed, so `Scope.Descend` now drops glob patterns whose anchor is fully consumed by the directory just resolved (patterns using `**`, or non-glob patterns that are meant to match at any depth, are left untouched). It's the mirror image of the existing `excluded` sticky flag: that one pins an exclude verdict forward past a resolved boundary; this pins the boundary itself.

Verified `TestStatusIgnored` passes again, added a new `Scope`-level test (`TestScopeOverriddenGlobDoesNotReclaimDescendants`) cross-checked against `git status --porcelain --ignored`, and hand-traced the change against every existing case in `scope_test.go` to confirm no behavior changes there. Also ran the full suite and diffed failures against a clean baseline — no new regressions beyond what's already broken on `main` today (e.g. `TestStatusMatchesReferenceGitForIgnoreLayouts`, which I confirmed also fails identically on a fresh, unmodified clone of `main` — unrelated to this PR).

KaizenShogun:
The bug you're fixing is real, your diagnosis matches what I measure, and the patch fixes more than the issue asks for. It also takes away one case that `main` currently gets right, and I'd rather say so before it lands than after. Numbers instead of an opinion, and a two-line repro at the end.

**Setup.** One Go helper (`ReadPatterns` + `Matcher.Match`, plus `Worktree.Status()` as a second entry point), built three times from the *same* source, changing only the tree it points at: `refs/heads/main` fetched today, `refs/pull/2311/merge` (main + both your commits, `4fc02ab` and `502b0e5`, as GitHub computes the merge), and `refs/pull/2318/merge` for contrast. I checked the drift rather than assuming it: the `+2311` tree differs from `main` in exactly the five files your PR touches.

The corpus is 33 real repositories' whole `.gitignore` trees materialised on disk — root plus nested rule files, files and directories variants — 4,463 matcher queries and 2,224 through `Status()`. Oracle is `git check-ignore -v --non-matching --no-index` (git 2.55.0), asked again in every tree.

| column | matcher | `Status()` |
|---|---|---|
| `main` | 2 / 4,463 | 1 / 2,224 |
| `+2311` | **4** / 4,463 — 0 fixed, **2 broken** | 1 / 2,224 |

Compared as sets, not counts: the two `main` already had are untouched, and two new ones appear.

**First, what it fixes.** #2112, verbatim, and a variant the issue doesn't mention:

| `.gitignore` | query | git | `main` | `+2311` |
|---|---|---|---|---|
| `*` + `!my-dir/` | `my-dir/sub/file.txt` | ignored | **not ignored** | ignored ✔ |
| `*` + `!my-dir/` | `my-dir/file.txt` | ignored | **not ignored** | ignored ✔ |
| `*.tmp` + `!my-dir/` | `my-dir/sub/a.tmp` | ignored | **not ignored** | ignored ✔ |
| `*.tmp` + `!my-dir/` | `my-dir/sub/file.txt` | not ignored | not ignored | not ignored |

That last row is the interesting one, and your patch gets it right.

**What it breaks.** Same shape, one thing changed — the pattern that excluded the directory now matches *only* the directory:

```sh
mkdir -p t/sub/.idea && cd t && git init -q
printf '.idea/\n!.idea/\n' > .gitignore
touch sub/.idea/f.txt
git check-ignore -v --no-index -- sub/.idea/f.txt   # no match: git does NOT ignore it
```

`Matcher.Match([]string{"sub", ".idea", "f.txt"}, false)` → `main`: `false` (agrees with git). `+2311`: `true`. Same answer through `Status()`. It doesn't depend on the name or on the depth (`sub/.idea/`, `sub/a/b/.idea/` behave alike), and it goes away if the negation loses its trailing slash (`!.idea` is correct in both columns) — which points at the `dirOnly` path your change touches.

The real-world case the bench found is the same thing across two rule files, which is what `python/cpython` ships today:

```
.gitignore                             ->  .idea/
Platforms/Android/testbed/.gitignore   ->  !.idea/
query: Platforms/Android/testbed/_gic_deep/.idea/inspectionProfiles/
git: not ignored   main: not ignored   +2311: ignored
```

**Where I think the line is.** Your write-up frames the fix as "a `dirOnly` re-include matching an *ancestor* component must not stand in as a match for the whole descendant path". The measurements above fit a slightly different rule: git re-includes the *directory*, and then re-evaluates what's inside against every pattern again. In #2112 the descendants stay ignored because `*` matches `sub` and `file.txt` on their own — not because `!my-dir/` stops counting. Swap `*` for `*.tmp` and git's answer splits down that line, which is the row your patch already gets right. The case that breaks is the one where nothing but the directory pattern ever matched the contents: after `!.idea/` re-includes the directory, no rule is left to ignore `f.txt`.

**How much is at stake.** Only 2 of 4,463 queries move, so this could read as a corner. But the construct isn't rare: 10 of the 33 repositories in the corpus write at least one `!something/` — airflow, elasticsearch, grafana, node, vscode, next.js, supabase, cpython, gradio, pathspec — 24 lines across 12 rule files. The bench only materialises the exact combination (excluded by a directory-only pattern, re-included by `!dir/`, queried below it) in cpython; that's a property of my corpus, not of the wild.

I'm an interested party, not a maintainer: I opened nothing here, I ran the bench because I keep one. The tests, the corpus and the go-git adapter are public if you want to reproduce any row: https://github.com/KaizenShogun/gitignore-conformance

PR #2318, opened 2026-08-12
plumbing: format/gitignore, Backtrack non-trailing ** expansions. Fixes #2317

### LINKED ISSUE #2317: gitignore: non-trailing `**` does not backtrack — `**/a/b` fails to match `a/x/a/b`
### Summary

In `plumbing/format/gitignore`, a non-trailing `**` locks onto the **first** path component that matches the segment following it. When a later segment then fails, the whole match is reported as a non-match instead of retrying the `**` expansion from the next candidate component. Real git backtracks here (wildmatch expands `**` recursively), so go-git disagrees with `git check-ignore` on common patterns like `**/a/b`.

### Expected behaviour

gitignore(5): *"A slash followed by two consecutive asterisks then a slash matches **zero or more directories**. For example, `a/**/b` matches `a/b`, `a/x/b`, `a/x/y/b` and so on."* — `**` is an existential "zero or more", so every possible expansion must be tried.

Verified against real git (2.48.1, same result on 2.39.5; these patterns keep `**` in the positions git documents, so they are unaffected by the `foo**/bar` wildmatch corner case git fixed in 2.52.0):

```console
$ git init -q repro && cd repro
$ echo '**/a/b' > .gitignore
$ git check-ignore -v a/x/a/b
.gitignore:1:**/a/b	a/x/a/b        # ignored
$ echo 'a/**/b/c' > .gitignore
$ git check-ignore -v a/b/x/b/c
.gitignore:1:a/**/b/c	a/b/x/b/c    # ignored
```

### Actual behaviour

```go
package main

import (
	"fmt"
	"strings"

	"github.com/go-git/go-git/v6/plumbing/format/gitignore"
)

func main() {
	for _, c := range []struct{ pattern, path string }{
		{"**/a/b", "a/x/a/b"},
		{"**/foo/bar", "foo/baz/foo/bar"},
		{"a/**/b/c", "a/b/x/b/c"},
		{"**/node_modules/lodash", "web/node_modules/x/node_modules/lodash"},
	} {
		p := gitignore.ParsePattern(c.pattern, nil)
		r := p.Match(strings.Split(c.path, "/"), false)
		fmt.Printf("%-24q vs %-42q => %v\n", c.pattern, c.path, r == gitignore.Exclude)
	}
}
```

```
"**/a/b"                 vs "a/x/a/b"                                  => false   // git: ignored
"**/foo/bar"             vs "foo/baz/foo/bar"                          => false   // git: ignored
"a/**/b/c"               vs "a/b/x/b/c"                                => false   // git: ignored
"**/node_modules/lodash" vs "web/node_modules/x/node_modules/lodash"   => false   // git: ignored
```

Because `Worktree.Status` consumes this matcher, files git ignores show up as untracked (the inverse of #2278).

### Root cause

`pattern.go`, `globMatch`: when `canTraverse` is set (a non-trailing `**` was just consumed), the loop scans forward and `break`s on the **first** component matching the next segment. If a later segment fails afterwards, the function returns `false` without ever retrying the remaining candidates:

https://github.com/go-git/go-git/blob/34770e01d7c205c5d27f0134c7b0b6b91ef43dd3/plumbing/format/gitignore/pattern.go#L523-L540

With `**/a/b` against `a/x/a/b`: the leading `a` is locked, `b` vs `x` fails, and the correct expansion (`**` → `a/x`) is never attempted.

The existing conformance cases all use paths where the segment after `**` occurs exactly once, which is why this never trips the suite.

### Environment

go-git `main` @ 34770e01d7c205c5d27f0134c7b0b6b91ef43dd3, Go 1.26.5, darwin/arm64. Oracle: git 2.48.1 and 2.39.5 (Apple).

I have a minimal fix with regression tests ready and will open a PR referencing this issue.

*Disclosure: this report was prepared with AI assistance (Claude, via Cursor). I reproduced the behaviour locally and reviewed every claim.*

codeAnqiang-ma:
Fixes #2317

### What

In `plumbing/format/gitignore`, `globMatch` handled a non-trailing `**` by locking onto the **first** path component that matched the following segment. When a later segment then failed, the whole match was rejected without retrying the remaining candidates, so patterns like `**/a/b` did not match `a/x/a/b` (or `**/node_modules/lodash` vs `web/node_modules/x/node_modules/lodash`) even though `git check-ignore` ignores them.

This PR threads the segment slice through `globMatch` and, at a non-trailing `**`, recurses over every candidate expansion instead of committing to the first one. All other behaviour is preserved line-for-line: empty segments still reset traversal, a trailing `**` still requires a remaining component or a directory endpoint (#2278), the "files matching dir globs don't match" rule and the final `dirOnly` check are untouched.

### Why this shape

`**` is documented as matching *zero or more directories* (gitignore(5)), i.e. an existential choice — a correct matcher must be able to retry longer expansions when a shorter one dead-ends. Canonical git implements this with recursion in [`wildmatch.c`](https://github.com/git/git/blob/v2.54.0/wildmatch.c) (the `**` branch of `dowild` tries the remainder at every position). Recursing on the existing segment matcher is the minimal equivalent here and keeps the subtle semantics that previous fixes (#2278, #154) encoded in this function.

The new conformance cases run against the `git check-ignore` oracle like the rest of the suite. These patterns keep `**` in the documented positions, so they are unaffected by the pre-2.52 wildmatch corner case the suite already guards against.

### Verification against real git

```console
$ git init -q repro && cd repro
$ echo '**/a/b' > .gitignore
$ git check-ignore -v a/x/a/b
.gitignore:1:**/a/b	a/x/a/b        # ignored (git 2.48.1; same on 2.39.5)
```

go-git on `main` returns no match for the same input; with this fix it matches.

### Tests

Regression tests fail on `main` before the fix and pass with it:

```console
$ go test -run 'TestConformanceSuite/(TestIgnoreDoubleStarBacktracking|TestIgnoreDoubleStarMiddleBacktracking)' -v ./plumbing/format/gitignore/
--- PASS: TestConformanceSuite (0.45s)
    --- PASS: TestConformanceSuite/TestIgnoreDoubleStarBacktracking (0.27s)
    --- PASS: TestConformanceSuite/TestIgnoreDoubleStarMiddleBacktracking (0.17s)

# before the fix (pattern.go stashed):
    --- FAIL: TestConformanceSuite/TestIgnoreDoubleStarBacktracking
    --- FAIL: TestConformanceSuite/TestIgnoreDoubleStarMiddleBacktracking
```

Full run: `go test -short ./...` passes on darwin/arm64 with Go 1.26.5, except three tests that need network access to `github.com/git-fixtures` and hit a transient TLS timeout; all three passed on a re-run (`TestFetchByHashThenResolveRevision`, `TestWorktreeSuite/TestCheckoutRelativePathSubmoduleInitialized`, `TestRepositorySuite/TestCloneMirror`). Not run: `plumbing/transport/git` (requires a local `git daemon`, per AGENTS.md). `go vet` and `gofmt` are clean. The full (non-`-short`) gitignore conformance suite passes against the git 2.52+ oracle expectations; on this machine's older git the pre-existing `TestIgnoreDoubleStarPrefix` oracle note applies, unrelated to this change.

### Disclosure

This contribution was produced with AI assistance (Claude Fable 5, via Cursor); the commit carries an `Assisted-by:` trailer per AI_POLICY.md. I reproduced the bug locally, verified the behaviour against the reference git implementation, and reviewed every line.

### DIFF
--- a/plumbing/format/gitignore/conformance_test.go
+++ b/plumbing/format/gitignore/conformance_test.go
@@ -765,6 +765,52 @@ func (s *ConformanceSuite) TestIgnoreDoubleStarPrefix() {
 	}
 }
 
+// TestIgnoreDoubleStarBacktracking verifies that a non-trailing ** retries
+// later expansions when the component first matching the next segment
+// dead-ends further down the pattern — **/a/b must match a/x/a/b even though
+// the leading a/x does not complete the match.
+func (s *ConformanceSuite) TestIgnoreDoubleStarBacktracking() {
+	patterns := []string{"**/a/b"}
+	m := s.createMatcher(patterns)
+
+	tests := []struct {
+		path    string
+		ignored bool
+		desc    string
+	}{
+		{"a/b", true, "zero directories before the segment"},
+		{"a/x/a/b", true, "retry after the leading a dead-ends"},
+		{"z/a/x/a/b", true, "retry below an unrelated prefix"},
+		{"a/x/b", false, "no expansion completes the pattern"},
+		{"a/x/a/c", false, "tail segment differs at every candidate"},
+	}
+	for _, tt := range tests {
+		s.assertIgnore(m, patterns, tt.path, false, tt.ignored, tt.desc)
+	}
+}
+
+// TestIgnoreDoubleStarMiddleBacktracking exercises the same retry for a **
+// between literal segments, where the segment after ** also names a component
+// that appears before the real match.
+func (s *ConformanceSuite) TestIgnoreDoubleStarMiddleBacktracking() {
+	patterns := []string{"a/**/b/c"}
+	m := s.createMatcher(patterns)
+
+	tests := []struct {
+		path    string
+		ignored bool
+		desc    string
+	}{
+		{"a/b/c", true, "zero directories consumed by **"},
+		{"a/b/x/b/c", true, "retry after the first b dead-ends"},
+		{"a/b/b/c", true, "adjacent duplicate segment"},
+		{"a/b/x/c", false, "no b directly above c"},
+	}
+	for _, tt := range tests {
+		s.assertIgnore(m, patterns, tt.path, false, tt.ignored, tt.desc)
+	}
+}
+
 // TestIgnoreTrailingWhitespace verifies t0008's rule that unescaped trailing
 // whitespace in a pattern is stripped.
 func (s *ConformanceSuite) TestIgnoreTrailingWhitespace() {
--- a/plumbing/format/gitignore/pattern.go
+++ b/plumbing/format/gitignore/pattern.go
@@ -87,7 +87,7 @@ func (p *pattern) Match(path []string, isDir bool) MatchResult {
 	}
 
 	path = path[len(p.domain):]
-	if p.isGlob && !p.globMatch(path, isDir) {
+	if p.isGlob && !p.globMatch(p.pattern, path, isDir) {
 		return NoMatch
 	} else if !p.isGlob && !p.simpleNameMatch(path, isDir) {
 		return NoMatch
@@ -491,17 +491,17 @@ func (p *pattern) simpleNameMatch(path []string, isDir bool) bool {
 	return false
 }
 
-func (p *pattern) globMatch(path []string, isDir bool) bool {
+func (p *pattern) globMatch(segs, path []string, isDir bool) bool {
 	matched := false
 	canTraverse := false
 	trailingStar := false
-	for i, pattern := range p.pattern {
+	for i, pattern := range segs {
 		if pattern == "" {
 			canTraverse = false
 			continue
 		}
 		if pattern == zeroToManyDirs {
-			if i == len(p.pattern)-1 {
+			if i == len(segs)-1 {
 				// A trailing `**` matches the entries below whatever the
 				// earlier segments consumed, so it needs either a remaining
 				// component or a directory candidate standing in for them.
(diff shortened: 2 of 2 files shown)


KaizenShogun:
I measured this PR against an outside bench before saying anything about it. This isn't a code review — I read the diff, I did not audit it. It's the behaviour, against `git`.

**Setup.** One Go helper (`ReadPatterns` + `Matcher.Match`, and `Worktree.Status()` as a second entry point), built three times from the *same* source, with only the tree it points at changing:

| column | tree |
|---|---|
| `main` | `refs/heads/main`, fetched today |
| `+2318` | `refs/pull/2318/merge` — main + this PR, as GitHub computes it (head `70665cd`) |
| `+2311` | `refs/pull/2311/merge`, for contrast |

I checked the drift rather than assuming it: against `main`, the `+2318` tree differs in exactly two files, `plumbing/format/gitignore/pattern.go` and `conformance_test.go`. So the column measures the patch and nothing else.

The corpus is 33 real repositories' full `.gitignore` trees (root plus nested rule files, materialised on disk), each in a files and a directories variant — 4,463 matcher queries and 2,224 through `Status()`. The oracle is `git check-ignore -v --non-matching --no-index` (git 2.55.0), asked again in every tree; nothing is remembered from a previous run.

**The result, compared as sets rather than counts:**

| column | matcher | `Status()` |
|---|---|---|
| `main` | 2 / 4,463 | 1 / 2,224 |
| `+2318` | 2 / 4,463 — *the same two* | 1 / 2,224 — the same one |

Zero fixed, zero broken, zero moved. The two that stay are the `.vscode` + `!.vscode/extensions.json` family (`go-git#694`, `go-git#877`) and a `Status()`-only regression, neither of which this patch claims.

**And it does fix what it claims.** Both repros from #2317, verbatim, with `git` asked in the same tree:

| tree | query | git | `main` | `+2318` |
|---|---|---|---|---|
| `**/a/b` | `a/x/a/b` | ignored | **not ignored** | ignored ✔ |
| `a/**/b/c` | `a/b/x/b/c` | ignored | **not ignored** | ignored ✔ |
| `**/a/b` (control) | `z/a/b/f.txt` | ignored | ignored | ignored |

**Why a real fix moves nothing in 4,463 queries** — worth stating, because "no change" could equally mean the bench never exercised it. It did: 78 of the corpus's 3,427 rule lines carry a non-trailing `**` (`/vendor/**/*.py`, `x-pack/.../**/golden_tests/**/*.actual`, and so on). None changes verdict, because the missing backtracking only bites when the segment *after* `**` also appears earlier in the path — `a/b/x/b/c`, not `a/b/c`. Real `.gitignore` files hit that shape rarely; when they do, they get a silently wrong answer. So take this as evidence of safety, not of impact: the patch fixes its two cases and disturbs nothing else I can measure.

The bench, the corpus and the go-git adapter are public if you want to re-run any of it: https://github.com/KaizenShogun/gitignore-conformance

PR #2325, opened 2026-08-19
worktree: only reject Windows reserved device names on Windows

### LINKED ISSUE #2322: Worktree checkout rejects Windows reserved device names (e.g. prn.sh) on Linux/macOS — regression since v5.19.1
`Worktree.Checkout` (and therefore `PlainClone`) refuses to check out a file whose base name matches a Windows reserved device name (CON, PRN, AUX, NUL, COM1-9, LPT1-9), even on Linux/macOS where such names are valid. Real `git` has no such restriction on non-Windows platforms.

**Regression:** working in v5.19.0, broken in v5.19.1, still broken in v5.19.2 and `main` (at time of writing, b15e610f).

<details>
<summary>Code to reproduce (Linux/macOS)</summary>

```go
package main

import (
	"os"
	"os/exec"

	git "github.com/go-git/go-git/v5"
)

func run(dir string, args ...string) {
	cmd := exec.Command("git", args...)
	cmd.Dir = dir
	cmd.Env = append(os.Environ(), "GIT_AUTHOR_NAME=t", "GIT_AUTHOR_EMAIL=t@t.com",
		"GIT_COMMITTER_NAME=t", "GIT_COMMITTER_EMAIL=t@t.com")
	if out, err := cmd.CombinedOutput(); err != nil {
		panic(string(out))
	}
}

func main() {
	src, _ := os.MkdirTemp("", "src")
	run(src, "init", "-q", "-b", "master")
	os.WriteFile(src+"/prn.sh", []byte("echo hi\n"), 0644)
	run(src, "add", "-A")
	run(src, "commit", "-q", "-m", "add prn.sh")

	_, err := git.PlainClone(os.TempDir()+"/dst", false, &git.CloneOptions{URL: src})
	if err != nil {
		panic(err) // invalid path: "prn.sh"
	}
}
```

</details>

**Expected:** clone succeeds — matches real `git`'s behavior on Linux/macOS.

**Actual:** `lstat: invalid path: "prn.sh"` (or `openfile: invalid path: ...`).

**Cause:** `worktreeFilesystem.validPath` (`worktree_fs.go`) calls `pathutil.WindowsValidPath`, which rejects reserved device names unconditionally regardless of OS, gated only by `core.protectNTFS` — and `defaultProtectNTFS()` now hardcodes `true` on every platform. Real `git`'s equivalent check (`is_valid_win32_path`) is compile-time restricted to Windows-native/Cygwin builds.

You already fixed this exact class of bug once: `pathutil.ValidTreePath` (the tree-read path) had the same unconditional reserved-name check and was scoped to Windows/Cygwin-only in ce4cca1861 ("align tree-path validation with upstream Git"). `worktreeFilesystem.validPath` — the checkout/write path — needs the same fix; it was missed.

SomSamantray:
## Summary

On Linux and macOS, `Worktree.Checkout` (and `PlainClone`) previously refused to check out files whose base name is a Windows reserved device name — `prn.sh`, `CON`, `aux.txt`, and so on — failing with `openfile: invalid path: "prn.sh"`. Real `git` accepts these names on non-Windows. Now the reserved-name check applies only on Windows, matching upstream Git, where `is_valid_win32_path` is compiled only into Windows-native and Cygwin builds.

The regression shipped in v5.19.1 when `defaultProtectNTFS()` began returning `true` on every platform and the worktree write path started applying `WindowsValidPath` unconditionally. That function conflated two distinct upstream checks: the NTFS-disguised `.git` family (gated by `core.protectNTFS` on every platform since CVE-2019-1353) and the Windows reserved device names (compile-time gated to Windows in upstream). The tree-read path was already aligned in ce4cca1; this fixes the checkout/write path the same way.

The NTFS `.git`-disguise protection is unchanged: `IsNTFSDotGit` still rejects disguised metadata paths on all platforms under `core.protectNTFS`.

Fixes #2322

## Reserved-name matching now delegates to Go's stdlib

Following review feedback from @basti-snyk, the Windows reserved-device-name check no longer hand-matches against a maintained list. `IsWindowsReservedName` now delegates to `path/filepath.IsLocal` (Go 1.20+), and the build-tagged `ntfs_reserved_windows.go` / `ntfs_reserved_other.go` split is gone — `filepath.IsLocal`'s own reserved-name check is already compiled only into Windows builds, so one cross-platform implementation covers both cases.

This is a UX/maintenance trade, not a git.exe-parity move: reading git v2.54.0's own `compat/mingw.c is_valid_win32_path` shows it is a static, version-invariant check that always treats an extension-bearing reserved name (`prn.sh`, `CON.txt`) as reserved, on every Windows version. `filepath.IsLocal` matches that for bare names, but for extension-bearing names it defers to a live `RtlIsDosDeviceName_U` syscall whose answer varies by Windows build (confirmed: rejected on Windows 10, accepted on Windows 11). Delegating trades exact `git.exe` parity for that narrow case in favor of the live OS's own answer and removes the list-maintenance burden — the trade-off `@basti-snyk` asked for on UX grounds, not a bug fix.

`WindowsReservedNames` stays exported as a test/reference fixture; production matching no longer reads it.

## Validation

- New regression test `TestCheckoutWindowsReservedDeviceName` clones a tree containing `prn.sh` (built via plumbing so the fixture is platform-independent): succeeds and materialises the file on non-Windows. Its Windows-side assertion is now split into two subtests — a bare reserved name (`PRN`, still deterministic) stays a hard assertion; the extension-bearing case (`prn.sh`) is deliberately left unasserted on Windows, since that answer is now OS-version-dependent by design.
- `TestCherryPickPathValidationMatchesGit` reserved-name rows assert go-git accepts `CON`/`NUL` on non-Windows (matching upstream) and rejects on Windows.
- `go test ./internal/pathutil/...` and the targeted worktree conformance/regression tests pass; `go vet`, `gofmt`, and `GOOS=windows`/`GOOS=linux` cross-compiles are clean; the repo CI matrix (`ubuntu-latest`, `macos-latest`, `windows-latest`) runs `make test` on all three.

Session-settled decisions carried from planning: split the two NTFS checks keeping `.git`-disguise rejection on all platforms (user-approved); gate the reserved-name check to Windows via build-tag files, later superseded by delegating to `filepath.IsLocal` per review feedback (user-directed); preserve the `WindowsValidPath` API surface (user-directed); reproduce with a real clone/checkout regression test (user-directed); delegate `IsWindowsReservedName` to `path/filepath.IsLocal` in place of the hand-maintained list (user-directed, per `@basti-snyk`'s review).

## Unapplied review findings

- [ ] P1 — `internal/pathutil/ntfs.go:1` — No commit on the branch carries the AGENTS.md-required `Assisted-by` trailer
  Suggested fix: if these commits were AI-assisted, add `Assisted-by: <tool/version> <noreply@address>` trailers to all four branch commits via interactive rebase. Requires rewriting already-reviewed commit history and a force-push; spans all 4 commits on the branch, not just the newest — not applied.

- [ ] P2 — `internal/pathutil/ntfs.go:1` — Commit messages don't follow AGENTS.md's required `<package>: <subpackage>, ... [Fixes #N]` format
  Suggested fix: interactive-rebase and rewrite each of the 4 titles to that shape, with no blank line after the title. Same history-rewrite caveat as above — decide both together; not applied.

- [ ] P2 — `internal/pathutil/ntfs.go:96` — Delegation accepts extension-bearing reserved names `git.exe` rejects, and the outcome varies by Windows build
  Suggested fix (report-only — this is the trade-off the session-settled decision above already accepted): pin the decorated-name case to a static table mirroring `git.exe`'s `is_valid_win32_path`, keeping the `filepath.IsLocal` delegation only for the deterministic bare-name case. Kept visible for a maintainer to reconsider; not applied.

- [ ] P2 — `internal/pathutil/ntfs_test.go:69` — Exported `IsWindowsReservedName`'s absolute-path branch has zero test coverage
  Suggested fix: add one deterministic, cross-platform subtest asserting `IsWindowsReservedName("/etc/passwd")` is `true` on every platform (the same broadening already covered for `".."` and colon-bearing components). Not applied — single-reviewer finding without cross-persona corroboration.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)


internal/pathutil/ntfs.go
```diff
@@ -46,50 +46,35 @@ func IsNTFSDotGit(part string) bool {
...
 
-// windowsReservedNames lists the Windows reserved device names.
+// WindowsReservedNames lists the Windows reserved device names.
 // A path component is reserved if its base name (ignoring trailing
 // spaces, extensions, and NTFS Alternate Data Streams) matches one of
 // these case-insensitively.
 //
 // See upstream Git compat/mingw.c is_valid_win32_path().
-var windowsReservedNames = []string{
+var WindowsReservedNames = []string{
```
basti-snyk, line 109: For go 1.20+ (and go-git uses 1.25 on the v5 versions), the following should work without having to specify it manually:

```
filepath.IsLocal("prn.sh")   // false on Win, true on Linux/macOS
filepath.Localize("prn.sh")  // Go 1.23+, returns an error on Windows
```
My test returns this:
```
== GOOS=windows, Win10 (RtlIsDosDeviceName_U > 0) ==
"prn.sh"     IsLocal=false  Localize="" err=invalid path
"PRN"        IsLocal=false  Localize="" err=invalid path
"con.txt"    IsLocal=false  Localize="" err=invalid path
"COM1 "      IsLocal=false  Localize="" err=invalid path
"COM¹"       IsLocal=false  Localize="" err=invalid path
"CONIN$"     IsLocal=false  Localize="" err=invalid path
"dollar$"    IsLocal=true   Localize="dollar$" err=<nil>
"COM0"       IsLocal=true   Localize="COM0" err=<nil>

== GOOS=windows, Win11 (RtlIsDosDeviceName_U == 0) ==
"prn.sh"     IsLocal=true   Localize="prn.sh" err=<nil>     <-- !
"PRN"        IsLocal=false  Localize="" err=invalid path
"con.txt"    IsLocal=true   Localize="con.txt" err=<nil>
"COM1 "      IsLocal=true   Localize="COM1 " err=<nil>
"COM¹"       IsLocal=false  Localize="" err=invalid path
"CONIN$"     IsLocal=false  Localize="" err=invalid path
```

basti-snyk: Apparently Windows 11 is more lenient than earlier versions

internal/pathutil/ntfs_reserved_windows.go
```diff
@@ -0,0 +1,29 @@
...
+
+package pathutil
+
+import "strings"
+
+// IsWindowsReservedName reports whether part is a Windows reserved
+// device name, matching upstream Git's is_valid_win32_path
+// (compat/mingw.c), which is compiled only into Windows-native and
+// Cygwin builds.
+func IsWindowsReservedName(part string) bool {
```
basti-snyk: see above, where the constants are defined. probably better to just use `filepath.IsLocal`

SomSamantray:
Hi @basti-snyk 

Thanks for digging into this — the Win10/Win11 divergence you found is real and it's actually documented in Go's own source: `internal/filepathlite/path_windows.go`'s `isReservedName` falls through to `windows.RtlIsDosDeviceName_U` for any reserved name with an extension (`prn.sh`, `con.txt`), and that syscall's behavior changed in Windows 11 per the comment right above it.

That's the reason I'd rather not switch: this PR's whole point is to match real `git.exe`'s behavior, which comes from a fixed compile-time list in `compat/mingw.c` — deterministic across every Windows version. If we switch to `filepath.IsLocal`, go-git's checkout behavior would itself become Windows-version-dependent, which would reintroduce a version of the same inconsistency bug in a different place (a repo with `prn.sh` clones fine on Win11 but fails on Win10, with no way to make that consistent). Keeping the static list, even though it's a bit more code, keeps go-git's behavior identical to real git on any Windows box. Happy to add a comment in the code calling out this exact tradeoff so it doesn't get "simplified" back to stdlib later.

bastiandoetsch:
How do you keep it synchronized if it changes with the next mingw version? This is a decision for manual effort and maintenance. I'm personally fine either way, just need the fix in.

SomSamantray:
@bastiandoetsch Fair question.

Yes, this list can change and it did before, when CONIN$/CONOUT$ got added to it upstream.

But I still think keeping it as a manual list is the better trade-off, for a simple reason: the alternative (filepath.IsLocal) doesn't read from a fixed list at all — it asks Windows itself, live, at runtime. And Windows gives different answers depending on which version you're on. So switching to it wouldn't remove maintenance work, it would just trade "occasionally update a 12-line list" for "checkout behaves differently depending on which Windows machine you're using" which is a wore problem since nothing would ever tell us it's happening.

Real git.exe has the same list, maintained by hand, for the same reason. So I'd rather match that approach than the OS-version-dependent one.

SomSamantray:
> Why would it be better to read from a fixed list? [...] So I think `filepath.IsLocal` is the better solution

Fixed in 5101cbc — `IsWindowsReservedName` now delegates to `path/filepath.IsLocal` and the two build-tag files (`ntfs_reserved_windows.go` / `ntfs_reserved_other.go`) are gone.

One correction to the framing, since it matters for review: this isn't a git.exe-parity move, and my earlier replies in this thread assumed the opposite. I went and read git v2.54.0's actual `compat/mingw.c is_valid_win32_path` — it's a static, version-invariant check with no OS call, and its own comment says an extension-bearing reserved name (`prn.sh`, `CON.txt`) is "still reserved" unconditionally, on every Windows version. `filepath.IsLocal` matches that for bare names, but for extension-bearing names it defers to the live `RtlIsDosDeviceName_U` syscall (the one @basti-snyk found earlier diverging Win10 vs Win11).

PR #2331, opened 2026-08-21
plumbing: protocol/packp/sideband, Test the muxer packet size limit

Soph:
Follow-up to #2329, which fixed `Muxer`'s chunk size but left no test that would catch a reintroduction.

The only test change there was `TestMuxerWrite`'s byte count, `2008` → `2013`. That number does not describe the constraint being violated, and the pre-fix code passed its *old* value while emitting packets no `Demuxer` accepts — so the assertion cannot distinguish correct chunking from broken chunking.

This asserts the invariant `Demuxer.nextPackData` already enforces on the wire: no pkt-line the `Muxer` emits exceeds the maximum packet size of its sideband type, counting the 4-byte length prefix. `pktline.Scanner.Len` reports exactly that, so the test walks the emitted stream and checks every packet, then reads the stream back through the `Demuxer`, which applies the same limit and so doubles as the wire-level check.

Against the pre-fix muxer both types fail, each on its own symptom:

```
--- FAIL: TestMuxerChunksWithinPacketSizeLimit/side-band
        Error:    "1004" is not less than or equal to "1000"
        Messages: packet 1 of 1000 exceeds the limit
--- FAIL: TestMuxerChunksWithinPacketSizeLimit/side-band-64k
        Error:    Received unexpected error: payload is too long
```

Those are the two distinct bugs #2329 fixed, and it is worth noting the first is not Go-version dependent: `side-band` (negotiated at `upload_pack.go:283` and `receive_pack.go:188`) emitted 1004-byte packets over git's 1000-byte `DEFAULT_PACKET_MAX`, which go-git's own `Demuxer` rejects with `ErrMaxPackedExceeded`.

`TestMuxerWriteFillsSinglePacket` guards the other direction, where too large a reservation would chunk a payload that fits: `max-5` bytes must fill exactly one packet and not spill into a second.

Test-only; no behaviour change. Merges cleanly with #2332 in either order.

## AI assistance

Claude Opus 5 assisted with writing the tests and verifying they fail against the pre-fix muxer. Commit carries `Assisted-by:` per AI_POLICY.md.

### DIFF
--- a/plumbing/protocol/packp/sideband/muxer_test.go
+++ b/plumbing/protocol/packp/sideband/muxer_test.go
@@ -2,7 +2,21 @@ package sideband
 
 import (
 	"bytes"
+	"io"
+
+	"github.com/go-git/go-git/v6/plumbing/format/pktline"
 )
+
+// sidebandTypes pairs each sideband type with the maximum pkt-line size it
+// permits, counting the 4-byte length prefix.
+var sidebandTypes = []struct {
+	name string
+	t    Type
+	max  int
+}{
+	{"side-band", Sideband, MaxPackedSize},
+	{"side-band-64k", Sideband64k, MaxPackedSize64k},
+}
 
 func (s *SidebandSuite) TestMuxerWrite() {
 	buf := bytes.NewBuffer(nil)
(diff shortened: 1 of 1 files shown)


pjbgf (approved):
@Soph Thanks for working on this. 🙇

PR #2337, opened 2026-08-23
plumbing: format/packfile, Break cyclic delta chains instead of recursing. Fixes #2249

### LINKED ISSUE #2249: (maybe already fixed) Stack overflow while processing delta pack
### Bug Description

Stack overflow when pushing to remote

```text
runtime: goroutine stack exceeds 2000000000-byte limit
runtime: sp=0x2ce7e9412398 stack=[0x2ce7e9412000, 0x2ce829412000]
fatal error: stack overflow

runtime stack:
runtime.throw({0xdfd830?, 0x200000001?})
        /usr/lib/go/src/runtime/panic.go:1229 +0x48 fp=0x7f6e08ff9ca0 sp=0x7f6e08ff9c70 pc=0x48f428
runtime.newstack()
        /usr/lib/go/src/runtime/stack.go:1178 +0x5fd fp=0x7f6e08ff9dd0 sp=0x7f6e08ff9ca0 pc=0x472b1d
runtime.morestack()
        /usr/lib/go/src/runtime/asm_amd64.s:681 +0x7d fp=0x7f6e08ff9dd8 sp=0x7f6e08ff9dd0 pc=0x4956fd

goroutine 145582 gp=0x2ce7a8bccd20 m=14 mp=0x2ce7a745b008 [running]:
type:.hash.github.com/go-git/go-git/v6/plumbing.ObjectID(0x2ce7e9412460, 0x66fdeec80083a365?)
        <autogenerated>:1 +0x46 fp=0x2ce7e94123a8 sp=0x2ce7e94123a0 pc=0x702546
runtime.mapaccess2(0xd1e0e0, 0x2ce829411c90, 0x2ce7e9412460)
        /usr/lib/go/src/internal/runtime/maps/runtime.go:156 +0x71 fp=0x2ce7e9412418 sp=0x2ce7e94123a8 pc=0x4158d1
github.com/go-git/go-git/v6/plumbing/format/packfile.(*DeltaSelector).fixAndBreakChainsOne(0x2ce7a74e4f70, 0x2ce829411c90, 0x2ce7a74e7580)
        /home/dixyes/kkbb/play/magic/cli/play/go-git/plumbing/format/packfile/delta_selector.go:178 +0xca fp=0x2ce7e94124a0 sp=0x2ce7e9412418 pc=0x70488a
github.com/go-git/go-git/v6/plumbing/format/packfile.(*DeltaSelector).fixAndBreakChainsOne(0x2ce7a74e4f70, 0x2ce829411c90, 0x2ce7a85b8700)
        /home/dixyes/kkbb/play/magic/cli/play/go-git/plumbing/format/packfile/delta_selector.go:185 +0xeb fp=0x2ce7e9412528 sp=0x2ce7e94124a0 pc=0x7048ab
github.com/go-git/go-git/v6/plumbing/format/packfile.(*DeltaSelector).fixAndBreakChainsOne(0x2ce7a74e4f70, 0x2ce829411c90, 0x2ce7a74e7580)
        /home/dixyes/kkbb/play/magic/cli/play/go-git/plumbing/format/packfile/delta_selector.go:185 +0xeb fp=0x2ce7e94125b0 sp=0x2ce7e9412528 pc=0x7048ab
github.com/go-git/go-git/v6/plumbing/format/packfile.(*DeltaSelector).fixAndBreakChainsOne(0x2ce7a74e4f70, 0x2ce829411c90, 0x2ce7a85b8700)
```

maybe already fixed (by side-effect?) in 21391261b9e68a2ab034bed172cc7f2eeb4f9fcf

### go-git Version

v6.0.0-alpha.4

### Steps to Reproduce

To generate test repo:
```bash
#!/usr/bin/env bash

set -xeo pipefail

rm -rf "testrepo12"
rm -rf "testrepo21"

# init repos
git init "testrepo12"
git init "testrepo21"
git -C "testrepo12" branch -M "test12"
git -C "testrepo21" branch -M "test21"

# (not important) config
git -C "testrepo12" config user.name "Test User"
git -C "testrepo12" config user.email "test@example.org"
git -C "testrepo21" config user.name "Test User 2"
git -C "testrepo21" config user.email "test2@example.org"

# (not important) make initial commit for both repo
echo initial12 > testrepo12/initial.txt
git -C "testrepo12" add initial.txt
git -C "testrepo12" commit -m "initial12"
echo initial21 > testrepo21/initial.txt
git -C "testrepo21" add initial.txt
git -C "testrepo21" commit -m "initial21"

# make a tree object
mkdir testrepo12/src
mkdir testrepo21/src
for i in $(seq 10)
do
    echo some content > "testrepo12/src/file$i"
    git -C "testrepo12" add "src/file$i"
    echo some content > "testrepo21/src/file$i"
    git -C "testrepo21" add "src/file$i"
done

# make state 1 -> 2 in repo12
echo state 1 > "testrepo12/src/file1"
git -C "testrepo12" add "src/file1"
git -C "testrepo12" commit -m "state 1"
echo state 2 > "testrepo12/src/file1"
git -C "testrepo12" add "src/file1"
git -C "testrepo12" commit -m "state 2"

# make state 2 -> 1 in repo12
echo state 2 > "testrepo21/src/file1"
git -C "testrepo21" add "src/file1"
git -C "testrepo21" commit -m "state 2"
echo state 1 > "testrepo21/src/file1"
git -C "testrepo21" add "src/file1"
git -C "testrepo21" commit -m "state 1"

# generate commits in both repo to let it generate packs
mkdir testrepo12/dummy
mkdir testrepo21/dummy
for i in $(seq 500)
do
    # echo "$i" > "testrepo12/dummy/$i.txt"
    # git -C "testrepo12" add "dummy/$i.txt"
    # git -C "testrepo12" commit -m "commit $i"
    # echo "$i" > "testrepo21/dummy/$i.txt"
    # git -C "testrepo21" add "dummy/$i.txt"
    # git -C "testrepo21" commit -m "commit $i"
    echo "$i" > "testrepo12/some.txt"
    git -C "testrepo12" add "some.txt"
    git -C "testrepo12" commit -m "commit $i"
    echo "$i" > "testrepo21/some.txt"
    git -C "testrepo21" add "some.txt"
    git -C "testrepo21" commit -m "commit $i"
done
git -C "testrepo21" gc
git -C "testrepo12" gc

# at this time
# git verify-pack -v testrepo12/.git/objects/pack/pack-*.idx
# shows
# ...
# <hash1> tree   42 55 660 1 <hash1>
# ...
# while
# git verify-pack -v testrepo21/.git/objects/pack/pack-*.idx
# shows
# ...
# <hash2> tree   42 55 662 1 <hash1>
# ...
# sizes may vary, but now we have two packs:
# one contains State 1 -> State 2, while another State 2 -> State 1

# fetch as remote, git will not extract pack by default
# normally, you may not fetch like this
# but I encountered this when using git subtree split
# it's hard to make git subtree split demo, so here use simple fetch
# disable maintenance and auto gc to avoid git extract pack automatically
git -C "testrepo12" config gc.auto 0
git -C "testrepo12" config maintenance.auto false
git -C "testrepo12" fetch ../testrepo21 test21:test21

# at this time 
# git verify-pack -v testrepo12/.git/objects/pack/pack-*.idx | grep <hash1>
# shows:
# the loop is constructed
# while git command is still ok:
# rm -rf testdest
# git init --bare testdest
# git -C "testrepo12" remote add remotedest "$(realpath "testdest")"
# git -C "testrepo12" push remotedest test21:test21
```

test source

```go
package git

import (
	"errors"
	"os"
	"path/filepath"
	"runtime/debug"
	"testing"

	"github.com/go-git/go-git/v6/config"
)

const sourceRepoPath = "/path/to/testrepo12"
const branch = "test21"

func TestGogitBug(t *testing.T) {
	tempDir := t.TempDir()

	// init a repo
	repo, err := PlainInit(filepath.Join(tempDir, "testrepo"), false)
	if err != nil {
		t.Fatalf("failed to init repo: %v", err)
	}
	repo.Close()

	// allow more stack space for the test (avoid false positive)
	debug.SetMaxStack(2 << 30)

	repo, err = PlainOpen(sourceRepoPath)
	if err != nil {
		t.Fatalf("failed to open repo: %v", err)
	}

	// create remote for push
	remote, err := repo.CreateRemoteAnonymous(&config.RemoteConfig{
		Name: "anonymous",
		URLs: []string{filepath.Join(tempDir, "testrepo")},
	})
	if err != nil {
		t.Fatalf("failed to create anonmymous remote: %v", err)
	}

	err = remote.Push(&PushOptions{
		RemoteName: "anonymous",
		RemoteURL:  filepath.Join(tempDir, "testrepo"),
		RefSpecs: []config.RefSpec{
			config.RefSpec("refs/heads/" + branch + ":refs/heads/" + branch),
		},
		Progress: os.Stderr,
	})
	if err != nil && !errors.Is(err, NoErrAlreadyUpToDate) {
		t.Fatalf("failed to push: %v", err)
	}
}

```

To trigger this:

```bash
# this occurs randomly (but mostly less than 100 times)
go test . -v --count=100 --run="TestGogitBug"
```

### Additional Information

The problem may be fixed after 21391261b9e68a2ab034bed172cc7f2eeb4f9fcf

But the stack is at

https://github.com/go-git/go-git/blob/v6.0.0-alpha.4/plumbing/format/packfile/delta_selector.go#L148

LD-RW:
Fixes #2249

## Problem

`DeltaSelector.fixAndBreakChainsOne` resolves a reused delta by recursing into its base first and assigning `ObjectToPack.Base` only *after* the recursive call returns:

```go
if otp.Base != nil {        // the only reentry guard
    return nil
}
...
if err := dw.fixAndBreakChainsOne(objectsToPack, base); err != nil {  // descend
    return err
}
otp.SetDelta(base, otp.Object)                                        // Base assigned here
```

Because `Base` is assigned after the descent, every object on the current resolution path still has `Base == nil` while the recursion is in flight, so the guard cannot distinguish an object being resolved from an unvisited one.

Given a delta chain that loops back on itself (`A → B → A`), the guard never fires and the function recurses until the goroutine stack is exhausted:

| depth | object | `otp.Base` | action |
|---|---|---|---|
| 1 | A | nil | base B found → descend |
| 2 | B | nil | base A found → descend |
| 3 | A | still nil | base B found → descend |
| … | | | until stack overflow |

This is an unrecoverable `fatal error: stack overflow` — not a returnable error — so callers cannot defend against it.

As noted in the issue thread, commit 2139126 (`storage: filesystem, MRU pack ordering`) changes which packfile is consulted first and so may mask a particular trigger, but it does not touch the recursion. The reporter's observation that there is "no recursive protection in `fixAndBreakChainsOne`" is the actual defect, and it is still present on `main`.

## Reproduction

Minimum working example (the delta objects are constructed directly so the cycle is deterministic; `newObject` is the existing test helper):

```go
// A delta object whose base hash we control.
type cyclicDelta struct {
	plumbing.EncodedObject
	self, base plumbing.Hash
}

func (d *cyclicDelta) Type() plumbing.ObjectType { return plumbing.OFSDeltaObject }
func (d *cyclicDelta) BaseHash() plumbing.Hash   { return d.base }
func (d *cyclicDelta) ActualHash() plumbing.Hash { return d.self }
func (d *cyclicDelta) ActualSize() int64         { return d.Size() }

a := newObject(plumbing.BlobObject, []byte("aaaaaaaa"))
b := newObject(plumbing.BlobObject, []byte("bbbbbbbb"))

// A's base is B, B's base is A.
otpA := &ObjectToPack{Object: &cyclicDelta{EncodedObject: a, self: a.Hash(), base: b.Hash()}}
otpB := &ObjectToPack{Object: &cyclicDelta{EncodedObject: b, self: b.Hash(), base: a.Hash()}}

NewDeltaSelector(store).fixAndBreakChains([]*ObjectToPack{otpA, otpB})
```

Before this change:

```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow

github.com/go-git/go-git/v6/plumbing/format/packfile.(*DeltaSelector).fixAndBreakChainsOne(...)
	plumbing/format/packfile/delta_selector.go:178
github.com/go-git/go-git/v6/plumbing/format/packfile.(*DeltaSelector).fixAndBreakChainsOne(...)
	plumbing/format/packfile/delta_selector.go:185
github.com/go-git/go-git/v6/plumbing/format/packfile.(*DeltaSelector).fixAndBreakChainsOne(...)
	plumbing/format/packfile/delta_selector.go:185      <- alternating between the two objects
```

This matches the frame pattern in the issue report (entry at `:178`, then `:185` repeating between two alternating object pointers).

## Fix

Track the objects on the current resolution path. Mark the current object *before* inspecting its base, and if the base is already on the path, break the chain by undeltifying instead of following the cycle.

Two details worth calling out:

- **The check is before the descent, not on entry.** Detecting the cycle on entry ("am I already being visited?") looks equivalent but is not: the inner frame undeltifies the object, then the outer frame — still holding its own `base` local — falls through to `SetDelta` and re-creates the loop. Testing whether the *base* is on the path means the cut is made by the frame that owns the offending edge.
- **The path set is unwound on return** (`defer delete`). This is required for correctness, not hygiene: without it, two deltas legitimately sharing one base would be treated as a cycle and the second would be needlessly undeltified.

Breaking the chain (rather than returning an error) matches what this function already does when a base is missing from the pack or is not a `DeltaObject`, and keeps a push working instead of failing it.

## Alignment with upstream git

git handles this in [`break_delta_chains()`](https://github.com/git/git/blob/master/builtin/pack-objects.c#L2490) using a three-state DFS, and does the same three things in the same order:

```c
cur->dfs_state = DFS_ACTIVE;                     /* mark self first */
if (DELTA(cur)->dfs_state == DFS_ACTIVE) {       /* is the BASE active? */
        drop_reused_delta(cur);                  /* undeltify this object */
        cur->dfs_state = DFS_DONE;
        break;
}
```

git's own comment confirms the ordering is deliberate — *"It's important to do this `_before_` we loop, because it impacts where we make the cut"* — and explains why cutting the **last** edge of the cycle is preferable: for `A → B → C → D → B`, cutting `D→B` keeps A's depth correct at 3, whereas cutting `B→C` would drop C and D from the chain. This change cuts the last edge for the same reason. `drop_reused_delta` is the direct counterpart of go-git's `undeltify` on the delta-reuse path.

## Tests

New `plumbing/format/packfile/delta_selector_cycle_test.go`:

| Test | Covers |
|---|---|
| `TestSelfReferencingDelta` | `A → A`; only escape is undeltifying |
| `TestTwoObjectCycle` | the reported case, `A → B → A`; asserts exactly one object undeltified |
| `TestThreeObjectCycle` | `A → B → C → A`; confirms the cut lands on the last edge (2 deltas kept) |
| `TestAcyclicChainIsPreserved` | shared base — guards against false positives |
| `TestDeepChainIsPreserved` | 50-long valid chain; every delta must survive |

A shared helper walks the resulting `Base` pointers and fails on a repeated hash or a delta left without a base, so the tests assert the output is a *writable* chain rather than merely that nothing crashed.

I verified the tests fail without the source change (the test binary dies with `fatal error: stack overflow`) and pass with it.

Also run locally: `make validate-lint` (golangci-lint v2.13.1) clean, `go test -race ./...`, `gofmt`, `go vet`.

## Notes for reviewers

- **On reachability:** the test constructs the cycle through the internal API rather than from a crafted packfile, so I would not claim a proven remote trigger. The issue reporter's stack trace is the field evidence; this change is defence-in-depth on the delta-*reuse* path, which is the same path upstream git guards (hence `drop_reused_delta`). Happy to be told this warrants a different framing.
- **Break vs. error:** I chose to break the chain for consistency with the function's existing behaviour, but if you would rather surface a cycle as an error (on the view that it indicates a corrupt local pack worth reporting), that is a small change.
- `fixAndBreakChainsOne` gains a parameter. It is unexported with a single caller, so there is no API impact — happy to thread the state differently if you prefer.
- **Backport:** targeting `main` per CONTRIBUTING. `releases/v5.x` looks to have the same shape; I can open a backport if useful rather than assume.

## AI assistance disclosure

Per [AI_POLICY.md](https://github.com/go-git/go-git/blob/main/AI_POLICY.md), this change was produced with AI assistance (Claude Opus 5), disclosed via an `Assisted-by:` trailer on the commit. I have reviewed and understand every line, verified the behaviour against reference git (2.55.0) and against git's own source, and confirmed the tests fail without the fix. I will respond to review feedback personally.

### DIFF
--- a/plumbing/format/packfile/delta_selector.go
+++ b/plumbing/format/packfile/delta_selector.go
@@ -147,8 +147,12 @@ func (dw *DeltaSelector) fixAndBreakChains(objectsToPack []*ObjectToPack) error 
 		m[otp.Hash()] = otp
 	}
 
+	// visiting holds the objects on the current resolution path, so that a
+	// delta chain looping back on itself can be detected and broken.
+	visiting := make(map[plumbing.Hash]bool)
+
 	for _, otp := range objectsToPack {
-		if err := dw.fixAndBreakChainsOne(m, otp); err != nil {
+		if err := dw.fixAndBreakChainsOne(m, otp, visiting); err != nil {
 			return err
 		}
 	}
@@ -156,7 +160,11 @@ func (dw *DeltaSelector) fixAndBreakChains(objectsToPack []*ObjectToPack) error 
 	return nil
 }
 
-func (dw *DeltaSelector) fixAndBreakChainsOne(objectsToPack map[plumbing.Hash]*ObjectToPack, otp *ObjectToPack) error {
+func (dw *DeltaSelector) fixAndBreakChainsOne(
+	objectsToPack map[plumbing.Hash]*ObjectToPack,
+	otp *ObjectToPack,
+	visiting map[plumbing.Hash]bool,
+) error {
 	if !otp.Object.Type().IsDelta() {
 		return nil
 	}
@@ -182,7 +190,21 @@ func (dw *DeltaSelector) fixAndBreakChainsOne(objectsToPack map[plumbing.Hash]*O
 		return dw.undeltify(otp)
 	}
 
-	if err := dw.fixAndBreakChainsOne(objectsToPack, base); err != nil {
+	// Mark this object as being resolved before looking at its base, so that
+	// a delta based on itself is caught by the check below.
+	h := otp.Hash()
+	visiting[h] = true
+	defer delete(visiting, h)
+
+	// A delta chain that loops back onto an object we are already resolving
+	// cannot be written: every delta needs its base written first. Break the
+	// chain here instead of following the cycle, which would recurse until
+	// the goroutine stack is exhausted.
+	if visiting[do.BaseHash()] {
+		return dw.undeltify(otp)
+	}
+
+	if err := dw.fixAndBreakChainsOne(objectsToPack, base, visiting); err != nil {
 		return err
 	}
 
--- a/plumbing/format/packfile/delta_selector_cycle_test.go
+++ b/plumbing/format/packfile/delta_selector_cycle_test.go
@@ -1,0 +1,179 @@
+package packfile
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+
+	"github.com/go-git/go-git/v6/plumbing"
+	"github.com/go-git/go-git/v6/storage/memory"
+)
+
+// cyclicDelta is a delta object whose base hash is set by the test, allowing
+// delta chains that loop back on themselves to be built.
+type cyclicDelta struct {
+	plumbing.EncodedObject
+	self plumbing.Hash
+	base plumbing.Hash
+}
+
+func (d *cyclicDelta) Type() plumbing.ObjectType { return plumbing.OFSDeltaObject }
+func (d *cyclicDelta) BaseHash() plumbing.Hash   { return d.base }
+func (d *cyclicDelta) ActualHash() plumbing.Hash { return d.self }
+func (d *cyclicDelta) ActualSize() int64         { return d.Size() }
+
+type DeltaSelectorCycleSuite struct {
+	suite.Suite
+	store *memory.Storage
+	ds    *DeltaSelector
+}
+
+func TestDeltaSelectorCycleSuite(t *testing.T) {
+	t.Parallel()
+	suite.Run(t, new(DeltaSelectorCycleSuite))
+}
+
+func (s *DeltaSelectorCycleSuite) SetupTest() {
+	s.store = memory.NewStorage()
+	s.ds = NewDeltaSelector(s.store)
+}
+
+// addObject stores a non-delta object so that undeltify can restore it.
+func (s *DeltaSelectorCycleSuite) addObject(content string) plumbing.EncodedObject {
+	o := newObject(plumbing.BlobObject, []byte(content))
+	_, err := s.store.SetEncodedObject(o)
+	s.Require().NoError(err)
+	return o
+}
+
+// deltaOn returns an ObjectToPack for a delta of o based on base.
+func (s *DeltaSelectorCycleSuite) deltaOn(o plumbing.EncodedObject, base plumbing.Hash) *ObjectToPack {
+	return &ObjectToPack{
+		Object: &cyclicDelta{EncodedObject: o, self: o.Hash(), base: base},
+	}
+}
+
+// assertResolvable walks each object's Base pointers and fails if a cycle is
+// still reachable or the chain does not terminate at a non-delta object.
+func (s *DeltaSelectorCycleSuite) assertResolvable(objs []*ObjectToPack) {
+	for _, otp := range objs {
+		seen := make(map[plumbing.Hash]bool)
+		for cur := otp; cur != nil; cur = cur.Base {
+			h := cur.Hash()
+			s.Require().False(seen[h], "delta chain still contains a cycle at %s", h)
+			seen[h] = true
+		}
+		// A delta must have a base to be writable.
+		if otp.Object.Type().IsDelta() {
+			s.Require().NotNil(otp.Base, "delta %s left without a base", otp.Hash())
+		}
+	}
+}
+
+// A delta whose base is itself must not recurse forever.
+func (s *DeltaSelectorCycleSuite) TestSelfReferencingDelta() {
+	a := s.addObject("aaaaaaaa")
+	otpA := s.deltaOn(a, a.Hash())
+
+	objs := []*ObjectToPack{otpA}
+	s.Require().NoError(s.ds.fixAndBreakChains(objs))
+
+	// The only way to break a self-cycle is to store the object whole.
+	s.False(otpA.Object.Type().IsDelta(), "self-referencing delta should be undeltified")
+	s.Nil(otpA.Base)
+	s.assertResolvable(objs)
+}
+
+// Two deltas based on each other must not recurse forever. This is the case
+// reported in issue #2249.
+func (s *DeltaSelectorCycleSuite) TestTwoObjectCycle() {
+	a := s.addObject("aaaaaaaa")
+	b := s.addObject("bbbbbbbb")
+
+	otpA := s.deltaOn(a, b.Hash())
+	otpB := s.deltaOn(b, a.Hash())
+
+	objs := []*ObjectToPack{otpA, otpB}
+	s.Require().NoError(s.ds.fixAndBreakChains(objs))
+
+	s.assertResolvable(objs)
+
+	// Exactly one of the two must have been undeltified to break the cycle;
+	// the other stays a delta so pack size is not sacrificed needlessly.
+	deltas := 0
+	for _, otp := range objs {
+		if otp.Object.Type().IsDelta() {
+			deltas++
+		}
+	}
+	s.Equal(1, deltas, "exactly one object should be undeltified")
+}
+
+// A longer cycle (A -> B -> C -> A) must also terminate.
+func (s *DeltaSelectorCycleSuite) TestThreeObjectCycle() {
+	a := s.addObject("aaaaaaaa")
+	b := s.addObject("bbbbbbbb")
+	c := s.addObject("cccccccc")
+
+	otpA := s.deltaOn(a, b.Hash())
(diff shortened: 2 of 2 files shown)


pjbgf (approved):
@LD-RW thanks for working on this. 🙇

PR #2338, opened 2026-08-24
storage: filesystem/dotgit, route symlink writes by link path.

Definition-f-Imaginative-Spring:
Fix RepositoryFilesystem.Symlink to choose the backing filesystem from the link path, not the symlink target. This keeps symlink creation routed to the correct dotgit root when target and link live under different trees.\n\nAdded a regression test that verifies symlink creation lands in the expected backing filesystem and does not touch the other root.\n\nVerification:\n- go test ./storage/filesystem/dotgit\n- go test ./storage/filesystem/...

### DIFF
--- a/storage/filesystem/dotgit/repository_filesystem_test.go
+++ b/storage/filesystem/dotgit/repository_filesystem_test.go
@@ -63,6 +63,24 @@ func (s *SuiteDotGit) TestRepositoryFilesystem() {
 	s.Require().NoError(err)
 	s.Equal("newfile", link)
 
+	crossRootLinkTarget := repositoryFs.Join("refs", "heads", "main")
+	err = repositoryFs.Symlink(crossRootLinkTarget, "cross-root-link")
+	s.Require().NoError(err)
+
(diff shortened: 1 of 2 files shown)


storage/filesystem/dotgit/repository_filesystem_test.go
```diff
@@ -63,6 +63,15 @@ func (s *SuiteDotGit) TestRepositoryFilesystem() {
 	s.Require().NoError(err)
 	s.Equal("newfile", link)
 
+	err = repositoryFs.Symlink("refs/heads/main", "cross-root-link")
+	s.Require().NoError(err)
+
+	_, err = dotGitFs.Lstat("cross-root-link")
+	s.Require().NoError(err)
```
pjbgf, line 75: Let's assert its content with `ReadLink`.

Definition-f-Imaginative-Spring: Done, thanks for the review. I added Readlink assertions so the test now verifies the cross-root symlink target, not just that the link exists.

PR #2341, opened 2026-08-24
plumbing: object, reject multi-signature tags at Verify

Sahana2524:
**Missing multi-signature guard in Tag.Verify**

Commit.Verify was recently taught to reject a payload carrying more than one armored signature block (c6f8c8d), matching upstream's parse_gpg_output. Tag.Verify never got the same check, so a signed tag whose trailing payload holds a valid signature followed by a second appended block still verifies: CheckArmoredDetachedSignature returns the entity of whichever block matches first, and the caller sees success for a tag whose signer is ambiguous. Reused the same countSignatureBlocks guard and generalized the shared ErrMultipleSignatures wording, which now applies to both object kinds.

### DIFF
--- a/plumbing/object/commit.go
+++ b/plumbing/object/commit.go
@@ -477,13 +477,13 @@ func (c *Commit) String() string {
 	)
 }
 
-// ErrMultipleSignatures is returned by Verify when the commit carries more
-// than one armored signature block. Mirrors upstream's parse_gpg_output
+// ErrMultipleSignatures is returned by Verify when a signed object carries
+// more than one armored signature block. Mirrors upstream's parse_gpg_output
 // rejection of GOODSIG/BADSIG status lines after the first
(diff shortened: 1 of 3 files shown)


pjbgf:
Please note that the current behaviour is aligned with upstream and enforced via conformance tests: https://github.com/go-git/go-git/blob/main/tests/objectverify/tag_test.go#L15

Sahana2524:
Fair point, I went back through the conformance suite. Since Decode keeps only the last armored block (matching parse_signed_buffer), a decoded tag can't reach Verify with two blocks in Signature, so this guard would only fire for hand-constructed Tag values, and the double-signature verdict is already pinned to git's per-version behavior by tests/objectverify. I don't think it buys much beyond that narrow case, so feel free to close this one if you'd rather keep Verify as is.

Sahana2524:
still interested?

PR #2344, opened 2026-08-25
plumbing: format/commitgraph, fix panic encoding a commit whose parent isn't in the index

HubertKal:
### Summary

`commitgraph.Encoder.Encode` panics with a nil-pointer dereference when the index contains a commit
whose recorded parent hash isn't part of that index. A shallow clone's boundary commit is the
natural way to reach it — its `ParentHashes` legitimately names a parent that was never fetched —
but any partially populated `MemoryIndex` does it.

### Root cause

Two discarded errors and an unchecked map read:

1. `prepare()` calls `idx.GetCommitDataByIndex`, discards the error, and dereferences the nil
   result on the next line. `encodeCommitData()` does the same, and also discards the error from
   `idx.GetIndexByHash`.
2. `encodeCommitData` resolves parent hashes with a bare `hashToIndex[hash]` read. An absent hash
   yields index `0`, so the encoder writes an arbitrary, wrong commit as the parent instead of
   failing. That corrupts the graph rather than crashing, and is reachable independently of the
   panic by any `Index` implementation whose `ParentHashes` names a commit outside `Hashes()`.

### Fix

- `prepare()` gains an error return; it and `encodeCommitData` propagate the lookup failures they
  were discarding, so a nil `CommitData` is never dereferenced. For a `MemoryIndex` this surfaces
  the `plumbing.ErrObjectNotFound` that `GetIndexByHash` was already producing and the encoder was
  throwing away.
- New `lookupParentIndex` helper returns `ErrParentNotInIndex` rather than encoding index `0`. This
  is the second defect and is independent of the first: it needs an `Index` whose `ParentIndexes`
  and `ParentHashes` agree but whose parent falls outside `Hashes()`, which `MemoryIndex` can't
  produce but a third-party implementation can.

No behaviour change for a well-formed index. Verified byte-identical: encoding an index containing a
root commit, two branches, a 2-parent merge and a 4-parent octopus — so the extra-edge and
generation-v2 paths are both exercised — produces the same 1692 bytes before and after
(`sha256=411537e7…`). The only new outcome is an error where there was previously a panic or a
silently wrong edge.

### Behaviour vs reference git

The panic has no git counterpart. Git's writer only encodes commits it has already collected, so it
can never be handed a parent it cannot resolve.

The second defect can be demonstrated, though it takes a purpose-built `Index` to reach it:
`MemoryIndex` fails earlier, so only an implementation that reports a parent outside its own
`Hashes()` gets that far. Given one, on a three-commit repository, go-git writes the file without
complaint and `git commit-graph verify` (git 2.55.0) exits 1 with:

```
commit-graph parent for 7bccae8d8abe3cd0d953622475a8a9ee5bb025c4 is 7bccae8d8abe3cd0d953622475a8a9ee5bb025c4 != 16dbf71e13ca56928da49c80a6b3e1236b0a9a8e
commit-graph generation for commit 7bccae8d8abe3cd0d953622475a8a9ee5bb025c4 is 3 < 4
```

Because the parent resolved to index `0` and the encoded hashes are sorted, the commit ends up
recorded as its own parent. With this change the same input is refused instead:

```
commitgraph: parent is not part of the index being encoded: 16dbf71e13ca56928da49c80a6b3e1236b0a9a8e
```

### Changes

- `encoder.go` — propagate the discarded lookup errors; checked parent-hash resolution via
  `lookupParentIndex`. (`prepare()` gained an error return; it's unexported with one call site.)
- `file.go` — new `ErrParentNotInIndex` sentinel, alongside the package's existing ones.
- `encoder_test.go` — three regression tests: an unresolvable single parent and a
  partially-resolvable merge both return an error instead of panicking, and a dangling edge from a
  custom `Index` returns `ErrParentNotInIndex` instead of silently encoding index `0`. All three
  were confirmed to fail against `main` — the first two panic, the third passes silently.

### AI assistance disclosure

Per [AI_POLICY.md](https://github.com/go-git/go-git/blob/main/AI_POLICY.md), this change was produced with AI assistance (Claude Opus 5), disclosed via an Assisted-by: trailer on the commit. I have reviewed and understand every line, verified the behaviour against reference git (2.55.0) and against git's own source, and confirmed the tests fail without the fix. I will respond to review feedback personally.

### DIFF
--- a/plumbing/format/commitgraph/encoder.go
+++ b/plumbing/format/commitgraph/encoder.go
@@ -2,6 +2,7 @@ package commitgraph
 
 import (
 	"crypto"
+	"fmt"
 	"io"
 	"math"
 
@@ -30,7 +31,10 @@ func (e *Encoder) Encode(idx Index) error {
 	hashes := idx.Hashes()
 
 	// Sort the input and prepare helper structures we'll need for encoding
-	hashToIndex, fanout, extraEdgesCount, generationV2OverflowCount := e.prepare(idx, hashes)
+	hashToIndex, fanout, extraEdgesCount, generationV2OverflowCount, err := e.prepare(idx, hashes)
+	if err != nil {
+		return err
+	}
 
 	chunkSignatures := [][]byte{OIDFanoutChunk.Signature(), OIDLookupChunk.Signature(), CommitDataChunk.Signature()}
 	chunkSizes := []uint64{szUint32 * lenFanout, uint64(len(hashes) * e.hash.Size()), uint64(len(hashes) * (e.hash.Size() + szCommitData))}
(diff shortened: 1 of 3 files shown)


pjbgf (approved):
@HubertKal thanks for working on this. :bow:

pjbgf:
x-ref: other in-flight commit graph changes #2183.

PR #2355, opened 2026-09-02
plumbing: transport/http, Compare origins when following redirects

hiddeco:
`url.URL.Host` was compared verbatim to decide whether a redirect had changed origin. That is both too loose and too strict: it ignores the scheme, and it treats `example.com` and `example.com:443` as different origins, dropping credentials the session still needed.

This adds one origin comparison over scheme, host and effective port, and uses it in the two places that need one: which headers may travel to a redirect target, and whether the session keeps its credentials. An http to https upgrade on the same host is allowed.

### How hosts compare

Registered names fold on ASCII case only. IP literals go through `netip.ParseAddr`, the same call `net`'s resolver uses to tell a literal from a name, so `[::1]` and `[0:0:0:0:0:0:0:1]` are one origin. Everything else is a different origin, including a trailing root dot and an IPv4-mapped literal against the IPv4 it dials. Both of those reach the same peer, but `net/http` sends the host as written in `Host`, so a server can route the two spellings to different virtual hosts. Same peer is not same authority.

No IDNA mapping is applied. It would only widen the equality, and go-git pins `x/net` in `go.mod` while `net/http` uses the copy vendored into the toolchain. The two are versioned separately, so a release that moved the Unicode tables under one and not the other would have the comparison merge origins `net/http` still dials apart.

### Behaviour change

When a redirect leaves the repository's origin, only the headers the transport sets itself are carried over. Headers added by an `Authorizer` are not, including non-credential ones such as a trace or tenant header. There is deliberately no opt-out: a caller who needs to authenticate to the origin a redirect moved the repository to can inject the credential from a `RoundTripper` on `Options.Client`, scoped to that origin. The transport keeps its own `CheckRedirect` on the client it copies, so the origin checks still apply. `Options.Client` now documents this.

### Deliberate deviations

Git does not carry `credential` authentication across a host, port or scheme change, and neither does this. But `http.extraHeader`, which is what an `Authorizer`-supplied header amounts to, *is* forwarded by Git to arbitrary hosts across a redirect. go-git now drops it.

A unicode hostname and its punycode are different origins here, so a redirect that only respells the host loses the credential and surfaces as `ErrAuthenticationRequired`. Git and curl keep it, because they convert to punycode when parsing the URL rather than when comparing. Normalising at parse time is the better fix, but it changes `Endpoint.Host`, error text and what `remote.Config().URLs` round-trips, so it does not belong here.

Allowing credentials across an http to https upgrade goes the other way, and departs from curl, Git and Fetch, which all count scheme as part of host identity. Auth is sent pre-emptively, so an http origin has already spent its credential in cleartext before any redirect exists.

### Tests

The origin comparison, including hosts where Unicode case-folding would merge two names that resolve to different servers; redirects within and across an origin; multi-hop chains; interaction with a caller-supplied `CheckRedirect`; and a redirected `git-upload-pack` POST on both sides of the origin boundary under `FollowRedirects: "true"`, where a 307 preserves method and body.

### DIFF
--- a/plumbing/transport/http/common.go
+++ b/plumbing/transport/http/common.go
@@ -218,6 +218,11 @@ func credentialsMayFollow(from, to *url.URL) bool {
 		effectivePort(from) == "80" && effectivePort(to) == "443"
 }
 
+// safeHeaders lists the headers go-git sets itself, none of which can carry a
+// caller credential. It has two consumers: trace.HTTP logs only these, and
+// stripCredentials keeps only these when a redirect leaves the credential's
+// origin. Adding a name here makes it both loggable and forwardable across an
+// origin boundary — do not add anything a caller can put a secret in.
 var safeHeaders = map[string]struct{}{
 	"User-Agent":        {},
 	"Host":              {},
--- a/plumbing/transport/http/common.go
+++ b/plumbing/transport/http/common.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"net/http"
+	"net/netip"
 	"net/url"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


pjbgf (approved):

PR #2361, opened 2026-09-07
internal: test, Isolate git from ambient config

hiddeco:
Break free of developer their machine configuration so tests always yield reliable results.

### DIFF
--- a/.github/workflows/git.yml
+++ b/.github/workflows/git.yml
@@ -59,8 +59,6 @@ jobs:
           echo "Git exec path mismatch: expected ${{ github.workspace }}/.git-dist/${{ matrix.git[0] }}, got $exec_path"
           exit 1
         fi
-        git config --global user.email "gha@example.com"
-        git config --global user.name "GitHub Actions"
 
     - name: Test
       run: make test-coverage
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,11 +43,6 @@ jobs:
         mkdir -p  ~/.ssh
         ssh-keyscan -H github.com > ~/.ssh/known_hosts
 
-    - name: Set Git config
-      run: |
-        git config --global user.email "gha@example.com"
-        git config --global user.name "GitHub Actions"
-
     - name: Validate
       run: make validate
 
--- a/_examples/common_test.go
+++ b/_examples/common_test.go
@@ -7,6 +7,8 @@ import (
 	"path/filepath"
 	"runtime"
 	"testing"
+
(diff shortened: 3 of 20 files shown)


Soph (approved):

Soph:
One unresolved regression from the local investigation on darwin/arm64 with Apple Git 2.50.1: concurrent-package testing exposes daemon readiness failures on this PR.

Command used:
```sh
go test -count=1 ./ ./plumbing/transport/http/... ./plumbing/transport/ssh/... ./plumbing/transport/git/... ./tests/...
```

Reported results, counting runs with a `transport/git` failure:

| Arm | Failed runs |
| --- | --- |
| Base `90d1a9a3` | 0/17 |
| PR `68602b783` | 12/17 |
| PR with only daemon invocation reverted to `exec.Command` | 0/6 |
| PR retaining real HOME/USERPROFILE/XDG_CONFIG_HOME | 0/6 |
| PR with existing 0700 isolated home beside test binary | 3/6 |
| PR with existing 0700 isolated home in TMPDIR | 3/6 |

Every observed failure was `startDaemon`'s one-second readiness budget expiring (`context canceled before the port is connectable`; 27 instances). Failure-path checks found the daemon alive, with no observed bind errors. The daemon-backed test sets are identical across the two commits, ruling out the changed `GitSupportsV2()` gate enabling extra suites.

The controlled reversions localise the trigger to home-directory redirection as seen by `git daemon`, but the mechanism remains unconfirmed. Keeping the real home passes even with `GIT_CONFIG_GLOBAL=""`, so this is not simply restoring global config. Creating or relocating the isolated home does not eliminate failures. Standalone startup measurements did not reproduce the delay; verbose daemon logging suppressed reproduction under concurrent tests. These observations are from one machine; impact on Linux CI is unknown.

I recommend resolving this before merge, without simply dropping the home isolation needed by pre-2.32 Git. The existing one-second readiness budget is a fragility, but increasing it alone would not explain the PR/base asymmetry. A useful next diagnostic would be sampling the timed-out process and its process tree before cleanup to locate where startup is delayed; an alive, delayed process need not emit a Git error message.

Separately, the `isolatedHome` documentation needs qualification: under `go test -c` or debugger-built tests, `os.Executable()` does not guarantee a private, unique, automatically removed build directory. Under normal `go test`, the 0700 protection comes from an ancestor of the leaf `bNNN` directory, not necessarily the leaf itself.

*AI-assisted write-up of the investigation results supplied by the contributor; the measurements above were not independently rerun by this assistant.*

pjbgf (approved):

Issue #2136, opened 2026-05-17
BasicAuth cleared permanently after cross-origin redirect in HTTP transport


dangerousbeans:
## Description

When `repo.Fetch()` (or any upload-pack session) encounters an HTTP redirect that changes the host or port, `ModifyEndpointIfRedirect` permanently sets `s.auth = nil` for the entire session. This means the subsequent `POST /git-upload-pack` request is sent without any authentication, even though valid credentials were provided via `BasicAuth`.

CLI `git` does not have this problem — it re-applies `http.extraHeader` credentials independently on each new request after updating the remote URL.

## Affected code

`plumbing/transport/http/common.go`, lines ~387-390:

```go
if host != s.endpoint.Host || effectivePort(r.URL.Scheme, port) != effectivePort(s.endpoint.Protocol, s.endpoint.Port) {
    s.endpoint.User = ""
    s.endpoint.Password = ""
    s.auth = nil  // ← clears auth for ALL subsequent requests in the session
}
```

## Steps to reproduce

1. Set up two HTTP servers on the same host but different ports (e.g., `127.0.0.1:A` and `127.0.0.1:B`)
2. Server A redirects `GET /info/refs?service=git-upload-pack` to Server B (HTTP 307)
3. Server B allows `/info/refs` without auth but requires Basic Auth on `POST /git-upload-pack`
4. Clone a repo from Server B directly, then set the remote to Server A
5. Call `repo.Fetch()` with valid `BasicAuth` credentials

**Expected:** Fetch succeeds — auth should be re-applied to the upload-pack POST after the redirect.

**Actual:** Fetch fails with `authentication required`. go-git clears `s.auth` when `ModifyEndpointIfRedirect` detects the port change, so the upload-pack POST is sent without credentials.

## Reproduction test

```go
func TestFetch_BasicAuth_ClearedOnRedirect(t *testing.T) {
    // Server B (backend): serves git protocol, requires auth on upload-pack only
    backend := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        isDiscovery := strings.Contains(r.URL.Path, "/info/refs")
        if !isDiscovery {
            user, pass, ok := r.BasicAuth()
            if !ok || user != "x-access-token" || pass != "token" {
                w.Header().Set("WWW-Authenticate", `Basic realm="git"`)
                http.Error(w, "authentication required", http.StatusUnauthorized)
                return
            }
        }
        // ... serve git-http-backend
    }))

[...]

PR #2362, opened 2026-09-07
plumbing: transport/http, Re-acquire credentials at a new origin

hiddeco:
Related to #2136 (closes for `v6`).

## Summary

HTTP credentials are now asked for **one origin at a time**, instead of being handed to the transport once and applied to every request it makes.

- A repository that has genuinely moved to another host can be cloned again — provided you can supply a credential for that host.
- `plumbing/client` callers are unaffected: `WithHTTPAuth` still works, and now means "this credential belongs to the host in the repository URL, and to no other".
- Direct transport callers must replace `http.Options.Authorizer` with `http.Options.Credentials` (`v6` is alpha, so we can still take the break).
- When authentication fails because a credential was withheld at an origin boundary, the error now says that, and names both origins.

## The problem

#2355 made the transport withhold the caller's credential as soon as a redirect leaves the repository's origin, which closed a credential leak. It also made a cross-origin redirect a dead end: the new host asks for authentication, and nothing was allowed to answer. A repository that has really moved (they exist — see #2136) could not be cloned at all, no matter how many credentials the caller held.

The missing piece was a way for the transport to ask a question it never had to ask before: *who authenticates for **this** host?* A single `Authorizer` cannot answer it, because it carries no idea of which host its credential belongs to.

<details>
<summary>What Git does here, and why it is not a template</summary>

Git has half of an answer, and only for a first request that went out anonymous: it re-derives a credential from the URL the chain arrived at, and the `HTTP_REAUTH` loop re-issues the request with it — up to twice, each attempt preceded by a fresh `credential_fill()`.

When a username and password are *already* held, `handle_curl_result()` answers the 401 by rejecting that credential and returning `HTTP_NOAUTH`, which `http_request_recoverable()` returns on before the redirect rewrite is ever reached. So with userinfo in the URL, or with `http.proactiveAuth` set, a cross-host redirect to a 401 is a dead end in Git too — and on the way out it erases the original host's credential from every configured helper.

go-git always sends auth pre-emptively, which is exactly that case. There is no Git precedent to copy for the retry below.

What Git does have that we do not is a credential store: `credential_fill()` can ask whatever `credential.helper` is configured about the host it just landed on, and prompt a human when no helper answers. We have no helper protocol, no keychain, and nothing to prompt — so the only thing that can answer for a new origin is the caller.
</details>

## What you need to change

**Using `plumbing/client`?** Nothing. This keeps working, and its scope is now explicit — the credential is offered to the repository's own origin only:

```go
c := client.New(client.WithHTTPAuth(auth))
```

**Using the transport directly?** `Authorizer` becomes a `CredentialsFunc`:

```go
// before
http.Options{Authorizer: auth}

// after
http.Options{Credentials: http.ForRepositoryOrigin(auth)}
```

**Want a clone to survive a redirect to another host?** Name that host as well. Each adapter answers only for the origin it stands for, so no credential can be drawn to a host you did not name:

```go
import (
        nethttp "net/http"

        "github.com/go-git/go-git/v6/plumbing/client"
        "github.com/go-git/go-git/v6/plumbing/transport/http"
)

gateway, _ := url.Parse("https://gateway.example")

c := client.New(
        // the host in the repository URL
        client.WithHTTPAuth(auth),
        // the host a redirect may move it to
        client.WithHTTPCredentials(http.ForOrigin(gateway, func(r *nethttp.Request) error {
                r.Header.Set("Authorization", "Bearer "+gatewayToken)
                return nil
        })),
)
```

For anything more dynamic — a `.netrc` reader, a keychain, a token service — write a `CredentialsFunc` yourself. It is handed a `CredentialRequest` describing what the credential is wanted for (`TargetOrigin`, `TargetPath`, `RepositoryURL`, `Redirected`) and returns a credential or declines. Multiple sources compose: every one is consulted, most-recently-configured first, and the first to answer for the origin in question wins. That is as close as we get to a helper chain.

## How the re-authentication works

When a 401 comes back from a chain that left the repository's origin, the transport:

1. validates the redirect target — it must still look like a repository, and the only scheme change allowed is `http` → `https` on the same host;
2. asks your `CredentialsFunc` for a credential belonging to the origin the chain actually reached;
3. re-issues the discovery request once with it;
4. keeps that credential for the rest of the session, but only while the session is still pointed at the origin it was acquired for.

The re-issued request is built by the same code path as the original, so nothing the server chose — userinfo, query, fragment — rides along on it, and it is sent through a client that refuses every redirect, so a new credential cannot be walked onwards hop by hop.

Both of those are deliberately stricter than Git, which retries up to twice and lets those retries keep following redirects. One attempt is enough here: your source has already answered for this origin, and asking it the same question again only makes sense when a store or a human might answer differently.

## When it still fails

If nobody can answer for the new origin, the failure now explains itself. `transport.CredentialsDroppedError` names the origin the withheld credential was issued for and the origin the chain ended at — the difference between *"wrong password"* and *"your credential was withheld at a boundary, and the host that wants one is not the host you named"*.

To be plain about what that means for #2136: **a caller using `WithHTTPAuth` alone still gets a 401 there.** That option answers only for the repository's origin, and the origin the chain reached is not it. What is new is that the error now names both origins instead of looking like a bad password, and that supplying a source for the second origin makes the clone work.

## Also fixed along the way

Each is its own commit.

- **A token in a query string is treated as a credential.** Several forges accept one (`?private_token=`, `?job_token=`). We copied the repository URL's query onto the session base URL, where it rode along on every later request — including to a redirect's new origin, where the credential itself was correctly withheld — and it was printed verbatim in every error and trace line. The query is now dropped exactly where a credential is dropped, and redacted except for the parameter names go-git sets itself. Such a token never worked on the discovery request anyway: the old code appended `?service=` to a URL that already had a query, producing `?private_token=X?service=git-upload-pack` — one parameter whose value swallowed the service name, and no `service` parameter at all.
- **A redirect target cannot plant a secret in an error message.** Where `net/http` builds the error itself it copies the `Location` header into `url.Error.URL` verbatim, so the raw target was printed beside go-git's own redacted rendering of the very same URL. It was not only the refusals: any transport-level failure after a redirect carries the target the same way — a DNS lookup that fails names the host it looked up — so the leak did not need a policy to be refusing anything. Every error from the one `client.Do` in the package now passes through the same redaction.
- **A hostile server cannot exhaust client memory through an error page.** The error path read the response body to EOF and put all of it in the returned error. A redirect target picks that body, so an endless 401 page was enough, from anywhere a redirect can reach. It is now bounded to 8 KiB. A `Location` header was the same opening from the other side — a 400 KB target became a 400 KB error string the caller then held — so every part of a URL a message renders is bounded too, at a kilobyte.
- **Host comparison matches `net/http` exactly.** We folded some host spellings that `net/http` does not. On a redirect between two of them `net/http` had already dropped `Authorization` while go-git recorded no origin crossing — so nothing was re-acquired, no error named the origin, and a caller holding a perfectly good credential was never asked for it.
- **A credential is scoped to the repository path, not just the host.** A redirect can move the path without leaving the origin, so every origin check passes and the credential is spent on whatever repository the redirect picked. `CredentialRequest.TargetPath` lets a source see that. Note what it can and cannot do: on a same-origin hop the credential has already been sent by the time your source is consulted, so what the comparison buys is that the move is visible and the credential is not reused for the rest of the session. Fixing this also turned up `applyRedirect` losing the correspondence between `URL.Path` and `URL.RawPath`, which sent every later request — the pack POST, the dumb protocol's object GETs, the retry — to a path neither the caller nor the redirect named.
- **A server's error text can no longer redraw your terminal.** `Err.Reason` is the response body, and it was interpolated into the error message raw, so up to 8 KiB of bytes chosen by the far end were printed beside a URL that was carefully quoted and redacted. An escape sequence there rewrites the reader's screen; a newline forges what looks like a separate log line. Anything that would not print as itself is now replaced with a space. Callers matching on the body text should know it is no longer byte-for-byte what the server sent.
- **The discovery response body is closed on error**, which was leaking a connection on every failed handshake.

<details>
<summary>Why the sanitising happens where it does</summary>

The replacement is decided by `unicode.IsPrint`, which covers the C0 and C1 controls that begin an escape sequence as well as the format characters that reorder what follows.

It happens in `checkError` rather than in `Error`, so the field carries the same guarantee the message does — as `URL` already did — and something that logs `Reason` on its own is covered too.

Nothing else in the package needed the same treatment: `url.Parse` refuses a control character and `http.NewRequest` refuses it again, so no `*url.URL` here can carry one however hostile the `Location` it came from, and the remaining server-supplied strings are rendered with `%q`.
</details>

## Documented, not changed

Under the non-default `FollowRedirects` policy, `net/http` replays a request body from `GetBody` on a 307 or 308, and we keep `Content-Type` and `Content-Length` — so a pack request arrives complete at an origin the server chose. For upload-pack that discloses which objects the caller already has; for receive-pack, the packfile being pushed.

curl and Git behave the same way under `http.followRedirects=true`, so this is not a defect to fix. It is stated on the option that opens it, together with the `CheckRedirect` a caller can set to allow the redirect for discovery while refusing it for a request carrying a body.

Three things do not happen after such a POST:

1. it is not retried;
2. `Credentials` is not consulted for the origin it reached;
3. its crossing is not recorded.

Those are now written there as a description of what the transport does today, not as guarantees — adding re-authentication to the pack POST later should extend that documentation rather than contradict it. The disclosure above does not rest on any of the three: `net/http` replays the body before a retry, a lookup or a crossing record could apply.

## API surface

New in `plumbing/transport/http`:

- `CredentialsFunc`, and the `CredentialRequest` it is given: `TargetOrigin`, `TargetPath`, `RepositoryURL`, `Redirected`, plus `IsOrigin`.
- `Credential` and `Authorizer`, which are the answer.
- `ForOrigin`, `ForRepositoryOrigin` and `Chain`, for the shapes most callers want.

New in `plumbing/client`: `WithHTTPCredentials`, taking a `CredentialsFunc` directly. It composes with `WithHTTPAuth` rather than overwriting it.

New in `plumbing/transport`: `CredentialsDroppedError`.

Changed in `plumbing/transport/http` — both exported fields of `Err` now carry what the message carries, so reading one is as safe as printing the error:

- `Err.URL` is an independent copy, redacted the way `Error` renders it: query values, password and fragment replaced, and any part longer than a kilobyte replaced whole. It no longer aliases the transport's live request URL.
- `Err.Reason` is bounded to 8 KiB, with every character that would not print as itself replaced by a space.

Removed: `http.Options.Authorizer`.

<details>
<summary>Two design notes on <code>CredentialRequest</code></summary>

**`IsOrigin` exists so callers do not compare hosts by hand.** A source standing in for a credential store knows many origins and none of them up front, and hand-rolled comparison is where that goes wrong: a suffix match accepts `evilexample.com` for `example.com`, and equality on `Host` alone forgets that a port makes another origin. `IsOrigin` exposes the transport's own comparison, including the permitted `http:80` → `https:443` upgrade.

**`TargetPath` is Git's `path` credential attribute** — the one `credential.useHttpPath` turns on — and it is re-derived from the redirect target exactly as Git re-derives it after `update_url_from_redirect`. It is kept in its escaped on-wire form, because a comparison has to be made on spelling: distinct paths can decode alike. A source doing a *lookup* should decode it and trim the leading slash; a source doing a *comparison* should not.

**How often a `CredentialsFunc` is consulted is deliberately unspecified.** Today it is asked once per origin, and again when a redirect moves the origin or the repository path — but that is not documented as the whole set. Git's default is challenge-driven: it makes an unauthenticated request first and fills only once a 401 comes back, with pre-emptive filling opt-in behind `http.proactiveAuth`. We would like to be able to move that way without breaking a source that prompts. The contract asks instead that an implementation be cheap to call and safe to call repeatedly.
</details>


plumbing/transport/http/discovery.go
```diff
@@ -0,0 +1,58 @@
...
+// it, so an escaped segment such as %2F survives — rather than from
+// base.String(), which would carry base's userinfo into the request URL, where
+// it reaches trace output and error strings, and would bring the clone URL's
+// query and fragment along too. Authentication is applied by the caller, never
+// from the URL.
+func (d discovery) request(ctx context.Context, base *url.URL) (*http.Request, error) {
+	origin := &url.URL{Scheme: base.Scheme, Host: base.Host, Path: base.Path, RawPath: base.RawPath}
+	infoURL := origin.JoinPath("info/refs").String()
+	if q := d.query(); q != "" {
+		infoURL += "?" + q
```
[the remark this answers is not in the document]
pjbgf: I don't think this is a supported feature for Smart HTTP upstream anyway.

plumbing/transport/http/reauth.go
```diff
@@ -0,0 +1,238 @@
...
+		// retryErr becomes its cause — the redirect refusal wrapped in a
+		// *url.Error, or whatever else client.Do returned.
+		cause := redactRetryError(retryErr)
+		if stopped(retryErr) {
+			// A clone the caller stopped is not a clone that needs
+			// credentials: leaving the 401 in the chain has a caller who
+			// classifies authentication first prompt for a password on a
+			// clone the user cancelled. Both errors still render.
+			return spent, resp, fmt.Errorf("%s: %w", err, cause)
+		}
```
pjbgf: This feels redundant.

```suggestion
```

plumbing/transport/http/reauth.go
```diff
@@ -0,0 +1,238 @@
...
+		return spent, resp, fmt.Errorf("%w: %w", err, cause)
```
pjbgf: For clarity:


```suggestion
		if stopped(retryErr) {
			// A clone the caller stopped is not a clone that needs credentials, so
			// the 401 renders in the message but leaves the error chain: %s, not %w.
			// Otherwise a caller that classifies authentication before cancellation
			// prompts for a password on a clone the user aborted.
			return spent, resp, fmt.Errorf("%s: %w", err, cause)
		}
		return spent, resp, fmt.Errorf("%w: %w", err, cause)
```

Soph (dismissed):

Soph (approved):

pjbgf (approved):

PR #2365, opened 2026-09-07
internal: test, Resolve Apple Git before isolating its environment

Soph:
## Why

Follow-up to #2361. On darwin/arm64 with Apple Git 2.50.1, HOME isolation exposed failures in the git transport tests' existing one-second daemon readiness budget under concurrent-package testing.

Failure-only instrumentation (no changes to startup or stdio) captured `/usr/bin/git` with an `xcodebuild -license check` child while readiness timed out, before the actual Git daemon started. Invoking the same selected Xcode Git executable directly, retaining isolation and the deadline, passed 6/6 concurrent runs; restoring the launcher failed 2/3 runs. This localises the delay to Apple's launcher/license-check path. The exact HOME-dependent behaviour inside Apple's launcher remains unconfirmed.

## Change

- On Darwin only, when exec resolves the requested command to `/usr/bin/git`, resolve the actual Git executable with `/usr/bin/xcrun --find git` before applying the isolated environment.
- Keep other PATH-selected Git binaries and non-Git commands unchanged.
- Preserve HOME/config isolation, GIT_EXEC_PATH, and the readiness deadline.
- Bound resolution to 30 seconds and report failures through `exec.Cmd.Err`, rather than silently falling back to the launcher.
- Do not cache resolution, so changes to PATH and Xcode selection between tests remain effective.

Regression tests cover explicit and PATH-selected Apple Git, custom PATH-selected Git, GIT_EXEC_PATH preservation, missing executables, unrelated commands, and resolution failure. The resolution tests were observed failing before the implementation.

## Validation

- Fixed concurrent workload: **6/6 runs passed**:
  ```sh
  go test -count=1 ./ ./plumbing/transport/http/... ./plumbing/transport/ssh/... ./plumbing/transport/git/... ./tests/...
  ```
- `go test -race -short ./...` passed.
- `go test -race -count=1 ./internal/test/gitenv` passed.
- Targeted golangci-lint v2.11.4: zero issues.
- gitenv test binaries cross-compiled for Linux/amd64 and Windows/amd64; not executed on those systems.
- Full `go test -race ./...` passed gitenv and the daemon package but was not entirely green: the Git-ignore oracle mismatch also reproduces on the unmodified PR with this Apple Git version; a repository clone pack-checksum failure did not recur in five targeted race reruns.

## Draft review points

This deliberately resolves only the exact Apple launcher path, not arbitrary wrappers or symlinks. Resolution runs per command, outside isolated Git execution. Please review that trade-off and the macOS-specific handling before merging. No Linux runtime validation has been performed locally.

AI-assisted implementation, investigation, tests, and write-up using OpenAI ChatGPT.

### DIFF
--- a/internal/test/gitenv/gitenv.go
+++ b/internal/test/gitenv/gitenv.go
@@ -12,8 +12,9 @@
 // built, so a stray value redirects the command rather than colouring it.
 //
 // Command returns a command with all of that removed and an identity supplied.
-// Env returns the environment on its own, for a caller that builds the command
-// itself. Every variable this package names is set unconditionally, whatever
+// CommandContext is the same command bound to a context, for a test that has
+// the harness's own to hand it. Env returns the environment on its own, for a
+// caller that builds the command itself. Every variable this package names is set unconditionally, whatever
 // the machine had: a value inherited from the environment is the thing being
 // isolated, so honouring it would defeat the purpose. A caller who needs
 // something different appends it, since the last entry for a variable is the
@@ -26,6 +27,7 @@
 package gitenv
 
 import (
+	"context"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


internal/test/gitenv/gitenv.go
```diff
@@ -169,8 +169,11 @@ var isolatedHome = sync.OnceValue(func() string {
 // read the machine's configuration or be pointed at another repository, where
 // one built with exec.Command has to remember not to. The caller sets Dir, and
 // may append to Env for anything it needs on top: later entries win.
+// On macOS, Apple's /usr/bin/git launcher is resolved before applying Env;
+// resolution failures are returned through Cmd.Err and prevent execution.
 func Command(name string, args ...string) *exec.Cmd {
 	cmd := exec.Command(name, args...)
+	resolveAppleGit(cmd)
```
hiddeco: Reviewing my own code here, but after a night of sleep I wonder if it might be nice to add a `CommandContext` and wire this all the way down to `resolveAppleGit`. That way tests can hypothetically inject the context from the testing harness.

pjbgf (approved):

PR #2366, opened 2026-09-08
plumbing: transport/http, Fix response handling

hiddeco:
Six things the HTTP transport got wrong about a response it did not expect. Each is its own commit.

## Summary

- **An HTML page served at `/info/refs` was accepted as a ref list.** An SSO interstitial or a 200 error page fails the smart content-type test and is handled as a dumb server, and the dumb decoder skipped every line it could not parse. `Remote.List` returned no error and one reference named after an HTML fragment; `Clone` then failed with `reference not found`, naming nothing about the response. `deadbeef\trefs/heads/master` decoded to a reference at a hash the server never sent, because `plumbing.FromHex` pads a short hex string. Decoding now fails on the first malformed line, as canonical git's `parse_info_refs` does.

- **A failed fetch leaked a socket and three goroutines, every time.** `Fetch` returned on a `NegotiatePack` error without closing the response — including when it returns *success*: `ErrNoChange` arrives after the POST has already fired, so an up-to-date fetch into an existing shallow clone over protocol v0/v1 leaked on every call. Five failed fetches held five sockets and took goroutines from 3 to 18, never released. A dumb walk leaked the same way through `downloadFile`, which returned on a copy failure without closing the body it was reading — one connection per object, and the walk asks for every object as a file of its own.

- **An error response body was never closed, and was shown whatever it turned out to be.** Left open it holds its connection for the life of the process, and a CDN's HTML error page was spliced into the Go error — `<head>`, `<hr>` and all, as a CI run against gitlab.com showed. The body is now closed once its message has been taken, the discard that keeps its connection is bounded, and the message reaches the error only when the server sent `text/plain` — the rule git applies in `show_http_message` — quoted, so a multi-line body cannot forge a record in your logs.

- **Every spent response dropped its connection.** The advertisement ends at a flush-pkt and a decoder stops at one, so the rest of the body — often just the terminating chunk — was still outstanding at the close, and net/http returns a connection to the pool only once its body has reached EOF. The smart advertisement, each negotiation round and each v2 command now discard what is left, under a bound, so the request that follows reuses the connection instead of paying a fresh handshake. Ten 404s carrying a 32 KiB error page: one connection with the discard, ten without.

- **`errors.Is(err, transport.ErrCommandUnsupported)` was false for a dumb push.** `dumbPackSession.Push` failed with a bare `fmt.Errorf`, so a caller falling back to another transport could not detect the refusal.

- **A `charset` parameter made a smart server look dumb.** The smart/dumb test compared the whole `Content-Type` header for equality; it now stops at the first `;`, trims and lower-cases, as git's `extract_content_type()` does. The accepted set only grows, so nothing that works today can start failing.

## Errors you get instead

| before | after |
|---|---|
| `reference not found` | `invalid response: https://host/repo.git/info/refs?service=git-upload-pack served content type "text/html": invalid info/refs: line 1 has no tab` |
| an HTML error page spliced into the error, 8 KiB of it | `http transport: unexpected requesting "https://host/repo.git/info/refs?service=git-upload-pack" status code: 502` |
| the same, for a message worth reading | `http transport: authorization failed: unexpected requesting "https://host/repo.git/info/refs?service=git-upload-pack" status code: 403: "ERROR: The project you were looking for could not be found."` |

## What you may need to change

- **`packp.InfoRefs.Decode` returns an error** on a line that is not hash-tab-refname, where it skipped it. Exported plumbing; one in-tree consumer. Empty input still decodes to no references, because that is what `git update-server-info` writes for a repository with nothing to advertise. A leading UTF-8 BOM now fails, as it does in git.
- **`Err.Reason` is empty unless the message arrived as `text/plain`**, and `Err.Error()` quotes what it holds. It is an exported field; anything matching on it should read its new doc.
- **A non-git `/info/refs` surfaces `transport.ErrInvalidResponse`** wrapping `packp.ErrInvalidInfoRefs`, where it previously surfaced nothing.

## Known limitation

`Decode` checks the hash and requires a non-empty name; it does not validate the name's spelling. A page that prints a full-length commit id before a tab can still yield a reference named after markup. `ReferenceName.Validate()` would catch it, but it rejects `refs/tags/v1^{}` — a peeled tag `git update-server-info` writes for every annotated tag — so it cannot be used here. Canonical git's `parse_info_refs` does not validate the name either, and neither does the smart path.

### DIFF
--- a/plumbing/transport/http/dumb.go
+++ b/plumbing/transport/http/dumb.go
@@ -134,7 +134,8 @@ func (r *fetchWalker) downloadFile(fp string) (rErr error) {
 
 	f, err := r.fs.TempFile(filepath.Dir(fp), filepath.Base(fp)+".temp")
 	if err != nil {
-		_, _ = io.Copy(io.Discard, res.Body)
+		// The body is an object or pack, so there is nothing to gain from
+		// draining a file this walk has given up on writing.
 		_ = res.Body.Close()
 		return err
 	}
@@ -145,6 +146,9 @@ func (r *fetchWalker) downloadFile(fp string) (rErr error) {
 	}()
 
 	if _, err := ioutil.CopyBufferPool(f, res.Body); err != nil {
+		// As above: an object or a pack, so a walk that has given up on
+		// writing it gains nothing from draining what is left.
+		_ = res.Body.Close()
 		return err
 	}
 	if err := res.Body.Close(); err != nil {
--- a/internal/transport/v2.go
+++ b/internal/transport/v2.go
@@ -224,24 +224,9 @@ func FetchV2(ctx context.Context, st storage.Storer, req *FetchRequest, round Fe
 
 		if out.Packfile {
 			streamErr := streamPackfile(ctx, st, packReader, req.Progress, req.Filter)
-			// Skip draining/closing on cancellation: streamPackfile wraps
-			// packReader in a NewContextReader, whose background goroutine
-			// can still be blocked in the underlying Read after the
-			// <-ctx.Done() branch returns (see its doc comment). Closing
-			// here would race that goroutine's in-flight Read -- the same
-			// class of race fixed in plumbing/transport/http's Fetch/Push
-			// for the v0/v1 path. On a non-cancellation error or success,
-			// streamPackfile's last Read has already returned via the
-			// result channel and the goroutine is quiescent, so closing is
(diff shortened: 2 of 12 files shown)


plumbing/transport/http/dumb.go
pjbgf, line 148: On error, `res.Body` is not closed.

pjbgf (approved):

PR #2372, opened 2026-09-10
git: worktree, Fix unreadable subtree handling and recovery

hiddeco:
Missing, corrupt, or malformed tree objects could cause several misleading or unrecoverable outcomes:

- **Incomplete results reported as success.** An unreadable subtree could end enumeration without an error, leaving checkout or archive output incomplete. An index rebuilt from partial results could cause a later commit to record omitted paths as deleted.
- **A failed operation could still move HEAD.** Checkout, reset, or pull could change references before failing to read the target tree. After a failed pull, retrying could incorrectly report that the repository was already up to date.
- **Hard reset could fail to recover.** Even with a healthy target commit, hard reset and forced checkout could fail because they needed to read the unreadable old HEAD tree.

This change reports unreadable subtrees as errors and updates references only after reset processing succeeds. Hard reset and forced checkout identify tracked deletions from the pre-reset index, allowing recovery without reading the old HEAD tree. `KeepReset` retains its old-tree conflict checks. Index and worktree updates are not atomic.

It also fixes path handling for malformed trees: lookup descends only through directory entries, and walking a directory whose name contains slashes no longer corrupts subsequent sibling paths. Directory-mode descent follows Git’s [find_tree_entry](https://github.com/git/git/blob/v2.54.0/tree-walk.c#L568-L633). Direct object access remains available for inspection.

Missing subtrees remain errors in partial clones because go-git cannot fetch them lazily. Documentation clarifies validation boundaries, error identities, iterator recovery, and the protections bypassed by direct use of `Worktree.Filesystem`.

Regression tests cover repeated failures, reference and index preservation, recovery after missing objects become available, and malformed-entry path handling

### DIFF
--- a/utils/merkletrie/index/node.go
+++ b/utils/merkletrie/index/node.go
@@ -29,6 +29,12 @@ type node struct {
 // RootNodeOptions contains configuration for the root node.
 type RootNodeOptions struct {
 	UpholdExecutableBit bool
+
+	// IgnoreSkipWorktree reports every entry to the diff, including those
+	// flagged SkipWorktree. Callers that bring the index to a tree need
+	// those entries; callers that compare the index with the worktree must
+	// leave it unset, because a skipped path is absent from disk by design.
+	IgnoreSkipWorktree bool
 }
 
 // NewRootNode returns the root node of a computed tree from a index.Index,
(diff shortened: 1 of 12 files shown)


utils/merkletrie/index/node.go
```diff
@@ -29,6 +29,12 @@ type node struct {
 // RootNodeOptions contains configuration for the root node.
 type RootNodeOptions struct {
 	UpholdExecutableBit bool
+
+	// IgnoreSkipWorktree reports every entry to the diff, including those
+	// flagged SkipWorktree. Callers that bring the index to a tree need
+	// those entries; callers that compare the index with the worktree must
+	// leave it unset, because a skipped path is absent from disk by design.
+	IgnoreSkipWorktree bool
```
hiddeco, line 37: Not too keen on the naming here, and open to alternatives.

PR #2373, opened 2026-09-10
git: remote, Keep locally held shallow boundaries reachable. Fixes #2367

### LINKED ISSUE #2367: push: false non-fast-forward rejection with multiple shallow boundaries
## Bug Description

A depth-one fetch of multiple related refs can cause `PushContext` to reject an update that is unambiguously fast-forward.

Given this graph:

```text
R -- B -- C  refs/heads/flipt/staging-test/blocker
     |
     +------  refs/heads/staging-test
     |
     +------  refs/heads/flipt/staging-test/target
```

fetching `staging-test` and `flipt/staging-test/*` with `Depth: 1` records both `B` and `C` as shallow boundaries. If the client then creates `N` with `B` as its direct parent and pushes `N` over the target branch, go-git rejects the update with:

```text
non-fast-forward update: refs/heads/flipt/staging-test/target
```

Expected behavior: the push succeeds because `N` is a direct descendant of the advertised remote head `B`.

Actual behavior: go-git rejects the update locally as non-fast-forward before sending it to `git-receive-pack`.

The equivalent operation using native Git with the same depth-one refspecs and commit graph succeeds.

## go-git Version

Reproduced with:

- `github.com/go-git/go-git/v6 v6.0.0-alpha.5`
- current `main`: `v6.0.0-alpha.5.0.20260907191556-57ed51864460`

Environment used for the reproduction:

```text
go version go1.27.1 darwin/arm64
git version 2.50.1 (Apple Git-155)
Darwin arm64
```

## Steps to Reproduce

Save the three files below, make the shell fixture executable, and run:

```sh
chmod +x testdata/setup_git_remote.sh
go mod tidy
go test -run '^TestPushDirectDescendantWithMultipleShallowBoundaries$' -count=1 -v
```

The test is written as a regression test for the expected behavior, so it fails on affected versions at the final `PushContext` call.

<details>
<summary>go.mod</summary>

```go
module example.com/go-git-shallow-regression

go 1.25.0

require (
	github.com/go-git/go-billy/v6 v6.0.0-alpha.2
	github.com/go-git/go-git/v6 v6.0.0-alpha.5
)

require (
	github.com/Microsoft/go-winio v0.6.2 // indirect
	github.com/ProtonMail/go-crypto v1.4.1 // indirect
	github.com/cloudflare/circl v1.6.3 // indirect
	github.com/emirpasic/gods v1.18.1 // indirect
	github.com/go-git/gcfg/v2 v2.0.2 // indirect
	github.com/kevinburke/ssh_config v1.6.0 // indirect
	github.com/klauspost/cpuid/v2 v2.3.0 // indirect
	github.com/pjbgf/sha1cd v0.6.0 // indirect
	github.com/sergi/go-diff v1.4.0 // indirect
	golang.org/x/crypto v0.54.0 // indirect
	golang.org/x/net v0.57.0 // indirect
	golang.org/x/sync v0.22.0 // indirect
	golang.org/x/sys v0.47.0 // indirect
)
```

</details>

<details>
<summary>shallow_push_test.go</summary>

```go
package shallowregression_test

import (
	"fmt"
	"os"
	"os/exec"
	"path/filepath"
	"strings"
	"testing"
	"time"

	"github.com/go-git/go-billy/v6/osfs"
	"github.com/go-git/go-git/v6/config"
	"github.com/go-git/go-git/v6/plumbing/cache"
	gitfilesystem "github.com/go-git/go-git/v6/storage/filesystem"
)

const (
	baseBranch   = "staging-test"
	targetBranch = "flipt/staging-test/target"
)

// TestPushDirectDescendantWithMultipleShallowBoundaries is a black-box
// regression test for go-git's false non-fast-forward rejection.
//
// The shell fixture constructs this graph and exposes it through git daemon:
//
//	R -- B -- C  flipt/staging-test/blocker
//	     |
//	     +------  staging-test and flipt/staging-test/target
//
// A depth-one fetch records B and C as shallow boundaries. The test creates N
// as a direct child of B and pushes N to target. That update is a fast-forward,
// but affected go-git versions reject it before contacting git-receive-pack.
func TestPushDirectDescendantWithMultipleShallowBoundaries(t *testing.T) {
	remoteURL, baseHash, blockerHash := setupGitRemote(t)

	clientPath := filepath.Join(t.TempDir(), "client")
	if err := os.Mkdir(clientPath, 0o755); err != nil {
		t.Fatalf("create client repository directory: %v", err)
	}

	storer := gitfilesystem.NewStorage(
		osfs.New(clientPath),
		cache.NewObjectLRUDefault(),
	)
	repository, err := git.Init(
		storer,
		git.WithDefaultBranch(plumbing.NewBranchReferenceName(targetBranch)),
	)
	if err != nil {
		t.Fatalf("initialize client repository: %v", err)
	}
	t.Cleanup(func() {
		if err := repository.Close(); err != nil {
			t.Errorf("close client repository: %v", err)
		}
	})

	if _, err := repository.CreateRemote(&config.RemoteConfig{
		Name: "origin",
		URLs: []string{remoteURL},
	}); err != nil {
		t.Fatalf("create origin remote: %v", err)
	}

	err = repository.FetchContext(t.Context(), &git.FetchOptions{
		RemoteName: "origin",
		RefSpecs: []config.RefSpec{
			"+refs/heads/staging-test:refs/remotes/origin/staging-test",
			"+refs/heads/flipt/staging-test/*:refs/remotes/origin/flipt/staging-test/*",
		},
		Depth: 1,
		Tags:  plumbing.NoTags,
	})
	if err != nil {
		t.Fatalf("depth-one fetch: %v", err)
	}

	shallows, err := storer.Shallow()
	if err != nil {
		t.Fatalf("read shallow boundaries: %v", err)
	}
	assertShallowBoundaries(t, shallows, baseHash, blockerHash)

	remoteTarget, err := repository.Reference(
		plumbing.NewRemoteReferenceName("origin", targetBranch),
		true,
	)
	if err != nil {
		t.Fatalf("resolve fetched target branch: %v", err)
	}
	if remoteTarget.Hash() != baseHash {
		t.Fatalf("target branch = %s, want base commit %s", remoteTarget.Hash(), baseHash)
	}

	parent, err := repository.CommitObject(baseHash)
	if err != nil {
		t.Fatalf("load target parent commit: %v", err)
	}
	newHash := storeDirectChild(t, storer, parent)

	localTarget := plumbing.NewBranchReferenceName(targetBranch)
	if err := storer.SetReference(plumbing.NewHashReference(localTarget, newHash)); err != nil {
		t.Fatalf("set local target branch: %v", err)
	}

	// This must succeed because newHash is a direct descendant of the remote
	// target. On affected versions this assertion fails with a false
	// "non-fast-forward update" error.
	err = repository.PushContext(t.Context(), &git.PushOptions{
		RemoteName: "origin",
		RefSpecs: []config.RefSpec{
			config.RefSpec(fmt.Sprintf("%[1]s:%[1]s", localTarget)),
		},
	})
	if err != nil {
		t.Fatalf(
			"push of direct descendant %s over %s was rejected: %v",
			newHash,
			baseHash,
			err,
		)
	}
}

func setupGitRemote(t *testing.T) (string, plumbing.Hash, plumbing.Hash) {
	t.Helper()

	fixtureRoot := t.TempDir()
	script, err := filepath.Abs(filepath.Join("testdata", "setup_git_remote.sh"))
	if err != nil {
		t.Fatalf("resolve fixture script: %v", err)
	}

	cmd := exec.CommandContext(t.Context(), script, "setup", fixtureRoot)
	if output, err := cmd.CombinedOutput(); err != nil {
		t.Fatalf("set up Git fixture: %v\n%s", err, output)
	}
	t.Cleanup(func() {
		cmd := exec.Command(script, "cleanup", fixtureRoot)
		if output, err := cmd.CombinedOutput(); err != nil {
			t.Errorf("clean up Git fixture: %v\n%s", err, output)
		}
	})

	return readFixtureValue(t, fixtureRoot, "remote-url"),
		plumbing.NewHash(readFixtureValue(t, fixtureRoot, "base-hash")),
		plumbing.NewHash(readFixtureValue(t, fixtureRoot, "blocker-hash"))
}

func readFixtureValue(t *testing.T, root, name string) string {
	t.Helper()

	value, err := os.ReadFile(filepath.Join(root, name))
	if err != nil {
		t.Fatalf("read fixture value %s: %v", name, err)
	}
	return strings.TrimSpace(string(value))
}

func storeDirectChild(
	t *testing.T,
	storer *gitfilesystem.Storage,
	parent *object.Commit,
) plumbing.Hash {
	t.Helper()

	now := time.Now().UTC()
	commit := &object.Commit{
		Author:       object.Signature{Name: "go-git regression test", Email: "repro@localhost", When: now},
		Committer:    object.Signature{Name: "go-git regression test", Email: "repro@localhost", When: now},
		Message:      "direct child of target head\n",
		TreeHash:     parent.TreeHash,
		ParentHashes: []plumbing.Hash{parent.Hash},
	}

	encoded := storer.NewEncodedObject()
	encoded.SetType(plumbing.CommitObject)
	if err := commit.Encode(encoded); err != nil {
		t.Fatalf("encode child commit: %v", err)
	}
	hash, err := storer.SetEncodedObject(encoded)
	if err != nil {
		t.Fatalf("store child commit: %v", err)
	}
	return hash
}

func assertShallowBoundaries(
	t *testing.T,
	shallows []plumbing.Hash,
	want ...plumbing.Hash,
) {
	t.Helper()

	if len(shallows) != len(want) {
		t.Fatalf("shallow boundaries = %v, want %v", shallows, want)
	}
	for _, expected := range want {
		found := false
		for _, actual := range shallows {
			if actual == expected {
				found = true
				break
			}
		}
		if !found {
			t.Fatalf("shallow boundaries %v do not contain %s", shallows, expected)
		}
	}
}
```

</details>

<details>
<summary>testdata/setup_git_remote.sh</summary>

```sh
#!/bin/sh
set -eu

command_name=${1:-}
fixture_root=${2:-}

if [ -z "$command_name" ] || [ -z "$fixture_root" ]; then
  echo "usage: $0 setup|cleanup FIXTURE_ROOT" >&2
  exit 2
fi

pid_file="$fixture_root/git-daemon.pid"

cleanup() {
  if [ ! -f "$pid_file" ]; then
    return
  fi

  daemon_pid=$(cat "$pid_file")
  if kill -0 "$daemon_pid" 2>/dev/null; then
    kill "$daemon_pid" 2>/dev/null || true
    attempt=0
    while kill -0 "$daemon_pid" 2>/dev/null && [ "$attempt" -lt 50 ]; do
      sleep 0.02
      attempt=$((attempt + 1))
    done
    kill -KILL "$daemon_pid" 2>/dev/null || true
  fi
  rm -f "$pid_file"
}

if [ "$command_name" = "cleanup" ]; then
  cleanup
  exit 0
fi

if [ "$command_name" != "setup" ]; then
  echo "unknown command: $command_name" >&2
  exit 2
fi

command -v git >/dev/null
command -v python3 >/dev/null

remote_repo="$fixture_root/remote.git"
seed_repo="$fixture_root/seed"
base_branch="staging-test"
target_branch="flipt/staging-test/target"
blocker_branch="flipt/staging-test/blocker"

git init --quiet --bare "$remote_repo"
git init --quiet --initial-branch="$base_branch" "$seed_repo"
git -C "$seed_repo" config user.name "go-git regression test"
git -C "$seed_repo" config user.email "repro@localhost"
git -C "$seed_repo" commit --quiet --allow-empty -m "root commit"
git -C "$seed_repo" commit --quiet --allow-empty -m "base commit"
git -C "$seed_repo" remote add origin "$remote_repo"
git -C "$seed_repo" push --quiet origin "HEAD:refs/heads/$base_branch"

git -C "$seed_repo" rev-parse HEAD > "$fixture_root/base-hash"
git -C "$seed_repo" push --quiet origin "HEAD:refs/heads/$target_branch"

git -C "$seed_repo" commit --quiet --allow-empty -m "shallow-boundary blocker"
git -C "$seed_repo" rev-parse HEAD > "$fixture_root/blocker-hash"
git -C "$seed_repo" push --quiet origin "HEAD:refs/heads/$blocker_branch"

daemon_port=$(python3 -c 'import socket; s = socket.socket(); s.bind(("127.0.0.1", 0)); print(s.getsockname()[1]); s.close()')
remote_url="git://127.0.0.1:$daemon_port/remote.git"

git daemon \
  --detach \
  --reuseaddr \
  --export-all \
  --enable=receive-pack \
  --pid-file="$pid_file" \
  --base-path="$fixture_root" \
  --listen=127.0.0.1 \
  --port="$daemon_port" \
  "$fixture_root"

attempt=0
while ! git ls-remote "$remote_url" "refs/heads/$base_branch" >/dev/null 2>&1; do
  attempt=$((attempt + 1))
  if [ "$attempt" -ge 100 ]; then
    cleanup
    echo "git daemon did not become ready at $remote_url" >&2
    exit 1
  fi
  sleep 0.02
done

printf '%s\n' "$remote_url" > "$fixture_root/remote-url"
```

</details>

## Additional Information

The failure appears to come from `isFastForward`. It builds one global `parentsToIgnore` list from the parents of every shallow commit. In this graph, `C` is shallow and `parent(C) == B`, so `B` is added to `parentsToIgnore`. During the ancestry walk from `N`, the iterator then skips `B` even though `B` is the exact `old` hash being checked.

This requires multiple shallow boundaries with this relationship; a depth-one fetch of only the target ref does not reproduce the failure.

The issue was initially observed through a downstream Git-backed application that fetches a base branch together with a wildcard set of proposal branches, but the reproducer uses only public go-git APIs and a loopback `git daemon`. It performs no external pushes.

kripa-sindhu-007:
Fixes #2367

## Problem

A depth-one fetch of two branches whose tips are `B` and its child `C` records **both** as shallow boundaries, so `B` is a boundary *and* `C`'s parent. `isFastForward` builds `parentsToIgnore` from the parent hashes of every shallow commit, which puts `B` in the ignore list even though it is stored locally and is the exact `old` the walk is looking for. The walk from the new commit never sees it and `found` stays `false`.

https://github.com/go-git/go-git/blob/d15c624290f14f5eec6f51a660b20b8449ee7d98/remote.go#L1260-L1273

The fallback added in bc7ce77b does not help: `boundedByShallow` only becomes true if the walk *reaches* a shallow commit, and the ignore list is what stops it getting there.

## Fix

Load the shallow commits once, then ignore a parent only when it is not a boundary we hold. The map also replaces `shallowsSet`, which was doing the same membership check with less information.

The smaller version — skip any parent that appears in `shallows` — **reintroduces #207's error**: if a marker names a commit we no longer hold and that commit is another boundary's parent, the walk tries to load it and fails. Measured against a store built for that case:

```
this PR:     isFastForward -> ok=false  err=<nil>
naive skip:  isFastForward -> ok=false  err=object not found
```

## Alignment with upstream git

git does not drop a boundary from the graph — it makes the commit parentless and leaves it in place. `register_shallow()` records the graft with `nr_parent = -1` and clears the commit's parents ([shallow.c#L34-L46](https://github.com/git/git/blob/e9019fcafe0040228b8631c30f97ae1adb61bcdc/shallow.c#L34-L46)); `parse_commit_buffer()` then skips its `parent` lines ([commit.c#L554-L577](https://github.com/git/git/blob/e9019fcafe0040228b8631c30f97ae1adb61bcdc/commit.c#L554-L577)). Confirmed with git 2.55.0 on the same graph: both boundaries recorded, `git rev-parse "$B^@"` empty, `merge-base --is-ancestor` yes, non-force push accepted.

## Tests

`RemoteSuite.TestPushToShallowBoundaryWithDependentBoundary` builds the graph with `PlainInit` and a depth-1 fetch over the file transport, so it needs no `git daemon`. Without the change it fails with the reported error, not a proxy assertion:

```
Error: Received unexpected error:
       non-fast-forward update: refs/heads/target
```

`TestFetchAfterShallowClone_NoForceRefspec` still passes, so #207's fallback is intact. `make validate` and `make test` are green.

## Notes for reviewers

- No backport needed: `git branch -r --contains bc7ce77b` does not list `releases/v5.x`, which still has the earlier single-`earliestShallow` implementation. The regression is v6 only.
- I left the absent-boundary corner above untested — it needs a hand-built store and reads synthetic next to the real one. Happy to add it if you would rather have it pinned.
- @Yiivgeny diagnosed this and offered to test a PR against the original downstream case (flipt-io/flipt#6478).

## AI assistance disclosure

Developed with AI assistance (Claude Code). I understand every line, verified the behaviour against reference git 2.55.0 as described, and confirmed the new test fails without the fix. I will answer review comments personally.

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -1250,16 +1250,10 @@ func isFastForward(s storer.EncodedObjectStorer, old, newHash plumbing.Hash, sha
 		return false, err
 	}
 
-	// Build a set of shallow commits so we can detect when the walk actually
-	// reaches a shallow boundary (as opposed to merely knowing shallows exist).
-	shallowsSet := make(map[plumbing.Hash]struct{}, len(shallows))
-	for _, sh := range shallows {
-		shallowsSet[sh] = struct{}{}
-	}
-
-	// For each known shallow commit, mark its parent hashes as boundaries so
-	// the walker never tries to load commits that are not stored locally.
-	parentsToIgnore := make([]plumbing.Hash, 0, len(shallows))
+	// Load the shallow commits we still hold, so the walk can both detect when
+	// it reaches a shallow boundary (as opposed to merely knowing shallows
+	// exist) and tell a boundary apart from a commit that is genuinely absent.
+	shallowCommits := make(map[plumbing.Hash]*object.Commit, len(shallows))
 	for _, sh := range shallows {
 		shallowCommit, err := object.GetCommit(s, sh)
 		if err != nil {
@@ -1269,14 +1263,36 @@ func isFastForward(s storer.EncodedObjectStorer, old, newHash plumbing.Hash, sha
 			}
 			return false, err
 		}
-		parentsToIgnore = append(parentsToIgnore, shallowCommit.ParentHashes...)
+
+		shallowCommits[sh] = shallowCommit
+	}
+
+	// For each known shallow commit, mark its parent hashes as boundaries so
+	// the walker never tries to load commits that are not stored locally.
+	parentsToIgnore := make([]plumbing.Hash, 0, len(shallows))
+	for _, sh := range shallows {
+		shallowCommit, ok := shallowCommits[sh]
+		if !ok {
+			continue
+		}
+
+		for _, parent := range shallowCommit.ParentHashes {
+			// git truncates at a boundary by making it parentless rather than by
+			// dropping it, so a boundary we hold stays reachable even when
+			// another boundary names it as a parent. See #2367.
+			if _, held := shallowCommits[parent]; held {
+				continue
+			}
+
+			parentsToIgnore = append(parentsToIgnore, parent)
+		}
 	}
 
 	found := false
 	boundedByShallow := false
 	iter := object.NewCommitPreorderIter(c, nil, parentsToIgnore)
 	err = iter.ForEach(func(c *object.Commit) error {
-		if _, isShallow := shallowsSet[c.Hash]; isShallow {
+		if _, isShallow := shallowCommits[c.Hash]; isShallow {
 			// The walk reached a shallow commit; history is truncated here.
 			boundedByShallow = true
 		}
--- a/remote_test.go
+++ b/remote_test.go
@@ -2130,6 +2130,76 @@ func (s *RemoteSuite) TestFetchAfterShallowClone_NoForceRefspec() {
 	s.Equal(sha4, head.Hash(), "local master must point to the new remote tip")
 }
 
+// TestPushToShallowBoundaryWithDependentBoundary is a regression test for
+// https://github.com/go-git/go-git/issues/2367.
+//
+// A depth-one fetch of two branches whose tips are base and its child records
+// both as shallow boundaries, which makes one boundary the parent of another.
+// git grafts a boundary's parents away but keeps the boundary itself reachable,
+// so a child of it is still a fast-forward.
+func (s *RemoteSuite) TestPushToShallowBoundaryWithDependentBoundary() {
+	tempDir := s.T().TempDir()
+	remoteURL := filepath.Join(tempDir, "remote")
+
+	remoteRepo, err := PlainInit(remoteURL, true)
+	s.Require().NoError(err)
+	defer func() { _ = remoteRepo.Close() }()
+
+	now := time.Now()
+	tree := writeEmptyTree(s.T(), remoteRepo)
+	writeCommitToRef(s.T(), remoteRepo, "refs/heads/base", tree, now)
+	base := writeCommitToRef(s.T(), remoteRepo, "refs/heads/base", tree, now.Add(time.Second))
+	s.Require().NoError(remoteRepo.Storer.SetReference(
+		plumbing.NewHashReference("refs/heads/target", base),
+	))
+	s.Require().NoError(remoteRepo.Storer.SetReference(
+		plumbing.NewHashReference("refs/heads/blocker", base),
+	))
+	blocker := writeCommitToRef(s.T(), remoteRepo, "refs/heads/blocker", tree, now.Add(2*time.Second))
+	s.Require().NoError(remoteRepo.Storer.SetReference(
+		plumbing.NewSymbolicReference(plumbing.HEAD, "refs/heads/base"),
+	))
+
+	repo, err := PlainInit(filepath.Join(tempDir, "repo"), false)
+	s.Require().NoError(err)
+	defer func() { _ = repo.Close() }()
+
+	_, err = repo.CreateRemote(&config.RemoteConfig{
+		Name: DefaultRemoteName,
+		URLs: []string{remoteURL},
+	})
+	s.Require().NoError(err)
+
+	err = repo.Fetch(&FetchOptions{
+		Depth: 1,
+		Tags:  plumbing.NoTags,
+		RefSpecs: []config.RefSpec{
+			"+refs/heads/target:refs/remotes/origin/target",
+			"+refs/heads/blocker:refs/remotes/origin/blocker",
+		},
+	})
+	s.Require().NoError(err)
+
+	shallows, err := repo.Storer.Shallow()
+	s.Require().NoError(err)
+	s.ElementsMatch([]plumbing.Hash{base, blocker}, shallows)
+
+	s.Require().NoError(repo.Storer.SetReference(
+		plumbing.NewHashReference("refs/heads/target", base),
+	))
+	child := writeCommitToRef(s.T(), repo, "refs/heads/target", tree, now.Add(3*time.Second))
+
+	err = repo.Push(&PushOptions{
+		RemoteName: DefaultRemoteName,
+		RefSpecs:   []config.RefSpec{"refs/heads/target:refs/heads/target"},
+	})
+	s.Require().NoError(err)
+
+	updated, err := remoteRepo.Reference("refs/heads/target", true)
+	s.Require().NoError(err)
+	s.Equal(child, updated.Hash())
+}
+
 func TestFetchFastForwardForCustomRef(t *testing.T) {
 	t.Parallel()
 	customRef := "refs/custom/branch"
--- a/remote_test.go
+++ b/remote_test.go
@@ -2200,6 +2200,27 @@ func (s *RemoteSuite) TestPushToShallowBoundaryWithDependentBoundary() {
 	s.Equal(child, updated.Hash())
 }
 
+func (s *RemoteSuite) TestIsFastForwardSkipsMissingShallowMarker() {
+	storage := memory.NewStorage()
+	missing := plumbing.NewHash("0000000000000000000000000000000000000001")
+
+	boundary := &object.Commit{ParentHashes: []plumbing.Hash{missing}}
+	encodedBoundary := storage.NewEncodedObject()
+	s.Require().NoError(boundary.Encode(encodedBoundary))
+	boundaryHash, err := storage.SetEncodedObject(encodedBoundary)
+	s.Require().NoError(err)
+
+	child := &object.Commit{ParentHashes: []plumbing.Hash{boundaryHash}}
+	encodedChild := storage.NewEncodedObject()
+	s.Require().NoError(child.Encode(encodedChild))
+	childHash, err := storage.SetEncodedObject(encodedChild)
+	s.Require().NoError(err)
+
+	fastForward, err := isFastForward(storage, missing, childHash, []plumbing.Hash{boundaryHash, missing})
+	s.Require().NoError(err)
+	s.True(fastForward)
+}
+
 func TestFetchFastForwardForCustomRef(t *testing.T) {
 	t.Parallel()
 	customRef := "refs/custom/branch"


kripa-sindhu-007:
Downstream verification came in from two directions.

@Yiivgeny rebuilt Flipt v2.12.0 against this branch and re-ran the scenario from his original report — create a branch from the base environment, advance the base, let the depth-1 fetch run, then edit the branch. The push that was rejected before now succeeds: https://github.com/go-git/go-git/issues/2367#issuecomment-5632908382

Separately, @markphelps wrote a Flipt-side regression test for the same graph. It fails on v6.0.0-alpha.5 and passes with this PR applied, and it is in their repo now, skipped until they can bump. He also checked whether pinning an earlier go-git would do instead — bc7ce77 is in every v6 alpha tag, so there is nothing earlier to pin to: https://github.com/flipt-io/flipt/issues/6478#issuecomment-5626944864

One thing that needs a maintainer: the five workflow runs are sitting at action_required because this is my first PR here, so the PR shows no checks rather than a failed one. At 5a03e446 locally: go test -race -count=1 ./... green across 70 packages, make validate clean, and the new test fails with the remote.go change reverted — with the real non-fast-forward update error, not a proxy assertion.

Happy to answer anything on the fix itself, including the missing-marker guard, which is the one part I chose rather than deduced.

kripa-sindhu-007:
Hi @pjbgf, could you please approve the first-time contributor workflow runs on this PR when you get a chance? They have remained at action_required, so no checks are visible yet. Both are ready for CI and review. Thanks!

PR #2375, opened 2026-09-10
git: worktree, Preserve nonempty directories on removal

### LINKED ISSUE #2374: git: worktree, Reset deletes nonempty submodule directories
Resetting to a commit before a submodule was added removes the submodule's entire worktree, including modified tracked files and untracked files, while returning success.

The removal helper recursively deletes the directory represented by the submodule's index entry. The same helper also deletes untracked contents when a tracked file has been replaced locally by a directory.

Expected behaviour: preserve nonempty directories when removing tracked entries, while continuing to remove ordinary files and empty parent directories. Git 2.54.0 preserves the submodule worktree in the same reset scenario; its removal code is nonrecursive: [entry.c](https://github.com/git/git/blob/v2.54.0/entry.c#L595-L613).

hiddeco:
Fixes #2374.

Reset could recursively delete a removed submodule's worktree, including uncommitted and untracked files. The same removal helper could also delete untracked contents when a tracked file had been replaced by a directory.

Removal now preserves nonempty directories while continuing to remove files, empty directories and empty parents. This follows [Git's nonrecursive removal behaviour](https://github.com/git/git/blob/v2.54.0/entry.c#L595-L613). Nested-repository handling in `Clean` remains a separate issue.

### DIFF
--- a/storage/filesystem/dotgit/dotgit.go
+++ b/storage/filesystem/dotgit/dotgit.go
@@ -1814,7 +1814,7 @@ func isHexAlpha(b byte) bool {
 func incBytes(in []byte) (out []byte, overflow bool) {
 	out = make([]byte, len(in))
 	copy(out, in)
-	for i := len(out) - 1; i >= 0; i-- {
+	for i := range slices.Backward(out) {
 		out[i]++
 		if out[i] != 0 {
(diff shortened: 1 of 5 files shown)


pjbgf (approved):

PR #2379, opened 2026-09-10
plumbing: packfile, Fix streaming delta base rewinds.

### LINKED ISSUE #2378: ReaderFromDelta silently returns incorrect bytes after a backward copy
### Bug Description

`packfile.ReaderFromDelta` can return incorrect content without an error when a delta copies backward through the base and then copies forward again.

For the base `abcdefghijklmnopqrst`, copying one byte from offsets `10`, `0`, and `13` should produce `kan`. The streaming decoder instead returns `kac`, with no error.

The same input produces `kan` through `packfile.PatchDelta`, go-git's regular pack parser, and reference Git.

### go-git Version

Reproduced on v6 `main` at `d15c624290f14f5eec6f51a660b20b8449ee7d98`.

### Steps to Reproduce

In an empty directory:

```sh
go mod init example.com/delta-repro
go get github.com/go-git/go-git/v6@d15c624290f14f5eec6f51a660b20b8449ee7d98
```

Save this as `main.go`:

```go
package main

import (
	"bytes"
	"fmt"
	"io"

	"github.com/go-git/go-git/v6/plumbing/format/packfile"
)

func main() {
	base := &plumbing.MemoryObject{}
	if _, err := base.Write([]byte("abcdefghijklmnopqrst")); err != nil {
		panic(err)
	}

	// Base size 20, target size 3; copy one byte at offsets 10, 0, 13.
	delta := []byte{20, 3, 0x91, 10, 1, 0x90, 1, 0x91, 13, 1}

	r, err := packfile.ReaderFromDelta(base, bytes.NewReader(delta))
	if err != nil {
		panic(err)
	}
	defer func() { _ = r.Close() }()

	out, err := io.ReadAll(r)
	fmt.Printf("%q, %v\n", out, err)
}
```

Run `go run .`.

Expected:

```text
"kan", <nil>
```

Actual:

```text
"kac", <nil>
```

### Additional Information

Tested with Go 1.26.3 on macOS/arm64. Independent checks using SHA-1 and SHA-256 packs matched reference Git's expected output.

When processing a backward copy, `ReaderFromDelta` reopens the base reader at position zero but retains the previous `basePos`. A subsequent forward copy therefore calculates its skip from the wrong position. Resetting `basePos` alongside the reader fixes the reproduction.

carsonfarmer:
Fixes #2378.

`ReaderFromDelta` mishandles the base reader when a copy instruction moves backward:

- It reopens the reader at position zero but leaves `basePos` unchanged. Subsequent forward copies can silently select incorrect bytes. Copying one byte at offsets `10`, `0`, and `13` from `abcdefghijklmnopqrst` returns `kac` instead of `kan`.
- If reopening fails and returns a nil reader, deferred cleanup calls `Close` on nil and panics after the decoder reports its error.

Reset `basePos` alongside the reopened reader, and guard deferred cleanup when reopening did not return a reader. This preserves the existing error result and Git's [base-object copy semantics](https://git-scm.com/docs/pack-format#_instruction_to_copy_from_base_object).

One regression covers repeated rewinds to zero and nonzero offsets; a separate test covers failure on the second `Reader` call. The failed-reopen test uses `synctest` so deferred goroutine cleanup completes before the test ends; it reproduces the panic without the guard. The rewind regression fails without the position reset and with an incomplete zero-offset-only reset.

Independent comparisons with reference Git confirm the corrected output for both SHA-1 and SHA-256 packs. Both changes received independent correctness review.

**Validation**

- Full packfile tests with race detection: `go test -race ./plumbing/format/packfile -count=1`
- Focused race tests passed again after consolidating the rewind tests.
- Repository `golangci-lint` checks passed.

**AI assistance:** Codex assisted with the implementation, tests, and independent review.

### DIFF
--- a/plumbing/format/packfile/patch_delta.go
+++ b/plumbing/format/packfile/patch_delta.go
@@ -182,6 +182,7 @@ func ReaderFromDelta(base plumbing.EncodedObject, deltaRC io.Reader) (io.ReadClo
 						return
 					}
 					baseBuf.Reset(baseRd)
+					basePos = 0
 					discard = offset
 				}
 				for discard > math.MaxInt32 {
--- a/plumbing/format/packfile/patch_delta.go
+++ b/plumbing/format/packfile/patch_delta.go
@@ -138,7 +138,11 @@ func ReaderFromDelta(base plumbing.EncodedObject, deltaRC io.Reader) (io.ReadClo
 			_ = dstWr.CloseWithError(ErrInvalidDelta)
 			return
 		}
-		defer func() { _ = baseRd.Close() }()
+		defer func() {
+			if baseRd != nil {
+				_ = baseRd.Close()
+			}
+		}()
 
 		baseBuf := bufio.NewReader(baseRd)
 		basePos := uint(0)
--- a/plumbing/format/packfile/patch_delta_test.go
+++ b/plumbing/format/packfile/patch_delta_test.go
@@ -181,57 +181,20 @@ func TestReaderFromDeltaRejectsOversizedCopies(t *testing.T) {
 		"ReaderFromDelta yielded more bytes than the declared target size")
 }
 
-func TestReaderFromDeltaCopyPositions(t *testing.T) {
+func TestReaderFromDeltaRepeatedRewinds(t *testing.T) {
 	t.Parallel()
-
-	tests := []struct {
-		name  string
-		delta []byte
-		want  string
-	}{
-		{
-			name: "forward copies",
-			delta: buildDelta(20, 3,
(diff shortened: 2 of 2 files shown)


carsonfarmer:
We missed #2295 when opening this PR... sorry for the duplicate work! It already proposes the same base-position reset for backward copies. This PR also fixes a related cleanup panic when reopening the base returns a nil reader and an error.

Could we combine the fixes from the two PRs into one change? We are happy to coordinate around #2295 and contribute the failed-reopen fix and regression test there, or consolidate here if preferred. We have left this PR intact for comparison.

PR #2383, opened 2026-09-11
plumbing: transport/git, Cover receive-pack in the connect test

hiddeco:
`TestGitTransport_Connect` names its two cases `UploadPack` and `ReceivePack`, but both sent `git-upload-pack`, so the receive-pack command was never requested.

Correcting the command alone does not cover it. The daemon answers a refused command with an `ERR` pkt-line, which is data like any other, so reading four bytes and asserting the count is above zero passes either way. With `--enable=receive-pack` dropped from the daemon the case still passed, on:

    003cERR access denied or repository not exported: /basic.git

Read the first pkt-line through `pktline.ReadLine`, which reports an `ERR` payload as `*pktline.ErrorLine`, and assert the payload is not empty. Both cases then distinguish an advertisement from a refusal: dropping `--enable=receive-pack` fails `ReceivePack` and leaves `UploadPack` passing.

### DIFF
--- a/plumbing/transport/git/git_test.go
+++ b/plumbing/transport/git/git_test.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/go-git/go-git/v6/internal/test/gitenv"
 	"github.com/go-git/go-git/v6/internal/transport/test"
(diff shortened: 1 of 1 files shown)


pjbgf (approved):

PR #2386, opened 2026-09-12
worktree: status, Preserve re-included files below /**

### LINKED ISSUE #2369: Status(): a directory matched by X/** is pruned from the walk, so !X/keep never re-includes (regression in 70ab8844)
`Worktree.Status()` reports a file as ignored when its directory matched `X/**` and a later rule re-includes the file. The pattern matcher gets it right on every version I tested, so this is the walk, not `gitignore`.

Bisected to **`70ab8844`** ("worktree: skip ignored directories during Status walk", #2048, 2026-05-04). Its parent `89091499` agrees with git; `70ab8844` does not; `main` today (`8372f31`) still does not.

### Repro

```sh
mkdir -p t/volumes/functions && cd t && git init -q
printf 'volumes/functions/**\n!volumes/functions/deno.json*\n' > .gitignore
touch volumes/functions/deno.jsonsample volumes/functions/otro.txt

git status --porcelain --ignored
# ?? .gitignore
# ?? volumes/
# !! volumes/functions/otro.txt      <- only otro.txt is ignored
```

`git check-ignore -v` agrees, rule by rule:

```
volumes/functions/deno.jsonsample  ->  .gitignore:2:!volumes/functions/deno.json*   (NOT ignored)
volumes/functions/otro.txt         ->  .gitignore:1:volumes/functions/**            (ignored)
```

Through go-git, opening that worktree and reading `Status()`, `deno.jsonsample` is reported as **not untracked**, i.e. ignored:

| version | `deno.jsonsample` via `Status()` | `otro.txt` |
|---|---|---|
| git (reference) | not ignored | ignored |
| v5.19.2 | not ignored ✔ | ignored ✔ |
| `89091499` (parent of the commit) | not ignored ✔ | ignored ✔ |
| **`70ab8844`** | **ignored ✘** | ignored ✔ |
| `main` `8372f31` | **ignored ✘** | ignored ✔ |

`gitignore.NewMatcher(ReadPatterns(...)).Match([]string{"volumes","functions","deno.jsonsample"}, false)` returns `false` — correct — on *all four*, including `main`. Only the `Status()` walk disagrees, which is why the pattern tests don't catch it.

### Why it happens

The walk prunes a directory when the matcher calls it ignored, and for `volumes/functions` the matcher says `true` — because `**` happily matches the empty remainder. `git check-ignore volumes/functions/` says the same thing (`.gitignore:1:volumes/functions/**`). But git's own walk does *not* treat that as a reason to stop descending: `X/**` excludes the *contents* of X, and a later `!X/keep` can still re-include one of them. Asking `Match(dir, true)` and pruning on the answer conflates "this directory's contents are excluded" with "nothing inside can be re-included".

### The commit also fixed the mirror case, so please don't just revert it

Same shape, but with the directory itself excluded (`X/` instead of `X/**`) — here git really does refuse to descend, and the re-inclusion is dead:

```sh
printf 'volumes/functions/\n!volumes/functions/deno.json*\n' > .gitignore
git status --porcelain --ignored
# !! volumes/          <- everything below is ignored, the negation cannot reach in
```

| version | `deno.jsonsample` via `Status()` |
|---|---|
| git (reference) | ignored |
| v5.19.2 | not ignored ✘ |
| `89091499` | not ignored ✘ |
| `70ab8844`, `main` | ignored ✔ |

So `70ab8844` traded one wrong family for another: it made `X/` correct and `X/**` wrong. What separates them isn't whether the matcher says the directory is ignored — it says yes to both — but whether the pattern excluded the directory itself or only its contents.

### How much it costs

I run a conformance bench that replays 33 real repositories' whole `.gitignore` trees against `git check-ignore` as the oracle: 4,463 matcher queries and 2,224 through `Status()`. `main` diverges on **1 of 2,224** through `Status()`, and this is it — the case above is `supabase/supabase`'s `docker/.gitignore`, verbatim. One query sounds like a corner, so here's the wider count from the same corpus: 5 of the 33 repositories write "ignore everything, keep this" with a negation reaching inside — elasticsearch, grafana, node, next.js, supabase — 14 rule lines between them. My corpus only materialises paths that hit the exact combination once; the construct itself is the one the gitignore docs recommend for the job.

I chased it past the matcher because `Status()` is the door a consumer actually reaches for: someone using go-git to find out what's untracked never calls `ReadPatterns` themselves, so a matcher that's right and a walk that isn't still hands them a wrong answer.

Bench, corpus and the go-git adapter, if you want to reproduce any row: https://github.com/KaizenShogun/gitignore-conformance

znnnnnnn-wil:
## Summary

Fixes #2369. `Worktree.Status()` prunes `volumes/functions` for a rule ending in `volumes/functions/**`, so a later `!volumes/functions/deno.json*` cannot expose a matching file. Git's trailing `/**` excludes the contents, not the parent itself.

Evaluate built-in ignore patterns against the current entry in `Scope`; excluded ancestors remain handled by `Descend`. This lets the walk reach re-included files without applying a directory's negation to its descendants. For example, with `f/**` and `!f/deno.json*`, a file named `f/deno.jsonsample` is visible, but `f/deno.jsonsample/keep` remains ignored when `deno.jsonsample` is a directory. Earlier explicit exclusions such as `f/` still take effect.

Public `Pattern.Match` and `Matcher.Match` behavior stays unchanged. No filesystem-walker special case is needed.

## Reference behavior

https://git-scm.com/docs/gitignore#_pattern_format documents trailing `/**` and the restriction on re-including files below excluded directories. The regression tests compare complete untracked sets with `git ls-files --others --exclude-standard`, including the reported directory-negation cases and nested `.gitignore` files.

## Validation

On Windows, Go 1.26.5:

- New reference-Git regression cases failed against the previous PR commit and pass with the fix.
- `go test . -run 'TestStatus|TestWorktree.*Status' -count=1`: passed.
- `go test ./utils/merkletrie/filesystem -count=1`: passed.
- `go test ./plumbing/format/gitignore ./utils/merkletrie/filesystem -short -skip 'TestMatcherSuite/TestDir_ReadRelativeGlobalGitIgnore' -count=1`: passed.
- `golangci-lint run`: 0 issues.
- `git diff --check`: passed.

The unfiltered gitignore package run has two failures also reproduced on the unchanged PR commit: `TestDir_ReadRelativeGlobalGitIgnore` on this Windows environment, and `TestIgnoreDoubleStarPrefix` with the installed Git 2.38.1 (the test documents the upstream Git fix in 2.52.0). These are not reported as passing.

Paired runs of [gitignore-conformance](https://github.com/KaizenShogun/gitignore-conformance) at `8d202f238ff3827e6e58bdb39a7cee5285f6317b`, using its adapter with v6 imports and repository cleanup:

- `Status()`: 6,714 file queries across 33 repositories, **5 → 4 divergences compared with `main`**: one existing divergence fixed, with no new regressions. The remaining four `apps/docs/.../generated/.gitkeep` divergences are pre-existing, identical to `main`, and untouched by this PR.
- `ReadPatterns` + `Matcher.Match`: 8,953 queries, **the same 12 divergences before and after** (the same paths, not just the same count).

## AI disclosure

Parts of the implementation and tests were assisted by Codex GPT-5 and GPT-6. The validation results above were obtained by running the tests and paired conformance measurements.

### DIFF
--- a/plumbing/format/gitignore/matcher.go
+++ b/plumbing/format/gitignore/matcher.go
@@ -25,11 +25,26 @@ type matcher struct {
 }
 
 func (m *matcher) Match(path []string, isDir bool) bool {
+	excluded, _ := m.matchForTraversal(path, isDir)
+	return excluded
+}
+
+func (m *matcher) matchForTraversal(path []string, isDir bool) (bool, bool) {
 	n := len(m.patterns)
 	for i := n - 1; i >= 0; i-- {
-		if match := m.patterns[i].Match(path, isDir); match > NoMatch {
-			return match == Exclude
+		var match MatchResult
+		var canPrune bool
+		if p, ok := m.patterns[i].(*pattern); ok {
+			match, canPrune = p.matchForTraversal(path, isDir)
+		} else {
+			match = m.patterns[i].Match(path, isDir)
+			// Preserve the historical treatment of custom Pattern
+			// implementations, for which no richer match information exists.
+			canPrune = match == Exclude
+		}
+		if match > NoMatch {
+			return match == Exclude, canPrune
 		}
 	}
-	return false
+	return false, false
 }
--- a/plumbing/format/gitignore/pattern.go
+++ b/plumbing/format/gitignore/pattern.go
@@ -77,26 +77,41 @@ func ParsePattern(p string, domain []string) Pattern {
 }
 
 func (p *pattern) Match(path []string, isDir bool) MatchResult {
+	match, _ := p.matchForTraversal(path, isDir)
+	return match
+}
+
+// matchForTraversal also reports whether a matching exclusion makes it safe
+// to prune a directory from a filesystem walk. A trailing /** may match its
+// parent directory only because isDir stands in for the otherwise empty
+// remainder; that match excludes the directory's contents, not the directory
+// itself, so a later pattern may still re-include a child.
+func (p *pattern) matchForTraversal(path []string, isDir bool) (MatchResult, bool) {
 	if len(path) <= len(p.domain) {
-		return NoMatch
+		return NoMatch, false
 	}
 	for i, e := range p.domain {
 		if path[i] != e {
-			return NoMatch
+			return NoMatch, false
 		}
 	}
 
 	path = path[len(p.domain):]
-	if p.isGlob && !p.globMatch(path, isDir) {
-		return NoMatch
-	} else if !p.isGlob && !p.simpleNameMatch(path, isDir) {
-		return NoMatch
+	canPrune := true
+	if p.isGlob {
+		matched, emptyTrailingStar := p.globMatchForTraversal(path, isDir)
+		if !matched {
+			return NoMatch, false
+		}
+		canPrune = !emptyTrailingStar
+	} else if !p.simpleNameMatch(path, isDir) {
+		return NoMatch, false
 	}
 
 	if p.inclusion {
-		return Include
+		return Include, false
 	}
-	return Exclude
+	return Exclude, canPrune
 }
 
 // The wildmatch implementation below ports the matcher from canonical Git's
@@ -491,10 +506,11 @@ func (p *pattern) simpleNameMatch(path []string, isDir bool) bool {
 	return false
 }
 
-func (p *pattern) globMatch(path []string, isDir bool) bool {
+func (p *pattern) globMatchForTraversal(path []string, isDir bool) (bool, bool) {
 	matched := false
 	canTraverse := false
 	trailingStar := false
+	emptyTrailingStar := false
 	for i, pattern := range p.pattern {
 		if pattern == "" {
 			canTraverse = false
@@ -508,7 +524,8 @@ func (p *pattern) globMatch(path []string, isDir bool) bool {
 				// Assigning matched rather than only raising it stops an
 				// exhausted path from inheriting the previous segment's
 				// result, which would make `a/**/*/**` match `a/f.txt`.
-				matched = len(path) > 0 || isDir
+				emptyTrailingStar = len(path) == 0 && isDir
+				matched = len(path) > 0 || emptyTrailingStar
 				trailingStar = matched
 				break
 			}
@@ -518,7 +535,7 @@ func (p *pattern) globMatch(path []string, isDir bool) bool {
 		// Note: If pattern contains ** but isn't exactly **, it's treated as a regular wildcard pattern
 		// (e.g., foo** or **bar) and wildmatch will handle it
 		if len(path) == 0 {
-			return false
+			return false, false
 		}
 		if canTraverse {
 			canTraverse = false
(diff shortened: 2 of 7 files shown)


KaizenShogun:
I opened #2369, so read this as an interested party's measurement rather than a review.

`Matcher.Match` doesn't move: 12 divergences from `git check-ignore` before and after, the same 12, over 8,953 queries — which is what the commit message promises. `Status()` goes **5 → 8** over 6,714 queries, compared as sets: it fixes the case I reported and breaks four. Base is `da885a30`, the parent of the PR commit.

The four share one shape — a negation that re-includes a directory now leaks into that directory's contents, which git doesn't do. Two lines are enough to see it (the real ones come from `supabase/supabase`'s `docker/.gitignore`):

```
$ git init -q repro && cd repro
$ printf 'f/**\n!f/deno.json*\n' > .gitignore
$ mkdir -p f/deno.jsonsample && touch f/deno.jsonsample/keep
$ git check-ignore -q f/deno.jsonsample/keep && echo ignored
ignored                     # .gitignore:1:f/**  — and `git status -uall` does not list it
```

`!f/deno.json*` re-includes the directory `f/deno.jsonsample`, so git descends into it, and `f/**` ignores what's inside. On `da885a30` `Status()` agrees with git here; with this patch the file comes back as untracked. Same for `main/_gic_keep` under the four-line original, while the case it fixes — `!volumes/functions/main/index.ts` — is right.

Worth one line: an independent patch for the same bug in another project (treeverse/dvc#11096, Python, different author) breaks those same four paths. Whatever is wrong is in the approach, not in this code.

Bench and adapter: https://github.com/KaizenShogun/gitignore-conformance

KaizenShogun:
Re-measured against the new head. The four paths I flagged are gone, and nothing else moved.

Method: `2797bb8f`, its parent `d0dbb214`, and base `da885a30` each built through a `replace` to a worktree of that commit, so the only thing changing between columns is the library. Bench pinned at `8d202f2`, the same commit you used. Oracle is git 2.55.0.

| build | `Matcher.Match` (8,953 queries) | `Status()` (6,714 queries) |
|---|---|---|
| `main` `da885a30` | 12 | 5 |
| PR `d0dbb214` | 12 | 8 |
| PR `2797bb8f` | 12 | **4** |

Compared as sets rather than counts: `Status()` on the new head fixes all four regressions I reported, introduces none, and keeps the #2369 case (`docker/volumes/functions/deno.jsonsample`) fixed. The remaining four are the pre-existing `**/*/generated/.gitkeep` rows, identical to `main` and untouched by this PR. `Matcher.Match` is the same 12 rows before and after — same paths, not just the same number.

The two-line repro from my earlier comment now agrees with git on both lines:

```
$ printf 'f/**\n!f/deno.json*\n' > .gitignore    # + f/deno.json, f/deno.jsonsample/keep
f/deno.json             not ignored   (git: not ignored)
f/deno.jsonsample/keep  ignored       (git: ignored)
```

One framing note, since it's easy to read the other way: the `8 -> 4` in the description is measured against the previous PR commit. Against `main` it's **5 -> 4** — one real fix, zero regressions. Which is the number I'd want if I were reviewing this.

znnnnnnn-wil:
Thanks for re-running the comparison and for pointing out the framing. I've updated the description to compare against `main`: 5 -> 4 divergences, with one existing divergence fixed and no new regressions.

PR #2389, opened 2026-09-13
plumbing: transport/http, Fix the dumb fetch walk for objects already present

### LINKED ISSUE #2380: dumb HTTP transport: fetch fails when object already present locally
While auditing `plumbing/transport/http` we found a set of related correctness bugs in the dumb-protocol path that cause fetch/clone to fail outright on inputs that should be valid. Filing these together since they share a root cause: the dumb-protocol path hasn't received the same hardening as the smart-HTTP path in this package.

## 1. `fetchObject` returns a zero-value object when it's already present, so the whole fetch aborts

`plumbing/transport/http/dumb.go`, `fetchObject` (~line 252):
```go
func (r *fetchWalker) fetchObject(objHash plumbing.Hash, obj plumbing.EncodedObject) (err error) {
	if r.st.HasEncodedObject(objHash) == nil {
		return nil
	}
	...
```
`HasEncodedObject` returns `nil` when the object already exists in local storage. In that case this function returns `nil` without ever populating `obj` (no `SetType`/`SetSize`/`Writer()`), so `obj.Type()` stays `plumbing.InvalidObject`.

The caller, `fetch()` (~line 309), doesn't special-case this — it unconditionally does:
```go
switch obj.Type() {
case plumbing.CommitObject: ...
case plumbing.TreeObject: ...
case plumbing.TagObject: ...
case plumbing.BlobObject: ...
default:
    return plumbing.ErrInvalidType   // ~line 407
}
```
Since `obj.Type()` is `InvalidObject`, this always hits `default` and the entire `Fetch` aborts with `ErrInvalidType`.

This is reachable in practice: `process()` (~line 241) unconditionally enqueues `head` with no `HasEncodedObject` gate — unlike the ref loop right below it, which does gate on `HasEncodedObject(ref.Hash()) != nil`. So **any second dumb-HTTP fetch against a repo already cloned this way** — including a no-op fetch, or an incremental fetch that shares history with what's already stored — hits this path for `head` (or any shared ancestor reached during the walk) and fails.

This matches the fact that `dumbUploadPackSuite` in `dumb_test.go` (~line 97) explicitly overrides `TestUploadPackNoChanges`, `TestUploadPackPartial`, `TestUploadPackFull`, `TestUploadPackMulti`, `TestUploadPack`, `TestUploadPackInvalidReq` with empty bodies — so this regression path is never exercised by the shared conformance suite.

**Suggested fix:** `fetchObject` should signal "already present, skip" distinctly (sentinel error or bool) so `fetch()` can `continue` without decoding an empty `obj`. `process()` should also gate `head` on `HasEncodedObject`, consistent with how other refs are already treated.

## 2. Inverted existence check means pack indexes/files are always re-downloaded

`plumbing/transport/http/dumb.go:231` and `dumb.go:359`:
```go
if _, err := r.fs.Stat(packIdx); errors.Is(err, fs.ErrExist) {
```
`Stat` returns `nil` (not `fs.ErrExist`) when the file exists, and an `fs.ErrNotExist`-wrapping error when it doesn't — it never returns `fs.ErrExist`. So this condition is always false, and the "already have it" branch is dead code: every fetch re-downloads every pack index and every full pack file over HTTP even when already cached on disk.

**Suggested fix:** should be `err == nil` (file exists) rather than `errors.Is(err, fs.ErrExist)`.

## 3. A 404 on `objects/info/packs` is treated as "repository not found"

`plumbing/transport/http/dumb.go:108` (`getInfoPacks`): a non-2xx status is propagated via `checkError`, which maps 404 to `transport.ErrRepositoryNotFound`. But a repo with no packs (all objects loose) legitimately 404s here. `checkError`'s own doc comment states "404 is ordinary control flow for the dumb walk," and `fetchObject` treats a 404 on a loose object the same way — but `getInfoPacks` doesn't get the same treatment, so a valid pack-less repository fails to fetch/clone entirely.

**Suggested fix:** `getInfoPacks` should treat 404 as "no packs" (empty list) rather than propagating `ErrRepositoryNotFound`.

---

Happy to send a PR for any/all of these if useful.

---

*Disclosure: these findings came from an AI-assisted code review (Claude) of this package. The core claims (the zero-value `obj` propagation, the inverted `errors.Is(err, fs.ErrExist)` check, and the 404-on-`info/packs` handling) were manually verified against the current source before filing. Happy to provide more detail or a reproduction if useful.*

thatssoheil:
Fixes #2380.

The dumb HTTP transport's fetch walk aborted on any second fetch of a
repository whose head - or whose shared ancestor - was already stored. All
three defects the issue reports reproduce on current `main` (29c4ef62), and
each is fixed here.

### 1. `fetchObject` returned a zero-value object for an object already present

`fetchObject` returned a bare `nil` for an object `HasEncodedObject` reported as
present, so `fetch` switched on `plumbing.InvalidObject` and returned
`plumbing.ErrInvalidType`. Object-present is now signalled distinctly and the
walk skips the object, the way it already skips references whose objects it
holds. `head` is also gated on `HasEncodedObject` like every other reference
(it was appended unconditionally, and twice).

### 2. The inverted existence check made the "already have it" branch dead

`Stat` returns `nil` for a file that is present and never `fs.ErrExist`, so
`errors.Is(err, fs.ErrExist)` was always false and every walk re-downloaded
every pack index and every pack. Both sites (`dumb.go:231`, `dumb.go:359`) now
test `err == nil`.

### 3. A 404 on `objects/info/packs` was reported as "repository not found"

`checkError` maps 404 to `transport.ErrRepositoryNotFound`, so a repository
whose objects are all loose could not be fetched at all. The dumb protocol
treats the file's absence as "no packs".

### Behaviour change: `FetchRequest.Wants` is now walked

The dumb walker never consulted `wants` - it walked `head` plus every
advertised reference. Consequently the shared conformance case
`TestFetchError` ("fetching an object the server does not have returns an
error") passed only as a side effect of the abort in (1), because the suite
hands the session a storer that already holds everything the server has. With
`head` no longer walked unconditionally that accidental error disappears, so
the wants are walked now: a want the server cannot serve reaches the walk and
is reported. This is the one part of the change the issue did not ask for, and
it is flagged here because it is the only reason `TestFetchError` still means
what it says. If you would rather keep the dumb walker wants-free, the
alternative is to leave that case overridden for this suite like its siblings;
I picked this way because it keeps the case honest.

## Tests

`plumbing/transport/http/dumb_fetch_test.go`, five regression tests, all
failing on `main` and passing here. Verified by reverting only `dumb.go`:
`--- FAIL: TestDumbFetchSkipsObjectsAlreadyPresent | TestDumbFetchSkipsAncestorsItAlreadyHas | TestDumbFetchKeepsPackIndexesItAlreadyHas | TestDumbFetchTreatsMissingInfoPacksAsNoPacks | TestDumbFetchReportsAWantTheServerCannotServe`.

They cover the two shapes the shared suite cannot reach - a storer that starts
empty never walks an object it already has - plus the incremental fetch, the
cached pack index, the missing `objects/info/packs`, and the unsatisfiable want.

`go test ./plumbing/transport/http/` passes. `go test ./...` fails only in
`x/plumbing/worktree`, identically with and without this change (a filemode
assertion that depends on the local umask).

## Reference git

Measured locally against `git` 2.5x over `python3 -m http.server` serving a bare
repository from `git update-server-info`: a dumb clone, a second fetch, an
incremental fetch after the server grew a commit (which is the shape that fails
in go-git today), and a clone with `objects/info/packs` removed - all succeed.

Upstream sources for the semantics:

- `objects/info/packs` absent is not an error: `http-walker.c`,
  `fetch_indices()` handles `HTTP_MISSING_TARGET` exactly like `HTTP_OK`.
- An object already present is not an error: `http-walker.c`, `fetch_object()`
  aborts the pending request and returns success when `odb_has_object()`
  reports the object as present.

## Disclosure

AI-assisted, per `AI_POLICY.md`: one commit, `Signed-off-by` + `Assisted-by`
trailers. I can explain any line on request.


thatssoheil:
Thanks - two of the three are addressed in 658cd92d, the third does not reproduce. Details, with what I measured rather than what I assumed.

**dumb.go: index file descriptor** - real, and it predates this branch: on main the file opened in the walk's pack-index loop is closed only on the decode-error path, so every successful decode leaks one descriptor per pack index per walk, and repeated fetches against a repository with packs could exhaust handles. The loop now closes it once on both paths (658cd92d). Measured with the GC disabled, 40 fetches into fresh on-disk client storages: 40 leaked descriptors before, 0 after. `idxfile.Decode` keeps no reference to the reader once it returns, so closing there cannot break a later read.

**dumb_fetch_test.go: `writeLooseRepository` fixture** - does not reproduce. Both tests that use that helper pass (`go test ./plumbing/transport/http/ -count=1`), so the first object write does not fail and no initialize call is needed. Two reasons: `blobObj.Writer()` cannot fail for a fresh object because `NewEncodedObject` returns an in-memory object and never touches the filesystem, and the directories appear lazily anyway - `objects/pack` when the first object is written, `objects/info` when the repository is prepared for the dumb protocol.

**dumb.go: the `packPath` Stat site** - kept as corrected, deliberately not covered. I instrumented it rather than reasoning about it. A fetch from a packed server by a fresh client does reach the index open and decode, then the Stat miss and the pack download, and that path now has a test: `TestDumbFetchReportsAWantTheServerCannotServeFromAPack` (instrumented hit count: 2 index opens for that test, 0 for every pre-existing test in the file - they either never enter this branch or enter it with no pack index loaded at all). The shortcut itself (pack already present, skip the download) was not reached by any flow I could build through the public API, including a client pre-seeded with the pack bytes but no usable index: once the walk downloads the index the object resolves and never gets here. It is reachable with a client whose storage cache was primed before the pack file appeared (hand-built probe: the shortcut fires, the fetch succeeds). So the corrected comparison stays - the previous form cannot be true on any path - and the shortcut has no test because I could not produce that state from the public API. If you would rather see the line deleted and the pack always re-downloaded, say so and I will drop it.

The code, the tests and the analysis above were verified against the real source with tooling assistance; the PR description carries the disclosure required by this repository.

PR #2391, opened 2026-09-14
plumbing: packp, Stabilize info/refs decoding and validate reference names

hiddeco:
Repeated trailing carriage returns could change reference names across decode/encode cycles or cause the decoder to reject its own output. The decoder also accepted malformed names that could reach callers unexamined.

Hold a decoded reference name to git's reference name rules, allowing the `^{}` suffix for peeled references, and skip the lines that fail. That settles the round trip at its source: a name carrying a carriage return is not one this package can write back. A body converted to CRLF twice therefore loses that reference, which is the reference canonical git loses at `get_fetch_map` for the same reason. Structural errors still reject the advertisement without modifying the receiver. Skipped references are traced with `%q` and their line number, matching the sites that already log a broken name.

Reference: [Git reference-name rules](https://git-scm.com/docs/git-check-ref-format).


pjbgf (approved):

PR #2395, opened 2026-09-15
plumbing: format/config, Index the lookups made while decoding

hiddeco:
`Config.Section` and `Section.Subsection` scan their whole collection, comparing every entry with the name being looked up. The decoder resolves the section of every header and of every option line, so decoding costs O(lines × sections × len(name)).

Resolve both lookups through a map instead, scoped to a single `Decode` call so that it cannot go stale and no exported API changes. The map is keyed on a case fold of the section name rather than its lower case form, because `strings.EqualFold` matches runes that lower case differently, U+017F and `s` for instance. Remembering the section the last header resolved to is part of the fix, not an optimisation: case folding and hashing the name once per option line costs O(lines × len(name)), which is quadratic in the size of an input the name dominates.


plumbing/format/config/decoder.go
```diff
@@ -19,19 +19,135 @@ func NewDecoder(r io.Reader) *Decoder {
 // Decode reads the whole config from its input and stores it in the
 // value pointed to by config.
 func (d *Decoder) Decode(config *Config) error {
+	idx := newDecodeIndex(config)
```
pjbgf, line 27: Add a `nil` check and return an error.

pjbgf (approved):

PR #2400, opened 2026-09-16
internal: url, Align endpoint parsing with canonical Git

hiddeco:
Fix SCP-like endpoint parsing to follow Git's `connect.c` and replace the
endpoint regexps with scanners. Rejecting a 512 KB non-URL drops from 15 ms to
9 µs; parsing representative URLs takes 86% less time with 25% fewer allocations.

## Compatibility changes

- SCP-like URLs no longer interpret a numeric path prefix as a port.
  `git@host:2222:repo.git` now uses the default SSH port and path `2222:repo.git`.
  Use `ssh://git@host:2222/repo.git` to specify a port.
- `host:`, `:path`, hosts containing whitespace, and paths containing backslashes
  or newlines are now treated as SSH endpoints, matching Git. Windows drive
  paths remain local on Windows; `C:\path\to\repo` is treated as SSH on Unix.
- `a:b://c` now fails as an unsupported scheme instead of opening an SSH connection.
- `FindScpLikeComponents` no longer returns a port. It reports whether parsing
  succeeded instead of panicking on non-SCP-like input.

## Other fixes

- Parse bracketed IPv6 hosts such as `[fe80::1]:repo.git` correctly.
- Omit the empty `@` when rendering userless SSH endpoints or resolving relative
  submodule URLs against them.

One intentional difference from Git remains: go-git rejects extra text between
`]` and the following colon. This avoids silently dropping host text when
resolving relative submodule URLs.

## Validation

- Checked parsing changes against Git 2.55 using a logging `GIT_SSH_COMMAND`.
- Deterministic and differential tests compare scanner results and all four
  submatches with the original regexps; differential fuzzing completed 1.8M
  executions without failure.
- `go test -race ./...`: one umask-dependent filemode failure in
  `x/plumbing/worktree`, also present on `main`; remaining tests pass.


plumbing/transport/url_test.go
```diff
@@ -70,13 +70,109 @@ func TestParseURL(t *testing.T) {
 			want:  "ssh://git@github.com/9999/user/repository.git",
 		},
 		{
+			// The SCP-like form has no port: everything after the
+			// first colon is the path, exactly as canonical Git reads
+			// it, so `8080:` here is the start of the path and not a
+			// port to dial.
 			input: "git@github.com:8080:9999/user/repository.git",
-			want:  "ssh://git@github.com:8080/9999/user/repository.git",
+			want:  "ssh://git@github.com/8080:9999/user/repository.git",
```
pjbgf, line 78: This does not align with:
```
ssh://[<user>@]<host>[:<port>]/<path-to-git-repo>
```

https://git-scm.com/docs/git-clone#_git_urls

hiddeco:
```
$ git fetch-pack --diag-url 'git@github.com:8080:9999/user/repository.git'
Diag: userandhost=git@github.com
Diag: port=NONE
Diag: path=8080:9999/user/repository.git
```

Git disagrees here because SCP-like URLs do not support ports.

